PR #31, opened 2015-03-26
Add C++ library compilation support.

KokaKiwi:


src/lib.rs
```diff
@@ -167,19 +175,35 @@ impl Config {
         println!("cargo:rustc-link-search=native={}", dst.display());
         println!("cargo:rustc-link-lib=static={}",
                  &output[3..output.len() - 2]);
+
+        // Add specific C++ libraries, if enabled.
+        if self.cpp {
+            println!("cargo:rustc-link-lib=static=stdc++");
```
alexcrichton: This seems a little surprising to me, I would expect this to be dynamically linked by default. Perhaps this could be its own configuration option?

src/lib.rs
```diff
@@ -167,19 +175,35 @@ impl Config {
...
+        }
+    }
+
+    fn compiler(&self, target: &str) -> String {
+        let compiler_fn: fn(&str) -> String = if self.cpp { gxx } else { gcc };
```
alexcrichton: Instead of creating a function pointer, It may also be ok to just say:

```rust
if self.cpp {gxx(target)} else {gcc(target)}
```

src/lib.rs
```diff
@@ -167,19 +175,35 @@ impl Config {
...
+        compiler_fn(target)
+    }
+
+    fn compile_flags(&self) -> Vec<String> {
+        let flags_fn: fn() -> Vec<String> = if self.cpp { cxxflags } else { cflags };
+        flags_fn()
```
alexcrichton: Similarly to above, I think it's fine to avoid the function pointer and do:

```rust
if self.cpp {cxxflags()} else {cflags()}
```

src/lib.rs
KokaKiwi: Actually it doesn't compile if I omit this line, so I suppose rustc doesn't link the C++ standard library by default. :p

alexcrichton: Yeah I'd definitely believe that, but I think it may be best done as a dynamic dependency (e.g. removing `static=`) instead of a static one. Statically linking libstdc++ is pretty rare I believe.

alexcrichton:
Thanks!

PR #33, opened 2015-03-28
Support stable Rust but make API less pleasant

SSheldon:
I'm not sure this should be merged; it depends on if `convert` will be stabilized before the beta.

In the event that `convert` won't be merged before the beta, the gcc crate can still be compiled on stable Rust by taking a `Path` instead of an `AsRef<Path>` for `include`, `object`, and `file`. This means that callers must construct a `Path` instead of directly passing a str, and it is a breaking change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,17 +33,18 @@
 //! ```no_run
 //! extern crate gcc;
 //!
+//! use std::path::Path;
+//!
 //! fn main() {
 //!     gcc::Config::new()
-//!                 .file("src/foo.c")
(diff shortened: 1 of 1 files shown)


SSheldon:
Alternatively, the API could be changed to only take `&str`, so it'd be as pleasant in the usual case but would be less correct.

alexcrichton:
Thanks for the issue! We plan to stabilize `convert` in one form or another for beta, however, so I'm gonna hold off on migrating away from it. If it turns out that the stabilization doesn't make the cut then I'll tweak some apis for beta.

PR #34, opened 2015-04-19
Enable unstable feature(convert)

alfiedotwtf:
Fixes unstable errors such as:

    Compiling gcc v0.3.4
    /root/.cargo/registry/src/github.com-1ecc6299db9ec823/gcc-0.3.4/src/lib.rs:115:23: 115:34 error: use of unstable library feature 'convert': recently added, experimental traits
    /root/.cargo/registry/src/github.com-1ecc6299db9ec823/gcc-0.3.4/src/lib.rs:115     pub fn include<P: AsRef<Path>>(&mut self, dir: P) -> &mut Config {

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,6 +44,7 @@
 
 #![doc(html_root_url = "http://alexcrichton.com/gcc-rs")]
 #![cfg_attr(test, deny(warnings))]
+#![feature(convert)]
 
 use std::env;
 use std::fs;


alfiedotwtf:
Hold up... this isn't my usual machine. I now see the possible problem:

    root@workhorse:~# rustc --version
    rustc 1.0.0-nightly (552080181 2015-03-27) (built 2015-03-28)

I'll re-try on beta.

alexcrichton:
Thanks for the PR! It looks like you may just need to update your compiler, however. Once you do so you should be good to go!

PR #37, opened 2015-04-29
Point Windows users to the Windows Notes section of readme.

### LINKED ISSUE #35: Steer Windows users towards correct information about installing gcc
Given the confusing naming of various mingw forks floating around Internet, it is unlikely that inexperienced Windows users will manage to install the right one ([example](http://www.reddit.com/r/rust/comments/33r0xu/good_windows_setup_howto)).
I think we should supplement [is 'gcc' not installed?](https://github.com/alexcrichton/gcc-rs/blob/master/src/lib.rs#L255) message with a link to the corrent information.   Not sure where exactly...  we used to have "Windows" page in Rust wiki, but that is now gone.

vadimcn:
Closes #35


alexcrichton:
Nice!

PR #40, opened 2015-05-10
Allow choosing which C++ standard library to use

### LINKED ISSUE #39: C++ Standard Library on OS X
(I'm no expert in these things, all information here is based on (recent) experience and some google-fu).

On OS X there are two difference C++ standard libraries: `libc++` and `libstdc++`. The clang shipped with OS X (which is also the default compiler) uses `libc++` by default, whereas gcc uses `libstdc++`. When compiling c++ code, `gcc-rs` automatically links to `libstdc++`.

This leads to errors at link time. One can work around this problem by either setting `CXX` to a real gcc (**not** the one shipped with OSX, which also uses `libc++`) or setting `CXXFLAGS` to `-stdlib=libstdc++`.

A third alternative would be if it were possible to instruct `gcc-rs` to link to `libc++` instead of `libstdc++`.

I'm not sure what exactly the best solution for `gcc-rs` would be, but I think following goes in the right direction:

* if `gcc-rs` automatically links to `libsdc++`, it should ensure that the compiler uses that specific standard library by automatically specifying `-stdlib=libstdc++`.
* `Config::cpp` should probably take a second argument, which forces the user to think about which standard library to use. Possible values could be `LibStdCpp`, `LibCpp`, `PlatformDefault` (`LibCpp` on OS X, `LibStdCpp` on anything else) and `NoStdLib` (don't specify `-stdlib` but neither automatically link to a standard library).

If this sounds good, I would be happy to implement these changes.

TimNN:
Closes #39.

This implements my proposed solution to issue #39 (with some slight changes). If you prefer a different solution, feel free to close this PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,6 +59,17 @@ pub struct Config {
     flags: Vec<String>,
     files: Vec<PathBuf>,
     cpp: bool,
+    cpp_stdlib: Option<String>,
+}
+
+/// Returns the default C++ standard library for the current target: `libc++`
+/// for OS X and `libstdc++` for anything else.
+fn target_default_cpp_stdlib() -> Option<&'static str> {
+    if getenv_unwrap("TARGET").contains("darwin") {
+        Some("c++")
+    } else {
+        Some("stdc++")
+    }
 }
 
 fn getenv(v: &str) -> Option<String> {
@@ -107,6 +118,7 @@ impl Config {
             flags: Vec::new(),
             files: Vec::new(),
             cpp: false,
+            cpp_stdlib: target_default_cpp_stdlib().map(|s| s.into()),
         }
     }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Awesome, thanks! I think perhaps though just a plain string may be best suited here? Unfortunately new variants of `Stdlib` can't be added backwards-compatibly, so it'd have to be frozen today (which I'm not super sure about). I also feel like the extra value of having an `enum` doesn't necessarily pay off in this case.

Perhaps something like:

```rust
pub fn cpp_stdlib(&mut self, stdlib: &str) -> &mut Self;
pub fn link_cpp_stdlib(&mut self, link: bool) -> &mut Self;
```

TimNN:
Sure that sounds good, although I don't think we really have to worry about adding new variants to `StdLib` because the creation of a new C++ standard library is probably a pretty rare occasion.

Anyway, instead of using two separate functions, I would use just one that takes an option. We can also define known std libs as constants and provide a function to retrieve the default stdlib for the current target. Something like:

```rust
const LibStdCpp: Option<&'static str> = Some("stdc++");
const LibCpp: Option<&'static str> = Some("c++");

pub fn target_default_libstd() -> Option<&'static str> { ... }

pub fn set_cpp_stdlib(&mut self, stdlib: Option<&str>) -> &mut Self {}
```

TimNN:
Alright, I implemented the new design and have to agree that it looks much cleaner & flexible than using the enum.

alexcrichton:
Perhaps this could avoid the `target_default_libstd` and public constants for now? I can see where there might be a use for them, but I would prefer to start out conservative and just add the ability to customize the C++ standard library in use.

TimNN:
Sure. Then we just default to `libstdc++`, no matter the target? And from an api design standpoint, should `Config::cpp_std_lib` take a `String` or a `&str` and clone it internally?

alexcrichton:
Oh I think it's fine to use `libc++` on OSX by default (as it's probably what you want anyway), and I'd take `Option<&str>` and clone internally (this API has basically no performance impact at all).

Perhaps also `cpp_stdlib` as the method name? (e.g. it's rarely broken up into "std lib")

TimNN:
I think it's either `stdlib` or `lib_std` (which I'm currently using). But either is fine with me.

TimNN:
Sure!

TimNN:
Alright, I removed the public constants / function and used `stdlib` everywhere.

alexcrichton:
Thanks!

PR #42, opened 2015-06-02
added file globbing support.

OxyDeadbeef:
File globbing is very useful and makes listing source files more practical, so I decided to add it to the library.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,3 +12,6 @@ C compiler to compile native C code into a static archive to be linked into Rust
 code.
 """
 keywords = ["build-dependencies"]
+
+[dependencies]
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! I think that this is easy enough to add to a build script itself, however, and I'd rather not pick up the extra dependency for now. For example:

```rust
extern crate glob;
extern crate gcc;

fn main() {
    let mut cfg = gcc::Config::new();
    for file in glob::glob("src/build/*.c").unwrap().filter_map(|a| a.ok()) {
        cfg.file(file);
    }
    cfg.compile("libfoo.a");
}
```

Issue #43, opened 2015-06-03
Command Line Option Not Supported In All Versions Of Gcc


OxyDeadbeef:
I get this error:
`
c++: error: unrecognized command line option ‘-stdlib=libstdc++’
`
When I try to compile c++ code from my rust build script with .cpp(true)
It appears as though `-stdlib=libstdc++` is not an accepted parameter by my version of gcc.

alexcrichton:
cc @TimNN, upon re-reading #39 I'm thinking the best option is to just not pass `-stdlib` by default (and instead rely on the compiler default). You mentioned that this can cause link errors on OSX though, so I'm wondering if you can elaborate on that?

TimNN:
The problem was that `gcc-rs` was unconditionally instructing cargo to link the rust code to `libstdc++` when compiling with `.ccp(true)`. [Original code](https://github.com/alexcrichton/gcc-rs/blob/b8da3f50a0a0b03e7fa0103e9dc8be9f776b9ce3/src/lib.rs#L178-L181).

On OS X the default compiler is clang, which however links to `libc++` thus resulting in linking errors.

As such the [`cpp_stdlib`](http://alexcrichton.com/gcc-rs/gcc/struct.Config.html#method.cpp_stdlib) option was added. It allows choosing which stdlib to use and if one is choosen, it will force the compiler to use this library (using the `-stdlib=` flag) and instructs cargo to link the rust code to this library.

@OxyDeadbeef which OS / compiler version are you using?

For the moment you can work around the problem by manually creating a [`Config`](http://alexcrichton.com/gcc-rs/gcc/struct.Config.html) and setting `config.cpp_stdlib(None)`. You will then have to manually link to `libstdc++` by writing `println!("cargo:rustc-link-lib=stdc++");` in your `build.rs` file.

TimNN:
Alright I did some digging and might have been a little careless when testing my original pull request:

OS X provides a `g++` in `/usr/bin`, which is just a wrapper around `clang` and thus supports `-stdlib=` and I have installed a 'real' `g++` in `/usr/local/bin` (which I usually have before `/usr/bin` in my path). For some reason while testing the code I have was using the OS X provided `g++` and thus everything worked.

[...]

PR #44, opened 2015-06-03
simplify cpp stdlib handling, don't use -stdlib=

TimNN:
As discussed in #43 my initial implementation of the `cpp_stdlib` option was based on some incorrect assumptions (that g++ supports the `-stdlib` flag).

From the discussion in #40 I believe the current goal is to keep things as simple as possible. As such all usages of `-stdlib` have been removed. `cpp_stdlib` now only determines which library should be linked to when compiling with C++ support and defaults to `libc++` on OS X and `libstdc++` on anything else.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -162,19 +162,16 @@ impl Config {
         self
     }
 
-    /// Set the standard library to use when compiling with C++ support.
+    /// Set the standard library to link against when compiling with C++
+    /// support.
     ///
     /// The default value of this property depends on the current target: On
     /// OS X `Some("c++")` is used and for other supported targets
     /// `Some("stdc++")` is used.
     ///
-    /// The choosen library is specified when compiling using the `-stdlib` flag
-    /// and Cargo is instructed to link to the given library.
-    ///
     /// A value of `None` indicates that no special actions regarding the C++
-    /// standard library should be take. This means that neither are linking
-    /// instructions provided to Cargo nor is the compiler told to use a
-    /// specific standard library.
+    /// standard library should be take. This means that you have to link
+    /// against the correct standard library yourself if required.
     ///
     /// The given library name must not contain the `lib` prefix.
     pub fn cpp_stdlib(&mut self, cpp_stdlib: Option<&str>) -> &mut Config {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm, if we're not passing the `-stdlib` flag then there doesn't actually seem like much point for the `cpp_stdlib` method at all? This is basically just adding a linked library at the end at that point.

Perhaps something like:

* `cpp_stdlib` stays, and indicates whether the `-stdlib` option will be passed, but it defaults to `None`
* The auto-linking behavior to the C++ standard library stays, defaulting to the current set of defaults you have. The auto-linked library changes if `cpp_stdlib` is `Some`.
* Eventually an option can be added to disable the auto-link behavior.

TimNN:
I think we have to decide what the `cpp_stdlib` option actually means: 

a) Automatically link against this stdlib or
b) Force the compiler to use this stdlib and link against it

After some deliberation I believe that b) will be the less common use case. It can also already be achieved using `Config::flag()`. The more common use case would be that a user wants to change which stdlib is linked against (or don't link to any stdlib at all). This behaviour would also be required to support the msvc target (which sets `cpp_stdlib` to `None` to prevent any automatic linking).

Regarding your proposal, let me paraphrase to ensure I understood correctly:

* If `cpp_stdlib` is `Some(..)` the `-stdlib=` flag will be passed to the compiler.
* `cpp_stdlib` defaults to `None`.
* If `cpp_stdlib` is `Some(..)` we link to the chosen stdlib, otherwise we link to the platforms default stdlib (`libc++` on OS X and `libstdc++` everywhere else).
* There will be no way to disable automatically linking to a given stdlib.

I believe that this makes things more complicated than they need to be and emphasises a less common use case (as mentioned above), especially because b) can easily be achieved using a) and `Config::flag` but disabling automatic linking would be impossible with b) without another option (which would be required anyway (at least internally) to support msvc).

alexcrichton:
Pretty much everything here can boil down to `flag`, but it's useful to give semantic meaning to each flag through a method instead. Your paraphrase of what I was thinking is correct, with the caveat that a method to disable auto-linking can be added later.

The `cpp_stdlib` function can't be removed now, so I figured a semantic meaning may as well be attached to it, and the `-stdlib` parameter seems reasonable at least.

TimNN:
Alright, I have implemented your suggested solution although I did already expose a way to disable (or change) auto-linking by providing two distinct options, `cpp_link_stdlib` and `cpp_set_stdlib`. This felt cleaner than using only one option which affected two different things especially since we would have needed two options internally anyway.

alexcrichton:
Awesome, thanks!

PR #47, opened 2015-06-27
RFC: Quasi-incremental builds

jdub:
**Objective:** Avoid lengthy, unnecessary compiler runs on every build by doing basic dependency checking.

Next steps:

- [x] Capture header dependencies from a preprocessor run, e.g. `cc -M -MG` (which works with `gcc` and `clang`).
- [x] Build a cache of dependencies and their modification times, to avoid duplicate work after each preprocessor run.
- [x] Add support for Windows systems, noting that the `MetadataExt`-provided modification time properties will be different. (Aha, [filetime](https://crates.io/crates/filetime)!)
- [ ] Support Rust 1.0 (which lacks `MetadataExt`), if @alexcrichton regards that as important. (Doesn't look like the `filetime` helps.)
- [ ] Add header dependency support for `msvc`.

Notes:

* Basic [performance testing](https://gist.github.com/jdub/9382c3e9ebcbb839d6f6) suggests that doing a single preprocessor run is not a significant win. This means we can skip the preprocessor run (and dependency modification checks) entirely if the input file itself is changed.
* Investigate preprocessor output of `msvc`. The `/showIncludes` parameter looks promising (though probably not as neat as the Makefile output).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,8 +52,8 @@ use std::io;
 use std::path::{PathBuf, Path};
 use std::process::{Command, Stdio};
 
-#[cfg(unix)]
-use std::os::unix::prelude::*;
+extern crate filetime;
+use filetime::FileTime;
 
 /// Extra configuration to pass to gcc.
 pub struct Config {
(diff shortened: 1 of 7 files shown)


jdub:
Requires Rust 1.1.0 for MetadataExt. Worth making it 1.0 compatible?

src/lib.rs
```diff
@@ -224,10 +227,28 @@ impl Config {
         let dst = PathBuf::from(getenv_unwrap("OUT_DIR"));
         let mut objects = Vec::new();
         for file in self.files.iter() {
+            let obj = dst.join(file).with_extension("o");
+
+            if cfg!(unix) {
+                let file_mtime = match fs::metadata(&file) {
+                    Ok(meta) => meta.mtime(),
+                    Err(_) => 0,
+                };
```
jdub: Hmm, am I really looking for `map_or` here?

alexcrichton:
Unfortunately I don't think this is quite sufficient for a few reasons:

* Windows support will be needed in addition to unix support.
* This doesn't take header files into account, which are often needed to trigger recompilations.

I'd certainly be opening to adding this, but those would need to be fixed head of time I believe.

jdub:
D'oh, I was hoping to sneak through without Window support. :wink: I'll muck around with that.

Very good point on the headers front. A nice bump in the degree of difficulty.

oprudkyi:
for those who still need this , use of ccache make like easier 
CC="/usr/local/lib64/ccache/gcc" CXX="/usr/local/lib64/ccache/g++" cargo build

alexcrichton:
Closing due to inactivity, but I'd still love to see progress on this!

PR #49, opened 2015-06-30
Perform debug builds under debug profile

jdub:


jdub:
Is this appropriate for the `ios` target? I'm assuming it would use `clang`, and `-g` is compatible and desirable.

alexcrichton:
Oh wow, this is like the one thing you expect gcc-rs to do and I forgot to do it, thanks!

I think `/DEBUG` on MSVC is just a flag to the linker, not the compiler itself, but I think `/Z7` should do the trick.

Also yeah calling gcc or clang with `-g` should be fine.

jdub:
D'oh! Fixed.

alexcrichton:
Thanks!

PR #54, opened 2015-09-09
Add the option to disable linking against the main crate

ngrewe:
Hi,

I have a suggestion for an improvement: At the moment, the archives produced by a build script using gcc-rs are unconditionally linked to the crate that is being built. But there are certain use cases where that is not desirable, for example when you only want to [link the built artifact for tests](https://users.rust-lang.org/t/cargo-how-to-build-external-static-lib-only-when-testing/1241). This is presently only possible by dropping down to constructing the compiler command line manually, which is not really desirable for portability reasons.
So I propose adding a `link()` option that you can add to the configuration in order to disable linking. The default would still be `true`, of course. Please have a look and tell me what you think. 

Thanks,

Niels

### DIFF
--- a/gcc-test/build.rs
+++ b/gcc-test/build.rs
@@ -72,5 +72,5 @@ fn main() {
 
     // This tests whether we  can build a library but not link it to the main crate.
     // The test module will do its own linking.
-    gcc::Config::new().link(false).file("src/opt_linkage.c").compile("libOptLinkage.a");
+    gcc::Config::new().cargo_metadata(false).file("src/opt_linkage.c").compile("libOptLinkage.a");
 }
--- a/gcc-test/build.rs
+++ b/gcc-test/build.rs
@@ -69,4 +69,8 @@ fn main() {
         println!("cargo:rustc-link-lib=msvc");
         println!("cargo:rustc-link-search={}", out.display());
     }
+
+    // This tests whether we  can build a library but not link it to the main crate.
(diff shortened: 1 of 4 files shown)


alexcrichton:
Sounds like a good idea to me, thanks! Could you also guard the print for `rustc-link-search` and the C++ standard library below on this flag?

I also think that `link` may not be the best name as it generally implies that the linker itself is being invoked, perhaps something like `cargo_metadata`?

ngrewe:
>  Could you also guard the print for rustc-link-search and the C++ standard library below on this flag?

Actually there's small convenience upside in keeping the println statement for the search path, but that might be trumped by consistency concerns (either you emit metadata for cargo or you don't), and it's not that difficult to get the reference to the `OUT_DIR` and add that yourself. 

> I also think that link may not be the best name as it generally implies that the linker itself is being invoked, perhaps something like `cargo_metadata`?

Yep, that's a good idea.

alexcrichton:
Yeah these are basically all pure convenience (e.g. doing the prints for you) so I'd prefer to be consistent and not print any metadata if this option is specified, and as you mentioned you can always print the `OUT_DIR` as a search path in the script afterwards.

ngrewe:
>  I'd prefer to be consistent and not print any metadata if this option is specified

Agreed. That's what the updated PR does now.

alexcrichton:
Thanks!

PR #57, opened 2015-09-20
If hard-linking fails, try to copy (in assemble())

kosta:
Fix for the failure described [here](https://users.rust-lang.org/t/openssl-sys-on-rust-32-bit-msvc-does-it-work/2893/2). I have my code checked out in a Virtual Machine on a Shared Drive (so I can e.g. use git outside the VM). Hard_linking doesn't work there.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -468,10 +468,10 @@ impl Config {
             // exist for now.
             let lib_dst = dst.with_file_name(format!("{}.lib", lib_name));
             let _ = fs::remove_file(&lib_dst);
-            if fs::hard_link(&dst, &lib_dst).is_err() {
-                println!("Hard-linking from {:?} to {:?} failed. Trying to copy...", dst, lib_dst);
-                fs::copy(&dst, &lib_dst).expect("Copying from {:?} to {:?} FAILED.");
-            };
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! I think it's fine to not worry about printing too much here so how about:

```rust
fs::hard_link(&dst, &lib_dst).or_else(|_| {
    fs::copy(&dst, &lib_dst)
}).unwrap()
```

kosta:
Ok I removed the println! but used expect() instead of unwrap() to get a better error message. Especially on Windows, where there's no (?) stack unwinding support, it's hard to find the underlying reason for a simple unwrap panic.

If you want this as a single commit, please say so and I'll rebase.

alexcrichton:
Could you also throw an `.ok()` in there before the `expect`? That was added recently and wasn't present in versions of older Rust (hence the CI failure)

kosta:
Done.

alexcrichton:
Thanks!

kosta:
Cool!

*writes Rust contributer on Resume* ;)

kosta:
Would you mind releasing this (e.g. as 0.3.16) so that openssl-sys can use the bugfix? :)

alexcrichton:
Certainly! (done)

PR #59, opened 2015-10-19
Clean whitespace

aarzee:
Remove leading newlines; replace lines containing only whitespace with empty lines; replace multiple trailing newlines with a single newline; remove trailing whitespace in lines.

This PR was created semiautomatically.


alexcrichton:
Thanks!

PR #61, opened 2015-10-26
Add Clone impl for `Config` and `Tool`

apoelstra:
I've got a project where I want to build a configuration most of the way (setting a bunch of defines, etc), then compile two files as separate libraries with it. To do so, I need to clone the `Config` object.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,6 +58,7 @@ mod registry;
 pub mod windows_registry;
 
 /// Extra configuration to pass to gcc.
+#[derive(Clone)]
 pub struct Config {
     include_directories: Vec<PathBuf>,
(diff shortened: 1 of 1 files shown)


alexcrichton:
I'd be fine adding `Clone` to `Tool`, but I'd prefer to leave it off `Config` for now. It may be the case in the future that something like a trait object makes its way in the configuration (e.g. some closure of some form) in which case `Clone` may be difficult to implement.

apoelstra:
Ah, that's very reasonable. I realized that for my use-case it's perfectly fine to just lump all my files into one library, since it's only used by my Rust project and is basically ephemeral anyway.

As for `Tool`, I don't actually have a use for making it be Clone; I just did it for symmetry :)

PR #67, opened 2015-12-22
Add `pic` method to override default emission of `-fPIC`

### LINKED ISSUE #64: Support custom target files
Right now the crate uses only the target _string_ for automatic configuration. For example, it compiles the dependencies with `-fPIC` if the target string does not contain `i686` or `windows-gnu`. This can lead to surprising behavior when using custom target files since their values are completely ignored.

For example, the resulting library uses position independent code (and a global offset table) even if the target json contains `"relocation-model": "static"`… except for target files that have `i686` somewhere in their file name. And renaming the file is the only option to override this automatic behavior.

phil-opp:
Adds a `pic` method to `Config` that allows overriding the default emission of `-fPIC`.

Closes #64

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,6 +76,7 @@ pub struct Config {
     compiler: Option<PathBuf>,
     archiver: Option<PathBuf>,
     cargo_metadata: bool,
+    pic: Option<bool>,
 }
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

PR #70, opened 2016-02-06
Preliminary support for building CUDA C files with nvcc.

peterhj:
This adds preliminary support for compiling CUDA C sources using the nvcc compiler shipped with the CUDA toolkit.

Currently, this only supports the typical "one-pass" compilation steps of: `*.cu` -> `*.o` -> `*.a`. The main differences from gcc consist of: passing different PIC args to the compiler, and omitting some args that nvcc does not support.

(Not implemented is a more involved "two-pass" compilation required for supporting dynamic parallelism, i.e. dynamic kernel launches. This requires some extra flags and linking with a certain shared object.)

I've personally used this on the CUDA 7.0 and CUDA 7.5 toolchains.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -397,6 +397,9 @@ impl Config {
         self.print(&format!("debug={} opt-level={}", debug, opt_level));
 
         let mut cmd = self.get_base_compiler();
+        let nvcc = cmd.path.to_str()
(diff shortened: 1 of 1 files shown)


alexcrichton:
Neat!

PR #73, opened 2016-02-19
Add support for MIPS little-endian targets

### LINKED ISSUE #72: Support MIPS little-endian targets
Currently when cross-compiling for MIPS the only supported prefix is the one for big-endian targets (mips-unknown-linux-gnu), little-endian targets have their own prefix that should be added to the supported prefix list (mipsel-unknown-linux-gnu).

Also, using an unsupported prefix silently uses the host compiler which doesn't seem like sane behaviour since it will cause objects in a project to be compiled for the wrong target.

gabrielesvelto:
Closes #72

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -625,6 +625,7 @@ impl Config {
                     "powerpc64-unknown-linux-gnu" => Some("powerpc-linux-gnu"),
                     "powerpc64le-unknown-linux-gnu" => Some("powerpc64le-linux-gnu"),
                     "mips-unknown-linux-gnu" => Some("mips-linux-gnu"),
+                    "mipsel-unknown-linux-gnu" => Some("mipsel-linux-gnu"),
                     "i686-pc-windows-gnu" => Some("i686-w64-mingw32"),
                     "x86_64-pc-windows-gnu" => Some("x86_64-w64-mingw32"),
                     "x86_64-unknown-linux-musl" => Some("musl"),


src/lib.rs
```diff
@@ -625,11 +625,12 @@ impl Config {
                     "powerpc64-unknown-linux-gnu" => Some("powerpc-linux-gnu"),
                     "powerpc64le-unknown-linux-gnu" => Some("powerpc64le-linux-gnu"),
                     "mips-unknown-linux-gnu" => Some("mips-linux-gnu"),
+                    "mipsel-unknown-linux-gnu" => Some("mipsel-linux-gnu"),
                     "i686-pc-windows-gnu" => Some("i686-w64-mingw32"),
                     "x86_64-pc-windows-gnu" => Some("x86_64-w64-mingw32"),
                     "x86_64-unknown-linux-musl" => Some("musl"),
                     "x86_64-rumprun-netbsd" => Some("x86_64-rumprun-netbsd"),
-                    _ => None,
+                    _ => fail(&format!("Unknown target prefix `{}`", &target[..])),
```
alexcrichton: This should stick around as `None` for now as it mirrors the behavior of many other standard cross-compilation setups. It's unfortunate, but it could always be the case that a cross compile with gcc actually works!

gabrielesvelto: I see, I'll revert that change then, thanks!

gabrielesvelto:
Amended the commit to not fail when the target is not known and changed the commit message accordingly.

alexcrichton:
Thanks!

PR #74, opened 2016-03-30
Split CC/CXX environment variable by space characters

rhenium:
People sometimes set CC="ccache gcc" or CC="distcc gcc", but gcc-rs currently doesn't recognize arguments in CC/CXX and treats the whole value as the compiler executable.
With this patch, CC/CXX environment variables will be splitted by whitespace charactors. The first string is used as the compiler, and the rest are used as arguments for the compiler.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -417,8 +417,10 @@ impl Config {
             cmd.args.push("-ffunction-sections".into());
             cmd.args.push("-fdata-sections".into());
         }
-        for arg in self.envflags(if self.cpp {"CXXFLAGS"} else {"CFLAGS"}) {
-            cmd.args.push(arg.into());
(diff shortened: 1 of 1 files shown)


rhenium:
I just noticed the current README explicitly says that no extra flags can be passed inside CC.
If it is intentional behavior, please reject this, sorry.

alexcrichton:
Ah yeah currently with behavior like this we unfortunately wouldn't have the ability to specify paths with spaces in them. For now we'll stick with the current behavior, but I'd be fine adding ccache-specific options if this is common enough!

rhenium:
I understand, thanks for the explaination!
I have a Makefile which calls `cargo build` inside it, and encountered this when I ran `make CC="ccache gcc"` as usual I do for other C projects.

PR #76, opened 2016-04-13
Add support for ARM MUSL cross-compilers.

timonvo:
This paves the way for adding support for statically built Rust
binaries for ARM using MUSL libc.

The standard suffixes for ARM MUSL cross-compilers are
musleabi/musleabihf, as opposed to gnueabi/gnueabihf.


alexcrichton:
Neat!

PR #79, opened 2016-04-30
Add .link option to compile to simplify linking

Firstyear:
Some helper libraries in rust require to be linked as well. For example:

rs/
   src/
       main.rs
       native.c

Where main.rs links to native, and native.c will be linked to some platform
library, such as netsnmp.

This allows build.rs to contain:

gcc::Config::new()
   .link("netsnmp")
   ....

Which hugely simplifies the build process, and is very reliable compared to
trying to set cflags.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,6 +77,7 @@ pub struct Config {
     archiver: Option<PathBuf>,
     cargo_metadata: bool,
     pic: Option<bool>,
+    links: Vec<String>,
 }
 
 /// Configuration used to represent an invocation of a C compiler.
@@ -137,6 +138,7 @@ impl Config {
             archiver: None,
             cargo_metadata: true,
             pic: None,
+            links: Vec::new(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
This could just be done as part of a build script, right? Is there a particular reason why gcc needs to be the crate that's printing this out?

Firstyear:
Hmmm that's true. I didn't think of that.

At the same time, the C++ stdlib parts also do the same thing?

Anyway, perhaps this pull request either is still valid, or, it is an example that can be added to the cargo build script page?

alexcrichton:
The C++ library changes name across platforms (and sometimes even with a platform), so to me it makes sense to help out with the concept of "I'd like to link C++". Here, however, there's not platform variance and it's just another way to write `println!("cargo:rustc-link-lib={}", "foo")` essentially?

Firstyear:
Yep, I think you are right. This change is irrelevant. It would be a good example on http://doc.crates.io/build-script.html#inputs-to-the-build-script I feel, as I didn't quite connect that I would need to do this for linking the c -> external lib.

Thanks for your time,

PR #85, opened 2016-07-11
Update MSVC logic for most recent VS

alexcrichton:
Mostly just matching what the compiler is currently doing.

Closes https://github.com/alexcrichton/gcc-rs/issues/83
cc https://github.com/alexcrichton/gcc-rs/issues/84

### DIFF
--- a/gcc-test/build.rs
+++ b/gcc-test/build.rs
@@ -46,6 +46,7 @@ fn main() {
     if target.contains("msvc") {
         let out = out.join("tmp");
         fs::create_dir(&out).unwrap();
+        println!("nmake 1");
         let status = gcc::windows_registry::find(&target, "nmake.exe").unwrap()
                             .arg("/fsrc/NMakefile")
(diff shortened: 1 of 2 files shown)


gcc-test/build.rs
```diff
@@ -46,6 +46,7 @@ fn main() {
     if target.contains("msvc") {
         let out = out.join("tmp");
         fs::create_dir(&out).unwrap();
+        println!("nmake 1");
```
retep998: Just checking, this `println!` is intentional?

alexcrichton: Yeah I just added these when debugging this test.

src/windows_registry.rs
```diff
@@ -15,6 +15,13 @@ use std::process::Command;
 
 use Tool;
 
+macro_rules! otry {
```
retep998: So, when are we going to finally get a standard way of doing this in Rust/libstd? :stuck_out_tongue:

PR #91, opened 2016-08-19
Emit cargo warnings when compilers emit warnings

### LINKED ISSUE #81: (optionally) emit warning messages from C(++) compiler
Once rust-lang/cargo#2630 lands, it would be awesome if there was an option to tell gcc-rs to forward the warning messages produced by the c/c++ compiler, but prefixed with `cargo:warning=`, to the build script's standard out. This would be useful for getting and fixing error messages in C++ and C code which is compiled through the build plugin as a companion to rust code.

mystor:
Fixes #81.

This is a quick implementation which captures stderr from the compiler process, and writes out the errors to the build plugin's stdout with the `cargo:warning=` prefix. I'm not sure if it's the nicest implementation, but it seems to do the trick from my manual testing.

I'm not sure how best to write tests for this feature, and I'd appreciate input as to how you would like something like that done.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,6 +52,7 @@ use std::fs;
 use std::io;
 use std::path::{PathBuf, Path};
 use std::process::{Command, Stdio};
+use std::io::{BufReader, BufRead};
 
 #[cfg(windows)]
 mod registry;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Nice! In general gcc-rs is a little light on tests, so I'm fine omitting tests here.

Issue #90, opened 2016-08-18
Support for parallel compilation


nipunn1313:
Would be nice to compile files in parallel (equivalent to `make -j4` for example).

alexcrichton:
Fixed by https://github.com/alexcrichton/gcc-rs/pull/92

nipunn1313:
Can you up the version number to include this?

alexcrichton:
Sure! Pushed 0.3.33

nipunn1313:
Thanks!

nipunn1313:
I tried applying the features flag to this project (https://github.com/nipunn1313/rust-rocksdb), but didn't seem to get any speedup. I confirmed that the rayon code path is being taken with a parallelism of 4, but I don't seem to see 4 tasks getting spun out concurrently. I confirmed that the majority of time is spent in the "compile_objects" function rather than the "assemble" function as well.

Still trying to get to the bottom of it, but I thought I'd see if you had any quick insight on this one.

PR #92, opened 2016-08-19
Use rayon for parallel compilation

nipunn1313:
Here's an attempt to add support parallel compilation (See #90).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,5 +13,8 @@ code.
 """
 keywords = ["build-dependencies"]
 
+[dependencies]
+rayon = "0.4"
+
 [dev-dependencies]
 tempdir = "0.3"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,12 +46,16 @@
 #![cfg_attr(test, deny(warnings))]
 #![deny(missing_docs)]
 
+extern crate rayon;
+
 use std::env;
 use std::ffi::{OsString, OsStr};
 use std::fs;
 use std::io;
 use std::path::{PathBuf, Path};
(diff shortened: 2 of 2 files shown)


alexcrichton:
Thanks for the PR! I find it kinda crazy that it amounts to a 2-line change when using crates like rayon to add parallelization here...

I'd like to also handle features like [`NUM_JOBS`](http://doc.crates.io/environment-variables.html#environment-variables-cargo-sets-for-build-scripts), however, to limit parallelism and I'd personally like to keep the dependencies of gcc-rs pretty lean (rayon pulls in rand/num_cpus/libc as well).

I think I'll probably just do this via a small "custom thread pool" which is a little bit more code but shouldn't be too bad! I'll post a commit in a few minutes.

alexcrichton:
Bah nevermind, too complicated, crates are easy!

nikomatsakis:
@alexcrichton note that Rayon does allow you to configure the number of helper threads

nikomatsakis:
I think you can either set it globally, or make a special pool of a fixed size and install it using TLS, I forget exactly :)

alexcrichton:
Ok hm I'm having second thoughts now, so I've moved rayon to a Cargo feature called `parallel` which is off by default for now (in the theory that most C code is actually very small). This can be enabled though via:

```toml
[build-dependencies]
gcc = { version = "0.3", features = ["parallel"] }
```

nipunn1313:
Yep makes sense! I also had some qualms about adding a dependency to a build step, but feature gating seems like a good tradeoff for now. Thanks!

PR #93, opened 2016-08-20
Add weight to the gcc parallel tasks

nipunn1313:
Tasks are not parallelized until their total "cost" exceeds 10k.
The weight of a vector is the length of the vector. The weight of
the task is increased by 5% for each additional task added (eg
for_each), but for compilation phases, because the step itself is
quite expensive, it is worth increasing the weight to force maximum
parallelism. This fixes the issue I mentioned in #90

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -367,7 +367,7 @@ impl Config {
         }
         drop(rayon::initialize(cfg));
 
-        objs.par_iter().for_each(|&(ref src, ref dst)| {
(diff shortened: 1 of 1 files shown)


nipunn1313:
For reference, these constants are tuned here
https://github.com/nikomatsakis/rayon/blob/master/src/par_iter/len.rs

alexcrichton:
Nice find!

Clearly I need to try out rayon more...

nipunn1313:
Thanks! Can you kick another version number with this one?

alexcrichton:
Sure! (done)

PR #96, opened 2016-08-26
Toolchain prefix mappings for NetBSD

jakllsch:
These prefixes are for toolchains created by NetBSD's `build.sh tools`.


alexcrichton:
Thanks!

PR #101, opened 2016-09-03
Add toolchain prefix mappings for s390x

uweigand:
This adds prefix mappings for s390x to allow cross-compiles without having to define environment variables.  Should match other Linux platforms.


alexcrichton:
Thanks!

PR #106, opened 2016-10-11
Support CROSS_COMPILE environment variable

jdub:
Used as last resort when HOST != TARGET.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -709,6 +709,7 @@ impl Config {
             } else if target.contains("android") {
                 format!("{}-{}", target, gnu)
             } else if self.get_host() != target {
+                let cross_compile = self.getenv("CROSS_COMPILE");
                 let prefix = match &target[..] {
                     "aarch64-unknown-linux-gnu" => Some("aarch64-linux-gnu"),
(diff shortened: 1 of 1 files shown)


jdub:
Intended as one approach to alexcrichton/gcc-rs#82.

This is kinda ugly, because I had to juggle an `Option<String>` into an `Option<&str>` and didn't want to muck around too much with the logic in the `host != target` if block.

No tests or docs because I'm spitballing an approach for now. 😳

alexcrichton:
Thanks! I wonder if perhaps this should be prioritized over the fallback prefixes though? It seems like it this variable is set it's higher priority

jdub:
Hmm, yeah. 😳

This is a bit ooky, but it looks better than `if cross_compile.is_some() { … } else { match … }`.

alexcrichton:
Thanks!

jdub:
Oh! That is a surprise. I hadn't written docs, tests, or extended it to the other tools yet. I'll follow up in a separate PR. 😀

alexcrichton:
Oh ok seems fine to me :)

PR #107, opened 2016-10-12
Another spin at incremental builds

jdub:
A fresh take on alexcrichton/gcc-rs#47. Hasn't got `msvc` support yet, but will look at it next.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -420,16 +420,14 @@ impl Config {
             objects.push(obj);
         }
 
-        // if objects are queued to build
-        if src_dst.len() > 0 {
+        // if objects are queued to build, or output doesn't exist
+        if src_dst.len() > 0 || !&dst.join(output).exists() {
             self.compile_objects(&src_dst);
             self.assemble(lib_name, &dst.join(output), &objects);
         }
 
-        for key in dependencies.keys() {
-            if let Some(file) = key.to_str() {
-                println!("cargo:rerun-if-changed={}", file);
-            }
+        for file in dependencies.keys() {
+            println!("cargo:rerun-if-changed={}", file.display());
         }
 
         self.print(&format!("cargo:rustc-link-lib=static={}",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,10 +46,14 @@
 #![cfg_attr(test, deny(warnings))]
 #![deny(missing_docs)]
 
+extern crate filetime;
+
 #[cfg(feature = "parallel")]
 extern crate rayon;
 
+use filetime::FileTime;
 use std::env;
+use std::collections::HashMap;
 use std::ffi::{OsString, OsStr};
 use std::fs;
 use std::io;
@@ -336,9 +340,12 @@ impl Config {
         assert!(output.ends_with(".a"));
         let lib_name = &output[3..output.len() - 2];
         let dst = self.get_out_dir();
+        let msvc = self.get_target().contains("msvc");
+        let compiler = self.get_compiler();
 
         let mut objects = Vec::new();
         let mut src_dst = Vec::new();
+        let mut dependencies = HashMap::new();
         for file in self.files.iter() {
             let obj = dst.join(file).with_extension("o");
             let obj = if !obj.starts_with(&dst) {
@@ -346,12 +353,84 @@ impl Config {
             } else {
                 obj
             };
-            fs::create_dir_all(&obj.parent().unwrap()).unwrap();
-            src_dst.push((file.to_path_buf(), obj.clone()));
+
+            let mut input_mtime = get_mtime(&file);
+            let output_mtime = get_mtime(&obj);
+
+            let mut cmd = compiler.to_command();
+            if msvc {
+                cmd.arg("/showIncludes").arg("/E");
+            } else {
+                cmd.arg("-M").arg("-MG");
+            }
+            cmd.arg(&file);
+            println!("running: {:?}", cmd);
+
+            let cpp_output = match cmd.output() {
+                Ok(ref out) if out.status.success() => {
+                    String::from_utf8(if msvc {
+                        out.stderr.clone()
+                    } else {
+                        out.stdout.clone()
+                    }).ok()
+                },
+                _ => {
+                    println!("warning: dependency check failed, building anyway: {:?}", file);
+                    input_mtime = output_mtime;
+                    None
+                }
+            };
+
+            if let Some(cpp) = cpp_output {
+                let files = if msvc {
+                    // Output is indented dependency per line, with "Note: ..." prefix 
+                    cpp
+                        .replace("Note: including file:", "")
+                } else {
+                    // Output is make rule with line continuations and multiple files per line
+                    cpp
+                        .replace("\\\n", "") // remove line continuations
+                        .replace("\\ ", "\0") // protect escaped spaces
+                        .replace(" ", "\n") // spaces to new lines
+                        .replace("\0", "\\ ") // bring escaped spaces back
+                };
+
+                input_mtime = files
+                    .lines().skip(1)
+                    .chain(vec![file.to_str().unwrap_or("")])
+                    .filter_map(|s| PathBuf::from(s.trim()).canonicalize().ok())
+                    .map(|f| dependencies
+                        .entry(f.clone())
+                        .or_insert(get_mtime(f))
+                        .clone())
+                    .max()
+                    .unwrap_or(FileTime::zero());
+            }
+
+            // build if a dependency is fresher than the output object
+            if input_mtime >= output_mtime {
+                println!("{:?} or one of its dependencies is {}s fresher than {:?}",
+                    file,
+                    input_mtime.seconds() - output_mtime.seconds(),
+                    obj);
+                src_dst.push((file.to_path_buf(), obj.clone()));
+                fs::create_dir_all(&obj.parent().unwrap()).unwrap();
+            }
+
             objects.push(obj);
         }
-        self.compile_objects(&src_dst);
-        self.assemble(lib_name, &dst.join(output), &objects);
+
+        // if objects are queued to build
+        if src_dst.len() > 0 {
+            self.compile_objects(&src_dst);
+            self.assemble(lib_name, &dst.join(output), &objects);
+        }
+
+        for key in dependencies.keys() {
+            if let Some(file) = key.to_str() {
+                println!("cargo:rerun-if-changed={}", file);
+            }
+        }
 
         self.print(&format!("cargo:rustc-link-lib=static={}",
                             &output[3..output.len() - 2]));
@@ -912,6 +991,12 @@ impl Tool {
     pub fn env(&self) -> &[(OsString, OsString)] {
         &self.env
     }
+}
+
+fn get_mtime<P: AsRef<Path>>(file: P) -> FileTime {
+    fs::metadata(file)
+        .map(|mtime| FileTime::from_last_modification_time(&mtime))
+        .unwrap_or(FileTime::zero())
 }
 
 fn run(cmd: &mut Command, program: &str) {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ code.
 keywords = ["build-dependencies"]
 
 [dependencies]
+filetime = "0.1"
 rayon = { version = "0.4", optional = true }
 
 [features]


jdub:
Here's a gist of Cargo output for `brotli-sys` when building with msvc. Hopefully it's clear that I'd already done a build, then touched `brotli\enc\backward_references.c`. 😄

https://gist.github.com/jdub/e61dd96dad50e2a227225eba2451369b

jdub:
Some timings, for giggles.

```
D:\GitHub\brotli2-rs\brotli-sys>cargo build
   Compiling libc v0.2.16
   Compiling filetime v0.1.10
   Compiling gcc v0.3.35 (file:///D:/GitHub/gcc-rs)
   Compiling brotli-sys v0.2.0 (file:///D:/GitHub/brotli2-rs/brotli-sys)
    Finished debug [unoptimized + debuginfo] target(s) in 7.8 secs

D:\GitHub\brotli2-rs\brotli-sys>cargo build
    Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs

D:\GitHub\brotli2-rs\brotli-sys>touch brotli\enc\backward_references.c

D:\GitHub\brotli2-rs\brotli-sys>cargo build
   Compiling brotli-sys v0.2.0 (file:///D:/GitHub/brotli2-rs/brotli-sys)
    Finished debug [unoptimized + debuginfo] target(s) in 1.25 secs

D:\GitHub\brotli2-rs\brotli-sys>touch build.rs

D:\GitHub\brotli2-rs\brotli-sys>cargo build
   Compiling brotli-sys v0.2.0 (file:///D:/GitHub/brotli2-rs/brotli-sys)
    Finished debug [unoptimized + debuginfo] target(s) in 1.84 secs
```

The log for the final run shows that no C object files were built.

src/lib.rs
```diff
@@ -907,6 +955,13 @@ impl Tool {
     }
 }
 
+fn get_mtime<P: AsRef<Path>>(file: P) -> u64 {
+    use filetime::FileTime;
+    fs::metadata(file)
+        .map(|mtime| FileTime::from_last_modification_time(&mtime).seconds())
```
alexcrichton: Sometimes the nanosecond precision here might matter, perhaps that could be preserved?

src/lib.rs
```diff
@@ -336,18 +339,63 @@ impl Config {
...
+                    cmd.arg("-M").arg("-MG");
+                    files = cmd.output()
+                        .map(|o| String::from_utf8(o.stdout))
+                        .unwrap().unwrap()
+                        .replace(" \\", "")
+                        .split_whitespace().skip(2) // skip input/output files
+                        .map(|s| PathBuf::from(s.trim()))
+                        .collect();
+                };
+                println!("running: {:?}", cmd);
```
alexcrichton: This is printed a little too late, right?

src/lib.rs
```diff
@@ -336,18 +339,63 @@ impl Config {
...
+                        .replace("Note: including file:", "")
+                        .lines().skip(1) // skip input files
+                        .map(|s| PathBuf::from(s.trim()))
+                        .collect();
+                } else {
+                    cmd.arg("-M").arg("-MG");
+                    files = cmd.output()
+                        .map(|o| String::from_utf8(o.stdout))
+                        .unwrap().unwrap()
+                        .replace(" \\", "")
```
alexcrichton: What's this dealing with?

jdub: The output of `-M -MG` is a multiline `make` rule with backslashes for line continuation. This strips them out.

Hmm, looking at this now, I realise I'm not sure what the output would be with spaces in parent directories. I'll have to handle that.

src/lib.rs
jdub: I'll move it up into the branches.

src/lib.rs
jdub: I thought that might be gilding the lily. But `FileTime` derives `Ord`, so it's a tiny change!

src/lib.rs
```diff
@@ -336,18 +340,66 @@ impl Config {
...
+            let output_mtime = get_mtime(&obj);
+            // If input file is older than output object, check its dependencies
+            if input_mtime < output_mtime {
+                let mut cmd = compiler.to_command();
+                let files: Vec<PathBuf>;
+                cmd.arg(&file);
+                if msvc {
+                    cmd.arg("/showIncludes").arg("/E");
+                    println!("running: {:?}", cmd);
+                    files = cmd.output()
```
alexcrichton: I think we'll want to check the result of this invocation as well. Additionally, we may want to be a bit more pessimistic about the `unwrap()` below. That is, if any error happens, seems like we should default to attempting to compile the file. For example if the command returns exit code 1, can't be found, or doesn't produce utf-8 output, we'd fall back to just always compiling.

src/lib.rs
```diff
@@ -336,18 +340,66 @@ impl Config {
...
+                        .lines().skip(1) // skip input file
+                        .map(|s| PathBuf::from(s.trim()))
+                        .collect();
+                } else {
+                    cmd.arg("-M").arg("-MG");
+                    println!("running: {:?}", cmd);
+                    files = cmd.output()
+                        .map(|o| String::from_utf8(o.stdout))
+                        .unwrap().unwrap()
+                        .replace("\\\n", "") // remove line continuations
```
alexcrichton: Could you add in a comment above this what the expected output is? (just a small gist of it)

src/lib.rs
```diff
@@ -336,18 +340,66 @@ impl Config {
...
+                        .replace("\\ ", "\0") // handle spaces in filenames
+                        .split_whitespace().skip(2) // skip input/output files
```
alexcrichton: With one file per line, shouldn't this use `.lines()`?

src/lib.rs
jdub: Yeah, the double `unwrap()` is offensive and embarrassing. I think I got way too caught up in the excitement of extreme chaining. I'll sort that out.

src/lib.rs
jdub: Unfortunately it's often multiple files per line, so it has to be split on non-escaped white space.

jdub:
Oops, I left `cargo:rerun-if-changed=brotli` in. That was for testing without the full `rerun-if-changed` list. This isn't ready to merge anyway. 😄

Here are some indicative numbers again:

```
D:\GitHub\brotli2-rs\brotli-sys>cargo clean

D:\GitHub\brotli2-rs\brotli-sys>cargo build
   Compiling libc v0.2.16
   Compiling filetime v0.1.10
   Compiling gcc v0.3.35 (file:///D:/GitHub/gcc-rs)
   Compiling brotli-sys v0.2.0 (file:///D:/GitHub/brotli2-rs/brotli-sys)
    Finished debug [unoptimized + debuginfo] target(s) in 6.98 secs

D:\GitHub\brotli2-rs\brotli-sys>cargo build
    Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs

D:\GitHub\brotli2-rs\brotli-sys>touch brotli

D:\GitHub\brotli2-rs\brotli-sys>cargo build
   Compiling brotli-sys v0.2.0 (file:///D:/GitHub/brotli2-rs/brotli-sys)
    Finished debug [unoptimized + debuginfo] target(s) in 1.19 secs

D:\GitHub\brotli2-rs\brotli-sys>touch brotli\dec\decode.c

D:\GitHub\brotli2-rs\brotli-sys>touch brotli

D:\GitHub\brotli2-rs\brotli-sys>cargo build
   Compiling brotli-sys v0.2.0 (file:///D:/GitHub/brotli2-rs/brotli-sys)
    Finished debug [unoptimized + debuginfo] target(s) in 1.27 secs
```

1.19s to check unchanged dependencies, and an extra 0.08s to build and assemble with one source file changed.

Here it is with the full `rerun-if-changed` list enabled, which appears to work:

```
D:\GitHub\brotli2-rs\brotli-sys>cargo clean

D:\GitHub\brotli2-rs\brotli-sys>cargo build
   Compiling libc v0.2.16
   Compiling filetime v0.1.10
   Compiling gcc v0.3.35 (file:///D:/GitHub/gcc-rs)
   Compiling brotli-sys v0.2.0 (file:///D:/GitHub/brotli2-rs/brotli-sys)
    Finished debug [unoptimized + debuginfo] target(s) in 6.87 secs

D:\GitHub\brotli2-rs\brotli-sys>cargo build
    Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs

D:\GitHub\brotli2-rs\brotli-sys>touch brotli

D:\GitHub\brotli2-rs\brotli-sys>cargo build
    Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs

D:\GitHub\brotli2-rs\brotli-sys>touch brotli\dec\decode.c

D:\GitHub\brotli2-rs\brotli-sys>cargo build
   Compiling brotli-sys v0.2.0 (file:///D:/GitHub/brotli2-rs/brotli-sys)
    Finished debug [unoptimized + debuginfo] target(s) in 1.29 secs
```

Interesting lifecycle problem with `rerun-if-changed` though: Right now, if a `gcc.file()` provided file changes, we don't check its dependencies, so the `rerun-if-changed` list may be incomplete. (Plus, I don't currently include those files in the `rerun-if-changed` list, which I should fix.)

jdub:
Pretty sure there's something wrong with the full `rerun-if-changed` list. Here's a sample with canonicalised file names on Windows:

```
cargo:rerun-if-changed="\\\\?\\C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\include\\vadefs.h"
cargo:rerun-if-changed="\\\\?\\D:\\GitHub\\brotli2-rs\\brotli-sys\\brotli\\enc\\ringbuffer.h"
cargo:rerun-if-changed="\\\\?\\D:\\GitHub\\brotli2-rs\\brotli-sys\\brotli\\dec\\prefix.h"
```

jdub:
Thoughts on where this should go next:

- Now I know cargo has `rerun-if-changed`, my initial goal (from months ago) to avoid dependency checks where possible is no longer as important.
- By producing an exhaustive `rerun-if-changed` list, we shift the work of "Should I build?" dependency checking to cargo itself. That gives us the bulk of the desired performance improvement.
- That frees us to always gather dependencies, necessary to inform the `rerun-if-changed` list, but we can still avoid doing builds.
- This will simplify the logic, but put even more pressure on dependency checking performance.

src/lib.rs
```diff
@@ -336,22 +340,84 @@ impl Config {
...
+                    .map(|f| dependencies
+                        .entry(f.clone())
+                        .or_insert(get_mtime(f))
+                        .clone())
+                    .max()
+                    .unwrap_or(FileTime::zero());
+            }
+
+            // build if a dependency is fresher than the output object
+            if input_mtime >= output_mtime {
```
alexcrichton: If the `/E` or `-M` execution failed (or our parsing failed), shouldn't `output_mtime` be zero here? That is, we want to always execute this in that case regardless of the state of the input/output.

Similarly below if we failed to learn about dependencies we'll want to avoid printing `cargo:rerun-if-changed` because we don't actually know the dependencies.

jdub: OK, I did some reading and feel less foolish about `std::process::Command` now. I'll sort out some fixes.

jdub:
I appear to have made a total mess of the commits here, but the latest changes fix the awful error handling in previous versions. What do you think?

src/lib.rs
```diff
@@ -336,22 +340,97 @@ impl Config {
...
+                    // Output is make rule with line continuations and multiple files per line
+                    cpp
+                        .replace("\\\n", "") // remove line continuations
+                        .replace("\\ ", "\0") // protect escaped spaces
+                        .replace(" ", "\n") // spaces to new lines
+                        .replace("\0", "\\ ") // bring escaped spaces back
+                };
+
+                input_mtime = files
+                    .lines().skip(1)
```
alexcrichton: This `skip(1)` seems necessary on Unix, right, as it's ignoring the name of the makefile target? Is this necessary on Windows, though? We're not accidentally skipping a header file?

src/lib.rs
```diff
@@ -336,22 +340,97 @@ impl Config {
...
+                    .chain(vec![file.to_str().unwrap_or("")])
```
alexcrichton: Perhaps the lines read off stdout/stderr could be converted to paths so we could avoid this `to_str` fallibility?

src/lib.rs
```diff
@@ -336,22 +340,97 @@ impl Config {
...
+                    .filter_map(|s| PathBuf::from(s.trim()).canonicalize().ok())
```
alexcrichton: How come the `canonicalize` was needed here?

alexcrichton: I'm also a little wary of ignoring all errors here. This means that if one header file has a problem then we'll ignore it even if it changed.

src/lib.rs
```diff
@@ -336,22 +340,97 @@ impl Config {
...
+
             objects.push(obj);
         }
-        self.compile_objects(&src_dst);
-        self.assemble(lib_name, &dst.join(output), &objects);
+
+        // if objects are queued to build
+        if src_dst.len() > 0 {
+            self.compile_objects(&src_dst);
+            self.assemble(lib_name, &dst.join(output), &objects);
```
alexcrichton: We may want to also check whether the output archive exists because otherwise it could be a case where a build was interrupted with all object files but the output wasn't created yet.

src/lib.rs
jdub: This is not a line read from stdout/stderr, but the input file chained into the iterator. It's not pretty. I could just do it separately from the iterator.

src/lib.rs
jdub: MSVC names the source file on the first line, while gcc names the target and doesn't include the source file at all. So, I skip the first line either way, and add the source file with the `.chain()` below.

src/lib.rs
jdub: Oh, now I remember the proper reason: I wanted to make sure that they were all the same in the `HashMap`.

jdub:
Sorry it's taken a while to get back to this, I've been taking the fight to bindgen.

- Rebased
- Compile/assemble if output doesn't exist
- I discovered `Path::display()`

jdub:
Ready for a detailed review / dissection. 😄

alexcrichton:
Thanks for the update! (sorry for being a bit slow to get around to this).

After thinking about this a bit more recently, I'm curious about another use case. Let's say you've got N object files you want to compile. Today we execute the compiler N times unconditionally to compile N object files. Presumably, though, with incremental support, we only execute a much smaller number of N compilers to generate far fewer than N object files (as most can be reused).

As written, though, we still execute at least N processes to learn about dependency information. This means that for a large project (presumably what this is targeted at) it may not have as great as a speedup as the number of processes spawned (which is relatively slow) is still on the order of the total number of objects.

Now solving this would be pretty tricky, though. We'd have to cache the output of the previous compilers and read that in to see if any files have been updated. I'm not even sure if that's possible on MSVC, but I do think it's possible on gcc at least.

Do you have thoughts about that? Do you think it's not worth caching the dependency lists and re-learning them every time?

jdub:
I like it! But yeah, tricky.

For now this PR dodges some build processes, so I'd be happy if we considered it as Step 1. But let's have a think about Step 2…

Let's assume we have a state file, probably living somewhere in `OUT_DIR`, that contains file names, their modification times, and for derived dependencies, their parent specified source file.

- Loop through the specified source files
  - If the output object doesn't exist, add source file to the dependency checking list (break)
  - If source file is newer than its object, add it to the dependency checking list (break)
  - If source file doesn't have cached state, add it to the dependency checking list (break)
  - If source file is newer than its cached state, add it to the dependency checking list (break)
  - Loop through the source file's dependencies
    - If dependency file is newer than its object, add source file to the dependency checking list (break)
    - If dependency file is newer than its cached state, add source file to the dependency checking list (break)

- If we have a dependency checking list, loop through it
  - Cache modified time of source file
  - Run compiler to fetch list of dependencies
  - Cache modified times of dependencies
  - Use the same rules as above, but push source files onto the build list (can break out rules into a helper)

- If we have a build list…
  - This is how the current code works, handing the list off to your lovely parallel build function

- Write state file from cache (we do this after the build because if something fails, we'd be better off leaving the state file as it is)

Make sense? Any suggestions for quick'n'dirty serialisation?

alexcrichton:
Yeah that sounds good to me! Ideally we'd combine the "run the compiler to fetch dependencies" and "run compile to generate object" steps, but that's fine to leave separate for now.

I don't think we'd need to store file times, but rather just use the file time of the dependency file itself.  In that sense we could just have a file, `dependencies`, which looks like:

* Each file name is emitted as bytes, followed by a 0 byte. 
* Each file is then followed by all of its dependencies, followed by a 0 byte.

For example if `foo` depended on `bar`, and `baz` didn't depend on anything, then we'd have a file that looks like:

```rust
b"foo\0bar\0\0baz\0\0"
```

Does that make sense?

jdub:
Yeah, nice and simple.

Hmm, combining the dependency generation and build... gcc's `-MD` will *"generate a dependency output file as a side-effect of the compilation process"* and we can use `-MF` to specify the file name or descriptor instead of guessing based on weird rules.

(Heh, now realising I could've used `-MF` to output to stderr, removing a silly difference between gcc and cl.exe.)

Doesn't look like cl.exe has an equivalent. I'll think upon it.

alexcrichton:
Ah it's ok to not combine those two steps (generating the dependency list and generating an object file) into one. If MSVC has to have them separate seems fine to keep them separate.

emoon:
Any more progress on this? This would be great to get in.

jdub:
I'd love to swing back through this some time, but if you'd like to pick it up sooner, go right ahead!

emoon:
@jdub Great to hear! I just have so much else going on so I won't be able to pick this up unfortunately so I really hope you will have the change of looking at it again :)

FaultyRAM:
I had a go at implementing the build cache, how would I go about getting the changes looked at?

jdub:
@FaultyRAM Nice! Fork my repo, add your changes to the `incremental` branch, and submit a new PR -- probably a good idea to mention this one. :-)

FaultyRAM:
Done!

alexcrichton:
I'm going to close this in favor of https://github.com/alexcrichton/cc-rs/pull/209

PR #110, opened 2016-10-20
Trim right - from CROSS_COMPILE

jdub:
Note the expected form of CROSS_COMPILE for future reference.

Fixes error introduced in pull request #106.


alexcrichton:
Thanks!

PR #114, opened 2016-11-13
Add default cpp_link_stdlib for FreeBSD target

paukstis:
FreeBSD defaults to Clang, libc++ would be a preferred lib.


alexcrichton:
Thanks!

PR #116, opened 2016-11-28
use armv6s-m instead of armv6-m

parched:
This allows use of the SVC instruction and matches what rustc gets from LLVM with thumbv6m.

From the GCC manual
> -march=armv6s-m is the ‘armv6-m’ architecture with support for the (now mandatory) SVC instruction.

I have also added a commit to specify the minimum FPU for the `thumbv7em` hard float target.


alexcrichton:
cc @japaric, look good to you?

japaric:
:+1: LGTM

PR #117, opened 2016-12-01
Add ATL/MFC detection

### LINKED ISSUE #115: Add ATL/MFC include directories to MSVC compiler invocation if available
The ATL/MFC include directory (`<VC directory>\atlmfc\include`) is currently not passed to `cl.exe`. The library directory (`<VC directory>\atlmfc\lib`, `<VC directory>\atlmfc\lib\arm` or `<VC directory>\atlmfc\lib\amd64`) are also not passed to the linker invocation. Mind you, the ATL and MFC libraries are not included with every installation of Visual Studio. 

If the idea is for the user to add these directories manually, it would be very useful if we could reuse the MSVC installation detection code in this crate. Maybe by adding a `find_vc_dir(target: &str)` function to `windows_registry`?

Jascha-N:
Fixes #115. Unfortunately, the linker workaround for `rustc` is quite messy.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -353,6 +353,24 @@ impl Config {
         self.compile_objects(&src_dst);
         self.assemble(lib_name, &dst.join(output), &objects);
 
+        if self.get_target().contains("msvc") {
+            let compiler = self.get_base_compiler();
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks good to me, thanks!

PR #121, opened 2016-12-13
Add support for clang

nagisa:
This does a little bit of refactoring in order to introduce abstraction-notion of different
families of compilers that may be used, so the gcc-rs didn’t rely so much on matching pathes and
targets.

I also added a ton of FIXME noting potential issues.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -102,8 +102,8 @@ pub struct Tool {
 /// Each family of tools differs in how and what arguments they accept.
 ///
 /// Detection of a family is done on best-effort basis and may not accurately reflect the tool.
-#[derive(Copy, Clone)]
-pub enum ToolFamily {
+#[derive(Copy, Clone, Debug)]
+enum ToolFamily {
     /// Tool is GNU Compiler Collection-like.
     Gnu,
     /// Tool is Clang-like. It differs from the GCC in a sense that it accepts superset of flags
@@ -492,10 +492,8 @@ impl Config {
         let target = self.get_target();
 
         let mut cmd = self.get_base_compiler();
-        // FIXME: this may detect nvcc if compiler happens to be in a directory containing nvcc but
-        // otherwise unrelated to nvcc, whatever it is.
-        let nvcc = cmd.path.to_str()
-            .map(|path| path.contains("nvcc"))
+        let nvcc = cmd.path.file_name()
+            .and_then(|p| p.to_str()).map(|p| p.contains("nvcc"))
             .unwrap_or(false);
 
         // Non-target flags
@@ -512,13 +510,9 @@ impl Config {
                 }
                 match &opt_level[..] {
                     "z" | "s" => cmd.args.push("/Os".into()),
-                    "2" => cmd.args.push("/O2".into()),
-                    // FIXME: -O1 does not exactly match the meaning of /O1 in MSVC. Rather than
-                    // being a flag for “generate sufficiently fast code but also compile fast”,
-                    // /O1 is instead “generate the smallest code in majority of cases”.
                     "1" => cmd.args.push("/O1".into()),
-                    // FIXME: this behaves badly if opt_level is set to 3 (which is an option on
-                    // Gnu and Clang-family tools) or more. Consider using /Ox or /O2 instead.
+                    // -O3 is a valid value for gcc and clang compilers, but not msvc. Cap to /O2.
+                    "2" | "3" => cmd.args.push("/O2".into()),
                     _ => {}
                 }
             }
@@ -640,7 +634,8 @@ impl Config {
                     cmd.args.push(format!("-stdlib=lib{}", stdlib).into());
                 }
                 _ => {
-                    // FIXME: perhaps a warning?
+                    println!("cargo:warning=cpp_set_stdlib is specified, but the {:?} compiler \
+                              does not support this option, ignored", cmd.family);
                 }
             }
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,43 @@ pub struct Tool {
     path: PathBuf,
     args: Vec<OsString>,
     env: Vec<(OsString, OsString)>,
+    family: ToolFamily
+}
+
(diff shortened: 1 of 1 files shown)


nagisa:
This is kinda a RFC, rather than PR too.

src/lib.rs
```diff
@@ -451,136 +488,165 @@ impl Config {
...
-            self.ios_flags(&mut cmd);
-        } else if !msvc {
-            if target.contains("i686") || target.contains("i586") {
-                cmd.args.push("-m32".into());
-            } else if target.contains("x86_64") || target.contains("powerpc64") {
-                cmd.args.push("-m64".into());
+        // Target flags
+        match cmd.family {
+            ToolFamily::Clang => {
+                cmd.args.push(format!("--target={}", target).into());
```
alexcrichton: I think we differ from Clang in the specifics for some targets, right? Something like NEON on armv7 or along those lines? (I forget the details here)

I like the idea though of just communicating via the target...

src/lib.rs
```diff
@@ -451,136 +488,165 @@ impl Config {
...
     pub fn get_compiler(&self) -> Tool {
         let opt_level = self.get_opt_level();
-        let debug = self.get_debug();
         let target = self.get_target();
-        let msvc = target.contains("msvc");
-        self.print(&format!("debug={} opt-level={}", debug, opt_level));
 
         let mut cmd = self.get_base_compiler();
+        // FIXME: this may detect nvcc if compiler happens to be in a directory containing nvcc but
+        // otherwise unrelated to nvcc, whatever it is.
```
alexcrichton: Yeah I'm not really sure what we can do here, I sorta forget what nvcc was even for as well (it was submitted awhile ago). I'd be fine to changing this to just check the basename, not the whole path.

src/lib.rs
```diff
@@ -451,136 +488,165 @@ impl Config {
...
+                }
+                match &opt_level[..] {
+                    "z" | "s" => cmd.args.push("/Os".into()),
+                    "2" => cmd.args.push("/O2".into()),
+                    // FIXME: -O1 does not exactly match the meaning of /O1 in MSVC. Rather than
+                    // being a flag for “generate sufficiently fast code but also compile fast”,
+                    // /O1 is instead “generate the smallest code in majority of cases”.
+                    "1" => cmd.args.push("/O1".into()),
+                    // FIXME: this behaves badly if opt_level is set to 3 (which is an option on
+                    // Gnu and Clang-family tools) or more. Consider using /Ox or /O2 instead.
```
alexcrichton: Yeah I'd be fine defaulting to something here like one of these flags!

src/lib.rs
```diff
@@ -451,136 +488,165 @@ impl Config {
...
+                    cmd.args.push("/MT".into());
+                } else {
+                    cmd.args.push("/MD".into());
+                }
+                match &opt_level[..] {
+                    "z" | "s" => cmd.args.push("/Os".into()),
+                    "2" => cmd.args.push("/O2".into()),
+                    // FIXME: -O1 does not exactly match the meaning of /O1 in MSVC. Rather than
+                    // being a flag for “generate sufficiently fast code but also compile fast”,
+                    // /O1 is instead “generate the smallest code in majority of cases”.
```
alexcrichton: Eh I figured we'd just call `/O1`, if this does something else than what's currently written though that's fine.

src/lib.rs
```diff
@@ -94,6 +94,43 @@ pub struct Tool {
...
+    family: ToolFamily
+}
+
+/// Represents the family of tools this tool belongs to.
+///
+/// Each family of tools differs in how and what arguments they accept.
+///
+/// Detection of a family is done on best-effort basis and may not accurately reflect the tool.
+#[derive(Copy, Clone)]
+pub enum ToolFamily {
```
alexcrichton: This can stay private, right?

src/lib.rs
```diff
@@ -451,136 +488,165 @@ impl Config {
...
+        }
+
+        if self.cpp {
+            match (self.cpp_set_stdlib.as_ref(), cmd.family) {
+                (Some(stdlib), ToolFamily::Gnu) |
+                (Some(stdlib), ToolFamily::Clang) => {
+                    cmd.args.push(format!("-stdlib=lib{}", stdlib).into());
+                }
+                _ => {
+                    // FIXME: perhaps a warning?
```
alexcrichton: Yeah printing a warning seems fine for now.

src/lib.rs
nagisa: I probably worded this FIXME sorta ambiguously, but primarily I’m concerned about pathes like these getting detected as nvcc compiler: `/home/nvccbbaa/local/bin/your-favourite-cc`. Instead only the filename should be checked.

src/lib.rs
nagisa: Yes, of course.

nagisa:
> I may have missed it though, but the intent here was to detect CC=clang and not use it for cross compiles, right? I didn't quite see that logic though...

The intent was to detect CC=clang *and* allow it to generate code for a target properly.

src/lib.rs
nagisa: I kinda feel that we could just deal with those cases as we go, although there’s also an option of querying rustc somehow for the target features for a specific target and generating relevant flags for clang.

src/lib.rs
alexcrichton: Yep, that'll work!

alexcrichton:
Ok that makes sense, I'd be fine landing with comments!

src/lib.rs
alexcrichton: Yeah I'd be ok with that (case-by-case basis).

nagisa:
So I’ve updated the PR, but cannot make tests run with `CC="clang"` set for some reason. I’m confused why `gcc` is able to find the `foo.c` file and `clang` is not.

nagisa:
The travis failures seem to be unrelated?

alexcrichton:
Looks good, thanks!

PR #127, opened 2017-01-09
Use arm toolchain for armv7 on Android

michaelfairley:
Android only has a single toolchain for compiling both `arm` and `armv7`, and the binaries are all prefixed with only `arm-`.


alexcrichton:
Thanks!

PR #128, opened 2017-01-24
Add a method to macro-expand source files

sfackler:
This is a somewhat bizarro thing to do; see https://github.com/sfackler/rust-openssl/issues/564 for rationale.


alexcrichton:
Thanks!

PR #130, opened 2017-01-30
Update README; `cargo build -jN` will set/override $NUM_JOBS for you

tatsuya6502:
This is a doc-only update. It took me some time to figure out why `NUM_JOBS=4 cargo build` does not limit the number of parallelism to 4, but to the number of cpu cores. (The correct way is `cargo build -j4`.)

Thanks!


alexcrichton:
Looks good!

Issue #131, opened 2017-02-03
Incorrect warning message


nitingupta910:
When compiling a C++ source to produce a library, I get spurious warning:
```
cpp_set_stdlib is specified, but the Gnu compiler does not support this option, ignored
```

Here's my build.rs:
```
extern crate gcc;

fn main() {
    gcc::Config::new()
        .file("src/foo.cc")
        .cpp(true)
        .compile("libfoo.a");
}
```

and here's src/foo.cc:
```
int addone(int n) {
    return n + 1;
}
```

and here's Cargo.toml
```
[package]
name = "testgcc"
version = "0.1.0"
authors = [""]

build = "build.rs"

[dependencies]

[build-dependencies]
gcc = "0.3"
```

This warning does not show up when compiling C library source.

uname -a
```
Darwin myiMac.local 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64
```

c++ --version
```
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin16.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

alexcrichton:
Closed in https://github.com/alexcrichton/gcc-rs/pull/132

PR #132, opened 2017-02-03
Fix suprious warning for cpp

nitingupta910:
Fix spurious warning:
warning: cpp_set_stdlib is specified, but the Gnu compiler does not support this option, ignored

Issue #131


alexcrichton:
Thanks!

PR #135, opened 2017-02-13
Default to gnu compilers on Solaris

### LINKED ISSUE #125: support gcc executable being named something other than "cc"
If I'm reading the code correctly, the gcc crate assumes (on a unix-like environment, at least) that it can find gcc with the executable name `cc`.  This is not the case on Solaris, at least, and while I've been able to get around this by setting `$CC` (and by patching the source to force it to look for `gcc`, in some cases, or creating a special directory to add to `$PATH` and putting a symlink in there), it seems like it should just work out of the box. I really don't have a good idea what that should look like, though. Two thoughts that come to mind are

  - simply make it use `gcc` on all platforms, since that seems to be the default executable name, at least on Linux
  - new feature or other form of conditional compilation that sets the compiler executable path to something whoever's building the gcc crate can specify

binarycrusader:
Fixes #125

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -825,11 +825,21 @@ impl Config {
         }
         let host = self.get_host();
         let target = self.get_target();
-        let (env, msvc, gnu, default) = if self.cpp {
-            ("CXX", "cl.exe", "g++", "c++")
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #146, opened 2017-03-17
Fix some style issues given by clippy

neosilky:


src/lib.rs
```diff
@@ -384,7 +384,7 @@ impl Config {
 
         let mut objects = Vec::new();
         let mut src_dst = Vec::new();
-        for file in self.files.iter() {
```
alexcrichton: I personally prefer `for x in foo.iter()` over `for x in &foo`, can this remain as it was?

src/lib.rs
```diff
@@ -963,9 +962,7 @@ impl Config {
             let target = self.get_target();
             if target.contains("msvc") {
                 None
-            } else if target.contains("darwin") {
-                Some("c++".to_string())
```
alexcrichton: Can these remain separate?

neosilky:
@alexcrichton Sorry for the late reply. I have removed those commits and force pushed. Should be good to go! :+1:

alexcrichton:
Looks like there's some errors on CI?

neosilky:
@alexcrichton Sorry about the super late reply... I forgot I had nightly as the default. Should be good to go now :+1:

alexcrichton:
Thanks!

PR #148, opened 2017-03-29
Use softfp float ABI on 32bit ARM Android GCC.

pietro:
The NDK docs[0] say:

> To target armeabi-v7a with GCC, you must set the following flags when invoking the compiler: -march=armv7-a -mfloat-abi=softfp -mfpu=vfpv3-d16.

I added `-mfloat-abi=softfp` because `-march=armv7-a` and `-mfpu=vfpv3-d16` are already set.

[0] https://developer.android.com/ndk/guides/standalone_toolchain.html#abi_compatibility


alexcrichton:
Thanks!

PR #149, opened 2017-04-01
Use armv6t2 instead of armv6

TimNN:
This is required to correctly build `compiler-rt` of [rust-lang/rust](https://github.com/rust-lang/rust) with LLVM 4.0.

I don't know what exactly the consequences of this change are, however in general `armv6t2` seems to be widely supported. However note that the raspberry pi 1, for example, apparently does not support Thumb 2. See also the [Wikipedia List of ARM Microarchitectures](https://en.wikipedia.org/wiki/List_of_ARM_microarchitectures).

cc @nagisa: You have done some research into this as well, I believe.

cc rust-lang/rust#40123

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -607,7 +607,7 @@ impl Config {
 
                 // For us arm == armv6 by default
                 if target.starts_with("arm-unknown-linux-") {
-                    cmd.args.push("-march=armv6".into());
+                    cmd.args.push("-march=armv6t2".into());
                     cmd.args.push("-marm".into());
                 }
 


nagisa:
Is there a way to disable both the thumbs? If not, then this seems like an obvious sol'n to me.

TimNN:
> Is there a way to disable both the thumbs?

I have no idea. Gcc accepts `armv6 armv6-m armv6j armv6k armv6kz armv6s-m armv6t2 armv6z armv6zk` for `-march` however I think they all imply at least Thumb1, with some implying only a subset of Thumb2.

alexcrichton:
[This comment](https://github.com/rust-lang/rust/issues/41004#issuecomment-290957055) seems quite relevent perhaps?

Currently `arm-unknown-linux-gnu` is sort of our "maxially compatible ARM" target, so if we're slicing out platforms (esp rpi) we may want to try to work around this another way unfortunately :(

nagisa:
So I tried to use a number of options to see if any of them would remove the define for thumb, sadly none of them except redefining the environment variable manually helped:

```
  -marm                       Generate code in 32 bit ARM state. (does not disable the THUMB define, though)
  -mno-thumb                  (CLANG ONLY: does not disable the thumb define either)
  -march=*                    Some thumb gets defined for all of them
  -D__ARM_ARCH_ISA_THUMB=0    (WORKS) Redefine manually to something else.
```

This is how I was testing it:

```
arm-none-eabi-gcc -dM -E -xc {PARAMS} /dev/null | grep THUMB
```

So I suppose simply adding this define for our own compiler-rt compilation is the way to go, leaving gcc-rs alone.

nagisa:
A small aside, I read a recommendation in multiple locations that if thumb2 is not available, `-marm` should be used, because `thumb1` by itself is terrible and everything.

TimNN:
Using a different fix for now (`-D__ARM_ARCH_ISA_THUMB=0 `).

MeinFuehrer:
@timnn
Stop running with scissors man! The only practical user of this target is the Rpi(A,B,zero). Learn the basics before proposing any more "fixes" like this.

MeinFuehrer:
https://archive.is/o2yuQ

PR #151, opened 2017-04-11
add '-march=armv7-a' to all armv7 linux targets

zachreizner:
This is related to my other changes that add support for chromeos specific target triples.


alexcrichton:
Thanks!

PR #152, opened 2017-04-16
lib.rs: use Display rather than Debug for ExitStatus

nodakai:
println!("{0}\n{0:?}", status) =>
exit code: 0
ExitStatus(ExitStatus(0))
 
I think the former makes more sense.


alexcrichton:
Thanks!

PR #154, opened 2017-04-20
Fix confusion with uclibc triplets

cactorium:
Hello! So I'm working on getting Rust to run on `mips-unknown-linux-uclibc`, and it looks like `gcc` gets confused before `mips-openwrt-uclibc-gcc` has `cl` in the name. I hope this is an okay fix


alexcrichton:
Oops, thanks!

PR #157, opened 2017-05-06
Fixing emscripten on windows to use the right files.

vvanders:
This fixes up a small issue where we don't resolve .bat files from PATH since windows is a bit picky.


alexcrichton:
Thanks!

PR #158, opened 2017-05-10
RFC: Handle custom targets

ppannuto:
For [Tock](https://www.tockos.org/), we use custom [target specifications](https://github.com/helena-project/tock/blob/master/boards/hail/sam4l.json) for each supported board. With the merging of https://github.com/rust-lang/rust/pull/35021, when I set up https://github.com/rust-lang-nursery/compiler-builtins for Tock, it invoked `cc` instead of `arm-none-eabi-gcc`. I traced that decision back to this crate, where compiler selection logic depends on matching against a list of known targets.

Unfortunately, our custom target causes the `TARGET` environment variable to be set to something like `sam4l.json`, which is not a known target. However, we can look inside the specs file at the `llvm-target` key to pull out the right target.

The PR is a quick proof-of-concept that always tries to treat the `TARGET` as a specs file, but falls back to treating it as a well-known target string if the file doesn't exist. I think this is a reasonable approach to handle custom targets, but am open to any other ideas you may have.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ appveyor = { repository = "alexcrichton/gcc-rs" }
 
 [dependencies]
 rayon = { version = "0.7", optional = true }
+serde_json = "1.0"
 
 [features]
 parallel = ["rayon"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,6 +48,7 @@
 
 #[cfg(feature = "parallel")]
 extern crate rayon;
+extern crate serde_json;
 
 use std::env;
 use std::ffi::{OsString, OsStr};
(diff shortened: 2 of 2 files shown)


alexcrichton:
Thanks for the PR! I know this is definitely one of the pain points about using custom targets right now.

Unfortunately though I fear that custom target specs are not fleshed out enough for this to use them just yet. For example, as you've found, just *finding* the target spec is no easy task. It's intended to be relative the cwd of the compiler invocation, which is different than the cwd of the build script invocations. As a result a build script doesn't actually have a way of finding where the target spec even is! While `PWD` here may work sometimes, I don't think it's a cross platform or tool-friendly solution (I think shells set that env var?)

In that sense I love the idea of making `gcc` more amenable to custom targets but I fear that we'd need to think through the custom target story more thoroughly before being able to add support :(

ppannuto:
With japaric's help, we've managed to get things working in Tock without relying on a custom target any more. While I suspect something like this may be useful someday, I don't have any great solutions to the issues that you've raised and am inclined to just close this for now.

PR #159, opened 2017-05-11
Explicitly set static crt

liranringel:
This PR allows to determine whether `/MD` or `/MT` will be passed to msvc targets (and not implicitly through the `CARGO_CFG_TARGET_FEATURE` environment variable).
Will help to resolve https://github.com/rust-lang/rust/issues/40246


alexcrichton:
Thanks!

PR #160, opened 2017-05-24
Add MSVC 2017 detection

brson:
This teaches gcc-rs to find tools installed by MSVC 2017. It's not
obvious to what extent the new COM interfaces are expected to be
used. This patch only uses it to find the product installation
directory, then infers everything else. A lot of COM scaffolding to do
very little. The toolchains seem to have a different directory
structure in this release to better support cross-compilation.

It looks to me like all the lib/include logic is pretty much the same
as it always has been.

This is tested to fix the rustup installation logic, and to fix
rustc in basic use cases on x86_64.

cc alexcrichton/gcc-rs#143
cc rust-lang-nursery/rustup.rs#1003
cc rust-lang/rust#38584
cc alexcrichton/curl-rust#161

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -15,6 +15,7 @@ use std::process::Command;
 
 use Tool;
 
+#[cfg(windows)]
 macro_rules! otry {
     ($expr:expr) => (match $expr {
         Some(val) => val,
@@ -53,7 +54,11 @@ pub fn find_tool(target: &str, tool: &str) -> Option<Tool> {
     use std::ffi::OsString;
     use std::mem;
     use std::path::{Path, PathBuf};
+    use std::fs::File;
+    use std::io::Read;
     use registry::{RegistryKey, LOCAL_MACHINE};
+    use com;
+    use setup_config::{SetupConfiguration, SetupInstance};
 
     struct MsvcTool {
         tool: PathBuf,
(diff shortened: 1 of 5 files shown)


src/windows_registry.rs
```diff
@@ -110,19 +114,96 @@ pub fn find_tool(target: &str, tool: &str) -> Option<Tool> {
...
+        otry!(add_sdks(&mut tool, target));
+        Some(tool.into_tool())
+    }
+
+    fn add_sdks(tool: &mut MsvcTool, target: &str) -> Option<()> {
         let sub = otry!(lib_subdir(target));
         let (ucrt, ucrt_version) = otry!(get_ucrt_dir());
 
+        tool.path.push(ucrt.join("bin").join(&ucrt_version).join(sub));
+
```
brson: This line may fix https://github.com/alexcrichton/curl-rust/issues/161. It found 'rc.exe' on my machine at least.

brson:
cc @alexcrichton @retep998 @Diggsey

brson:
The COM code is copied from https://github.com/retep998/msvc-bunny

src/windows_registry.rs
```diff
@@ -110,19 +115,96 @@ pub fn find_tool(target: &str, tool: &str) -> Option<Tool> {
...
+            Some((atl_path.join("lib").join(sub), atl_path.join("include")))
+        } else {
+            None
+        }
+    }
+
+    // For MSVC 14 and 15 we need to find the Universal CRT as well as either
     // the Windows 10 SDK or Windows 8.1 SDK.
-    fn find_msvc_latest(tool: &str, target: &str, ver: &str) -> Option<Tool> {
+    fn find_msvc_14_or_15(tool: &str, target: &str, ver: &str) -> Option<Tool> {
```
retep998: This code only works for 14. The only reason this was used for 15 was due to the early preview versions of 15 before they switched over to the new installer, but those preview versions have all expired so there's nothing left to support  there.

src/windows_registry.rs
```diff
@@ -110,19 +115,96 @@ pub fn find_tool(target: &str, tool: &str) -> Option<Tool> {
...
+    return find_msvc_17(tool, target)
+        .or_else(|| find_msvc_14_or_15(tool, target, "15.0"))
+        .or_else(|| find_msvc_14_or_15(tool, target, "14.0"))
         .or_else(|| find_msvc_12(tool, target))
         .or_else(|| find_msvc_11(tool, target));
 
-    // For MSVC 14 or newer we need to find the Universal CRT as well as either
+    // In MSVC 17 MS once again changed the scheme for locating the tooling.
+    // Now we must go through some COM interfaces, which is super fun for Rust.
+    fn find_msvc_17(tool: &str, target: &str) -> Option<Tool> {
```
retep998: Note that while it is called VS 2017, the version number is actually 15, not 17.

alexcrichton:
I barely understand all the COM business or what's going on here really, it'd be awesome if there was a document we could link to on MSDN or something like that saying "here's the structure and here's how to run tools", but alas!

In the meantime if it works it works, thanks @brson!

alexcrichton:
I've published this as 0.3.47

brson:
Thanks for the review @retep998. I'll circle back later to clean up.

luser:
> I barely understand all the COM business or what's going on here really, it'd be awesome if there was a document we could link to on MSDN or something like that saying "here's the structure and here's how to run tools", but alas!

See https://blogs.msdn.microsoft.com/vcblog/2017/03/06/finding-the-visual-c-compiler-tools-in-visual-studio-2017/. Specifically the "Identifying the VC++ compiler tools version" paragraph.

alexcrichton:
That's actually perfect, thanks @luser!

PR #163, opened 2017-05-25
Update msbuild discovery for 2017

### LINKED ISSUE #162: MSBuild detection for VS 2017 is incorrect
With VS 2017 the path to MSBuild is no longer in the same location. See https://stackoverflow.com/questions/328017/path-to-msbuild

Since cmake-rs and gcc-rs both need this info, seems like cmake should defer to gcc.

brson:
This also adds a find_vs_version function to share with
the cmake crate.

Fixes #162

Lightly tested.


alexcrichton:
👍

PR #166, opened 2017-06-01
Only add the 'host dylib path' to the MSVC 2017 PATH

brson:
This passed bors and should be the last update to support MSVC 2017.

It is just removing one path from PATH. I think it will not affect any behavior in practice, but this is more correct, so I'd like to get the correct version into all the tools.

Can you publish another build? Sorry again for all the churn.


alexcrichton:
Sure thing!

PR #167, opened 2017-06-08
config: add shared_flag and static_flag

### LINKED ISSUE #165: musl: hardcoded `-static` causes issues with library objects
Capturing an issue first reported at https://github.com/briansmith/ring/issues/544

It looks like https://github.com/alexcrichton/gcc-rs/commit/f308ef2121a812c07718e15e689cbb1c90778513 introduced an hardcoded `-static` flag for musl-gcc invocation.
This seems to be introducing linking issues when building objects for native libraries, due to non-relocatable symbols.

lucab:
This adds two configuration options to optionally enable or disable
the `-shared` and `-static` flags. If omitted, musl targets enable
`-static` by default.

Fixes #165

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -92,6 +92,8 @@ pub struct Config {
     cargo_metadata: bool,
     pic: Option<bool>,
     static_crt: Option<bool>,
+    shared_flag: Option<bool>,
+    static_flag: Option<bool>,
 }
 
 /// Configuration used to represent an invocation of a C compiler.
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

lucab:
Thank you. I see this crate has a pretty steady release cycle, so I'll busy-wait for .51 to wire this in ring.

alexcrichton:
Ah sorry just saw this message! I'll do a release to make sure this gets out there.

lucab:
In the meanwhile we realized the issue this was meant to fix was actually a PEBCAK due to https://github.com/rust-lang/cargo/issues/4133, so no hurry for this to be released.

PR #200, opened 2017-07-13
Implement Clone, Debug and Default for Config

### LINKED ISSUE #175: Implement `Clone`, `Debug` and `Default` for `Config`

opilar:
Resolves #175

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -72,6 +72,7 @@ mod setup_config;
 pub mod windows_registry;
 
 /// Extra configuration to pass to gcc.
+#[derive(Clone, Debug, Default)]
 pub struct Config {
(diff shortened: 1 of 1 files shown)


alexcrichton:
👍

opilar:
@alexcrichton can I ask a question? 
`Config::default()` is not the same as `Config::new()`. Maybe it more appropriate to make `default` method to call `new` as [here](https://doc.rust-lang.org/nightly/src/std/ffi/os_str.rs.html#308). Some boolean fields of `Config` struct is `true`, but default implementation for `bool` returns `false`.

alexcrichton:
Oh oops good catch! Yeah let's have a manual implementation of `Default` that delegates to `new`

PR #207, opened 2017-07-15
cpp_link_stdlib/cpp_set_stdlib examples

### LINKED ISSUE #179: Add examples for `cpp_link_stdlib` and `cpp_set_stdlib`

opilar:
Resolves #179

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -329,6 +329,16 @@ impl Config {
     /// otherwise cargo will link against the specified library.
     ///
     /// The given library name must not contain the `lib` prefix.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// gcc::Config::new()
+    ///             .file("src/foo.c")
+    ///             .shared_flag(true)
+    ///             .cpp_set_stdlib(Some("stdc++"))
+    ///             .compile("libfoo.so");
+    /// ```
     pub fn cpp_link_stdlib(&mut self, cpp_link_stdlib: Option<&str>) -> &mut Config {
         self.cpp_link_stdlib = Some(cpp_link_stdlib.map(|s| s.into()));
         self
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -329,6 +329,16 @@ impl Config {
...
     /// The given library name must not contain the `lib` prefix.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// gcc::Config::new()
+    ///             .file("src/foo.c")
+    ///             .shared_flag(true)
+    ///             .cpp_set_stdlib(Some("stdc++"))
+    ///             .compile("libfoo.so");
```
alexcrichton: Hm I'm realizing now these have `shared_flag`, but how come? I think this as-is will fail at runtime because this doesn't support creating `libfoo.so` afaik.

opilar: Why? Static library doesn't link. Or you talk about different header files?

alexcrichton: Oh well so this option is intended for the C++ standard a piece of code is compiled with, but `cpp_set_stdlib` and such aren't really relevant for linkage other than telling rustc it may want a different C++ standard library.

Perhaps the `shared_flag` could be removed and `libfoo.so` could be changed to `libfoo.a`?

opilar: Oh year. You're right! It's not only about linking.

src/lib.rs
```diff
@@ -329,6 +329,16 @@ impl Config {
...
     ///
     /// The given library name must not contain the `lib` prefix.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// gcc::Config::new()
+    ///             .file("src/foo.c")
+    ///             .shared_flag(true)
+    ///             .cpp_set_stdlib(Some("stdc++"))
```
dtolnay: I would expect the example for `cpp_link_stdlib` to call `cpp_link_stdlib`, not `cpp_set_stdlib`.

opilar: Thank you! My bad.

opilar:
@alexcrichton @dtolnay thank you for your help.

alexcrichton:
👍

PR #209, opened 2017-07-17
Yet another incremental compilation PR

FaultyRAM:
Continuation of https://github.com/alexcrichton/gcc-rs/pull/47 & https://github.com/alexcrichton/gcc-rs/pull/107

This augments the existing incremental compilation code with a `BuildCache` type that contains compiler and dependency information.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ travis-ci = { repository = "alexcrichton/gcc-rs" }
 appveyor = { repository = "alexcrichton/gcc-rs" }
 
 [dependencies]
+filetime = "0.1"
 rayon = { version = "0.8", optional = true }
 
 [features]
--- a/gcc-test/build.rs
+++ b/gcc-test/build.rs
@@ -1,21 +1,33 @@
 extern crate gcc;
 
+use std::time::Duration;
 use std::env;
 use std::fs;
 use std::path::PathBuf;
+use std::thread;
 
 fn main() {
     let out = PathBuf::from(env::var_os("OUT_DIR").unwrap());
     fs::remove_dir_all(&out).unwrap();
     fs::create_dir(&out).unwrap();
 
-    gcc::Build::new()
-        .file("src/foo.c")
-        .flag_if_supported("-Wall")
-        .flag_if_supported("-Wfoo-bar-this-flag-does-not-exist")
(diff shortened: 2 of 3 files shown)


FaultyRAM:
Couple of implementation notes:

* Compiler info is needed since compiler flags can affect the way code is compiled or, in the case of preprocessor flags, what dependencies get pulled in. We could probably track the version as well in case the compiler changes to a different version without anything else changing. I'm not sure if we need linker info, but getting that is kind of tricky at the moment because it's split across multiple methods depending on the build target.
* Currently this relies on serde_json for reading and writing the build cache (and flate2, since the cache files for OpenChaos took up several megabytes uncompressed). On one hand I'm not a fan of writing parser code, but on the other hand serde bumps the minimum required Rust to 1.15.0. Not sure how much breakage that would cause?

FaultyRAM:
Travis failed because flate2 depends on an older version of gcc-rs, which confused rustdoc because it was picking up two instances of gcc-rs and couldn't decide which one to use. I switched to libflate 0.1.8, which doesn't have this problem, but is slower and produces larger files (I didn't see a way to specify compression level for zlib).

alexcrichton:
Thanks for the PR! This is picking up some pretty weighty dependencies, though. Would it be possible to slim this down to not pick up new dependencies?

FaultyRAM:
Done!

alexcrichton:
Thanks! Since this is a pretty weighty feature as well, would it be possible to add tests?

FaultyRAM:
Looks like the environment of the test profile is confusing things somehow? Need to investigate this one.

alexcrichton:
@FaultyRAM any luck investigating the test here?

FaultyRAM:
Well, I moved the tests from `tests/test.rs` into gcc-test and that seems to work, so I guess I just had them in the wrong place to begin with. 🙈

BusyJay:
Is it better to use ccache to enable incremental compilation? I feel more safe to keep gcc-rs as a simple wrapper of compiler itself.

By the way, does the environment variables like "LIBRARY_PATH" need to be taken into consideration?

alexcrichton:
I'm going to close this since it's quite old at this point, but it can always be resubmitted if rebased!

PR #211, opened 2017-07-18
Add some panic conditions

rap2hpoutre:
Just a try in order to help on https://github.com/alexcrichton/gcc-rs/issues/181. I don't know gcc and I'm not english, so the quality of this PR may be a bit poor!

I found another `unwrap` in a public function (`expand`) so I added a doc for this one too. I'm not sure about the text "compiler path has an invalid file name" though.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -501,6 +501,12 @@ impl Config {
     /// Run the compiler, generating the file `output`
     ///
     /// The name `output` must begin with `lib` and end with `.a`
+    ///
+    /// # Panics
+    ///
+    /// Panics if `output` is not formatted correctly or if one of the underlying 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Could you also mention that panics can originate if the compiler can't be found or otherwise fails?

rap2hpoutre:
Thanks! So you mean replacing: 

 > Panics if compiler path has an invalid file name.

By:

> Panics if compiler path has an invalid file name, if it can not be found or otherwise fails

Am I right?

rap2hpoutre:
(build failed but I don't think it's related to my commit)

alexcrichton:
Thanks!

Issue #185, opened 2017-07-13
Change `compile` to take just a bare name, not a full file name


brson:
This method takes a library name of the form "libfoo.a". This is a weird API because the proper way to name libraries differs across platforms. Change this to take just "foo", and make it clear how that will be translated to a static lib file name behind the scenes.

Remove any assertions about the form of this library - if the users happens to write "libfoo.a" anyway the resulting artifact will just be named "liblibfoo.a.a", which should still work.

There's a weird backcompat concern here since probably _everybody_ is going to quietly upgrade without changing this string, and is going to get weirdly named libraries. Should we care about this cc @alexcrichton @sfackler @dtolnay @burntsushi.

alexcrichton:
In some sense the string here doesn't actually matter, `gcc` could always infer a name and pick it up automatically. In light of that I see this string purely as a debugging tool when things go awry. I think the crate should just be "smart" here, if the string looks like `libfoo.a` then generate `libfoo.a`, and if it looks like `foo` then also generate `libfoo.a`. In both cases it should detect duplicates and handle that gracefully (in theory), but probably doesn't have to do it yet.

AndyGauge:
Should it warn when the old syntax is used?

alexcrichton:
Fixed in https://github.com/alexcrichton/gcc-rs/pull/212

PR #212, opened 2017-07-18
Changed compile to accept name of library and updated examples

AndyGauge:
Addresses #185

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,7 +24,7 @@
 //! extern crate gcc;
 //!
 //! fn main() {
-//!     gcc::compile_library("libfoo.a", &["src/foo.c"]);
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -296,7 +296,7 @@ impl Config {
     ///             .file("src/foo.c")
     ///             .shared_flag(true)
     ///             .static_flag(true)
-    ///             .compile("libfoo.so");
```
AndyGauge: Can you review this line.  I'm really not sure if this is in scope.  IDK how .so passes that assert!

alexcrichton: Nah yeah this is fine!

alexcrichton:
Thanks!

PR #205, opened 2017-07-15
Config::files method

opilar:
Resolves #199

PR #213, opened 2017-07-18
Remove `compile_library`

### LINKED ISSUE #188: Remove `compile_library` function
This is not actually that much shorter to use than `Config::new().file("foo.c").compile("foo");`. Update docs accordingly.

omh1280:
Fixes #188.

Pretty straightforward removal. I use the syntax introduced in #205 in the README.

Let me know if it needs work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,9 @@ Next up, you'll want to write a build script like so:
 extern crate gcc;
 
 fn main() {
-    gcc::compile_library("libfoo.a", &["foo.c", "bar.c"]);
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! In terms of coordination though could this avoid deleting it just yet? Instead can this delete all references and just slap a `#[deprecated]` and `#[doc(hidden)]` on it? 

When a new major release comes out I can clean out the deprecated pieces.

omh1280:
Updated! Thanks Alex

PR #214, opened 2017-07-19
Preventing bug if liberace makes a library using new syntax

AndyGauge:
Realized my code merged (#212) would cause the liberace library to produce a file liberace.a rather than libliberace.a


alexcrichton:
Thanks!

Issue #173, opened 2017-07-12
RFE: flag_if_supported


ronnychevalier:
In [Tectonic](https://github.com/tectonic-typesetting/tectonic/), we have a lot of legacy C code for the moment (the end goal being having as less C code as possible). In the mean time, we want to improve the C code quality by using warning flags while not triggering errors on compilers not supporting such flags.

We [added](https://github.com/tectonic-typesetting/tectonic/commit/574621cdae6adaaa544c6c70f9ce28c60d829311) a wrapper around `gcc-rs`, called `flag_if_supported` to check whether a flag exists or not and add it to the list of flags if this is the case. Similar to [CHECK_C_COMPILER_FLAG](https://cmake.org/cmake/help/v3.0/module/CheckCCompilerFlag.html) from cmake.

The problem is that since we do not have access to the fields of `Config`, we have to duplicate some work to find the target, add the output file to the command, and add a duplicate of the `cpp` flag.

Would you be interested in having this feature directly in `gcc-rs`?

This way we would have direct access to the cpp flag, the target and add transparently the output file (for each target supported by `gcc-rs`). In addition, this would avoid implementing it in each `build.rs` needing this feature.

Related: #48 have a proposal about `try_compile`

alexcrichton:
Sounds good to me!

PR #215, opened 2017-07-19
config: add flag_if_supported

ronnychevalier:
It allows one to add a flag to the compiler on the condition that the
compiler supports it.
We check if the compiler supports the flag by trying to compile a
simple executable. If there is some output on stderr we consider that
the compiler failed and the flag is not supported.

See #173

### DIFF
--- a/gcc-test/build.rs
+++ b/gcc-test/build.rs
@@ -11,6 +11,8 @@ fn main() {
 
     gcc::Config::new()
         .file("src/foo.c")
+        .flag_if_supported("-Wall")
+        .flag_if_supported("-Wfoo-bar-this-flag-does-not-exist")
         .define("FOO", None)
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks!

ronnychevalier:
I just saw that the CI [failed](https://travis-ci.org/alexcrichton/gcc-rs/jobs/255398486) due to the use of `?` which is only available and stable since [1.13.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1130-2016-11-10), while the minimum version supported is 1.6.0.

Should I send a patch to use `try!()` instead or increase the minimum version supported?
The former seems a better idea.

alexcrichton:
Ah yeah it's probably easy enough to just stick to `try!` for now, want to send a PR with that?

PR #216, opened 2017-07-19
Add warnings control

### LINKED ISSUE #190: Add warning control
Do a few things:

- Turn on maximum warnings by default, in the compiler-specific way
- Add a method `warnings_as_errors` that turns warnings into errors in a compiler specific way

We haven't discussed this, but if we enable maximum warnings by default, it seems like some users will _not_ want that and a way to turn it off. So it seems like we should also add a `default_warnings` method that declines to set the maximum warnings.

opilar:
Resolves #190

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,6 +91,8 @@ pub struct Config {
     shared_flag: Option<bool>,
     static_flag: Option<bool>,
     check_file_created: bool,
+    warnings_into_errors: bool,
+    warnings: bool,
 }
 
 /// Configuration used to represent an invocation of a C compiler.
@@ -151,6 +153,28 @@ impl ToolFamily {
             ToolFamily::Clang => "-E",
         }
     }
+
+    /// What the flags to enable all warnings
+    fn warnings_flags(&self) -> &'static [&'static str] {
+        static MSVC_FLAGS: &'static [&'static str] = &["/Wall"];
+        static GNU_FLAGS: &'static [&'static str] = &["-Wall", "-Wpedantic", "-Wextra"];
+        static CLANG_FLAGS: &'static [&'static str] = &["-Weverything"];
+
+        match *self {
+            ToolFamily::Msvc => &MSVC_FLAGS,
+            ToolFamily::Gnu => &GNU_FLAGS,
+            ToolFamily::Clang => &CLANG_FLAGS,
+        }
+    }
+
+    /// What the flag to turn warning into errors
+    fn warnings_to_errors_flag(&self) -> &'static str {
+        match *self {
+            ToolFamily::Msvc => "/WX",
+            ToolFamily::Gnu |
+            ToolFamily::Clang => "-Werror"
+        }
+    }
 }
 
 /// Compile a library from the given set of input C files.
@@ -206,6 +230,8 @@ impl Config {
             pic: None,
             static_crt: None,
             check_file_created: false,
+            warnings: true,
+            warnings_into_errors: false,
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -146,6 +148,28 @@ impl ToolFamily {
...
+    fn warnings_flags(&self) -> &'static [&'static str] {
+        static MSVC_FLAGS: &'static [&'static str] = &["/Wall"];
+        static GNU_FLAGS: &'static [&'static str] = &["-Wall", "-Wpedantic", "-Wextra"];
+        static CLANG_FLAGS: &'static [&'static str] = &["-Weverything"];
+
+        match *self {
+            ToolFamily::Msvc => &MSVC_FLAGS,
+            ToolFamily::Gnu => &GNU_FLAGS,
+            ToolFamily::Clang => &CLANG_FLAGS,
+        }
```
opilar: Can it be simpler?

alexcrichton:
Thanks! Could you be sure to add some tests for this as well?

Also, is it worth having different sets of default warnings for gcc vs clang?

src/lib.rs
ronnychevalier: I am not sure using `-Wpedantic` is a good idea... It rejects (an error not a warning) some programs using [non standard features](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html). Since this is by default it will break many crates using `gcc-rs` since it is not common for C code to strictly follow ISO C standard.

Also, using `-Weverything` with clang is not necessarily a good idea either. There is some warnings that do not make sense in some programs. It will just spam the output when compiling for no good reason and it will probably get people used to see warnings, thus they will ignore them...

src/lib.rs
```diff
@@ -146,6 +148,28 @@ impl ToolFamily {
...
+    }
+
+    /// What the flag to turn warning into errors
+    fn warnings_to_errors_flag(&self) -> &'static str {
+        match *self {
+            ToolFamily::Msvc => "/WX",
+            ToolFamily::Gnu |
+            ToolFamily::Clang => "-Werror"
+        }
```
ronnychevalier: I think turning warnings into errors only make sense if you are a developer of the crate using gcc-rs (i.e., enabling the warnings as errors only in a "developer mode" or something). Any user of this crate, or any other crate depending on it, could fail during compile time. Some warnings only appear on some architecture. Some people might have a version of the compiler that have a new warning, but not the developers. Hence, if the developers use another architecture or another compiler they will not have any warning, thus no error. But another person might have the error.

I think it would make sense to put some comment on the risk of using this function in your `build.rs`?

What do you think?

alexcrichton: Sounds fine by me!

src/lib.rs
opilar: Thank you!

alexcrichton:
Thanks!

src/lib.rs
```diff
@@ -91,6 +91,8 @@ pub struct Config {
     shared_flag: Option<bool>,
     static_flag: Option<bool>,
     check_file_created: bool,
+    warnings_into_errors: bool,
+    warnings: bool,
```
gnzlbg: I'd like to be able to turn off -Wextra only, many projects compile with -Wall just fine, but they don't compile silently with -Wextra, and this is on purpose.

gnzlbg:
Why was `-Wextra` added here? `-Wextra` is basically the set of warnings that aren't good enough to be enabled by default, due to either false-positives, or because they are opinionated. Enabling them for all projects by default, and providing only a way to either use `-Wall -Wextra` or no warnings at all, does not appear to be reasonable behavior to me.

PR #217, opened 2017-07-19
Add a compile_binary function

zrneely:
Sometimes, the output of a program which must be written in C/C++ is useful for a build script. This enables building such a program, which can then be called later in the build script with `Stdio::piped()` to retrieve that output. Alternatives include asking the user to manually build the program ahead of time in a README, or calling the C/C++ compiler by hand with `std::process::Command`.

In order to keep the implementation simple, `compile_binary` only supports single-object-file programs, and panics if any more than one than one file is provided. I believe panic-ing to be an acceptable result here since providing multiple files is always a programming mistake and failing silently could produce unexpected behavior and hard-to-find bugs. This decision could be changed later if there is demand.

I've tested this on Linux, but I don't have easy access to a Windows box to test it on. I'm reasonably confident that it should work there, and appveyor will prove it one way or another.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "gcc"
-version = "0.3.51"
+version = "0.3.52"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT/Apache-2.0"
 repository = "https://github.com/alexcrichton/gcc-rs"
--- a/gcc-test/build.rs
+++ b/gcc-test/build.rs
@@ -91,4 +91,9 @@ fn main() {
         .expand();
     let out = String::from_utf8(out).unwrap();
     assert!(out.contains("hello world"));
+
+    // Tests that we can build a binary.
+    gcc::Config::new()
+        .file("src/hello.c")
+        .compile_binary("hello");
 }
--- a/gcc-test/src/hello.c
+++ b/gcc-test/src/hello.c
@@ -1,0 +1,7 @@
+#include <stdio.h>
+
+int main() {
+  printf("Hello World!\n");
+
+  return 0;
+}
--- a/gcc-test/src/hello.c
+++ b/gcc-test/src/hello.c
@@ -1,7 +1,7 @@
 #include <stdio.h>
 
 int main() {
-  printf("Hello World!\n");
+  printf("Hello World!");
 
   return 0;
 }
--- a/gcc-test/tests/all.rs
+++ b/gcc-test/tests/all.rs
@@ -19,7 +19,7 @@ fn hello_works() {
     let output = child.wait_with_output().expect("Failed to wait on hello.").stdout;
     let output = String::from_utf8(output).unwrap();
 
-    assert_eq!(output, "Hello World!\n");
+    assert_eq!(output, "Hello World!");
 }
 
 #[test]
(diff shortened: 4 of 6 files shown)


zrneely:
The Travis failure happens on Rust 1.6.0 only from the use of `#[deprecated]` on `compile_library` - it's unrelated to this change. All the other checks pass.

zrneely:
The Travis failure happens on Rust 1.6.0 only from the use of `#[deprecated]` on `compile_library` - it's unrelated to this change. All the other checks pass.

alexcrichton:
Thanks for the PR! I've personally been hesitant to add this in the past, however, because this isn't possible with cross compilation which this crate tries pretty hard to support. Mind detailing a bit the use case you have in mind?

zrneely:
In my case, I'm writing bindings for libmediainfo, which has functions which can take one of hundreds of stringly-typed arguments, and a function which informs the client of each possible option by returning the strings as CSV data. I'd like to call that function, and automatically generate enums from them (I actually have this working already using my branch).

Cross-compilation is a problem, yeah. In my case, the library's options available depend only on the version of the library used, and not the operating system, so there's no issues - you can just always run the compiled program on your native architecture and use those results on every cross-compilation target. `compile_binary` isn't intended to build a shippable artifact, it's meant to allow to build part of your compilation process without requiring a multi-stage or manual build, so cross-compilation hopefully isn't an issue, but the documentation should reflect that more clearly.

Another issue is dynamic linking. Since my project is temporarily Linux-only, I just used `.to_command()` to append a `-lmediainfo` argument after all the ones generated. (On Windows, you'd have to append `/link mediainfo.lib`, and have the `.lib` somewhere the linker could find it.) I could also add a `fn dynamic_link(&mut self, library: &str)` to this PR which would generate those arguments.

alexcrichton:
This seems to me like something that's better dealt with as code generation before a crate is published rather than something that happens at build time for the crate? Otherwise in cross-compile situations you're compiling the library twice where you'd otherwise only need to compile it once?

zrneely:
In my scenario, you'd use `gcc-rs` to compile your dynamically-linked-against-the-native-libxyz C/C++ program once to generate its output, which would then be used by the build script to build the statically-linked-against-a-cross-compiled-libxyz Rust code once. My assumption is that you have a system libxyz installed already, so there's no need to compile it the "first time".

Generating the data and checking it in to git prior to crate publication is also a perfectly valid alternative, but updating that data could be complex and interferes with the simple "`git clone` → `cargo build`" process. I personally think the reduced complexity for those building the crate from scratch is worth the requirement of having a system copy of the library you're linking against installed, but I agree it may not always be possible.

Also, apologies for leaving this branch in a broken state for so long - I'll hopefully push some new commits to get it working soon™.

alexcrichton:
Hm well I guess my main worry is that adding a function like this I feel implies that it works in more scenarios than it's likely to. Even in libraries which have their own set of constants, the chance of various values always being constant across all architectures seems like it may be a bit slim? In that if *any* libc constants are mixed in, for example, the output may be wrong.

I wonder though if this could be added externally? Could there be a separate crate to compile and run executables using `gcc` as a support library?

zrneely:
I think it would be reasonable to have a separate crate. I'll close this and work on that instead.

PR #219, opened 2017-07-21
Add note that crate can compile assembly.

### LINKED ISSUE #193: Mention in the crate docs that this crate also supports assembly
This crate also works, via the C compiler, with assembly files with extension `.S` with gcc/clang, and `.asm` with MSVC. Mention this in the crate docs.

chrisvittal:
Closes #193. I tried to keep the brief summary style of the rest of the main crate doc. If you would like me to include an example I can do that, I just need a place to put them. The top of the crate docs didn't seem like the right place.


alexcrichton:
Thanks!

Issue #187, opened 2017-07-13
Clarify behavior of `expand`


brson:
This is a convenience function for doing macro expansion. It only makes sense to use with a single file though, so add an assert that there's only a single file in the config, and document that it will panic if there are multiple files configured.

PR #221, opened 2017-07-24
Clarify behavior of `expand`

jehiggs:
Resolves [#187](https://github.com/alexcrichton/gcc-rs/issues/187).

I wasn't too sure what this should do if there's no files in the config. Expanding no files seems like a strange thing to do, but it wasn't clear to me that attempting to do so should also be a panic condition.


alexcrichton:
Looks good to me, thanks!

Issue #192, opened 2017-07-13
Add try_compile and try_expand


brson:
As a build script the default mode is to just panic on error, but sometimes that is not appropriate. Add `try_compile` and `try_expand` methods that return a Result with a new enum type for the possible errors cases.

It probably makes sense to have the new error type`s `description` method return the current panic messages, and change the existing `try_compile`, `try_expand` methods to simply `panic!("{}", e.description())`.

This may require some internal refactoring to convert internal panics into Results.

marco9999:
Hi, is this appropriate for someone new to rust to work on? I've finished reading through the second book, and would like to further improve my skills.

alexcrichton:
@marco9999 sure! This'd mostly involve adding a custom error type, changing all panics to returning an error, and then propagating that result outwards to the API boundaries. We'd then rejigger the top-level APIs to have `compile` and `expand` panic while the `try_*` variants would return an error.

If you've got any questions just let me know!

marco9999:
Great, I'll give it a go. Thanks!

pornel:
Does the `try_compile` set any cargo flags?

I'm interested in testing whether certain compiler features or flags are supported, such as OpenMP (`-fopenmp`) or Intel-compiler-specific optimization settings. In old school configure scripts that's done by using a minimal file like `int main(void){}` and checking whether it compiles with the flags.

I would try to do the same with `gcc-rs`, but if `try_compile` implicitly prints instructions for Cargo, then I'm afraid I'd end up linking the actual Rust product with garbage from test files.

alexcrichton:
@pornel I think a recent addition of `flag_if_supported` would suffice for that?

pornel:
Yes, it's perfect!

alexcrichton:
Fixed in https://github.com/alexcrichton/gcc-rs/pull/222

Issue #48, opened 2015-06-29
Add try_compile, returning Result<(), String>


seanmonstar:
The `String` could be `stderr` from the command.

This could be used to try to provide help messages, by checking the stderr, noticing an error about something like a missing header file, and suggesting to the user instructions on how to install those headers.

sfackler:
I'd like this as well.

alexcrichton:
One thing I realized recently is that the [`get_compiler` method](http://alexcrichton.com/gcc-rs/gcc/struct.Config.html#method.get_compiler) may actually suffice for this pretty well, can you perhaps test out using that API and see if it works well? If you end up having to pass lots of your own flags and such I can see where that may not quite work out so well though :)

dtolnay:
This is also tracked in #192.

alexcrichton:
Fixed in https://github.com/alexcrichton/gcc-rs/pull/222

PR #222, opened 2017-07-29
Added in error checking for the compile() and expand() functions.

marco9999:
Initial attempt to resolve issue #192.

New functions try_compile() and try_expand() have been introduced that returns a result. I have introduced the `Error` struct type, which is used within the result in the form of `Result<_, Error>`. The error contains the type of error (`enum ErrorKind`, unused for now) and a detailed message (`String`) to display to the user.

I also had to change the signature of the pubic function get_compiler(), in order to propagate errors (in this specific case, it is because of environment vars missing) - if you don't want to do this, maybe I can put in try_get_compiler and get_compiler?

This was pretty fun to work on, I might have a go at some others. One problem I ran into was you can't propagate a `Result<>` from a closure (used within `Option<>::unwrap_or_else()` calls), so I moved the relevant code out into a match statement instead. Not sure if there was another way to do this, but I at least understand why it's not allowed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -319,7 +319,7 @@ impl Config {
            .host(&target)
            .debug(false)
            .cpp(self.cpp);
-        let compiler = cfg.get_compiler()?;
+        let compiler = cfg.try_get_compiler()?;
         let mut cmd = compiler.to_command();
         command_add_output_file(&mut cmd, &obj, target.contains("msvc"), false);
         cmd.arg(&src);
@@ -712,7 +712,7 @@ impl Config {
         let (mut cmd, name) = if msvc && is_asm {
             self.msvc_macro_assembler()?
         } else {
-            let compiler = self.get_compiler()?;
+            let compiler = self.try_get_compiler()?;
             let mut cmd = compiler.to_command();
             for &(ref a, ref b) in self.env.iter() {
                 cmd.env(a, b);
@@ -736,7 +736,7 @@ impl Config {
     ///
     /// This will return a result instead of panicing; see expand() for the complete description.
     pub fn try_expand(&self) -> Result<Vec<u8>, Error> {
-        let compiler = self.get_compiler()?;
+        let compiler = self.try_get_compiler()?;
         let mut cmd = compiler.to_command();
         for &(ref a, ref b) in self.env.iter() {
             cmd.env(a, b);
@@ -794,8 +794,20 @@ impl Config {
     /// environment variables (a number of which are inspected here), and then
     /// falling back to the default configuration.
     ///
-    /// An error may occur while determining the architecture.
-    pub fn get_compiler(&self) -> Result<Tool, Error> {
+    /// # Panics
+    ///
+    /// Panics if an error occurred while determining the architecture.
+    pub fn get_compiler(&self) -> Tool {
+        match self.try_get_compiler() {
+            Ok(tool) => tool,
+            Err(e) => fail(&e.message),
+        }
+    }
+
+    /// Get the compiler that's in use for this configuration.
+    ///
+    /// This will return a result instead of panicing; see get_compiler() for the complete description.
+    pub fn try_get_compiler(&self) -> Result<Tool, Error> {
         let opt_level = self.get_opt_level()?;
         let target = self.get_target()?;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,6 +93,30 @@ pub struct Config {
     check_file_created: bool,
 }
 
+/// Represents the types of errors that may occur while using gcc-rs.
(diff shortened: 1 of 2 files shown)


marco9999:
Hmm, I don't know how to properly handle error checking within a parallel loop... (didn't have the parallel feature turned on for myself)

src/lib.rs
```diff
@@ -266,29 +290,47 @@ impl Config {
         self
     }
 
-    fn is_flag_supported(&mut self, flag: &str) -> io::Result<bool> {
+    fn is_flag_supported(&mut self, flag: &str) -> Result<bool, Error> {
```
alexcrichton: I think the original source of the error here is `io::Error`, so perhaps this could continue to return a `io::Result`?

src/lib.rs
```diff
@@ -93,6 +93,30 @@ pub struct Config {
...
+    ToolExecError,
+    /// Error occurred due to missing external tools.
+    ToolNotFound,
+}
+
+/// Represents an internal error that occurred, with an explaination.
+#[derive(Clone, Debug)]
+pub struct Error {
+    /// Describes the kind of error that occurred.
+    pub kind: ErrorKind,
```
alexcrichton: We typically try to keep the fields of a struct private for maximal forwards compatibility, so could the `ErrorKind` enum and the fields of this struct be left private? We could always publicly expose them later, however!

src/lib.rs
```diff
@@ -266,29 +290,47 @@ impl Config {
...
+        let compiler = cfg.get_compiler()?;
         let mut cmd = compiler.to_command();
         command_add_output_file(&mut cmd, &obj, target.contains("msvc"), false);
         cmd.arg(&src);
 
-        let output = try!(cmd.output());
+        let output = match cmd.output() {
+            Err(_) => return Err(Error {
+                kind: ErrorKind::ToolExecError,
+                message: "Flag check failed to obtain output.".to_string(),
```
alexcrichton: I think it'd be possible to cut down on the verbosity here by adding:

```rust
impl From<io::Error> for Error {
    // ...
}
```

and that way you can just use `?`!

src/lib.rs
```diff
@@ -569,19 +602,39 @@ impl Config {
...
             let obj = dst.join(file).with_extension("o");
             let obj = if !obj.starts_with(&dst) {
-                dst.join(obj.file_name().unwrap())
+                dst.join(match obj.file_name() {
+                    Some(s) => s,
+                    None => return Err(Error {
+                        kind: ErrorKind::IOError,
+                        message: "Getting object file details failed.".to_string(),
+                    }),
+                })
```
alexcrichton: This might be a little neater as a helper function, so you could do something like:

```rust
dst.join(obj.file_name().ok_or_else(|| err("failed to get details"))?);
```

(or something like that)

src/lib.rs
dtolnay: :+1: for private for now. Specifically #48 is asking for the content of stderr to be available in the error, which would need to be added here later.

marco9999: Sure. What happens when someone tries to use this outside of the crate though? As in, will there be some sort of visibility issue when using the `Result<_, Error>` from one of the try functions?

src/lib.rs
marco9999: This function calls `get_target()` and `try_get_compiler()`, in which at least `get_target()` tries to determine if an environment variable exists. The `std::env::var()` function it calls doesn't return any kind of `io::Error` (it used to panic if there the result was `None`), so I don't think its possible to use this?

src/lib.rs
marco9999: Sure, I will take a look into this.

src/lib.rs
marco9999: Sure, I will take a look into this.

src/lib.rs
alexcrichton: Ah yes indeed, good point!

src/lib.rs
alexcrichton: Oh so for now you just won't be able to look inside of `Error` outside this crate. Eventually we can add accessors for various fields and such in a backwards compatible fashion, but for now external users will just see that an `Error` type exists.

marco9999:
Are you still supporting rustc 1.9.0?

alexcrichton:
Everything looks great to me, thanks so much @marco9999! Want to update the Travis config to inrease the minimum version to a Rust with the `?` operator? Othert than that I believe this is ready to merge.

marco9999:
Updated the rust version to 1.13.0. I think everything is ready...

alexcrichton:
Alright merged and good to go, thanks again @marco9999!

PR #225, opened 2017-08-04
Make ar outputs deterministic

metajack:
ar entry headers include timestamps and uid/gid which are not deterministic outputs. While binutils added command line flags to zero these fields out to make deterministic builds possible, macOS ar does not have these. This PR makes all invocations of ar deterministic by rewriting the archive and zeroing out the appropriate fields.

The immediate benefit of doing this is that sccache (and potentially other future Rust compiler caches) will be able to cache crates that include static C libraries are part of their build process.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ travis-ci = { repository = "alexcrichton/gcc-rs" }
 appveyor = { repository = "alexcrichton/gcc-rs" }
 
 [dependencies]
+ar = "0.3"
 rayon = { version = "0.8", optional = true }
 
 [features]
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! [According to the documentation](https://docs.rs/ar/0.3.0/ar/) though I think that the `ar` crate only supports writing BSD archives (those on OSX), so I wonder if that's related to the test failures?

metajack:
I think that the testing of the fake compiler means ar doesn't output a file. And I don't handle that case: https://travis-ci.org/alexcrichton/gcc-rs/jobs/261141868#L401

I'll put in a thing that skips if there's no archive and repush.

metajack:
Well, that fixed half of them :(

alexcrichton:
I'm going to close this due to inactivity, but I'm more than willing to merge when tests pass!

metajack:
I decided to make sccache hash ar files better instead. I probably won't reopen this unless that path hits a roadblock.

PR #226, opened 2017-08-08
Rename `Config` type to `Build`

### LINKED ISSUE #189: Rename `Config` type
This is not all that descriptive a name, though we're not sure what a better name is. I suggest `Builder`. If nothing better comes up, pick that and we'll bikeshed it on the PR.

cc @alexcrichton @sfackler @dtolnay @kimundi @burntsushi bikeshedding.

laumann:
From the discussion it seems that `gcc::Build` was the preferred name.

Fixes #189

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Next up, you'll want to write a build script like so:
 extern crate gcc;
 
 fn main() {
-    gcc::Config::new()
+    gcc::Build::new()
         .file("foo.c")
         .file("bar.c")
         .compile("foo");
@@ -141,13 +141,13 @@ required varies per platform, but there are three broad categories:
 ## C++ support
 
 `gcc-rs` supports C++ libraries compilation by using the `cpp` method on
-`Config`:
+`Build`:
 
 ```rust,no_run
 extern crate gcc;
 
 fn main() {
-    gcc::Config::new()
+    gcc::Build::new()
         .cpp(true) // Switch to C++ library compilation.
         .file("foo.cpp")
         .compile("libfoo.a");
--- a/gcc-test/build.rs
+++ b/gcc-test/build.rs
@@ -9,7 +9,7 @@ fn main() {
     fs::remove_dir_all(&out).unwrap();
     fs::create_dir(&out).unwrap();
 
-    gcc::Config::new()
+    gcc::Build::new()
         .file("src/foo.c")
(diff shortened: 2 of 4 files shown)


alexcrichton:
Thanks! Mind also leaving behind a temporary `pub use Build as Config` with a `#[doc(hidden)]` so we can land this in the 0.3 series?

laumann:
Sure! Should there be a deprecation warning to alert users to change `Config` to `Build`?

laumann:
If I do:
```rust
#[doc(hidden)]
#[deprecated(since="0.3.51", notes="gcc::Config has been renamed to gcc::Build")]
pub type Config = Build;
```
I get deprecation warnings, but that doesn't work with `pub use Build as Config`

@alexcrichton Which do you prefer?

alexcrichton:
@laumann ah unfortunately I think they'll effectively be the same which is to say functionally equivalent and `#[deprecated]` likely ignored on both.

In any case though this looks great to me, thanks!

laumann:
> ah unfortunately I think they'll effectively be the same which is to say functionally equivalent and #[deprecated] likely ignored on both.

I tried using `gcc::Config` in `gcc-test/build.rs` and only with the type alias did I get a deprecation warning when running `cargo test` (not with `pub use`).

Like this:
```
$ pwd
/home/tj/rust/gcc-rs/gcc-test
$ cargo check
   Compiling gcc v0.3.51 (file:///home/tj/rust/gcc-rs)
   Compiling gcc-test v0.1.0 (file:///home/tj/rust/gcc-rs/gcc-test)
warning: use of deprecated item: gcc::Config has been renamed to gcc::Build
  --> build.rs:12:5
   |
12 |     gcc::Config::new()
   |     ^^^^^^^^^^^^^^^^
   |
   = note: #[warn(deprecated)] on by default
    Finished dev [unoptimized + debuginfo] target(s) in 2.19 secs
```
But I don't know if there is anything special about `gcc-test`...

alexcrichton:
Oh hey! I guess some bugs in rustc have gotten fixed since I last looked!

laumann:
Always a nice surprise :-) But should deprecation on `pub use Foo as Bar` also generate a warning?

alexcrichton:
Last I checked it didn't, but maybe it's changed!

PR #227, opened 2017-08-08
Config: accept Into<Option<&str>> instead of Option<&str>

### LINKED ISSUE #224: Accept Into<Option<&str>> rather than Option<&str> in Config methods
There are currently three Config methods that take Option\<&str\>. It seems like they are commonly going to be called with Some string.

```rust
define("FOO", Some("BAR"))
```

Into\<Option\<&str\>\> could make the calls look neater.

```rust
define("FOO", "BAR")
define("FOO", None) // uncommon
```

laumann:
Per @dtolnay's suggestion, change methods on Config that accepts Option<&str> to
accept Into<Option<&str>> to make the calls look neater.

Fixes #224

### DIFF
--- a/gcc-test/build.rs
+++ b/gcc-test/build.rs
@@ -14,7 +14,7 @@ fn main() {
         .flag_if_supported("-Wall")
         .flag_if_supported("-Wfoo-bar-this-flag-does-not-exist")
         .define("FOO", None)
-        .define("BAR", Some("1"))
+        .define("BAR", "1")
         .compile("libfoo.a");
 
     gcc::Config::new()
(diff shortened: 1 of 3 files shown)


laumann:
I don't know what happened with AppVeyor :confused:

opilar:
@laumann maybe we can close and reopen the issue to restart AppVeyor?

laumann:
@opilar Thanks for the tip :-) Seems to be running now

alexcrichton:
Thanks @laumann!

laumann:
Happy to be of help :smile:

PR #229, opened 2017-08-08
Check for DEBUG env var

### LINKED ISSUE #220: Debug info should depend on the DEBUG var, not PROFILE
Currently, `gcc` [checks](https://github.com/alexcrichton/gcc-rs/blob/501a5bfb4add0f4f04bb6829bf3b8848fa08b207/src/lib.rs#L1230) if `PROFILE` is "debug" and enables debug info if so. I'm not sure if this was added recently, but Cargo [appears](http://doc.crates.io/environment-variables.html) to set the `DEBUG` env var to the `debug` setting in the active profile, which seems like a better variable to check.

If this is done, the compiled C code has debug info iff the Rust code has debug info (this isn't currently the case).

opilar:
Resolves #220

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1242,7 +1242,7 @@ impl Config {
     }
 
     fn get_debug(&self) -> bool {
-        self.debug.unwrap_or_else(|| self.getenv_unwrap("PROFILE") == "debug")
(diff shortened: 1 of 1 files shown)


alexcrichton:
Awesome, thanks!

PR #231, opened 2017-08-11
Fix non-delimited library names

FaultyRAM:
These were accepted in https://github.com/alexcrichton/gcc-rs/pull/212, but never actually worked and weren't being tested. This fixes the logic errors that prevented the new style from working, and also updates the tests to use the new style.


alexcrichton:
Thanks!

PR #238, opened 2017-08-23
Fix flag_if_supported for C++ only flags

### LINKED ISSUE #236: flag_if_supported does not work for C++ flags
`flag_if_supported()` does not work for flags that are C++ only. From what I can tell, this is because the file used to test the flag is a C file, which causes the compiler to throw up.

I use this to set the C++ standard, which would essentially solve #191 (unless you want to support this explicitly).

**Example configuration**

```rust
    gcc::Build::new()
        .cpp(true)
        .flag_if_supported("-std=c++14")
        .file("main.cpp")
        .compile("libtest.a");
```

jan-auer:
Fixes #236. 

Note that in order to test with `cpp(true)`, I had to add a shim to `c++`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,7 +98,6 @@ pub struct Build {
     static_crt: Option<bool>,
     shared_flag: Option<bool>,
     static_flag: Option<bool>,
-    check_file_created: bool,
     warnings_into_errors: bool,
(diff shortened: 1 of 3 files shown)


alexcrichton:
Looks great to me, thanks!

PR #243, opened 2017-09-16
Revamped support for CUDA C++ compilation.

peterhj:
This adds slightly more robust support for compiling CUDA (*.cu) files using the CUDA compiler (nvcc) if it's available.

What this does:
1. Add a method `cuda` to `Build`, similar to `cpp`.
2. If CUDA compilation is enabled, replace the path of the inferred toolchain with the path to nvcc: this is either "nvcc", or it looks for the environment variable $NVCC.
3. Add the path of the inferred toolchain via a flag: "-ccbin=<original_compiler_path>".
4. Before some flags, add an extra "-Xcompiler" flag to signal to nvcc to pass them to the underlying C/C++ compiler.


src/lib.rs
```diff
@@ -967,15 +1009,13 @@ impl Build {
...
-                    if self.pic.unwrap_or(!target.contains("windows-gnu")) {
-                        cmd.args.push("-fPIC".into());
-                    }
-                } else if self.pic.unwrap_or(false) {
-                    cmd.args.push("-Xcompiler".into());
-                    cmd.args.push("\'-fPIC\'".into());
+                cmd.nvcc_wrap_arg(self.cuda);
+                cmd.args.push("-ffunction-sections".into());
+                cmd.nvcc_wrap_arg(self.cuda);
+                cmd.args.push("-fdata-sections".into());
```
alexcrichton: Could this idiom perhaps get wrapped up into a method? Something which internally pairs these two invocations together?

src/lib.rs
```diff
@@ -450,6 +476,22 @@ impl Build {
...
+    /// Enabling CUDA will pass the detected C/C++ toolchain as an argument to
+    /// the CUDA compiler, NVCC. NVCC itself accepts some limited GNU-like args;
+    /// any other arguments for the C/C++ toolchain will be redirected using
+    /// "-Xcompiler" flags.
+    ///
+    /// If enabled, this also implicitly enables C++ support.
+    pub fn cuda(&mut self, cuda: bool) -> &mut Build {
+        self.cuda = cuda;
+        if cuda {
+            self.cpp(true);
```
alexcrichton: I'm pretty unknowledgeable when it comes to everything cuda, but is there maybe C code which gets compiled as well? Or does this suffice for all compilations?

alexcrichton:
Thanks! This is looking like it's getting complicated enough, though, that perhaps there could be some tests for the cuda compilation?

alexcrichton:
I was also thinking, I wonder if we should automatically detect cuda files to not require the explicit `cuda` option?

src/lib.rs
peterhj: Hmm, originally I had a `Tool::wrapped_arg` function that did just that, but I thought this sort of thing might not be exclusive to CUDA, i.e. it might be possible there are other C-like extensions where we want to prefix a subset of possible args but leave the rest untouched, and the subset will invariably differ between these extensions. For example it seems the "-Xcompiler" convention originates at least from libtool. I don't know, what's your take on this rationale?

src/lib.rs
peterhj: AFAIK the "CUDA" dialect is really a superset of C++11 (with extensions for writing and calling GPU code in something that looks like C++). Although NVCC should be able to indirectly compile plain C files if you pass it `-ccbin=gcc`.

peterhj:
It's a good idea! More generally, if you have a library of mixed C, C++, CUDA C++, etc. files, it could also make sense to compile them with their respective compilers?

src/lib.rs
alexcrichton: Ok makes sense!

peterhj:
Apologies for the delay -- this preserves the `.cuda(true)` behavior, and also adds a `Tool::push_cc_arg` method to automatically add the nvcc-required `-Xcompiler` before certain flags.

One additional change I made was adding an `Object` struct to wrap pairs of (source file path, destination object path).

alexcrichton:
Thanks! (and sorry for the delay myself!)

Could the new apis added here be private instead of `pub`? (to keep the API surface area of this crate smaller). Also could the `push_cc_arg` function perhaps avoid taking a `cuda` argument and instead take the `cuda` boolean on construction and that way `push_cc_arg` just takes the parameter to push?

peterhj:
New methods are now private. `Tool` has a `cuda` field, as well as a new constructor `Tool::with_features` which takes a `cuda` argument.

alexcrichton:
Thanks!

PR #244, opened 2017-09-20
Use Thumb-2 as Android/armv7-a's default

makotokato:
Actually, gcc-es doesn't generate thumb2 option as default even if Android/armv7-a.

From https://developer.android.com/ndk/guides/standalone_toolchain.html#abi_compatibility

> We recommend using the -mthumb compiler flag to force the generation of 16-bit Thumb-2 instructions (Thumb-1 for armeabi). If omitted, the toolchain will emit 32-bit ARM instructions.

Since Android is mobile and embedded platform, so we should use -mthumb as android's default for footprint.


alexcrichton:
Looks good to me, thanks!

Issue #245, opened 2017-09-28
Using `flag_if_supported` forces recompile with `msvc`


Jake-Shadle:
It seems that using `flag_if_supported` causes an issue when using `msvc` (where it seems to affect the compilers dependency calculations) causing it to *always* rebuild, even when there are no changes to either the C/C++ code, nor the crate where the build.rs and accompanying Rust wrapper code is.

Compiling the same exact code with `gcc` yields the expected results of only compiling when an actual change has been made, and not calling `flag_if_supported` when using `msvc` does the same, so I'm pretty confident it's isolated to the combination.

I think this function is mainly for more painlessly supporting either gcc or clang, as msvc's compiler flags are so wildly different from them it doesn't really make sense to call it in the msvc context, so it might be easiest to put `flag_if_supported` behind `[cfg(not(target_env = "msvc"))]` or similar to stop people from accidentally shooting themselves in the foot with regard to compile times.

alexcrichton:
Oh dear this sounds bad an unintended! Do you have an example which is recompiling too much that I could play with?

Jake-Shadle:
Sure! Try out https://github.com/Jake-Shadle/cc_repro

alexcrichton:
Ah ok! Could you also set `RUST_LOG=cargo::ops::cargo_rustc::fingerprint=info` for a compilation which rebuilds too much when it shouldn't?

Jake-Shadle:
Ahh nice, didn't know about that 😄 
```
INFO:cargo::ops::cargo_rustc::fingerprint: fingerprint error for cc_repro v0.1.0 (file:///D:/git/cc_repro): precalculated components have changed: 13151091367.686187000s (D:/git/cc_repro/flag_check.exe) != 13151091344.312350600s (D:/git/cc_repro/flag_check.exe)
INFO:cargo::ops::cargo_rustc::fingerprint:   cause: precalculated components have changed: 13151091367.686187000s (D:/git/cc_repro/flag_check.exe) != 13151091344.312350600s (D:/git/cc_repro/flag_check.exe)
```
So yah, definitely the executable that gets generated to do the flag check.

alexcrichton:
Oh dear looks like that file is being generated in the wrong place for sure! I'll look to fixing this soon unless you beat me to it!

Jake-Shadle:
Ahh ok, I see what is happening, I think I can fix it, just to confirm, the flag_check is only intended to compile the object file, not link, correct?

PR #246, opened 2017-09-28
fix(flag_check): Send flag to MSVC telling it not to link

Jake-Shadle:
Using `flag_if_supported` with `msvc` was causing an actual executable to be produced in the root directory of the compiling crate, causing dependency checks to fail and forcing recompilation without any changes.

This change adds the [`/c`](https://docs.microsoft.com/en-us/cpp/build/reference/c-compile-without-linking) flag to the compiler args so it only compiles to the object file as is the default in gcc and clang.

Resolves: #245


alexcrichton:
Looks great to me, thanks!

PR #248, opened 2017-10-05
Add custom flags after -Wall

illicitonion:
Right now we add them before, and we add -Wall by default. This allows
users to manually turn off specific warnings, without having to resort
to a dance of:

```
 .warnings(false)
 .flag("-Wall")
 .flag("-Wno-foo")
```

This strikes a good balance of keeping -Wall on by default, while
minimising the surprising experience observed in
https://github.com/alexcrichton/cc-rs/issues/235.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1093,6 +1093,12 @@ impl Build {
             cmd.args.push(directory.into());
         }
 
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks!

PR #251, opened 2017-10-11
Add x86_64-unknown-linux-gnux32 support

malbarbo:
This is the first step in adding x86_64-unknown-linux-gnux32 target to rust. I was was able to build the  libstd using this patch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "cc"
-version = "1.0.0"
+version = "1.0.1"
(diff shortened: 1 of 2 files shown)


malbarbo:
Without this patch, the native dependencies of libstd are build with -m64, generating a linking error like this:

```
/usr/bin/ld: i386:x86-64 architecture of input file `/tmp/rustc.qSNX6EnkT2FV/liballoc_jemalloc-5c483b777c53a191.rlib(jemalloc.pic.o)' is incompatible with i386:x64-32 output
```

alexcrichton:
👍

Issue #254, opened 2017-10-17
Clarify use of `cpp_set_stdlib` flag


grovesNL:
Thanks for the great library.

In the [documentation for `cpp_set_stdlib`](https://github.com/alexcrichton/cc-rs/blob/master/src/lib.rs#L522) it states that the caller should ensure the flag is supported by the compiler. What is the recommended way of doing this? I was hoping the library could assist me here, because it already parses target from the environment.

I was interested in `try_get_compiler` and [using the returned `Tool`'s `ToolFamily`, but the `family` member is not public](https://github.com/alexcrichton/cc-rs/blob/master/src/lib.rs#L153). Is this intentional?

Alternatively, is there a way to accomplish this using `flag_if_supported`? Is there a reason why the semantics of `cpp_set_stdlib` couldn't allow for a similar `cpp_set_stdlib_if_supported` (just curious)?

alexcrichton:
Hm I haven't personally used this flag much, but it typically comes up for linux/osx/msvc I believe, not for clang/gcc inside of linux

grovesNL:
@alexcrichton Yes, this is for clang in OS X.

alexcrichton:
Here I was thinking that gcc didn't exist any more on OSX...

grovesNL:
I was trying to use gcc on OS X originally to get around this issue for now, but [it doesn't seem to work well with other libraries containing certain OS X types](https://travis-ci.org/gfx-rs/gfx/jobs/288817761#L1826). So I ended up using clang, which seems to be the happy path on OS X anyway, but [clang requires `cpp_set_stdlib` ](https://travis-ci.org/gfx-rs/gfx/jobs/288892138#L813).

grovesNL:
This has been mostly resolved in #256, because we can decide whether or not to call `cpp_set_stdlib` based on the tool family. Although it could be interesting to follow the `*_if_supported` pattern and avoid the need to check entirely.

alexcrichton:
Ok! I'll close this now that #256 is merged

PR #256, opened 2017-10-18
RFC: expose tool family

grovesNL:
As described in #254, [`try_get_compiler`](https://github.com/alexcrichton/cc-rs/blob/master/src/lib.rs#L904) doesn't seem to be very useful when called externally, because `Tool`'s fields are all private. This PR exposes the field `family` and enum `ToolFamily` so it can be used in a build script (i.e. in conjunction with `cpp_set_stdlib` or other tool specific flags).

I only expose `family` in this PR because that is immediately useful to me for my build script. If any other fields from `Tool`/`Error`/any other structs should be marked `pub`, let me know and I can update this PR.

Let me know what you think (or whether this should be handled another way entirely 😃)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1674,6 +1674,21 @@ impl Tool {
     pub fn env(&self) -> &[(OsString, OsString)] {
         &self.env
     }
+
+    /// Whether the tool is GNU Compiler Collection-like.
+    pub fn is_like_gnu(&self) -> bool {
+        self.family == ToolFamily::Gnu
+    }
+
+    /// Whether the tool is Clang-like.
+    pub fn is_like_clang(&self) -> bool {
+        self.family == ToolFamily::Clang
+    }
+
+    /// Whether the tool is MSVC-like.
+    pub fn is_like_msvc(&self) -> bool {
+        self.family == ToolFamily::Msvc
+    }
 }
 
 fn run(cmd: &mut Command, program: &str) -> Result<(), Error> {


alexcrichton:
Thanks for the PR! I'm not a huge fan of exposing this as I don't think we're really doing robust enough detection to say this is accurate right now, but could you elaborate on your use case for using it?

grovesNL:
@alexcrichton I'm trying to detect whether Clang is used to conditionally [pass `cpp_set_stdlib` in my build.rs](https://github.com/grovesNL/spirv_cross/blob/master/spirv_cross/build.rs). If I mistakenly pass this flag to gcc then my build fails. I'm using my fork of cc-rs to do this at the moment. I can hack in my own compiler detection as a workaround, or even copy the logic directly from cc-rs, but I was hoping to leverage the same detection logic as cc-rs (they're both used for flags). Otherwise `try_get_compiler` is not useful externally and should probably not be marked `pub`.

alexcrichton:
Hm ok, thanks for the extra info. I'm still very wary to do this, I fear that this sort of compiler detection is going down a road that basically won't end.

If we *were* to expose this information the API would also look quite different. I think we'd probably just have an `is_like_clang` method to the `Tool` struct most likely.

grovesNL:
`is_like_clang` would work great for me as well. I thought about something similar this morning although I wasn't sure if eventually this would expand to `is_like_gcc` etc. Alternatively if you're ok with exposing `ToolFamily` itself, we could do have `is_like_tool_family(family: ToolFamily)`

alexcrichton:
Nah I'd prefer to keep `ToolFamily` private for now (otherwise tough to add to in the future), but adding accessors seems fine!

grovesNL:
Sounds good. I updated my commit to instead add methods for GNU/Clang/MSVC. Please take a look. If you're not sure about the inclusion of GNU/MSVC for now, let me know and I can remove them.

alexcrichton:
Thanks!

grovesNL:
@alexcrichton Thanks for merging! Will you be publishing a new crate sometime soon?

alexcrichton:
Yeah I can do that now

PR #259, opened 2017-10-18
Separated CC/CFLAGS and command args, especially for "ccache" case.

not-fl3:
All apis and previous results of ```.path()``` and ```.args()``` appears unchanged, the only thing added - ```cc_env()``` and ```cflags_env``` Tool functions. This will work for backtrace-rs case. 

I am unsatisfied with introducing new vec of arguments, but run out of ideas how to avoid it :( 
 
What do you think about that methods? maybe that thing should be done in an entirely different way?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,6 +148,7 @@ impl From<io::Error> for Error {
 #[derive(Clone, Debug)]
 pub struct Tool {
     path: PathBuf,
+    path_args: Vec<OsString>,
     args: Vec<OsString>,
     env: Vec<(OsString, OsString)>,
     family: ToolFamily
@@ -1285,7 +1286,7 @@ impl Build {
             .map(|(tool, args)| {
                 let mut t = Tool::new(PathBuf::from(tool));
                 for arg in args {
-                    t.args.push(arg.into());
+                    t.path_args.push(arg.into());
                 }
                 t
             })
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! I think as well for backtrace-rs though we ideally need to change the return value of `Tool::path` as well though. I believe backtrace-sys guesses the name of the `objcopy` binary based on the name of the compiler binary. I think it shouldn't be too hard as well, right? I think that it's just the first element of `path_args` if present? (that field could also be renamed)

not-fl3:
Now Tool storing internally both compiler path and that cc-thing path - ccache for example. 
And return as .path() path to the compiler itself, while compilation command starts with ccache. 

So normal ```cc-rs``` users will have ccache enabled, and backtrace-rs should use ```.cc_env()``` and ```.cflags_env()``` for correct compilation CC and CFLAGS. Actually, even using ```.path()``` and ```.args()``` is correct - it will be just a normal compilator, without ccache.

I am not sure about naming and implementation details, but at least it works for me :)

alexcrichton:
Thanks and looks great! It looks like CI may be failing though?

Also, could the `cc_path` field be perhaps named `cc_wrapper_path`? (I think that's the wrapper around the compiler, right?

not-fl3:
I am not exactly sure how to fix CI. 
I have two ideas:
- add "ccache" to travis apt plugins and test how its compiling. But what shall we do on appveyor/windows?
- or test only simplest things without compiling, like I did in the last commit. I hope it will fix CI, but tests will not be really good.

alexcrichton:
Nah yeah these changes look good to me, in general most of the testing of this crate happens externally, so that's ok!

not-fl3:
Awesome! Will you publish a new crate sometime soon? Going to make a PR for backtrace-rs than.

alexcrichton:
Published!

Issue #196, opened 2017-07-13
Document the `rayon` feature


brson:
This crate can optionally enable `rayon`. Unfortunately we don't have guidelines on the proper way to do this.

I'd suggest adding a "Parallelism" section to crate docs, and then writing a short paragraph explaining how to turn on the feature and what it does. Depending on the outcome of https://github.com/alexcrichton/gcc-rs/issues/195 there might end up being other things to say too.

naoNao89:
The issue should be closed as completed because:

Existing documentation 
https://github.com/rust-lang/cc-rs/blob/9ec00e4bf2a8b087760245b0ec721fb3bd59731f/src/lib.rs#L133-L147

No rayon dependency
https://github.com/rust-lang/cc-rs/blob/9ec00e4bf2a8b087760245b0ec721fb3bd59731f/Cargo.toml#L33

NobodyXu:
Thanks! I'm closing this

Issue #195, opened 2017-07-13
Consider how to expose parallelism


brson:
From @briansmith:

> Note that gcc-rs has a way of doing parallel building, but we were not able to use it efficiently because we had other build steps that we also wanted to run in parallel, and gcc-rs's parallel build support doesn't compose well with additional build parallelism implemented in the build.rs. It would be good to factor parallel build support and dependency tracking out into a more general (not just asm/C/C++) library.

This seems like an important consideration (I didn't even realize there was an optional rayon feature). What should we do here?

At the least we could add a `num_jobs` knob so the user can deal with it themselves. This issue of coordinating build system parallelism though is quite cross-cutting. Whatever the ultimate solution here is seems like it will entail interaction between make, cargo, this crate, the cmake crate, etc.

Needs more design work before we take action.

brson:
cc @sfackler @burntsushi

sfackler:
The new jobserver stuff in Cargo might help here right?

alexcrichton:
Perhaps yeah, it'd want to integrate with that, but it wouldn't necessarily be an automatic interface to expose. Sort of what needs to be returned here is futures almost?

alexcrichton:
This was done with jobserver integration in https://github.com/alexcrichton/cc-rs/pull/438

PR #260, opened 2017-10-19
Docs for parallelism - Addresses #195 and #196

rrichardson:
These tasks were in the Crate Evaluation Tracking Issue and the issue was closed.  Perhaps because it was deemed that the README contained enough info regarding parallelism.  Either way I've gone ahead and added a bit more. 

This just adds a basic write-up on how to enable parallelism in top level module docs. It also adds one minor categorical hint in the README.  

Also, apologies for the rabid rustfmt spam.  I can figure out how to undo that if it is unwelcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,8 @@ and `HOST` variables.
 
 ## Optional features
 
+### Parallel
+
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

PR #261, opened 2017-10-19
Fix build parallelization

rrichardson:
The compile_object step was occurring within the scope of the mutex lock, so rayon wasn't actually able to execute in parallel.  This change just moves the compile out of the scope of the lock then pushes the result. 

Tested on librocksdb-sys which is a beast. Works quite well now.


rrichardson:
Not sure why Travis is spinning up duplicate jobs for each arch and then stalling on them. Did that for the previous PR as well.

alexcrichton:
Heh, thanks!

PR #262, opened 2017-10-23
Add sparc64-unknown-linux-gnu cross-compile target

glaubitz:
Hi Alex!

This small change adds the cross-compile target sparc64-linux-gnu. This is necessary to resolve issue https://github.com/rust-lang/rust/issues/45456 in Rust.

Thanks!


alexcrichton:
Thanks!

PR #272, opened 2017-12-08
Add aarch64-linux-musl

asonix:
This one-line change fixes my build so here you go.

In the future, it would be nice if cc-rs could read a config file for paths to binaries, but this fixes my builds


alexcrichton:
Thanks!

PR #275, opened 2017-12-26
Pick the right compiler for CloudABI.

EdSchouten:
CloudABI is a sandboxed UNIX-like runtime environment. For CloudABI, the
compiler is always named <target>-cc or <target>-c++. The target name is
never mangled. As we use Clang, not GCC, we use -cc and -c++, as opposed
to -gcc and -g++.


alexcrichton:
Thanks!

PR #278, opened 2018-01-01
Don't treat 'x86_64-unknown-cloudabi-cc' as being MSVC.

EdSchouten:
Though the name contains 'cl', it should not be treated as Microsoft's
cl.exe.


alexcrichton:
Thanks!

PR #279, opened 2018-01-03
Make is_flag_supported public

### LINKED ISSUE #277: Expose is_flag_supported
Would it be possible to expose `is_flag_supported` in the public API?

I want to set `-march={}`, but the compiler may be older than the desired target architecture and reject it. I can't just use `flag_if_supported`, because if the flag is rejected, I don't want to ignore it. I want to set other flags as a fallback instead, and the fallback flags shouldn't be set unconditionally. (related to #268)

kornelski:
Closes #277

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -391,7 +391,11 @@ impl Build {
         Ok(src)
     }
 
-    fn is_flag_supported(&self, flag: &str) -> Result<bool, Error> {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Could this also expand a bit about why it returns a `Result`, for example missing comiplers, intermittent I/O errors, etc.

kornelski:
Added

PR #282, opened 2018-01-08
Cache is_flag_supported

Mrmaxmeier:
Hey,

recently I've investigated been annoyed by slow build times of [tectonic-typesetting/tectonic](https://github.com/tectonic-typesetting/tectonic).
The project compiles 95 [object files](https://github.com/tectonic-typesetting/tectonic/blob/master/build.rs#L201) with and sets about 25 build flags using `flag_if_supported`.
Even with <50ms per internal call to `is_flag_supported` this outweighs the actual time spent compiling the object files.

In the case of `tectonic` this change reduces the build_script phase from 90s to 18s.
I'm not particularly satisfied with my implementation though as it's pretty hacky ~and pollutes the _astonishingly_ clean set of dependencies~.
_Alternatively one could fix the problem in the build script by using the recently exposed `is_flag_supported`._

If you'd prefer a different solution to this problem, let me know!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -67,9 +67,9 @@ use std::path::{PathBuf, Path};
 use std::process::{Command, Stdio, Child};
 use std::io::{self, BufReader, BufRead, Read, Write};
 use std::thread::{self, JoinHandle};
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
 
-#[cfg(feature = "parallel")]
-use std::sync::Mutex;
 
 // These modules are all glue to support reading the MSVC version from
 // the registry and from COM interfaces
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Could this cache perhaps be inside `Config` rather than global?

Mrmaxmeier:
Sure! This also gets rid of the additional dependency.
Travis seems to be a bit flaky at the moment.

alexcrichton:
Thanks! I think the `Arc` and `Mutex` may not be necessary here though, perhaps just `RefCell`?

alexcrichton:
Ok no worries makes sense to me!

PR #286, opened 2018-02-18
Microsoft ARM/ARM64 compiler support

makotokato:
VS2017 supports upcoming ARM64 desktop Windows support.  But cc-rs does't support it yet, so I would like to support it.

armasm and armasm64 don't support -c options and thes uses -o for output file option.


alexcrichton:
Thanks!

PR #288, opened 2018-02-27
Only pass explicitly specified custom flags to compiler

peterhj:
This adds a function `explicit_flags_only` to `Build`. Enabling this option prevents `Build` from adding implicit flags to the compiler command. Instead, only flags specified through the `flag` function will be added (as well as `-o` before the output, which is a common enough idiom).

The main motivation is to provide one-off support for builds using certain non-standard but still somewhat common tools, e.g. assembling with NASM/YASM.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,6 +95,7 @@ pub struct Build {
     include_directories: Vec<PathBuf>,
     definitions: Vec<(String, Option<String>)>,
     objects: Vec<PathBuf>,
+    explicit_flags: bool,
     flags: Vec<String>,
     flags_supported: Vec<String>,
     known_flag_support_status: Arc<Mutex<HashMap<String, bool>>>,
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Would this perhaps be better managed though through the `Build::object` method with custom-compiled objects?

peterhj:
It looks like `Build::object` currently just takes in a path to the object, so the compilation would still have to be done separately. How about an API like `Build::custom_object(obj: Path, src: Path, custom_cmd: Vec<&str>)`?

alexcrichton:
That's possible yeah but in theory this is a totally custom object, right? Could the nasm/yasm build happen in your build script? (you can also spawn processes...)

peterhj:
The actual library I'm building is libjpeg-turbo, which has a nice %age of simd assembly sources. What I was hoping to do was to reuse the part of `Build` that lets you just specify source files and not worry too much about paths or running your own commands, sort of treating `Build` as a make-like API.

alexcrichton:
Sure yeah, but it seems like you're running a custom tool and you're not using almost any of the logic in this crate, so the only real benefit here is reducing the boilerplate of using `Command`?

peterhj:
Sorry this dropped off my radar; closing as it's better to wrap `Command` as suggested

PR #293, opened 2018-03-03
Pass lib.exe arguments though a file when the arg list is too long.

### LINKED ISSUE #292: "Argument list too long" with many files on Windows
https://github.com/servo/mozangle/tree/ee3f46cc4e6a990634d3cab4d49d1ebfdb785d22 uses the `cc` crate, and `cargo test` works on both Linux and Windows for me.

When adding `--features egl` however, compilation fails on Windows with the error below. This feature is Windows-only, but as far as the `cc` crate is concerned it mostly adds a bunch more source files.

> Internal error occurred: Command "C:\\[…]\\lib.exe" [many arguments…] with args "lib.exe" failed to start.

When I copy/paste the command into a shell, the output is:

> Argument list too long

Could this crate do something like https://github.com/rust-lang/rust/pull/47507 ?

SimonSapin:
Fixes #292

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1331,12 +1331,46 @@ impl Build {
                     )
                 }
             };
+
             let mut out = OsString::from("/OUT:");
             out.push(dst);
-            run(
-                cmd.arg(out).arg("/nologo").args(&objects).args(&self.objects),
-                "lib.exe",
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #294, opened 2018-03-08
Some tweaks for iOS

### LINKED ISSUE #291: Don't pass -ffunction-sections or -fdata-sections on Darwin targets
On GNU targets, these flags are necessary to allow the linker to strip unused symbols.  On Darwin, they aren't necessary for that, and in fact don't seem to have any effect.  Normally they're harmless, but if the user passes `-fembed-bitcode` in CFLAGS, Xcode's version of clang gets upset:

```
clang: error: -ffunction-sections is not supported with -fembed-bitcode
clang: error: -fdata-sections is not supported with -fembed-bitcode
```

Thus, it would be best to simply omit these flags on Darwin.

alexcrichton:
* Don't pass `-ffunction-sections -fdata-sections` by default
* Pass `-fembed-bitcode` unconditionally
* Pass `-fembed-bitcode-marker` in debug mode

Closes #291

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1082,8 +1082,10 @@ impl Build {
                     cmd.args.push(format!("-O{}", opt_level).into());
                 }
 
-                cmd.push_cc_arg("-ffunction-sections".into());
-                cmd.push_cc_arg("-fdata-sections".into());
+                if !target.contains("-ios") {
+                    cmd.push_cc_arg("-ffunction-sections".into());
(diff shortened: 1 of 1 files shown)


alexcrichton:
@comex mind taking a look at this to make sure it's right?

cc @michaeleiselsc

comex:
Looks good, though I haven't tried testing it.

alexcrichton:
Ok!

PR #304, opened 2018-03-15
Ensure supported flags are resolved before compilation starts.

### LINKED ISSUE #303: Supported flags are nondeterministic when using parallel compilation
Imagine something like:
```rust
let cfg = cc::Build::new();
cfg.file("file1.cpp");
cfg.file("file2.cpp");
cfg.flag_if_supported("-std=gnu++11");
cfg.cpp(true);
cfg.compile();
```
With parallel compilation in use, two threads will be spun up that proceed to call `compile_object`. This invokes `try_get_compiler`, which fetches the flags stored in the `flags_supported` vector and calls `is_flag_supported` for each one. If the second thread enters `is_flag_supported` before the first thread has stored the result of compilation in the `known_flag_support_status` hashtable, the check will be run twice. `is_flag_supported` creates the `flag_check` file, but first checks if it exists. If both threads pass that check and attempt to create the file, one thread will lose and pass the error back up the stack, where it will be interpreted as meaning that the flag is not supported.

There are a couple ways this could be resolved:
* resolve all flag support checks before starting parallel compilation
* perform one single compilation completely before performing the remainder in parallel
* don't fail the flag check on that particular error
* extend the critical in which the hashtable is locked to the entirely of `is_flag_supported`

jdm:
This removes all shared mutable state from the actual process of compilation, and sequentializes some operations that can otherwise be nondeterministic because of TOCTTOU. Fixes #303.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -67,8 +67,6 @@ use std::path::{Path, PathBuf};
 use std::process::{Child, Command, Stdio};
 use std::io::{self, BufRead, BufReader, Read, Write};
 use std::thread::{self, JoinHandle};
-use std::collections::HashMap;
-use std::sync::{Arc, Mutex};
 
 // These modules are all glue to support reading the MSVC version from
 // the registry and from COM interfaces
@@ -96,7 +94,6 @@ pub struct Build {
     objects: Vec<PathBuf>,
     flags: Vec<String>,
     flags_supported: Vec<String>,
-    known_flag_support_status: Arc<Mutex<HashMap<String, bool>>>,
     files: Vec<PathBuf>,
     cpp: bool,
     cpp_link_stdlib: Option<Option<String>>,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -881,12 +874,24 @@ impl Build {
     /// Panics if `output` is not formatted correctly or if one of the underlying
     /// compiler commands fails. It can also panic if it fails reading file names
     /// or creating directories.
-    pub fn compile(&self, output: &str) {
+    pub fn compile(&mut self, output: &str) {
```
SimonSapin: Unfortunately taking `&mut self` here is a breaking change (even if I suspect it won’t affect many if any users in practice). Could `resolve_supported_flags` return a new Vec instead of mutating the existing one?

alexcrichton:
Thanks! I agree with @SimonSapin though in that this unfortunately is a breaking change, but perhaps some `Mutex`-like synchronization could be helpful to resolve the flags up front and then use them afterwards?

src/lib.rs
jdm: This suggestion ended up requiring even greater interface changes, so I've abandoned it.

jdm:
Closing in favour of #306.

PR #305, opened 2018-03-16
iOS embed bitcode fix, with testcase

kali:


kali:
Bug manifests with the first commit, see here:

https://travis-ci.org/snipsco/cc-rs/jobs/354228120#L231

alexcrichton:
Awesome, thanks!

volodg:
Fix does not work for `arm` arch :-(

PR #306, opened 2018-03-16
Extend critical section to encompass full supported flags check.

### LINKED ISSUE #303: Supported flags are nondeterministic when using parallel compilation

jdm:
Resolving whether a flags is supported by the compiler needs to be an atomic operation because it interacts with the filesystem. This change extends the scope of the mutex to cover the entire operation, ensuring that all threads will see a consistent version of the hashtable storing supported flags. Fixes #303.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -397,11 +397,10 @@ impl Build {
     /// `known_flag_support` field. If `is_flag_supported(flag)`
     /// is called again, the result will be read from the hash table.
     pub fn is_flag_supported(&self, flag: &str) -> Result<bool, Error> {
-        let known_status = self.known_flag_support_status
+        let mut known_status = self.known_flag_support_status
             .lock()
-            .ok()
-            .and_then(|flag_status| flag_status.get(flag).cloned());
-        if let Some(is_supported) = known_status {
+            .unwrap();
+        if let Some(is_supported) = known_status.get(flag).cloned() {
             return Ok(is_supported);
         }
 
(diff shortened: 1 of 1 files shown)


jdm:
This is an alternative solution to #304 that does not require any public interface changes. My attempts to remove the mutability requirements from the other PR were leading me in a direction similar to this one, and this change is about as small as it can be.

alexcrichton:
Thanks!

PR #307, opened 2018-04-01
Pass -static depending on crt-static on linux too.

mixi:
This is the done the same way as it's already done for msvc.

This is one piece still missing to compile rustc on a musl host with internal llvm (the llvm build process builds at least one shared library, LLVMHello.so, that fails to link with -static in CFLAGS).


alexcrichton:
Thanks!

PR #311, opened 2018-04-13
Allow overwriting the default cpp_link_stdlib

### LINKED ISSUE #310: The automatic stdc++ linking makes it impossible to statically link stdc++ on Android
In order to statically link the standard C++ library, I pass `-lc++_static` to the linker by tweaking my `.cargo/config` file, like this:

```rust
[target.armv7-linux-androideabi]
linker = "arm-linux-androideabi-clang"
ar = "arm-linux-androideabi-ar"
rustflags = ["-C", "link-arg=-lc++_static", "-C", "link-arg=-lc++abi", "-C", "link-arg=-landroid_support"]

[target.arm-linux-androideabi]
linker = "arm-linux-androideabi-clang"
ar = "arm-linux-androideabi-ar"
rustflags = ["-C", "link-arg=-lc++_static", "-C", "link-arg=-lc++abi", "-C", "link-arg=-landroid_support"]
```

However `cc-rs` automatically adds `-l stdc++` at the end of the arguments because of the `cpp_link_stdlib()` method.
For an undetermined reason, this `-l stdc++` flag seems to have taken priority over `-l c++_static`, which means that compiled libraries now depend on `c++_shared`. The Android build of Parity has recently broken because of that.

Ideally there should be a way to overwrite the default value for `cpp_link_stdlib()` from the `.cargo/config.toml`, however that's not really possible.

Do you have any suggestion?

tomaka:
Close #310 

Adds an environment variable named `CXXSTDLIB` that can be used to overwrite the automatically detected C++ standard library.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -582,6 +582,8 @@ impl Build {
     /// The default value of this property depends on the current target: On
     /// OS X `Some("c++")` is used, when compiling for a Visual Studio based
     /// target `None` is used and for other targets `Some("stdc++")` is used.
+    /// If the `CXXSTDLIB` environment variable is set, its value will
+    /// override the default value.
     ///
     /// A value of `None` indicates that no automatic linking should happen,
     /// otherwise cargo will link against the specified library.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #315, opened 2018-05-01
Use lowercase for clang-cl arch argument

mati865:
It doesn't accept uppercase argument.
Reference: https://clang.llvm.org/docs/UsersManual.html#id8

Bug exposed by https://github.com/rust-lang/rust/pull/50200

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1126,7 +1126,7 @@ impl Build {
             }
             ToolFamily::Msvc { clang_cl } => {
                 if target.contains("i586") {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! Unfortunately this wasn't quite sufficient as for clang-cl it looks like you *also* need to pass in `-m32`, which means a [bit more restructuring needed to happen here](https://github.com/alexcrichton/cc-rs/commit/f9c9f648ee28929c4ec3fd86601d99d8ae068dd7), but hopefully that'll help the Rust PR land!

PR #320, opened 2018-05-28
Add new cppwinrt/ directory to SDK includes for Windows 10

bwalderman:
C++/WinRT is included in the Windows SDK starting in version 10.0.17134.0 (Windows 10, version 1803). The header files are located in the "cppwinrt" directory under the SDK path. This change adds "cppwinrt" to the include path so that the tool can compile libs that want to use C++/WinRT.

Users will also need to add the "/std:c++17" flag when compiling since C++/WinRT require C++17.


alexcrichton:
Thanks!

PR #324, opened 2018-06-22
Don't override host when checking supported flags.

jdm:
This has an impact on cross-compilation, since it affects whether HOST_CFLAGS/HOST_CXXFLAGS/HOST_CC/HOST_CXX will be selected and can impact whether a compilation succeeds or not.


alexcrichton:
👍

PR #326, opened 2018-06-28
special case for devenv.exe

Geobert:
Fix https://github.com/alexcrichton/cc-rs/issues/325

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -638,7 +638,6 @@ mod impl_ {
             .and_then(|key| key.query_str("15.0").ok())
             .map(|path| {
                 let path = PathBuf::from(path).join(r"Common7\IDE\devenv.exe");
-                println!("*** path: {:?}", path);
                 let mut tool = Tool::new(path);
                 if target.contains("x86_64") {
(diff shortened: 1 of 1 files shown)


src/windows_registry.rs
```diff
@@ -622,6 +626,27 @@ mod impl_ {
...
+
+    fn find_devenv_vs15(target: &str) -> Option<Tool> {
+        let key = r"SOFTWARE\WOW6432Node\Microsoft\VisualStudio\SxS\VS7";
+        LOCAL_MACHINE
+            .open(key.as_ref())
+            .ok()
+            .and_then(|key| key.query_str("15.0").ok())
+            .map(|path| {
+                let path = PathBuf::from(path).join(r"Common7\IDE\devenv.exe");
+                println!("*** path: {:?}", path);
```
alexcrichton: stray println?

alexcrichton:
Thanks!

src/windows_registry.rs
Geobert: oops, fixed ^^

Geobert:
Note that it's only for VS2017, if we need other version as well, I need to figure out their path and code the missing parts :)

PR #333, opened 2018-08-08
Filter out -v from compiler args when checking if a flag is supported.

### LINKED ISSUE #329: Strip out -v from compile flags when checking if a flag is supported
If `-v` is present in CFLAGS/CXXFLAGS during the compilation check to see if a given flag is supported, this makes `clang` add information to stderr that is not actually an error. This means that we get erroneous and extremely confusing results when usually the presence of `-v` means that we're trying to debug something in a native build.

jdm:
Fixes #329. I didn't write an automated test because I didn't see a way to inject the args outside of environment variables, and doing that in unit tests that run concurrently seems dangerous.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 
 matrix:
   include:
-    - rust: 1.13.0
+    - rust: 1.16.0
       install:
       script: cargo build
     - rust: stable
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Looks like CI may need a slight tweak though? Feel free to bump the minimum version here as well!

jdm:
1.16 is the first Rust release that can successfully compile these changes.

alexcrichton:
Thanks!

apoelstra:
It should really be a major version bump to break compilation.

PR #334, opened 2018-08-09
Use clang for target android

edelangh:
android ndk r18 remove gnu support, we should use clang instead of gcc

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1549,10 +1549,10 @@ impl Build {
         }
         let host = self.get_host()?;
         let target = self.get_target()?;
-        let (env, msvc, gnu, traditional) = if self.cpp {
-            ("CXX", "cl.exe", "g++", "c++")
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! In order to keep supporting older SDKs though (which rust-lang/rust may be using?) could this default to using clang but use gcc as a fallback if it's not detected?

edelangh:
I've updated my commit.
It will check for gnu first as it always has done

alexcrichton:
Thanks! Could the rustfmt changes be left for a later time as well? Additionally, could the logic of replacing strings in the target be deduplicated between the gcc/clang case?

edelangh:
Ok sorry for fmt, I used `git add -u` because of time :/
Thank you for seeing the duplication, everything should be right now

alexcrichton:
Thanks!

Issue #336, opened 2018-08-21
Doesn't build on Rust 1.14.0


jeandudey:
Error:

```rust
error[E0308]: mismatched types
    --> /cc-1.0.19/src/lib.rs:2021:18
     |
2021 |         cmd.args(self.args.iter().filter(|a| !self.removed_args.contains(a)));
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected &[_], found struct `std::iter::Filter`
     |
     = note: expected type `&[_]`
     = note:    found type `std::iter::Filter<std::slice::Iter<'_, std::ffi::OsString>, [closure@/cc-1.0.19/src/lib.rs:2021:42: 2021:76 self:_]>`
```

This library is used by the [`rust-bitcoinconsensus`](https://github.com/rust-bitcoin/rust-bitcoinconsensus) which has an strict policy on rustc versions used, this project defaults to use Rust 1.14.0 because it's the only one that is shipped on Debian Stable (`i686`, `arm64`). The main concern for using a fixed rustc version is for deterministic builds and such things.

alexcrichton:
Thanks for the report! Right now this crate doesn't promise anything other than the 2 latest stable releases, but PRs are welcome to add compatibility to older versions where possible!

jeandudey:
The `Command::args` signature has changed since 1.14.0, isn't it a breaking change in the `std` library?

PR #337, opened 2018-08-22
Compiling with Rust 1.14.0

crw5996:
Regarding issue #336... 

The following changes allow for cc-rs to be built on Rust 1.14.0, however it fails to pass "cargo test" on Rust 1.14.0. These changes build and test successfully on the most recent version or Rust.

It seems like the macro #![cfg_attr(test, deny(warnings))] overshadows #[allow(non_camel_case_types)] under "__Nonexhaustive_do_not_match_this_or_your_code_will_break".

I'm not quite sure what is the best way to approach this issue, as it seems like any changes could break things for the more recent versions of Rust.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2019,8 +2019,8 @@ impl Tool {
         };
         cmd.args(&self.cc_wrapper_args);
 
-        let value = self.args.iter().filter(|a| !self.removed_args.contains(a)).map(|a| a).collect::<Vec<_>>();
-        cmd.args(&value[..]);
+        let value = self.args.iter().filter(|a| !self.removed_args.contains(a)).collect::<Vec<_>>();
+        cmd.args(&value);
 
         for &(ref k, ref v) in self.env.iter() {
             cmd.env(k, v);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2018,7 +2018,10 @@ impl Tool {
             None => Command::new(&self.path),
         };
         cmd.args(&self.cc_wrapper_args);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2018,7 +2018,10 @@ impl Tool {
             None => Command::new(&self.path),
         };
         cmd.args(&self.cc_wrapper_args);
-        cmd.args(self.args.iter().filter(|a| !self.removed_args.contains(a)));
+
+        let value = self.args.iter().filter(|a| !self.removed_args.contains(a)).map(|a| a).collect::<Vec<_>>();
```
alexcrichton: I think the `.map(|a| a)` shouldn't be needed here?

src/lib.rs
```diff
@@ -2018,7 +2018,10 @@ impl Tool {
...
+        cmd.args(&value[..]);
```
alexcrichton: I believe the `[..]` here should be superfluous

apoelstra:
Since this project's policy is only to support the last couple of rustc versions, and downstream users don't care if the tests compile or not, maybe it's OK if building works on 1.14 but tests don't?

Though this would mean not Travis-testing 1.14 so future breakage wasn't noticed, which is unfortunate.

alexcrichton:
Thanks @crw5996!

@apoelstra I do not personally wish to invest the effort in maintaining support for versions as old as 1.14 (e.g. with CI). I find it far too onerous myself, but I'm more than willing to take PRs like this.

PR #342, opened 2018-08-27
don't enable warnings by default if CFLAGS/CXXFLAGS defined in environment

mykmelez:
If CFLAGS/CXXFLAGS is defined in the environment, then presumably the consumer has already determined the set of warnings flags to pass to the compiler, so we shouldn't enable warnings by default in that case.

I'd hoped to use this change to help address [bug 1482810](https://bugzilla.mozilla.org/show_bug.cgi?id=1482810), but I'm not sure it'll actually work for that, since [this Makefile snippet](https://searchfox.org/mozilla-central/rev/55da592d85c2baf8d8818010c41d9738c97013d2/config/rules.mk#908-939) shows that Mozilla's build system doesn't pass CFLAGS/CXXFLAGS to Cargo on Windows, so this change won't disable warnings by default there.

It might still be worth taking, which is why I'm opening this pull request (while I continue to look for a more robust and general solution in that bug).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,8 +116,8 @@ pub struct Build {
     shared_flag: Option<bool>,
     static_flag: Option<bool>,
     warnings_into_errors: bool,
-    warnings: bool,
-    extra_warnings: bool,
+    warnings: Option<bool>,
+    extra_warnings: Option<bool>,
 }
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks good to me, thanks!

PR #346, opened 2018-09-06
msvc: find tools correctly when target is thumbv7a

jordanrh1:
This change enables cc-rs to find the MSVC tools when the target architecture is thumbv7a, which is the target architecture for windows/ARM.


alexcrichton:
Thanks!

PR #349, opened 2018-09-19
Msvc/Arm: Pass /D_ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE=1 to cl.exe

jordanrh1:
This flag must be defined to avoid a compilation error.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1159,6 +1159,19 @@ impl Build {
                         cmd.args.push("/ARCH:IA32".into());
                     }
                 }
+
+                // There is a check in corecrt.h that will generate a
+                // compilation error if
+                // _ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE is
+                // not defined to 1. The check was added in Windows
+                // 8 days because only store apps were allowed on ARM.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Is there perhaps some documentation we could point to to say why we're passing this flag?

jordanrh1:
The documentation is the source code :) From corecrt.h:

```
// Verify that the ARM Desktop SDK is available when building an ARM Desktop app
#ifdef _M_ARM
    #if _CRT_BUILD_DESKTOP_APP && !_ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE
        #error Compiling Desktop applications for the ARM platform is not supported.
    #endif
#endif
```

The check was added in Windows 8 days because only store apps were allowed on ARM, but this changed with the release of Windows 10 IoT Core. This check will be going away in future versions of the SDK, but for all released versions of the Windows SDK it is required.

alexcrichton:
Ok cool! Could you add that all in a comment next to the source here?

jordanrh1:
Sure, done.

alexcrichton:
Thanks!

PR #351, opened 2018-09-27
Add support for `-fno-plt`

GabrielMajeri:
Similar to rust-lang/rust#54592, this PR enables the `-fno-plt` compiler flag by default when PIC is used. Should provide a small performance improvement with no obvious downsides.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -112,6 +112,7 @@ pub struct Build {
     archiver: Option<PathBuf>,
     cargo_metadata: bool,
     pic: Option<bool>,
+    use_plt: Option<bool>,
     static_crt: Option<bool>,
     shared_flag: Option<bool>,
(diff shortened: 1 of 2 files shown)


GabrielMajeri:
Uh, I've run into a problem: how do I check if `flag_is_supported` from inside `try_get_compiler`? Because it seems it causes an infinite loop.

alexcrichton:
Hm so unfortunately `is_flag_supported` is somewhat expensive, could this perhaps default to off for now?

GabrielMajeri:
@alexcrichton Ok, I've made it so that this flag is not enabled by default. Another alterantive would've been to look at GCC's / Clang's version directly, and hard-code a version check, but it looks like we don't retrieve the compiler's version at any point.

alexcrichton:
Ok thanks!

PR #358, opened 2018-10-28
exclude CI files

ignatenkobrain:


alexcrichton:
FWIW this isn't the first time I've seen this PR throughout the ecosystem, and I don't really understand why this is being done. It would be helpful to fill out the PR description with motivation for why this is done in case anyone wants to track this back.

ignatenkobrain:
@alexcrichton well, the distributions are shipping source code from crates.io and when that bloody thing contains  CI specific files -- we have to develop some downstream code.

alexcrichton:
@ignatenkobrain er... ok, thanks for the explanation. Would it be possible though to not use the term "bloody thing" when referring to crates that users maintain? We're just trying to provide good-quality crates to the ecosystem?

ignatenkobrain:
@alexcrichton yeah, sure. I just meant that I was so tired of doing this for 90 crates I was touching recently. And more are coming ;)

ignatenkobrain:
Didn't mean that crates / ecosystem is bad

froydnj:
> well, the distributions are shipping source code from crates.io and when that bloody thing contains CI specific files

Why does this matter, though?  We vendor crates via crates.io that contain CI-specific files without any problems.  Is the concern over the space that they consume (we've certainly asked crates to make changes to exclude particularly large test-only files) or something else?

PR #361, opened 2018-12-12
Generate valid intructions on ARMv6 (see rust-lang/rust issue #50583)

piersfinlayson:
I have tested this by generating a test C program containing an ARMv6 illegal instruction as follows:

```
void main()
{
  asm("vmov.f64        d10, #1.0");
}
```

Then compiling it via the cc crate - without this fix the program compiles find (and shouldn't).  After the fix the compilation fails (as it should).  Obviously this is a slightly contrived scenario, but couldn't immediately invent some C which would end up generating the illegal instruction.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1249,6 +1249,7 @@ impl Build {
                 if target.starts_with("arm-unknown-linux-") {
                     cmd.args.push("-march=armv6".into());
                     cmd.args.push("-marm".into());
+                    cmd.args.push("-mfpu=vfp".into());
                 }
 
                 // We can guarantee some settings for FRC
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1249,7 +1249,11 @@ impl Build {
                 if target.starts_with("arm-unknown-linux-") {
                     cmd.args.push("-march=armv6".into());
                     cmd.args.push("-marm".into());
-                    cmd.args.push("-mfpu=vfp".into());
+                    if target.ends_with("hf") {
+                        cmd.args.push("-mfpu=vfp".into());
+                    } else {
+                        cmd.args.push("-mfloat-abi=soft".into());
+                    }
                 }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Should this only be passed when the target doesn't end with `hf` as well?

piersfinlayson:
Don't think so - hf is when you need it.  Should probably pass -mfloat-abi=soft if it doesn't end in hf, otherwise it's likely to break soft floating point implementations.  (The other alternative is to leave non hf as is.)

alexcrichton:
Ah ok! Should this only be passed in the presence of that flag then? The arm flags are sort of all over the place, but I wouldn't be surprised if we're not passing mfloat-abi=soft and we should be!

piersfinlayson:
I haven't quite got my head around what compiler will/won't/may be used in what cases, but it seems to me it would be viable to have a single arm cross compiler which supports multiple arm versions, and hf/not-hf and to use that for each of the different arm targets.

In that case you'd need the -mfloat-abi=soft (because I know that the gcc armv6 hf compiler will compile with hard float enabled by default - and as we now know if will even use an armv6 incompatible set of instructions unless given -mfpu=vfp!).

However, if you're using a compiler specifically built for one target (say armv6, non-hf) then I don't think you need -mfloat-abi=soft.

For my own uses I'm tending towards a single ARM cross compiler which requires the correct and complete set of gcc flags to be passed, so prefer explicit setting the -mfloat-abi=soft.  Before I add that to this PR I want to do some testing of it though, to be fairly sure it isn't going to break anything pre-existing.

alexcrichton:
Ok sure, testing sounds good! And yeah while I know the default build of gcc can often affect lots of flags and things it's probably best that we pass as many flags as we know of in this crate to ensure that we ignore the defaults and select what the rust target matches up with

alexcrichton:
👍

PR #363, opened 2018-12-18
Add the Armv8-M targets

hug-dev:
The first commit is to add support to the Arm targets that were recently added in Rust.
The second one, I am less sure about, is to replace a future deprecating call. I faced a compilation error about that deprecation problem when trying to produce the artifacts for the targets using `./x.py dist --target thumbv8m.[].-none-eabi[hf]`. Maybe this last commit is not needed in which case I am happy to remove it.

This pull request is mainly made to be able to add those targets in the Rust CI to make those artifacts available for everybody, as `cc-rs` is used to assemble/compile the files in `compiler-builtins`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 
 matrix:
   include:
-    - rust: 1.16.0
+    - rust: 1.30.0
       install:
       script: cargo build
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks!

PR #370, opened 2019-01-31
Use --target with non x86/x86_64 clang-cl targets

glandium:


glandium:
As @froydnj has mentioned in https://phabricator.services.mozilla.com/D18186, there hasn't been a release in the past month and a half. Could you do a new release (including this fix)?

alexcrichton:
Certainly! Published now

Issue #371, opened 2019-01-31
Allow to skip the automatic addition of compiler flags


glandium:
External build systems like Firefox's may want to pass down its own flags for target, arch, ABI and more, and all they can do at the moment is pass CFLAGS/CXXFLAGS, but then cc-rs goes on to add its own flags after those, which may override what the explicit CFLAGS/CXXFLAGS wanted. One example is that we're passing the correct target flags for clang-cl, but cc-rs was adding -m32 for --target aarch64-windows-msvc, effectively making it a 32-bits arm build (which failed in some ucrt header complaining that ARM desktop builds are not allowed). Some other flags that have an impact on ABI (fpu, etc.) could be at odds with what the top-level build system really wants.
So I think it should be possible to bypass the entire section that deals with adding platform-specific flags .

alexcrichton:
I think we could go either way on this because the `cc` crate shouldn't add any flags that actually conflict with other flags, modulo bugs. It's a bug that `-m32` was passed for MSVC as this crate only handled x86 and x86_64 there.

This could be useful as a bug prevention mechanism but this crate is already *very* carefully constructed and complicated, so I'd sort of prefer that it sticks to passing a minimal set of working flags and then fix bugs as they arise.

alexcrichton:
> that are added and are at odds with what we'd actually want for Firefox

I think in that case you want a different target. The purpose of the `-march` style flags is to match the Rust code. If you want different flags, then you likely want to compile the Rust code differently.

> As a matter of fact, the flags that cc-rs adds are also at odds with how rust's target itself might be configured through -C target_feature.

Yes and there's various bugs open to read flags like this and process them in `cc`. It's obviously a nontrivial undertaking. That's where I think an escape hatch can possibly work, but I think it's worthwhile to acknowledge it's a hack of an escape hatch that should not be necessary, it's only because work isn't being diverted elsewhere.

GeorgeHahn:
This should be fixed in version 1.0.30 - the presence of `-m` flags in `CFLAGS/CXXFLAGS` will automatically disable the default flags. If this is insufficient, the environment variable `CRATE_CC_NO_DEFAULTS` may be set to manually disable default flags.

alexcrichton:
I'm going to close this since various changes have happened internally here and for entirely skipping flags compilation can always be done in build scripts themselves instead of using this crate.

PR #374, opened 2019-02-21
Avoid flag conflicts and allow default flag generation to be disabled

GeorgeHahn:
This PR addresses https://github.com/alexcrichton/cc-rs/issues/309 and https://github.com/alexcrichton/cc-rs/issues/371.

The default flags may not be appropriate for all possible cross-compiling scenarios. Many flag conflicts can be caught automatically by checking `CFLAGS`/`CXXFLAGS` and disabling default flags if necessary. Some conflicts may not be caught automatically - to address this, an environment variable is added to disable default flags. This variable is (arbitrarily named & quickly chosen) `CRATE_CC_NO_DEFAULTS`.

A special case is added for optimization flags. No optimization flag is added if one is set in `CFLAGS`/`CXXFLAGS`. With the current implementation/code organization, an optimization flag is not added when default flags are disabled.

This is a first pass - I'll address changes & rebase as necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ number of different environment variables.
          certain `TARGET`s, it also is assumed to know about other flags (most
          common is `-fPIC`).
 * `AR` - the `ar` (archiver) executable to use to build the static library.
+* `CRATE_CC_NO_DEFAULTS` - the default compiler flags may cause conflicts in some cross compiling scenarios. Setting this variable will disable the generation of default compiler flags.
 
 Each of these variables can also be supplied with certain prefixes and suffixes,
 in the following prioritized order:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1098,8 +1098,13 @@ impl Build {
             cmd.push_cc_arg(arg.into());
         }
 
-        // Disable default flag generation via environment variable
-        let use_defaults = self.getenv("CRATE_CC_NO_DEFAULTS").is_none();
+        // Disable default flag generation via environment variable or when
+        // certain cross compiling arguments are set
+        let use_defaults = self.getenv("CRATE_CC_NO_DEFAULTS").is_none()
+            && !(cmd.args.iter().any(|ref a| {
+                let arg = a.to_str().unwrap_or("");
+                arg.starts_with("-m") || arg.starts_with("/arch") || arg.starts_with("--target")
+            }));
 
         if use_defaults {
             // Non-target flags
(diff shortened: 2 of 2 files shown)


alexcrichton:
Thanks for the PR! 

Thinking about this again and seeing the recent opening of https://github.com/alexcrichton/cc-rs/issues/376, it actually reminds me that I wonder if we could solve this by simply placing user-defined arguments at the end of the CLI instead of the front? I thought that was what `cc` already!

I don't think that strategy would solve #371, but I think it would solve your use case and #376 I thinkg? (afaik C compilers always prefer later CLI flags)

GeorgeHahn:
In our case, the consuming crate sets `-werror`; this causes the conflicting `-march` flags to fail the build. I agree that user defined flags should be at the end - I'll make that change here.

The `-werror` case requires either a strategy for automatic conflict handling or a configurable option to disable default flags. This PR implements a fairly simple conflict handling strategy, but there are definitely other options.

GeorgeHahn:
69940e0a389f0a7e7da7474e12ac2560af93acb3 inserts user defined flags after default flags but before flags specified by the dependent crate. This minimizes the amount of change while still solving #376.

Inserting the user's flags after the dependent crate's flags would require further re-ordering. I'm not certain about the level of impact, but I think that would belong in a separate PR and may deserve a new minor version.

alexcrichton:
I'm curious, is the explicit checking for `-O` and such flags still necessary if flags are placed at the end? I was under the impression that later optimization settings always override previous ones, similarly for target configuration.

GeorgeHahn:
Duplicate optimization flags shouldn't impact gcc/clang, but may impact MSVC ([ref](https://docs.microsoft.com/en-us/cpp/build/reference/o-options-optimize-code?view=vs-2017)).

Duplicate target related flags (in my case, the `-march` flag) can give a warn level message that turns into an error if the consumer crate sets `-werror`.

alexcrichton:
Ok!

fanzeyi:
:( This change is breaking our build system since we are passing some flags in `CFLAGS` that are starting with `-m` while not doing cross-compiling. Can we please have an environment variable to tell cc skip these checks?

alexcrichton:
Ok I've turned off the automatic inference of turning off defaults in https://github.com/alexcrichton/cc-rs/commit/3d0abff4846a14d9770230d7790b4b4ab3ddbf6a, sorry for the breakage!

PR #381, opened 2019-03-14
use clang for wasm32-unknown-unknown target

pravdomil:
see https://github.com/alexcrichton/cc-rs/issues/378


alexcrichton:
Thanks!

Issue #385, opened 2019-03-26
[Request] Support Visual Studio 2019


TheGoddessInari:
Apparently as a prerequisite for alexcrichton/cmake-rs#70, cc-rs needs to be updated for Visual Studio 2019.

It uses the same installation path scheme as 2017, but cc-rs hardcodes a lot of the VS2017 values.

Visual Studio 2019 releases officially on April 2nd, but has had a go-live RC version out for some time.

This is currently preventing, for instance, servo from building with VS2019,

alexcrichton:
I think this was covered in https://github.com/alexcrichton/cc-rs/pull/386 so closing.

PR #386, opened 2019-03-26
Minimally support Visual Studio 2019.

TheGoddessInari:
This would be the bare minimum for fixing #385.

It'd need more fleshing out later, but should support the case when the environment variables are already set.


alexcrichton:
Thanks!

PR #387, opened 2019-04-02
enable using AR to modify archiving behavior on windows

### LINKED ISSUE #384: Allow to use something else than lib.exe to create libraries on Windows
See https://bugzilla.mozilla.org/show_bug.cgi?id=1537643

`LIB` is however a variable that already has a meaning, so it can't be used. The Gecko build system, as well as some other build systems tend to call it `AR` because lib.exe actually does create ar archives.

froydnj:
Fixes #384.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1487,12 +1487,7 @@ impl Build {
         let objects: Vec<_> = objs.iter().map(|obj| obj.dst.clone()).collect();
         let target = self.get_target()?;
         if target.contains("msvc") {
-            let mut cmd = match self.archiver {
-                Some(ref s) => self.cmd(s),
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Since the two are now being interpreted as basically the same, could this msvc branch be updated to call `get_archiver`? That should naturally pick up the usage of AR (and any future changes to that). The current code using `windows_registry` would then just need to move into there as well as a fallback case where the target contains `msvc`

froydnj:
> Since the two are now being interpreted as basically the same, could this msvc branch be updated to call `get_archiver`?

Good call!  I tried keeping the `unwrap_or_else` logic, but eventually had to admit that `match` was much nicer here.

alexcrichton:
Thanks!

PR #392, opened 2019-04-10
Use emscripten compiler for wasm32-unknown-unknown

tmfink:
Does this make sense?

Related issue with `wasm-pack` building Rust dependency that builds C code with `cc-rs`:
https://github.com/rustwasm/wasm-pack/issues/621

With this fix, I no longer need to manually set `CC=emcc`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1708,7 +1708,7 @@ impl Build {
                 t
             })
             .or_else(|| {
-                if target.contains("emscripten") {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! Unfortunately though I don't think this PR can be landed because the `emcc` compiler, as far as I know, cannot be used with our wasm32-unknown-unknown target. That's intended for the `{asmjs,wasm32}-unknown-emscripten` targets.

PR #396, opened 2019-04-19
Don't add default flags that conflict with _FLAGS

GeorgeHahn:
This is an evolution of https://github.com/alexcrichton/cc-rs/pull/374. That PR was too naive with duplicate flag checking and ended up causing problems.

This change organizes flags into two groups: flags that do not carry a value and are mutually exclusive (`-m64, -m32`) and flags that carry a value that should only be set once (`-march=, -mcpu=`). Both groups are handled separately.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1095,18 +1095,17 @@ impl Build {
         let mut cmd = self.get_base_compiler()?;
         let envflags = self.envflags(if self.cpp { "CXXFLAGS" } else { "CFLAGS" });
 
-        // Disable default flag generation via environment variable or when
-        // certain cross compiling arguments are set
-        let use_defaults = self.getenv("CRATE_CC_NO_DEFAULTS").is_none();
+        for arg in envflags {
+            cmd.push_cc_arg(arg.into());
+        }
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! After the first iteration though I'm pretty wary about this in the sense that it seems like it may cause undue difficulty keeping it working over time. Since we place user flags at the end and later flags should take precedence, is this still needed?

GeorgeHahn:
I'm really sorry about that. I should have better considered the implications of the naive first implementation.

The main pain point this solves is successfully building crates that set `warnings_into_errors` - these fail on the flag conflict warning.

I had hoped that the `CC_NO_DEFAULTS` environment variable would offer a useful tool to help solve these situations, but my experience so far has been that setting it causes a lot more trouble than it solves.

alexcrichton:
Oh no worries at all! This area of compiling is notoriously tricky and really tough to ever know if it's right or wrong, so sometimes it's just a "throw things out there and see what sticks" sort of situation :)

That's an interesting point though about warnings to errors, would it be possible to specifically disable warnings related to overriding command line flags though? (since it's the intended behavior that we want)

PR #399, opened 2019-04-29
Automatically check if README.md examples are working when running "cargo test"

GuillaumeGomez:
Since rustdoc nightly now provides "cfg(test)" when running on test mode, we can now use this macro on test mode only to check if README.md examples are working as expected.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,4 @@ parallel = ["rayon"]
 
 [dev-dependencies]
 tempdir = "0.3"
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
It failed at this step:

```
Install g++-multilib
```

Doesn't seem related to my changes. :)

alexcrichton:
Thanks for the PR, but I would personally prefer to keep it manually up-to-date and would prefer to avoid taking on a new dev-dependency.

PR #401, opened 2019-04-29
Run the `Minimum Rust` CI job on Windows and fix it on Linux

### LINKED ISSUE #400: Minimum supported Rust version differs between Windows and other OSs
`windows_registry.rs` uses a number of features not supported on Rust 1.16.0 which appears to be [the official minimum version](https://github.com/alexcrichton/cc-rs/blob/master/azure-pipelines.yml#L10). These include:

- `Iterator::find_map` (Rust 1.30.0)
- `str::trim_start_matches` (Rust 1.30.0)
- `Option::filter` (Rust 1.29.0)

This seems to be slipping through CI as the `Minimum Rust Job` target [only runs on Linux](https://dev.azure.com/alexcrichton/cc-rs/_build/results?buildId=278&view=logs&jobId=8faf42a8-5946-50a5-f557-7e1d982634e3) but most of the contents of `windows_registry.rs` is only built for `cfg(windows)` targets.

ebarnard:
Closes #400

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,12 +2,24 @@ trigger:
   - master
 
 jobs:
-  - job: min
-    displayName: Minimum Rust
+  - job: min_linux
+    pool:
+      vmImage: ubuntu-16.04
+    displayName: Minimum Rust (Linux)
+    variables:
+      TOOLCHAIN: 1.16.0
     steps:
       - template: ci/azure-install-rust.yml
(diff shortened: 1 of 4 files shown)


ebarnard:
@alexcrichton If you're ok with this fix could you release a new version when you merge. It will fix some CI issues I'm having.

alexcrichton:
Thanks for the PR!

PR #402, opened 2019-04-30
Recognize the "wasm32-wasi" target tuple.

sunfishcode:
To minimize confusion with [multiarch tuples](https://wiki.debian.org/Multiarch/Tuples) which omit the vendor field of the target tuple, and to save a small amount of typing, we're preparing to switch the "wasm32-unknown-wasi" target to be spelled "wasm32-wasi". LLVM already accepts both forms. This is similar to how the vendor field is omitted in rustc tuples such as "x86_64-fuchsia" and "aarch64-linux-android".

The patch here just changes the cc-rs crate to recognize either form in anticipation of this change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1751,7 +1751,9 @@ impl Build {
                     }
                 } else if target.contains("cloudabi") {
                     format!("{}-{}", target, traditional)
-                } else if target == "wasm32-unknown-wasi" || target == "wasm32-unknown-unknown" {
+                } else if target == "wasm32-wasi" ||
+                          target == "wasm32-unknown-wasi" ||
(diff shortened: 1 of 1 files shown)


alexcrichton:
I think that this target here is the one passed to `TARGET` which is in turn passed to rustc, and rustc doesn't currently accept `wasm32-wasi` (that's a whole separate discussion...), so I'm not sure that this'll actually ever get exercised? Did this show up in practice though?

sunfishcode:
https://github.com/rust-lang/rust/pull/60585 has now landed and is in the latest nightly builds, so wasm32-unknown-wasi is now wasm32-wasi.

alexcrichton:
👍

PR #403, opened 2019-05-24
Skip compile when already built output file.

higumachan:
Hi,

I implement skip mode for build speedup.
If you need this function, please please give me an opinion about the API.
I think this API is still in development.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,4 @@ parallel = ["rayon"]
 
 [dev-dependencies]
 tempdir = "0.3"
+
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for the PR! 

This is unfortunately really tricky to get right though. In addition to C files we need to account for changes to header files which typically the C compiler can tell us about, but it's pretty nontrivial. Additionally I personally value the quick (ish) compile time of this crate, and pulling in Serde/`serde_json` as dependencies is probably a bit much for simply compiling C code.

I think it would be interesting though to build a layer on top of the `cc` crate which builds in caching perhaps, and that'd be best done as a separate crate.

higumachan:
Thanks for comment.

I understand for your opinion.
I will create other create that caching.

alexcrichton:
Oh sorry for the late response, but I don't personally have time to maintain more crates so I would prefer if it were created externally. I'm going to close this in the meantime however.

PR #404, opened 2019-05-30
Remove ppa:ubuntu-toolchain-r/test from Azure Pipelines

spl:
Remove the `ubuntu-toolchain-r/test` PPA, which is added by default. Some packages were removed, and this is causing the `g++multilib` install to fail.

See https://github.com/scikit-learn/scikit-learn/issues/13928 for a similar experience.


alexcrichton:
Thanks!

PR #405, opened 2019-05-30
Remove master branch trigger on Azure Pipelines

spl:
This PR removes a [`trigger` configuration](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers) that prevents me from testing non-`master` branches on a fork. I'm curious about the motivation behind it. If there's a good reason for it, feel free to close this PR without merging; it's easy enough to work around.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,3 @@
-trigger:
-  - master
-
 jobs:
   - job: min_linux
     pool:


alexcrichton:
This is here so when I push to branches of this repository it doesn't do a double build, but if you fork it should be easy enough to remove from the fork!

spl:
> This is here so when I push to branches of this repository it doesn't do a double build,

I'm new to Azure Pipelines. Why is that a problem? It happens all the time on Travis-CI. I suppose it's a minor annoyance sometimes but not enough to be disabled.

> but if you fork it should be easy enough to remove from the fork!

True, but I should add that it's a pain to remember to include the change when dealing with multiple branches that change different things.

Would you consider an alternative `trigger` configuration using branch wildcards?

alexcrichton:
Oh the issue is the same with all CI services, if you push to a branch of a repository and then send a PR from that repository to itself, two builds are scheduled. One is the PR build and one is the branch build. For larger CI configurations this can tend to cause issues.

Different triggers seem reasonable to me!

spl:
> Different triggers seem reasonable to me!

Would you want to choose some branch prefix for branches you don't want to CI? For example, you could exclude all branches matching the wildcard pattern `skipci-*`.

alexcrichton:
I think I'd probably prefer to have the inverse, only run CI on a whitelist of branches, would that be possible? Something like `ci-*` seems reasonable.

PR #407, opened 2019-05-30
Set default compiler to clang on apple hosts

spl:
The change in this PR (615c53bf23b3a026b90a613c6364dd6ade869cd1) _seems_ to be the right choice to make. `clang`/`clang++` should be considered the default on macOS, and it allows the `ToolFamily` to be chosen correctly based on the name.

Many tests fail, however, and I believe that's because (a) they are duplicated for Linux and macOS targets and (b) each uses the same target and host. For example, consider this test from `tests/test.rs`:

```rust
#[test]
fn gnu_x86_64() {
    for vendor in &["unknown-linux-gnu", "apple-darwin"] {
        let target = format!("x86_64-{}", vendor);
        let test = Test::gnu();
        test.gcc()
            .target(&target)
            .host(&target)
            .file("foo.c")
            .compile("foo");
        test.cmd(0).must_have("-fPIC").must_have("-m64");
    }   
}
```

So, first, is my premise about choosing the default compiler on macOS is correct? If so, how should the tests change?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1426,8 +1426,6 @@ impl Build {
         }
 
         if target.contains("-ios") {
-            // FIXME: potential bug. iOS is always compiled with Clang, but Gcc compiler may be
-            // detected instead.
             self.ios_flags(cmd)?;
         }
 
@@ -1682,9 +1680,12 @@ impl Build {
             ("CC", "cl.exe", "gcc", "cc", "clang")
         };
 
-        // On Solaris, c++/cc unlikely to exist or be correct.
         let default = if host.contains("solaris") {
+            // On Solaris, cc/c++ is unlikely to exist or be correct.
             gnu
+        } else if host.contains("apple") {
+            // On macOS, this allows us to correctly identify the ToolFamily.
+            clang
         } else {
             traditional
         };


spl:
This is how one test for the target and host `x86_64-unknown-linux-gnu` fails on Azure Pipelines:

```
test gnu_i686 ... cargo:warning=clang: error: no such file or directory: 'foo.c'
cargo:warning=clang: error: no input files


error occurred: Command "clang" "-O2" "-ffunction-sections" "-fdata-sections" "-fPIC" "--target=i686-apple-darwin" "-Wall" "-Wextra" "-o" "/tmp/gcc-test.TJJ1vYB38xne/foo.o" "-c" "foo.c" with args "clang" did not execute successfully (status code exit code: 1).
```

alexcrichton:
Seems reasonable to me! I'm not sure why CI didn't fire here, but this plus with fixed tests sounds reasonableto me.

spl:
What's the best way to fix the tests?

It's not clear to me why the host and target are both set to the same or even why the host is set at all? For one thing, I think you'd want any current or future logic in `cc-rs` to be able to assume the host has the expected environment, whereas using `x86_64-apple-darwin` as the host on a Linux machine does not fit that expectation.

Would it be acceptable to let the tests assume the actual host instead of the target as the host?

spl:
Okay, I have better understanding of the test infrastructure. Ignore my musings above. I'm currently performing some renovations to the tests.

alexcrichton:
Ok thanks for that!

I wonder if in the meantime that https://github.com/alexcrichton/cc-rs/issues/408 is a better solution for this? I'm somewhat wary of changing defaults as that can have weird unintended rippling effects, but if we can detect that `/usr/bin/cc` is a symlink to a binary called `clang` that seems like a pretty clear indicator that it's clang-by-default rather than gcc

thomcc (changes requested):
This seems reasonable, if someone wants to use GCC on these targets, they can set `CC` or set the compiler via the API, but I think we shouldn't ever default to it on Apple targets.

That said, this needs a rebase, and I'll probably go over it again after that's done.

spl:
Replaced by #709.

PR #412, opened 2019-06-04
Add find msbuild vs16 logic.

crlf0710:
`sass-sys` crate doesn't build on MSVC 2019 because the build script can't find 2019 msbuild.exe to build the solution. This commit fixes this problem.


alexcrichton:
Thanks!

Issue #419, opened 2019-07-06
File::create in gcc-shim sometimes fails with multiple test threads


spl:
On macOS 10.14.5 (Mojave) with `rustc` 1.35.0 and 1.36.0, if I do the following with `cc-rs` at d2ba46f65cbf56b6b8852a3ab277c45259e38811:

```
$ export RUST_BACKTRACE=full
$ while cargo test; do sleep 0.00001; done
```

I eventually get the following:

```
cargo:warning=thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:997:5
cargo:warning=stack backtrace:
cargo:warning=   0:        0x10b4928c3 - std::sys::unix::backtrace::tracing::imp::unwind_backtrace::hf2949dcf16da83bd
cargo:warning=   1:        0x10b48ed52 - std::sys_common::backtrace::_print::h4b87ed6ba0cd5304
cargo:warning=   2:        0x10b491986 - std::panicking::default_hook::{{closure}}::hde39fed432870543
cargo:warning=   3:        0x10b49172f - std::panicking::default_hook::hfce8820d41dbdcbf
cargo:warning=   4:        0x10b49203f - std::panicking::rust_panic_with_hook::hfcf2d0777bc6c409
cargo:warning=   5:        0x10b491b6c - std::panicking::continue_panic_fmt::h4d668741ea600293
cargo:warning=   6:        0x10b491a58 - rust_begin_unwind
cargo:warning=   7:        0x10b4a21a1 - core::panicking::panic_fmt::hdf67eedb88644e1e
cargo:warning=   8:        0x10b487a1e - core::result::unwrap_failed::h72e112a4d5b9862c
cargo:warning=   9:        0x10b487c2c - core::result::Result<T,E>::unwrap::hd3ca7ccf5659392e
cargo:warning=  10:        0x10b48500a - gcc_shim::main::h533f983c23ce4b8d
cargo:warning=  11:        0x10b485841 - std::rt::lang_start::{{closure}}::h02cd9a674d8a658f
cargo:warning=  12:        0x10b491a37 - std::panicking::try::do_call::h1252fc9a2ff235eb
cargo:warning=  13:        0x10b493a6e - __rust_maybe_catch_panic
cargo:warning=  14:        0x10b4923fd - std::rt::lang_start_internal::h4c054360e442146c
[...]

PR #421, opened 2019-07-08
Fix occasional test failure due to env::set_var with multiple test threads

spl:
Tests that modify environment variables (e.g. `CFLAGS` and `CXXFLAGS`) can cause tests to fail when run in parallel because they change a shared context.

This change moves the problematic tests into separate modules. Since each `tests/*.rs` is compiled as a separate crate, these tests are not run in parallel with other tests.

### DIFF
--- a/ci/azure-steps.yml
+++ b/ci/azure-steps.yml
@@ -16,9 +16,9 @@ steps:
 
   - script: cargo build
     displayName: "Normal build"
-  - bash: cargo test $NO_RUN -- --test-threads 1
+  - bash: cargo test $NO_RUN
     displayName: "Crate tests"
-  - bash: cargo test $NO_RUN --features parallel -- --test-threads 1
+  - bash: cargo test $NO_RUN --features parallel
     displayName: "Crate tests (parallel)"
   - bash: cargo test $NO_RUN --manifest-path cc-test/Cargo.toml --target $TARGET
     displayName: "cc-test tests"
--- a/tests/cflags.rs
+++ b/tests/cflags.rs
@@ -1,0 +1,18 @@
+extern crate cc;
+extern crate tempdir;
+
+mod support;
+
+use std::env;
+use support::Test;
+
+/// This test is in its own module because it modifies the environment and would affect other tests
+/// when run in parallel with them.
+#[test]
+fn gnu_no_warnings_if_cflags() {
+    env::set_var("CFLAGS", "-arbitrary");
+    let test = Test::gnu();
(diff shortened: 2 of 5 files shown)


alexcrichton:
Thanks! Could this also remove the [`--test-threads 1`](https://github.com/alexcrichton/cc-rs/blob/d2ba46f65cbf56b6b8852a3ab277c45259e38811/ci/azure-steps.yml#L19-L22) argument we pass on CI?

spl:
> Could this also remove the [`--test-threads 1`](https://github.com/alexcrichton/cc-rs/blob/d2ba46f65cbf56b6b8852a3ab277c45259e38811/ci/azure-steps.yml#L19-L22) argument we pass on CI?

This and #419 are the only problems I've encountered with parallel test threads. Unfortunately, the work I've been doing on `cc-rs` exacerbated these issues, which is why I've been chasing them down.

alexcrichton:
It seems the intention of this PR is to basically enable `cargo test` though, right? If that's the case we should be able to run that on CI to verify it?

spl:
I guess you're suggesting that I remove `--test-threads 1` in CI. I don't think this PR is enough to allow that, since #419 is unresolved, but at least you can see the result now (1f81b89fc9bc5e48e2df088b0c048e1f58b8e963).

spl:
It seems like my hack (aa968a0f70b59509280bc60ff4959aa6e313e4af) is not enough to “fix” #419 on Azure ([build results](https://dev.azure.com/alexcrichton/cc-rs/_build/results?buildId=631)).

alexcrichton:
Ah ok, if that's the case mind if we hold off on this? Right now the test suite just needs to be run with one thread, but if there's more lurking then I'd prefer if it were all fixed instead of only having half of it fixed

spl:
Alternatively, with d09804846d77cbcbc74e01e4a4b63203f11bf1bf implemented, it seems like you only need `--test-threads 1` on macOS. Then, you only have one outstanding known issue with multiple test threads rather than two. In my view, I don't see the downside of merging this even if you don't remove `--test-threads 1` from CI.

spl:
Okay, I think I've fixed all the test threads issues. The failures in the [build results](https://dev.azure.com/alexcrichton/cc-rs/_build/results?buildId=634) appear to be network-related.

spl:
Huh. I accidentally deleted the branch, which closed the PR. Then, I restored it and re-opened the PR. And that had the unexpected result of re-running the Azure build. 🤷‍♂

PR #425, opened 2019-07-25
Change ucrt and windows10sdk detection to match vcvars.

jrmuizel:
I had an installation of the 10.0.17763 sdk that did had files
in Lib/10.0.17763.0 but not in Include/10.0.17763.0. vcvars handled
this correctly but cc-rs did not.

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -523,7 +523,7 @@ mod impl_ {
         let key = r"SOFTWARE\Microsoft\Windows Kits\Installed Roots";
         let key = otry!(LOCAL_MACHINE.open(key.as_ref()).ok());
         let root = otry!(key.query_str("KitsRoot10").ok());
-        let readdir = otry!(Path::new(&root).join("lib").read_dir().ok());
+        let readdir = otry!(Path::new(&root).join("include").read_dir().ok());
         let max_libdir = otry!(readdir
             .filter_map(|dir| dir.ok())
             .map(|dir| dir.path())
@@ -539,19 +539,13 @@ mod impl_ {
         Some((root.into(), version))
     }
 
-    // Vcvars finds the correct version of the Windows 10 SDK by looking
-    // for the include `um\Windows.h` because sometimes a given version will
-    // only have UCRT bits without the rest of the SDK. Since we only care about
-    // libraries and not includes, we instead look for `um\x64\kernel32.lib`.
-    // Since the 32-bit and 64-bit libraries are always installed together we
-    // only need to bother checking x64, making this code a tiny bit simpler.
     // Like we do for the Universal CRT, we sort the possibilities
     // asciibetically to find the newest one as that is what vcvars does.
     fn get_sdk10_dir() -> Option<(PathBuf, String)> {
         let key = r"SOFTWARE\Microsoft\Microsoft SDKs\Windows\v10.0";
         let key = otry!(LOCAL_MACHINE.open(key.as_ref()).ok());
         let root = otry!(key.query_str("InstallationFolder").ok());
-        let readdir = otry!(Path::new(&root).join("lib").read_dir().ok());
+        let readdir = otry!(Path::new(&root).join("include").read_dir().ok());
         let mut dirs = readdir
             .filter_map(|dir| dir.ok())
             .map(|dir| dir.path())
@@ -560,7 +554,7 @@ mod impl_ {
         let dir = otry!(dirs
             .into_iter()
             .rev()
-            .filter(|dir| dir.join("um").join("x64").join("kernel32.lib").is_file())
+            .filter(|dir| dir.join("winsdkver.h").is_file())
             .next());
         let version = dir.components().last().unwrap();
         let version = version.as_os_str().to_str().unwrap().to_string();


jrmuizel:
I'm not sure why my installation lacked Include files. Reinstalling it fixed the problem so maybe it's not worth taking this, but I figured it would be good to at least put it out there.

alexcrichton:
This is unfortunately pretty finnicky code, and it's also basically untesable in the sense that I've got no idea if this regresses environments which work today. (although it does look like VS2019 regressed on CI?)

If this ended up being fixable on your end though maybe it's best then to not take it just yet?

nico-abram:
@jrmuizel Were you able to figure out why you were missing the directory in your installation? I wonder if there is some optional component in the VS installer for the .libs . Then again, if you didn't have the .libs I don't think things would link correctly even if cc-rs found the folders.
Also, is it possible your comment in the original message is backwards w.r.t "Lib" and "Include"? (Looking at the code change it looks like you fixed it by making cc-rs look for include instead of dir, so I assume you had include and you didn't have lib)

> and it's also basically untesable in the sense that I've got no idea if this regresses environments which work today.

 @alexcrichton If this is actually a problem I wonder if we could keep the detection via /lib but fall back to /include if that fails?
 I'm thinking something like this
 ```rs
        let max_libdir = |sub: &'static str| -> Option<(PathBuf, String)> {
	          let readdir = Path::new(&root).join(sub).read_dir().ok()?;
	          let max_libdir = readdir
	              .filter_map(|dir| dir.ok())
	              .map(|dir| dir.path())
	              .filter(|dir| {
	                  dir.components()
	                      .last()
	                      .and_then(|c| c.as_os_str().to_str())
	                      .map(|c| c.starts_with("10.") && dir.join("ucrt").is_dir())
	                      .unwrap_or(false)
	              })
	              .max()?;
        };
        let max_libdir = max_libdir("lib").unwrap_or_else(|| max_libdir("include"));
```

**Comment by jrmuizel** [comment]
I wasn't able to figure out why I was missing the directory. I believe my original comment is correct, by checking for include files instead of lib files `cc-rs` skips the candidate sdk that's missing the include files. From what I recall, this matches what vcvars checks.

**Comment by nico-abram** [comment]
Ohh, I understand now. So you had multiple SDKs and you wanted the skip the one without the include files. That explains why you didn't get linker errors, since that second SDK presumably had both the lib and include folders. Checking both "lib" and "include" as I proposed above would then not fix this, since it would still find the one without the include folder first.

Although maybe what we could do is look for "include" first, and when we find one we *also* find the lib folder with the same mechanism as now for that same SDK version. If we find no "include" folder or we find one without a lib folder, we fall back to searching for just "lib". I think that should prioritize SDKs with an include folder over ones without one, but it should still work for some obscure case where someone is currently compiling with an SDK without include files correctly.

The current `get_ucrt_dir` looks this

```rs
    fn get_ucrt_dir() -> Option<(PathBuf, String)> {
        let key = r"SOFTWARE\Microsoft\Windows Kits\Installed Roots";
        let key = LOCAL_MACHINE.open(key.as_ref()).ok()?;
        let root = key.query_str("KitsRoot10").ok()?;
        let readdir = Path::new(&root).join("lib").read_dir().ok()?;
        let max_libdir = readdir
            .filter_map(|dir| dir.ok())
            .map(|dir| dir.path())
            .filter(|dir| {
                dir.components()
                    .last()
                    .and_then(|c| c.as_os_str().to_str())
                    .map(|c| c.starts_with("10.") && dir.join("ucrt").is_dir())
                    .unwrap_or(false)
            })
            .max()?;
        let version = max_libdir.components().last().unwrap();
        let version = version.as_os_str().to_str().unwrap().to_string();
        Some((root.into(), version))
    }

```

With the method I'm proposing it could look like this. I'm not sure if this is too much complexity or can be simplified.
```rs
    fn get_ucrt_dir() -> Option<(PathBuf, String)> {
        let key = r"SOFTWARE\Microsoft\Windows Kits\Installed Roots";
        let key = LOCAL_MACHINE.open(key.as_ref()).ok()?;
        let root = key.query_str("KitsRoot10").ok()?;

        let read_incdir = Path::new(&root).join("include").read_dir().ok()?;
        let read_libdir = Path::new(&root).join("lib").read_dir().ok()?;
        // vcvars finds the SDK using the include dir. cc-rs used to look for the lib dir
        // What we do now is look for include dirs, but also make sure that SDK has a library directory
        // If that fails, we fall back to the previous cc-rs behaviour and find the SDK using the lib folder
        let filter_libdir = |dir: &PathBuf| {
            dir.components()
                .last()
                .and_then(|c| c.as_os_str().to_str())
                .map(|c| c.starts_with("10.") && dir.join("ucrt").is_dir())
                .unwrap_or(false)
        };
        let max_incdir = read_incdir
            .filter_map(|dir| dir.ok())
            .map(|dir| dir.path())
            .filter(|dir| {
                dir.components()
                    .last()
                    .and_then(|c| c.as_os_str().to_str())
                    .map(|c| {
                        c.starts_with("10.")
                            && dir.join("ucrt").is_dir()
                            && filter_libdir(&Path::new(&root).join("lib").join(c))
                    })
                    .unwrap_or(false)
            })
            .max();
        let max_libdir = read_libdir
            .filter_map(|dir| dir.ok())
            .map(|dir| dir.path())
            .filter(|dir| filter_libdir(dir))
            .max();

        let sdk_dir = max_incdir.or(max_libdir)?;
        let version = sdk_dir.components().last().unwrap();
        let version = version.as_os_str().to_str().unwrap().to_string();
        Some((root.into(), version))
    }
```

ChrisDenton:
As Alex said, this sort of thing isn't really tested atm so I do think we need a fallback because we don't want to break backwards compatibility (maybe someone only needs the libs). And I would want to try testing locally with a no-includes install (presumably the installer has options not to include the includes?). But otherwise I'm happy for us to check for includes.

Would you be interested in rebasing and adding a fallback as suggested? Or maybe @nico-abram would be willing to submit a PR? Sorry this has been open so long; there are more maintainers now and we're slowly working to reduce the backlog.

ChrisDenton:
Ok, I'm going to close this as inactive but if anyone wants to continue working on this then feel very free to do so.

PR #426, opened 2019-07-28
Add CUDA support for MSVC

trolleyman:
Tested building https://github.com/trolleyman/cuda-macros with this (the `cuda-macros-test` crate) and it builds & links correctly. Haven't had a chance to test that this runs yet, but will in the morning.

I wasn't sure that this way is the most elegant, but this seemed like the way that did the least amount of changes. I am also not sure that this is the correct way of doing this, especially regarding cross-compiling, but it gets it up and running at least.

To test you can do a `cargo build` in the root of the repo linked above. The build stuff is a bit hacky, but essentially it generates the CUDA function below & calls it.

```c
extern "C" __global__ void hello(int32_t* x, int32_t y) {
    printf("Hello from block %d, thread %d (y=%d)\n", blockIdx.x, threadIdx.x, y);
    *x = 2;
}
```

```rust
extern "C" unsafe fn hello(x: *mut i32, y: i32);
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -420,7 +420,14 @@ impl Build {
 
         let mut cmd = compiler.to_command();
         let is_arm = target.contains("aarch64") || target.contains("arm");
-        command_add_output_file(&mut cmd, &obj, self.cuda, target.contains("msvc"), false, is_arm);
+        command_add_output_file(
+            &mut cmd,
+            &obj,
+            self.cuda,
+            target.contains("msvc"),
+            false,
+            is_arm,
+        );
 
         // We need to explicitly tell msvc not to link and create an exe
         // in the root directory of the crate
@@ -1183,7 +1190,10 @@ impl Build {
                     cmd.push_cc_arg("-fdata-sections".into());
                 }
                 // Disable generation of PIC on RISC-V for now: rust-lld doesn't support this yet
-                if self.pic.unwrap_or(!target.contains("windows-gnu") && !target.contains("riscv")) {
+                if self
+                    .pic
+                    .unwrap_or(!target.contains("windows-gnu") && !target.contains("riscv"))
+                {
                     cmd.push_cc_arg("-fPIC".into());
                     // PLT only applies if code is compiled with PIC support,
                     // and only for ELF targets.
@@ -1234,7 +1244,8 @@ impl Build {
                 // the SDK, but for all released versions of the
                 // Windows SDK it is required.
                 if target.contains("arm") || target.contains("thumb") {
-                    cmd.args.push("-D_ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE=1".into());
+                    cmd.args
+                        .push("-D_ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE=1".into());
                 }
             }
             ToolFamily::Gnu => {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! It looks like the main difference here is `/X` vs `-X` but I think that MSVC considers them equivalent for C compilation, could the MSVC bits just be updated to use the dash instead?

trolleyman:
Oh cool, didn't know that. Does that mean we can remove all of those `*_flag` functions from the compiler family struct & just use raw string literals everywhere?

alexcrichton:
Perhaps yeah! I forget which target those were added for, so I'd just follow the various logic internally to see if it's necessary to keep it or not.

trolleyman:
Alright, there we go. I also tested `cuda-macros` & it compiles & runs a simple test program. Should be ready to merge

trolleyman:
Ah. Broken on linux.

trolleyman:
There we go

src/lib.rs
```diff
@@ -1276,8 +1248,11 @@ impl Build {
                 // the SDK, but for all released versions of the
                 // Windows SDK it is required.
                 if target.contains("arm") || target.contains("thumb") {
-                    cmd.args
-                        .push("/D_ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE=1".into());
+                    if self.cuda {
```
alexcrichton: Are changes like this necessary? This PR feels a bit overreaching in terms of changing everything it finds?

src/lib.rs
```diff
@@ -1253,16 +1219,22 @@ impl Build {
             ToolFamily::Msvc { clang_cl } => {
                 if clang_cl {
                     if target.contains("x86_64") {
-                        cmd.args.push("-m64".into());
+                        if self.cuda {
+                            cmd.args.push("-m64".into());
+                        }
+                        cmd.push_cc_arg("-m64".into());
```
alexcrichton: I'm not really sure what's going on here, how come the flag is pushed twice?

alexcrichton:
Hm I was sort of hoping that everything that was `/foo` could be changed to `-foo`, but this seems like it's sort of just a smattering of some one way and some another way...

src/lib.rs
trolleyman: I wasn't too sure about this define, so I added it when compiling with nvcc as well. We can just simplify it to a `-D` option instead and it will define it for both nvcc & non-nvcc versions?

src/lib.rs
trolleyman: I wasn't too sure about if the nvcc compiler passed through the -m32 and -m64 options, so I added it for both. I think it does though, so I'll modify these lines to just add the options onto the command args directly, as both msvc and nvcc support these options.

Also, for context, the nvcc compiler [only supports 32/64 bit cross-compilation](https://stackoverflow.com/a/10582406), and it's phasing out 32 bit compilation altogether as well [it seems](https://devtalk.nvidia.com/default/topic/1026144/cuda-9-0-nvcc-does-not-support-compiling-cuda-code-inside-32-bit-projects-on-visual-studio-2015-/).

trolleyman:
Ah, ok - should I try and remove all of the `/opt` options?

alexcrichton:
I think that should work yeah, and I think it's probably better than sprinkling `if cuda` everywhere

trolleyman:
Ok, removed all `/-` options everywhere, and simplified `command_add_output_file`. I've tested that `-Foout.obj` is the same as `-Fo out.obj` when compiling assembly.

Removed the edits that are mentioned in https://github.com/alexcrichton/cc-rs/pull/426#discussion_r309270871 as cross-compilation is not supported with nvcc, so it's probably better not to even pretend that we support it.

Simplified the change in https://github.com/alexcrichton/cc-rs/pull/426#discussion_r309270590

alexcrichton:
Thanks @trolleyman!

PR #428, opened 2019-08-07
Add RISC-V support

### LINKED ISSUE #397: Missing RISCV support
Currently, RISCV support is missing. The target is not recognized and the system compiler is invoked, which will of course fail. Since RISCV is an supported RUST target in stable, I think it would make sense to support this platform.
For my current project on RISCV32, I could add support (at least for GCC) by simply adding:

    "riscv32imac-unknown-none-elf" => Some("riscv32-unknown-elf"),

in src/lib.rs:Build::get_base_compiler (around line 1800). I guess for complete support you should add similar lines for:

    $ rustup run stable rustc --print target-list | grep risc
    riscv32imac-unknown-none-elf
    riscv32imc-unknown-none-elf
    riscv64gc-unknown-none-elf
    riscv64imac-unknown-none-elf

Which would probably be something like:

    "riscv32imac-unknown-none-elf" => Some("riscv32-unknown-elf"),
    "riscv32imc-unknown-none-elf" => Some("riscv32-unknown-elf"),
    "riscv64gc-unknown-none-elf" => Some("riscv64-unknown-elf"),
    "riscv64imac-unknown-none-elf" => Some("riscv64-unknown-elf"),

Please note that I didn't test this!

laanwj:
Automatically choose the most common name for the toolchain, and add the command-line arguments to generate code that is compatible with the chosen architecture.

Fixes #397 by @XvR-Technolution.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1428,6 +1428,17 @@ impl Build {
                         cmd.args.push("-mfloat-abi=soft".into());
                     }
                 }
+                if target.starts_with("riscv32") || target.starts_with("riscv64") {
+                    // get the 32i/32imac/32imc/64gc/64imac/... part
+                    let mut parts = target.split('-');
+                    if let Some(arch) = parts.next() {
+                        let arch = &arch[5..];
+                        cmd.args.push(("-march=rv".to_owned() + arch).into());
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

Issue #383, opened 2019-03-21
Can we please have full control over the minimum iOS version?


ForceBru:
I was trying to see why my build of `rustc` for iOS was trying to pass `-miphoneos-version-min=7.0` to the compiler, even though I specifically set the following flags before compilation:

```
CFLAGS_aarch64_apple_ios="-miphoneos-version-min=12.0" CRATE_CC_NO_DEFAULTS=1
``` 

The output of the compilation process is the following:

```
TARGET = Some("aarch64-apple-ios")
OPT_LEVEL = Some("2")
HOST = Some("x86_64-apple-darwin")
CXX_aarch64-apple-ios = Some("/Users/forcebru/Desktop/Test/TOOLCHAIN/bin/clang++")
CXXFLAGS_aarch64-apple-ios = Some("-fPIC -miphoneos-version-min=12.0 --target=aarch64-apple-ios -arch arm64 -miphoneos-version-min=7.0 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS12.1.sdk -fembed-bitcode")
DEBUG = Some("false")
Detecting iOS SDK path for iphoneos
```

As one can see, the `-miphoneos-version-min=12.0` flag _is_ included in `CXXFLAGS_aarch64-apple-ios`, but it's immediately discarded by the auto-inserted `-miphoneos-version-min=7.0` flag that comes right after it.

I think it should be possible to have control over that flag because some features, like thread-local storage (which apparently [requires iOS 8.0+](https://github.com/nvzqz/Threadly) or [`__IPHONE_OS_VERSION_MIN_REQUIRED > 70100` to be true](https://github.com/ericniebler/range-v3/issues/407)) are not available with this setting.

This setting is hard-coded here:
https://github.com/alexcrichton/cc-rs/blob/caf78d56c7101847304724cca229ad0b14eec26b/src/lib.rs#L1601

It may be better to first check if the `IPHONEOS_DEPLOYMENT_TARGET` environment variable is set or `CFLAGS` contains `-miphoneos-version-min=<stuff>` before switching to default behaviour.

alexcrichton:
I think these were hardcoded once and simply never changed, seems reasonable to read env vars and add a method too!

turboladen:
@alexcrichton when you say 

> and add a method too!

...what were you thinking there? Something as explicit as 

```rust
// On `Tool`
[...]

PR #433, opened 2019-08-12
Detect and use `IPHONEOS_DEPLOYMENT_TARGET` for the `-mi*os-version-min` flag

turboladen:
...If it isn't set, fallback to the previous setting of `-miphoneos-version-min=7.0` for real devices and `mios-simulator-version-min=7.0` for simulators.

----

I realize this doesn't quite fulfill the request in #383 (defining a method on `Tool`), but it at least allows you to specify your targeted minimum version in what seems to be a conventional way (I'm not super familiar with building Xcode projects from the CLI). In my case, I'm using [`cmake-rs`](https://github.com/alexcrichton/cmake-rs) to build a C project, which uses cmake to build an iOS project, all for my Rust project, and this change allowed me to get that project building.

I'd really like to implement that method on `Tool`, but after a few days now I realize I won't have the time in the near future and don't want to let my changes go stale. I hope this is helpful...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1627,16 +1627,19 @@ impl Build {
             }
         };
 
+        let min_version = std::env::var("IPHONEOS_DEPLOYMENT_TARGET")
+                .unwrap_or_else(|_| "7.0".into());
+
         let sdk = match arch {
             ArchSpec::Device(arch) => {
                 cmd.args.push("-arch".into());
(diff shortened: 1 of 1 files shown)


alexcrichton:
This is at least a good start, thanks @turboladen!

turboladen:
And thank you, @alexcrichton, for all the goodies you put out!

PR #443, opened 2019-09-24
document the lack of automatic rebuilding after env var changes

oconnor663:
Tracking issue: https://github.com/alexcrichton/cc-rs/issues/230


alexcrichton:
Thanks for this! I wonder though, if we're going to go to the effort to document this, would you be up for perhaps improving the current state of things? While I don't think we can emit `rerun-if-*` directives by default, it seems reasonable to have an option to opt-in to rerun-if directives perhaps? That way `cc` could print out all the env vars it reads, users could print out source files, and the note here in the README could mention the opt-in to use for watching changed env vars?

PR #460, opened 2019-12-01
Default to double-float ABI on RISC-V Linux

msizanoen1:
This is the default ABI used by GCC and rustc on linux (https://github.com/rust-lang/rust/blob/4007d4ef26eab44bdabc2b7574d032152264d3ad/src/librustc_target/spec/riscv64gc_unknown_linux_gnu.rs#L20).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1588,9 +1588,11 @@ impl Build {
                     if let Some(arch) = parts.next() {
                         let arch = &arch[5..];
                         cmd.args.push(("-march=rv".to_owned() + arch).into());
-                        // ABI is always soft-float right now, update this when this is no longer the
-                        // case:
-                        if arch.starts_with("64") {
+                        if target.contains("linux") && arch.starts_with("64") {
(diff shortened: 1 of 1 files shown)


msizanoen1:
Also RISC-V ELF psABI documentation says that the default ABI is always double-float: https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md#default-abis
Maybe we can only use soft-float ABI for `*-none-*` target?

msizanoen1:
Also there is `riscv64gc-unknown-none-elf` target and `riscv32imafdc-unknown-none-elf` target that supports hard float. Should we default to double-float for these target in rustc?

alexcrichton:
I'm not personally too familiar with the various naming conventions around the riscv targets, but I'm happy to merge PRs to incrementally improve support!

msizanoen1:
Now a new cc version should be released to crates.io?

alexcrichton:
Sure thing, done now!

PR #461, opened 2019-12-05
Enable -fPIC on non-bare-metal RISC-V

msizanoen1:
Rust default to position-independent executables so this is needed.


msizanoen1:
Can you publish a new version on crates.io so I can include it into rust-lang/rust#66899?

alexcrichton:
Sure thing, published now!

PR #464, opened 2019-12-29
Implement std::error::Error for Error

### LINKED ISSUE #456: Implement `std::error::Error` for `Error`
Hey, is it possible to implement `std::error::Error` for the `Error` type? I have some parts where I want to simplify my errors into a `Box<dyn Error + 'static>`.

Since `Debug` and `Display` are already implemented and this library always uses `std`, the empty trait implementation `impl std::error::Error for Error {}` should be completely sufficient and would not break anything?

dtolnay:
Fixes #456.

Would it be possible to get a patch release when this lands? I have an upcoming project that would want to use the new impl. cc @KizzyCode

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -167,6 +167,8 @@ impl Display for Error {
     }
 }
 
+impl std::error::Error for Error {}
+
 /// Configuration used to represent an invocation of a C compiler.
 ///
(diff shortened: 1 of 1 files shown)


alexcrichton:
👍

PR #475, opened 2020-01-31
Detect and use sccache by introspecting RUSTC_WRAPPER

mqudsi:
If no other C/C++ caching tool is found by inspecting `CC` and `CXX`,
`RUSTC_WRAPPER` is tested to see if an output-caching wrapper for
`rustc` is in use. If that is the case and it is a wrapper known to also
support C/C++ caching, use it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1867,7 +1867,7 @@ impl Build {
         cmd.args.push(sdk_path.trim().into());
         cmd.args.push("-fembed-bitcode".into());
         /*
-         * TODO we probably ultimatedly want the -fembed-bitcode-marker flag
+         * TODO we probably ultimately want the -fembed-bitcode-marker flag
          * but can't have it now because of an issue in LLVM:
          * https://github.com/alexcrichton/cc-rs/issues/301
          * https://github.com/rust-lang/rust/pull/48896#comment-372192660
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

kyren:
Just an FYI in case anybody else comes across this bug, this PR can actually break your build if you're already setting RUSTC_WRAPPER and also happen to also have sccache -> gcc symlinks in your PATH with an error similar to the following:

```
running: "sccache" "gcc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-I" "/nix/store/kpc7x7mvb38di1711n6n2s1pgp9zwl7k-openssl-1.1.1d-dev/include" "-Wall" "-Wextra" "-E" "build/expando.c"
cargo:warning=error: failed to execute compile
cargo:warning=caused by: Compiler not supported: ""
exit code: 2

--- stderr
thread 'main' panicked at '
Header expansion error:
Error { kind: ToolExecError, message: "Command \"sccache\" \"gcc\" \"-O3\" \"-ffunction-sections\" \"-fdata-sections\" \"-fPIC\" \"-g\" \"-fno-omit-frame-pointer\" \"-m64\" \"-I\" \"/nix/store/kpc7x7mvb38di1711n6n2s1pgp9zwl7k-openssl-1.1.1d-dev/include\" \"-Wall\" \"-Wextra\" \"-E\" \"build/expando.c\" with args \"gcc\" did not execute successfully (status code exit code: 2)." }
```

Since pretty much everything in rust land that builds C code uses the `cc` crate, simply *not* having those symlinks in the PATH when invoking builds via cargo seems like a reasonable fix (since it'll be now be added by the `cc` crate automatically).

PR #478, opened 2020-02-23
Add prefix autodetecting for 'armv5te-unknown-linux-musleabi' target

antage:
Rustup supports `armv5te-unknown-linux-musleabi` target but `cc-rs` can't autodetect the prefix of CLI tools (\*-gcc/\*-g++). This PR fixes it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1994,6 +1994,7 @@ impl Build {
                         "arm-unknown-linux-gnueabi" => Some("arm-linux-gnueabi"),
                         "armv4t-unknown-linux-gnueabi" => Some("arm-linux-gnueabi"),
                         "armv5te-unknown-linux-gnueabi" => Some("arm-linux-gnueabi"),
+                        "armv5te-unknown-linux-musleabi" => Some("arm-linux-gnueabi"),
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1994,6 +1994,7 @@ impl Build {
                         "arm-unknown-linux-gnueabi" => Some("arm-linux-gnueabi"),
                         "armv4t-unknown-linux-gnueabi" => Some("arm-linux-gnueabi"),
                         "armv5te-unknown-linux-gnueabi" => Some("arm-linux-gnueabi"),
+                        "armv5te-unknown-linux-musleabi" => Some("arm-linux-gnueabi"),
```
alexcrichton: Should this be `arm-linux-musleabi`, or is intentional that gcc is used for the musl target?

antage: `*-musleabi` is the same target as `*-gnueabi` but it uses a static linking with [musl libc](https://musl.libc.org/). Both targets use gcc.
I tested some `*-sys` crates and it works fine.
Neither Debian nor Ubuntu have `arm-linux-musleabi` toolchain packages.

alexcrichton: Ok, thanks for confirming!

PR #479, opened 2020-03-03
Add tool autodetection for armv7a targets

jonas-schievink:


jonas-schievink:
@alexcrichton Okay, this should be ready now I think

PR #481, opened 2020-03-03
Do not set the -fPIC flag for Windows targets when using clang for CX…

### LINKED ISSUE #480: CXX build fails with clang as compiler for Windows MSVC
Maybe this is a pretty far out edge case, but I am using clang on windows for the msvc target and when compiling in cpp mode I receive this error:
cargo:warning=clang: error: unsupported option '-fPIC' for target 'x86_64-pc-windows-msvc'

I noticed in the cc-rs source that you are not setting -fPIC for windows-gnu targets when using the clang toolchain. I removed just the -gnu to also target windows-msvc (and any other Windows targets) and that works. I was thinking of also being more explicit and adding windows-msvc, but practically I don't see any other targets besides windows-gnu & windows-msvc.

I will submit a PR, if it's too edge casey or has other unforseen consequences, let me know! 

Also, thanks for a very useful crate :)

datalus:
Fixes #480

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1374,7 +1374,7 @@ impl Build {
                 // Disable generation of PIC on bare-metal for now: rust-lld doesn't support this yet
                 if self
                     .pic
-                    .unwrap_or(!target.contains("windows-gnu") && !target.contains("-none-"))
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #488, opened 2020-04-21
fix find_tool and find_vs_version might returns incompatible values

Speedy37:
On my PC, I have both MSVC 15 and 16 and found out that find_vs_version returns MSVC 16 while find_tool return MSVC 15 linker.

This caused rust to use MSVC 15 linker, while cmake generated file where made with MSVC 16.

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -80,7 +80,7 @@ pub fn find_tool(target: &str, tool: &str) -> Option<Tool> {
     // environment variables like `LIB`, `INCLUDE`, and `PATH` to ensure that
     // the tool is actually usable.
 
-    return impl_::find_msvc_15(tool, target)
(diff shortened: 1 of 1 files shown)


Speedy37:
Maybe a better approach to find the latest version would be to implement something like: https://github.com/microsoft/vswhere/blob/f078b21784b53dd811ad72f0334b7b70f22aa9cf/src/vswhere.lib/InstanceSelector.cpp#L34

alexcrichton:
Thanks for this! I agree though yeah if we're doing comparisons we should probably do some pattern matching here, because otherwise I think this is doing lexicographical ordering which isn't numerical ordering? (or semver ordering)

Speedy37:
I now convert installation version to Vec<u16> for the ordering.

alexcrichton:
Nice!

PR #489, opened 2020-04-21
Fix Android clang compiler path when building with Windows host

Osspial:
On Windows, the Android NDK provides the `%ANDROID_TARGET%-clang` executable as a `.cmd` file rather than an `.exe` file. `std::process::Command` can only resolve `.exe` commands when the command extension isn't explicitly provided (see https://github.com/rust-lang/rust/issues/37519#issuecomment-390311261), so `cc-rs` mistakenly thinks the `%ANDROID_TARGET%-clang` compiler isn't installed on the system even when it's in the path. This PR fixes that.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1980,7 +1980,11 @@ impl Build {
                         .replace("thumbv7neon", "arm")
                         .replace("thumbv7", "arm");
                     let gnu_compiler = format!("{}-{}", target, gnu);
-                    let clang_compiler = format!("{}-{}", target, clang);
+                    let clang_compiler = if host.contains("windows") {
+                        format!("{}-{}.cmd", target, clang)
+                    } else {
+                        format!("{}-{}", target, clang)
+                    };
                     // Check if gnu compiler is present
                     // if not, use clang
                     if Command::new(&gnu_compiler).spawn().is_ok() {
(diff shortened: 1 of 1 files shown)


alexcrichton:
To confirm, this has always been the case, right? Should we perhaps check for `.exe` and fall back to `.cmd` if it's available?

Osspial:
As far as I understand this has always been the case, but there's no harm in checking for the `.exe` file. I've updated the code to do so.

alexcrichton:
Thanks! Actually though while you're at it, this is something that I missed previously. Could you replace `.spawn()` with `.output()`? Otherwise I think this runs a risk of creating a zombie process.

alexcrichton:
Ah it's ok I'll tweak that locally

Osspial:
It looks like 4469060c5a3ae21aa7b3dccec6aaef65655f9b9d changed the check at [`lib.rs:1994`](https://github.com/alexcrichton/cc-rs/blob/master/src/lib.rs#L1994) from `is_err` to `is_ok`, which broke this fix.

alexcrichton:
Oops sorry about that! I mistakenly assumed which compiler was being tested. I pushed up a new release which I think should fix this.

Osspial:
Looks like that's fixed things. Thanks!

PR #495, opened 2020-05-09
Implement autodetection for default compiler from NDK

Dushistov:
ref #459

Long time ago you had to extract compiler from Android NDK to make it standalone and usable as linker/c/c++ compiler (for example process described here https://mozilla.github.io/firefox-browser-architecture/experiments/2017-09-21-rust-on-android.html). 

At now you can use default C/C++ compiler without efforts. But naming for compilers are different now. For example before this patch, cc-rs try to find `arm-linux-androideabi-clang`, but default compiler now is `armv7a-linux-androideabi16-clang`. Also new clang compiler doesn't define things like `ANDROID/__ANDROID_API__` so it should be defined manually.

Without this patch I can not compile rusqlite for android:
```toml
//Cargo.toml
[dependencies]
rusqlite = { version = "0.17.0", features = ["chrono", "functions", "bundled"] }
//.cargo/config
[target.armv7-linux-androideabi]
linker = "armv7a-linux-androideabi16-clang"
```

it prints:
```
error occurred: Failed to find tool. Is `arm-linux-androideabi-clang` installed?
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1370,6 +1370,15 @@ impl Build {
                     cmd.push_opt_unless_duplicate(format!("-O{}", opt_level).into());
                 }
 
+                if cmd.family == ToolFamily::Clang && target.contains("android") {
+                    // For compatibility with code that doesn't use pre-defined `__ANDROID__` macro.
+                    // If compiler used via ndk-build or cmake (officially supported build methods)
+                    // this macros is defined.
+                    // See https://android.googlesource.com/platform/ndk/+/refs/heads/ndk-release-r21/build/cmake/android.toolchain.cmake#456
+                    // https://android.googlesource.com/platform/ndk/+/refs/heads/ndk-release-r21/build/core/build-binary.mk#141
+                    cmd.push_opt_unless_duplicate("-DANDROID".into());
+                }
+
                 if !target.contains("-ios") {
                     cmd.push_cc_arg("-ffunction-sections".into());
                     cmd.push_cc_arg("-fdata-sections".into());
@@ -1406,7 +1415,11 @@ impl Build {
         // Target flags
         match cmd.family {
             ToolFamily::Clang => {
-                cmd.args.push(format!("--target={}", target).into());
+                if !(target.contains("android")
+                    && android_clang_compiler_uses_target_arg_internally(&cmd.path))
+                {
+                    cmd.args.push(format!("--target={}", target).into());
+                }
             }
             ToolFamily::Msvc { clang_cl } => {
                 // This is an undocumented flag from MSVC but helps with making
@@ -1974,29 +1987,7 @@ impl Build {
                         format!("{}.exe", gnu)
                     }
                 } else if target.contains("android") {
-                    let target = target
-                        .replace("armv7neon", "arm")
-                        .replace("armv7", "arm")
-                        .replace("thumbv7neon", "arm")
-                        .replace("thumbv7", "arm");
-                    let gnu_compiler = format!("{}-{}", target, gnu);
-                    let clang_compiler = format!("{}-{}", target, clang);
-                    // On Windows, the Android clang compiler is provided as a `.cmd` file instead
-                    // of a `.exe` file. `std::process::Command` won't run `.cmd` files unless the
-                    // `.cmd` is explicitly appended to the command name, so we do that here.
-                    let clang_compiler_cmd = format!("{}-{}.cmd", target, clang);
-
-                    // Check if gnu compiler is present
-                    // if not, use clang
-                    if Command::new(&gnu_compiler).output().is_ok() {
-                        gnu_compiler
-                    } else if host.contains("windows")
-                        && Command::new(&clang_compiler_cmd).output().is_ok()
-                    {
-                        clang_compiler_cmd
-                    } else {
-                        clang_compiler
-                    }
+                    autodetect_android_compiler(&target, &host, gnu, clang)
                 } else if target.contains("cloudabi") {
                     format!("{}-{}", target, traditional)
                 } else if target == "wasm32-wasi"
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2722,3 +2706,79 @@ fn command_add_output_file(
...
+];
+
+fn defines_for_android_compiler(clang_path: &Path) -> Vec<&'static str> {
+    match NEW_STANDALONE_ANDROID_COMPILERS.iter().find(|x| {
+        let x: &OsStr = x.as_ref();
+        x == clang_path.as_os_str()
+    }) {
+        Some(new_clang) => {
+            let mut ret = Vec::with_capacity(2);
+            ret.push("-DANDROID");
```
alexcrichton, line 2724: I'd personally sort of prefer to follow whatever the official compilers are doing, if they've stopped passing this I'm not sure if this should explicitly start passing them by default?

src/lib.rs
```diff
@@ -2722,3 +2706,79 @@ fn command_add_output_file(
         cmd.arg("-o").arg(&dst);
     }
 }
+
+static NEW_STANDALONE_ANDROID_COMPILERS: [&str; 4] = [
+    "aarch64-linux-android21-clang",
+    "armv7a-linux-androideabi16-clang",
+    "i686-linux-android16-clang",
+    "x86_64-linux-android21-clang",
```
alexcrichton, line 2714: Are these embedded version numbers something that will change with each release? Is there perhaps other env vars we can probe for or a well-known executable which doesn't change which we can look for and then find sibling compilers?

src/lib.rs
Dushistov: > if they've stopped passing this I'm not sure if this should explicitly start passing them by default?

1. There are two official way to build code with NDK (as I know): include file for GNU make Android.mk
2. cmake

And both of them is set `-DANDROID`. I suppose this is for compatibility. There are predefined variable `__ANDROID__` that embedded into compiler, but I suppose there are a lot of code that do: `#ifdef ANDROID` instead of `#ifdef __ANDROID__`.

And I am sure that adding dependency to cmake-rs crate is good idea, or generate make file, search `make` is not good idea.

I also thought that __ANDROID_API__ is not defined by default, because of sqlite3.c compilation failed (mmap64 not defined), but looks like problem in cc-rs, not with not seted __ANDROID_API__. The line:
```
                cmd.args.push(format!("--target={}", target).into());
```
should be disabled in cc-rs for new android toolchain. I am comeback with fix.

src/lib.rs
Dushistov: > Are these embedded version numbers something that will change with each release?

No, this is minimum API level. 16 for armv7a, i686 and 21 for x86_64 /aarch64 is minium API version. Android 4.0.3 (API level 15) is not supported by this new clang toolchain. But I don't think that this is problem, rustc will also drop support for such low API level rust-lang/rust#71123 .

So this numbers changed when Google drop support of 4.0.3 in their SDK/NDK, have no idea when this happens.

src/lib.rs
Dushistov: I've updated code. And `__ANDROID_API__` is not used anymore. Though I leave `-DANDROID` for compability with code around, because of ndk-build/NDK's cmake toolchain files define it.

src/lib.rs
```diff
@@ -2722,3 +2710,64 @@ fn command_add_output_file(
...
+        if Command::new(new_clang).output().is_ok() {
+            return (*new_clang).into();
+        }
+    }
+
+    let target = target
+        .replace("armv7neon", "arm")
+        .replace("armv7", "arm")
+        .replace("thumbv7neon", "arm")
+        .replace("thumbv7", "arm");
```
alexcrichton, line 2755: FWIW this normalization isn't used when trying to find a new compiler, could this be refactored a bit such that these rustc targets would still find the armv7a android compiler?

src/lib.rs
```diff
@@ -1406,7 +1411,12 @@ impl Build {
         // Target flags
         match cmd.family {
             ToolFamily::Clang => {
-                cmd.args.push(format!("--target={}", target).into());
+                // clang from Android NDK target names are different from rustc.
+                // For example, armv7a-linux-androideabi16-clang passes
+                // --target=armv7a-linux-androideabi16 to clang.
+                if !target.contains("android") || !new_clang_android_compiler(&cmd.path) {
```
alexcrichton, line 1417: Can you expand a bit on this? Why does android choke if this `--target` argument is passed?

src/lib.rs
```diff
@@ -1370,6 +1370,11 @@ impl Build {
                     cmd.push_opt_unless_duplicate(format!("-O{}", opt_level).into());
                 }
 
+                if cmd.family == ToolFamily::Clang && target.contains("android") {
+                    // for compatibility with code that doesn't use predefined `__ANDROID__`
+                    cmd.push_opt_unless_duplicate("-DANDROID".into());
```
alexcrichton, line 1375: Can you expand the comment here to point to official locations how the official tooling is passing a flag like this?

src/lib.rs
alexcrichton: Hm ok, is there perhaps an official location where we can point to which has the names of these various compilers?

Dushistov: https://android.googlesource.com/platform/ndk/+/refs/heads/ndk-release-r21/docs/BuildSystemMaintainers.md#Clang

See 
> Second, a target-specific Clang can be used. In addition to the clang and clang++ binaries, there are also <triple><API-level>-clang and <triple><API-level>-clang++ scripts. For example, to target API 26 32-bit ARM, invoke armv7a-linux-androideabi26-clang or armv7a-linux-androideabi26-clang++ instead of clang or clang++.

src/lib.rs
Dushistov: I intentionally did not touch this code. Tier 1,2 platforms have only linux+android support. So it is hard test for Windows and MacOS, but code bellow works with windows, so I avoid changing anything that I can not test. 

Plus I don't think that this is refactoring is possible. IMHO new target naming that include android API level was introduced recently, so if somebody do not want update Android NDK and uses old `clang` it is easy enough to break things for them.

src/lib.rs
Dushistov: For example:
```
cat ~/Android/Sdk/ndk-bundle/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi16-clang
#!/bin/bash
if [ "$1" != "-cc1" ]; then
    `dirname $0`/clang --target=armv7a-linux-androideabi16 "$@"
else
    # Target is already an argument.
    `dirname $0`/clang "$@"
fi
```

so final command line generated by scripts looks like
```
clang --target=armv7a-linux-androideabi16  --target=<What cc-rs takes from cargo>
```

So for example definition of `__ANDROID_API__` disappear,
because of clang uses the last `--target=`, not the first one.
That's why I have to re-define `__ANDROID_API__` in the first version of my patch.

src/lib.rs
Dushistov: I added this link into code's comments.

src/lib.rs
Dushistov: I added note about "double" `--target` into code's comment section.

src/lib.rs
Dushistov: I added links to last NDK release where `-DANDROID` is used.

src/lib.rs
alexcrichton: Could this comment be updated to indicate that the "new" clang compilers automatically pass `--target` so we don't have to?

src/lib.rs
alexcrichton: My point is that the `armv7-linux-androideabi` target, if folks are using it, will not hit the new clang compiler when it very likely should. I don't think there's a lot of risk here, it's basically just moving this `let target` block up to the top of the function instead of having it at the bottom

Dushistov: > My point is that the armv7-linux-androideabi target, if folks are using it, will not hit the new 
> clang compiler when it very likely should.

@alexcrichton, not sure that I understand you. `armv7-linux-androideabi` mapped to Some("armv7a") and then to `armv7a-linux-androideabi16-clang` call.
So `armv7-linux-androideabi` hit the new clang compiler, it is my main testing configuration,
I build all my code with `cargo build --target=armv7-linux-androideabi` and run on android device before submitting this PR.

src/lib.rs
Dushistov: Done and also I changed function name to make it clear.

src/lib.rs
alexcrichton: Ah ok I think I was misreading a bit. In any case we can always tweak this later!

alexcrichton:
Thanks again!

PR #496, opened 2020-05-11
Fix "Argument list too long" when building with `windows-gnu` target

yangby-cryptape:
When building `rust-rocksdb` with target `x86_64-pc-windows-gnu` (`mingw-w64-x86_64-rust` in msys2), the compiler throw the error `ar: Argument list too long`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1702,55 +1702,20 @@ impl Build {
 
         let objects: Vec<_> = objs.iter().map(|obj| obj.dst.clone()).collect();
         let target = self.get_target()?;
-        if target.contains("msvc") {
+        if target.contains("windows") {
+            let is_msvc = target.contains("msvc");
             let (mut cmd, program) = self.get_ar()?;
-            let mut out = OsString::from("-out:");
-            out.push(dst);
-            cmd.arg(out).arg("-nologo");
+            if is_msvc {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! Would it be possible to perhaps take a different strategy though? I think we can call the `ar` command multiple times (although maybe with a different set of flags) to incrementally add objects to an archive. I think that may be a bit easier than trying to lump it all into one go?

yangby-cryptape:
> call the `ar` command multiple times

OK, I will try this.

VZout:
I'm running into this issue as well. @alexcrichton could you elaborate on your feedback?

As I understand it `cc-rs` supports response files for the `msvc` target but no other targets. This PR adds that.
Do you want to "redesign/move" how response files are handled in general?

Minor side note: https://github.com/alexcrichton/cc-rs/pull/496/files#diff-b4aea3e418ccdb71239b96952d9cddb6R1705
Should be `host = windows` not `target = windows` to support cross compiling.

alexcrichton:
I don't think it's necessary to use response files here, I think we can call the archiver multiple times to add objects, which should simplify this PR

NobodyXu:
#564 should have already fixed this issue, closing this PR

Issue #498, opened 2020-05-16
`windows_registry::find_tool` can report the `Tool` as being in the gnu family


ChrisDenton:
`windows_registry::find_tool` should surely return a MSVC family tool as that's the point of the function. However, although it finds the tool correctly, `is_like_msvc()` is returning `false`.

```Rust
use cc::windows_registry;

fn main() {
    let linker = windows_registry::find_tool("x86_64-pc-windows-msvc", "link.exe").unwrap();
    println!("{}", linker.path().to_string_lossy());
    dbg!(
        linker.is_like_msvc(),
        linker.is_like_gnu(),
    );
}
```
Output:
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.25.28610\bin\HostX64\x64\link.exe
[src\main.rs:6] linker.is_like_msvc() = false
[src\main.rs:6] linker.is_like_gnu() = true
```

As far as I can tell it falls back to the default tool family detection code, which in turn seems to mostly assume it's GNU unless it's `cl`. But I'm still finding my way round the code so I could be wrong.

alexcrichton:
Definitely seems like a bug! A PR would be most welcome to help and fix.

PR #499, opened 2020-05-18
In windows_registry::find_tool, set the ToolFamily to Msvc

ChrisDenton:
Ensures that the `Tool` returned from `windows_registry::find_tool` will always return true for `is_like_msvc()`. Addresses issue #498.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2393,6 +2393,7 @@ impl Tool {
         Self::with_features(path, clang_driver, false)
     }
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! I think there's an error on CI though?

ChrisDenton:
Yes, sorry. It took me a minute to figure out why I wasn't seeing any issues locally. Of course, I forgot the items would be unused outside of Windows.

PR #502, opened 2020-05-19
Find prefixed AR when cross compiling.

### LINKED ISSUE #330: Use a prefixed archiver if using a prefixed compiler
Original report: https://github.com/japaric/cortex-m-rt/issues/80

We are using `cc` to cross assemble some external ARM assembly files for the `thumbv*` targets. `cc` uses by default `arm-none-eabi-gcc` as the assembler and `ar` as the archiver and that works fine on Linux. But on Windows `ar` is usually not available; however, the user will have `arm-none-eabi-ar` installed because they installed `arm-none-eabi-gcc`.

Could we make `cc` use `$prefix-ar` if `$prefix-gcc` is being used for compiling / assembling? `ar` doesn't do anything architecture specific so using `ar` or `$prefix-ar` would give the same results but it's more likely that user has `$prefix-ar` installed if they have `$prefix-gcc` installed.

tkonolige:
This PR uses a prefixed ar when cross compiling (for example `arm-none-eabi-ar`). It will fall back to `ar` if a prefixed version does not exist. The code for computing the prefix is the same between the compiler and the archiver. However, the code does not check if both a prefixed compiler and archiver are being used together (maybe this is a reasonable thing to do?).

Fixes #330.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1998,83 +1998,7 @@ impl Build {
                 } else if target.contains("vxworks") {
                     "wr-c++".to_string()
                 } else if self.get_host()? != target {
-                    // CROSS_COMPILE is of the form: "arm-linux-gnueabi-"
-                    let cc_env = self.getenv("CROSS_COMPILE");
-                    let cross_compile = cc_env.as_ref().map(|s| s.trim_right_matches('-'));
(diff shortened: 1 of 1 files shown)


alexcrichton:
Seems reasonable to me, thanks!

PR #507, opened 2020-06-01
Fix for multi-component Android compiler paths

alyssais:
NEW_STANDALONE_ANDROID_COMPILERS is a list of file names, but
clang_path might be a path (containing path separators).  So only the
file name should be compared to the elements in
NEW_STANDALONE_ANDROID_COMPILERS.


alexcrichton:
Thanks!

Issue #505, opened 2020-05-25
compiling multiple cuda files


YoshikawaMasashi:
I'm trying to compile multiple Cuda files that have dependencies. and I got an error message.
Is there a way to compile multiple CUDA files using cc-rs?

build.rs
```rust
extern crate cc;

fn main() {
    cc::Build::new()
        .cuda(true)
        .flag("-gencode")
        .flag("arch=compute_60,code=sm_60")
        .flag("-gencode")
        .flag("arch=compute_61,code=sm_61")
        .flag("-gencode")
        .flag("arch=compute_62,code=sm_62")
        .file("cuda_src/a.cu")
        .file("cuda_src/b.cu")
        .compile("kernel.o");

    println!("cargo:rustc-link-search=native=/usr/local/cuda/lib64");
    println!("cargo:rustc-link-lib=cudart");
}
```

a.cu
```cpp
#include "b.h"

__device__ int a_func(int x) {
    return b_func(x) + 1;
}
```

b.cu
```cpp
#include "b.h"

__device__ int b_func(int x) {
    return x + 1;
}
```

b.h
```cpp
__device__ int b_func(int x);
```

error
```
warning: ptxas fatal   : Unresolved extern function '_Z6b_funci'
error: failed to run custom build command for `cc-rs-test v0.1.0
```

alexcrichton:
Sorry I do not know anything about cuda, so I am not able to help you here.

PR #508, opened 2020-06-02
Support for Cuda Separate Compilation

YoshikawaMasashi:
related to #505

This PR change the code to support for cuda separate compilation.
https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#using-separate-compilation-in-cuda

Before change, cc-rs run these command

```
$ nvcc {some option} -o {out_dir}/a.o -c a.cu
$ nvcc {some option} -o {out_dir}/b.o -c b.cu
$ ar crs {out_dir}/libkernel.o.a {out_dir}/a.o {out_dir}/b.o
```

But, this command fail by unresolved extern function error.
To perform separate compilation, we need to compile with `-dc` option, and to link object file with `-dlink` option.

After change, cc-rs run these command

```
$ nvcc {some option} -dc -o {out_dir}/a.o -c a.cu
$ nvcc {some option} -dc -o {out_dir}/b.o -c b.cu
$ nvcc {some option} -dlink {out_dir}/a.o {out_dir}/b.o -o {out_dir}/link.o
$ ar crs {out_dir}/libkernel.o.a {out_dir}/a.o {out_dir}/b.o {out_dir}/link.o
```

## code

build.rs(we need `println!("cargo:rustc-link-lib=cudadevrt");` for separate compilation)
```rust
extern crate cc;

fn main() {
    cc::Build::new()
        .cuda(true)
        .file("a.cu")
        .file("b.cu")
        .compile("kernel.o");

    println!("cargo:rustc-link-search=native=/usr/local/cuda/lib64");
    println!("cargo:rustc-link-lib=cudart");
    println!("cargo:rustc-link-lib=cudadevrt");
}
```

a.cu
```cpp
#include "b.h"

__device__ double a_func(double x) {
    return b_func(x) + 1;
}


extern "C" {
    __global__ void
    batch_a_func(
        const double* x, double* y
    ) {
        int i = threadIdx.x;

        y[i] = a_func(x[i]);
    }
}
```

b.cu
```cpp
#include "b.h"

__device__ double b_func(double x) {
    return x + 1;
}
```

b.h
```cpp
__device__ double b_func(double x);
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -929,22 +929,6 @@ impl Build {
             objects.push(Object::new(file.to_path_buf(), obj));
         }
         self.compile_objects(&objects)?;
-
-        if self.cuda {
-            let compiler = self.try_get_compiler()?;
-            let mut cmd = compiler.to_command();
-            cmd.arg("-dlink");
-            for obj in objects.clone() {
-                cmd.arg(obj.dst);
-            }
-            cmd.arg("-o");
-            let out_dir = self.get_out_dir()?;
-            let out_dir = out_dir.join("link.o");
-            cmd.arg(out_dir);
-            println!("running {:?}", cmd);
-            run(&mut cmd, "nvcc")?;
-        }
-
         self.assemble(lib_name, &dst.join(gnu_lib_name), &objects)?;
 
         if self.get_target()?.contains("msvc") {
@@ -1155,6 +1139,21 @@ impl Build {
         for obj in objs {
             self.compile_object(obj)?;
         }
+
+        if self.cuda {
+            let compiler = self.try_get_compiler()?;
+            let mut cmd = compiler.to_command();
+            cmd.arg("-dlink");
+            for obj in objs {
+                cmd.arg(obj.dst.clone());
+            }
+            cmd.arg("-o");
+            let out_dir = self.get_out_dir()?;
+            let out_dir = out_dir.join("link.o");
+            cmd.arg(out_dir);
+            run(&mut cmd, "nvcc")?;
+        }
+
         Ok(())
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -929,6 +929,22 @@ impl Build {
             objects.push(Object::new(file.to_path_buf(), obj));
         }
         self.compile_objects(&objects)?;
+
+        if self.cuda {
+            let compiler = self.try_get_compiler()?;
+            let mut cmd = compiler.to_command();
+            cmd.arg("-dlink");
+            for obj in objects.clone() {
+                cmd.arg(obj.dst);
+            }
+            cmd.arg("-o");
+            let out_dir = self.get_out_dir()?;
+            let out_dir = out_dir.join("link.o");
+            cmd.arg(out_dir);
+            println!("running {:?}", cmd);
+            run(&mut cmd, "nvcc")?;
+        }
+
         self.assemble(lib_name, &dst.join(gnu_lib_name), &objects)?;
 
         if self.get_target()?.contains("msvc") {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1139,6 +1139,21 @@ impl Build {
...
+        if self.cuda {
+            let compiler = self.try_get_compiler()?;
+            let mut cmd = compiler.to_command();
+            cmd.arg("-dlink");
+            for obj in objs {
+                cmd.arg(obj.dst.clone());
+            }
+            cmd.arg("-o");
+            let out_dir = self.get_out_dir()?;
+            let out_dir = out_dir.join("link.o");
```
alexcrichton, line 1152: Could `"link.o"` shared between here and below be refactored to a common definition? Additionally could it be named something like `__cc_internal_link.o` to try to avoid name clashes with normal objects?

YoshikawaMasashi:
To test compilation of cuda, we need `nvcc`, which is compiler of CUDA. And to test execution of compiled code, we need an environment that has Nvidia GPU.

`nvcc` is included in the CUDA Toolkit, the SDK for CUDA. And we can use this in [the nvidia/cuda
 docker image](https://hub.docker.com/r/nvidia/cuda/).
And this SDK may be introduced into Github Action's virtual environment in the future. ([PR](https://github.com/actions/virtual-environments/issues/116))

However, nvidia GPU is not available in [Github Action's virtual environment](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners) now. We can use only 2-core CPU, 7 GB of RAM memory and 14 GB of SSD disk space. So it is difficult to test execution cuda code...

To build CI of cuda execution test, other OSS are using third party CI service, for example [CircleCI](https://circleci.com/). But there is a cost to using GPUs in CircleCI...

src/lib.rs
YoshikawaMasashi: Thanks, I renamed to `__cc_internal_link.o`.

alexcrichton:
Oh I'd be fine not actually executing the CUDA code, it should be ok to just get everything to a linked state I think? Would that be possible to do on CI today? (just exercising this library, which builds, rather than end-to-end testing)

YoshikawaMasashi:
We can build CI that has nvcc(cuda compiler) by also using github actions.
Fow example, this yaml perform `cargo test` at environment that has nvcc.

```
name: CI
on: [push, pull_request]

jobs:
  cuda_test:
    name: CUDA Test
    runs-on: ubuntu-latest
    container: nvidia/cuda:10.2-cudnn7-devel-ubuntu18.04
    steps:
    - uses: actions/checkout@master
    - run: nvcc --version # cuda install check
    - run: apt-get update && apt-get install -y curl
    - name: Install rustup
      run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
    - name: Install Rust (rustup)
      run: source $HOME/.cargo/env && rustup update stable --no-self-update && rustup default stable
      shell: bash
    - name: rustup target
      run: source $HOME/.cargo/env && rustup target add x86_64-unknown-linux-gnu
      shell: bash
    - name: Install g++-multilib
      run: |
        set -e
        # Remove the ubuntu-toolchain-r/test PPA, which is added by default.
        # Some packages were removed, and this is causing the g++multilib
        # install to fail. Similar issue:
        # https://github.com/scikit-learn/scikit-learn/issues/13928.
        sudo add-apt-repository --remove ppa:ubuntu-toolchain-r/test
        sudo apt-get install g++-multilib
      if: matrix.build == 'linux32'
    - name: cargo build
      run: source $HOME/.cargo/env && cargo build
      shell: bash
    - name: cargo test 
      run: source $HOME/.cargo/env && cargo test
      shell: bash 
```

Of course we have to write test codes to perform cuda compile test. I am trying to do. And I need some time to understanding current cc-rs test code, and write cuda test code.

alexcrichton:
Yeah that looks perfect! I'd be happy to have that on CI here

PR #511, opened 2020-06-06
Name wrangling c++ caveat in Readme

Narsil:
Try to add a small help for people not super familiar in C++ and name wrangling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ fn main() {
 When using C++ library compilation switch, the `CXX` and `CXXFLAGS` env
 variables are used instead of `CC` and `CFLAGS` and the C++ standard library is
 linked to the crate target.
-Remember that C++ does name wrangling so `extern "C {...}` might be required to enable rust linker to find your functions.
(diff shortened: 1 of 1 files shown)


slyshykO:
Maybe better to use word "mangling" despite wrangkung like in wikipedia?
https://en.wikipedia.org/wiki/Name_mangling

README.md
```diff
@@ -150,6 +150,7 @@ fn main() {
 When using C++ library compilation switch, the `CXX` and `CXXFLAGS` env
 variables are used instead of `CC` and `CFLAGS` and the C++ standard library is
 linked to the crate target.
+Remember that C++ does name mangling so `extern "C {...}` might be required to enable rust linker to find your functions.
```
alexcrichton, line 153: I think there's a missing `"` here, and perhaps the `{...}` can be dropped? There's a few different styles of doing this in C++ I believe but `extern "C"` should be the common denominator.

Narsil: I updated the sentence ! Thanks.

Issue #513, opened 2020-06-09
Use `--` to separate flags and options from input filenames


ncalexan:
I'm in the unusual position of trying to cross-compile from macOS to Windows with `clang-cl` (building Firefox, but that's not really relevant).  I'm running into the `[-Wslash-u-filename]` warning, like so:
```
running: "/Users/nalexander/.mozbuild/clang/bin/clang-cl" "-Xclang" "-std=gnu99" "-nologo" "-MD" "-O1" "-Z7" "-Brepro" "-m64" "-Xclang" "-ftrivial-auto-var-init=pattern" "-guard:cf" "-DDEBUG=1" "-I/Users/nalexander/Mozilla/gecko/toolkit/library/rust" "-I/Users/nalexander/Mozilla/objdirs/objdir-windows-compile/toolkit/library/rust" "-I/Users/nalexander/Mozilla/objdirs/objdir-windows-compile/dist/include" "-I/Users/nalexander/Mozilla/objdirs/objdir-windows-compile/dist/include/nspr" "-I/Users/nalexander/Mozilla/objdirs/objdir-windows-compile/dist/include/nss" "-MD" "-FI" "/Users/nalexander/Mozilla/objdirs/objdir-windows-compile/mozilla-config.h" "-DMOZILLA_CLIENT" "-Qunused-arguments" "-D_HAS_EXCEPTIONS=0" "-W3" "-Gy" "-Zc:inline" "-Gw" "-Wno-unknown-pragmas" "-Wno-ignored-pragmas" "-Wno-deprecated-declarations" "-Wno-invalid-noreturn" "-Z7" "-O2" "-Oy-" "-DMOZILLA_CONFIG_H" "-DMDB_IDL_LOGN=9" "-Fo/Users/nalexander/Mozilla/objdirs/objdir-windows-compile/x86_64-pc-windows-msvc/debug/build/lmdb-rkv-sys-e159e1f997d95e72/out/mdb.o" "-c" "/Users/nalexander/Mozilla/gecko/third_party/rust/lmdb-rkv-sys/lmdb/libraries/liblmdb/mdb.c"
cargo:warning=clang-9: warning: '/Users/nalexander/Mozilla/gecko/third_party/rust/lmdb-rkv-sys/lmdb/libraries/liblmdb/mdb.c' treated as the '/U' option [-Wslash-u-filename]
cargo:warning=clang-9: note: Use '--' to treat subsequent arguments as filenames
[...]

PR #514, opened 2020-06-09
For non-MSVC, separate flags/options from the input file. (#513)

ncalexan:
This avoids a `clang-cl` issue common when cross-compiling from macOS
to Windows: the `-Wslash-u-filename` where a `/Users/...` path is
interpreted as the `cl.exe` flag `/U`.

In general it's somewhere between harmless and good form to separate
flags/options from input files, so `--` is used for all compilers save
MSVC.

There are no existing tests mocking out `clang-cl` and surface efforts
didn't succeed, so that will have to wait for follow-up.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1171,6 +1171,13 @@ impl Build {
         if !msvc || !is_asm || !is_arm {
             cmd.arg("-c");
         }
+        if compiler.family == (ToolFamily::Msvc { clang_cl: true }) {
+            // #513: For `clang-cl`, separate flags/options from the input file.
+            // When cross-compiling macOS -> Windows, this avoids interpreting
+            // common `/Users/...` paths as the `/U` flag and triggering
+            // `-Wslash-u-filename` warning.
+            cmd.arg("--");
+        }
         cmd.arg(&obj.src);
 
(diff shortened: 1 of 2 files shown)


ncalexan:
@alexcrichton two relevant questions/things to check:

#.  I elected to separate with `--` for all compilers save MSVC.  That's not the most conservative choice, but I do think it's the right one.  Happy to be proven wrong here.
#.  I tried to mock out `clang-cl`, but the test code is difficult to work with and can't really reach into the internals such as the `FamilyType`.  Happy to be shown the way here.

roblabla:
So it seems like even if `compiler.family == (ToolFamily::Msvc { clang_cl: true })`, the assembler may be MSVC, and thus not accept the `--` to separate the outputs.

A quick fix would be to add an `&& !is_asm` to the `if` statement.

roblabla:
Got a fix over at https://github.com/rust-lang/cc-rs/pull/747

thomcc:
Fix released in https://github.com/rust-lang/cc-rs/releases/tag/1.0.76

glandium:
FYI, this broke building with sccache, because sccache doesn't handle the "--" properly.

Issue #516, opened 2020-06-18
armasm64.exe on Windows doesn't support -D flags


tleasor:
The aarch64 assembler on Windows, `armasm64.exe` doesn't support -D flags set using .define(). This causes a build failure to occur when targeting aarch64-pc-windows-msvc using `cc`. I'm happy to commit a fix using the equivalent -predefine flag (http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.kui0100a/armasm_cacbjgdb.htm) for this target if that's an acceptable solution.

alexcrichton:
Sounds good to me! A PR would be most welcome :)

Stammark:
I also ran into this :) (but for my purposes I just disabled cc-rs from passing the -D option because my assembler snippet didn't care)

I'd be wary of calling --predefine an equivalent flag, though. (https://developer.arm.com/docs/101754/0614/armasm-legacy-assembler-reference/armasm-command-line-options/-predefine-directive : "--predefine defines a global variable, whereas -Dname defines a macro that the preprocessor expands" (think #define , #ifdef etc.), but I don't know enough about the legacy assembler syntax to know the direct impact in data allocation or symbol creation.

This link suggests "Although you can use predefined global variables in combination with assembly control directives, for example IF and ELSE to control conditional assembly, they are not intended to provide the same functionality as the C preprocessor in armasm. If you require this functionality, Arm recommends you use the compiler to pre-process your assembly code."
Can that be done with MSVC, though.... I'm not sure!

tleasor:
@Stammark Thanks for the link, that helps to clarify things. When I ran into this issue myself (trying to build a native rustc) the [assembly file in question](https://github.com/rust-lang/stacker/blob/master/psm/src/arch/aarch64_armasm.asm) didn't make use of the -D flags so I used the same workaround you described.

It looks like MSVC might support that since it's not listed in the [differences of their armasm version](https://docs.microsoft.com/en-us/cpp/assembler/arm/arm-assembler-directives?view=vs-2019). However, given that the flags aren't equivalent and assembly authors targeting Windows ARM64 would most likely be aware of MSVC limitations, the best solution would seem to be to skip defining any -D flags for this target. To that end, I've opened a simple PR with a fix in #518.

tleasor:
Fix merged in #518

PR #518, opened 2020-06-23
Don't set definition flags for MSVC ARM targets

tleasor:
This is a simple fix for #516

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1690,11 +1690,13 @@ impl Build {
         for directory in self.include_directories.iter() {
             cmd.arg("-I").arg(directory);
         }
-        for &(ref key, ref value) in self.definitions.iter() {
-            if let Some(ref value) = *value {
-                cmd.arg(&format!("-D{}={}", key, value));
-            } else {
-                cmd.arg(&format!("-D{}", key));
+        if !(target.contains("aarch64") || target.contains("arm")) {
+            for &(ref key, ref value) in self.definitions.iter() {
+                if let Some(ref value) = *value {
+                    cmd.arg(&format!("-D{}={}", key, value));
+                } else {
+                    cmd.arg(&format!("-D{}", key));
+                }
             }
         }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1690,7 +1690,9 @@ impl Build {
         for directory in self.include_directories.iter() {
             cmd.arg("-I").arg(directory);
         }
-        if !(target.contains("aarch64") || target.contains("arm")) {
+        if target.contains("aarch64") || target.contains("arm") {
+            println!("cargo:warning=The MSVC ARM assemblers do not support -D flags");
+        } else {
             for &(ref key, ref value) in self.definitions.iter() {
                 if let Some(ref value) = *value {
                     cmd.arg(&format!("-D{}={}", key, value));


alexcrichton:
Thanks for this! I'm wondering though if it's perhaps best to panic or error in these cases? It seems sort of bad to silently ignore user input?

I'm not really sure of the best way to handle this unfortunately if it's a feature that's not supported by just one assembler...

tleasor:
That's a good point, I guess it ultimately comes down to where one thinks this quirk should be taken into account (i.e. at the `cc-rs` level or at the level of a project that depends on `cc-rs`).

On the one hand someone that's writing assembly for this target presumably knows that they can't use these `-D` flags with the relevant assemblers so would avoid using `.define()` for these targets in their Rust code. On the other hand, they wouldn't be able to use text macros in MSVC ARM assembly anyway so might assume that `cc-rs` should "just work" and compile without any issues (and ignore any `-D` flags in doing so).

I think both viewpoints are valid, and would be happy to go the route of a panic or error in this (or a new) PR if you think that silently ignoring these flags isn't the right way to go.

> I'm not really sure of the best way to handle this unfortunately if it's a feature that's not supported by just one assembler...

I agree it doesn't help that it's just one (relatively niche) assembler that has this issue. Ideally I'd look at existing assembly code for this target to see how this change would affect people, but GitHub search doesn't make it easy to filter that granularly, and it's also unlikely that's there's much code for this target to start with.

alexcrichton:
To be clear I have zero experience with this assembler, so I'm not sure what the best route is! Is the syntax different enough that you wouldn't try to shove multiple architectures/platforms into one file? If that's the case then if you're already not trying to rely on any `-D` thing in your code then I think it might make sense to simply ignore it since no one familiar with the assembler really relies on its existence anyway.

Stammark:
Agree with everything that has been said here. 

This quite a nice case of someone wanting to compile assembler for ARM on windows, for windows, exclusively through an MSVC environment (where this would be the only assembler available to them).

For context this came in because of 

And I guess the other



I guess the slightly more "middle of the road" solution would be to emit a warning: If the user did try to use a -Dmacro in their armasm assembler, they would get an error anyway, either when assembling or a "symbol not found" when linking. Then they'd be able to see the warning and realise what's happening.
If on the other hand they weren't using the macros, they (and also any project using that crate as a dependency, even if it is 5 dependency layers down), would get a ping of "something fishy is be happening over there".

alexcrichton:
Hm @Stammark if what you're saying is right then I don't think this PR will help that use case. This is only avoiding flags explicitly defined with `.define("A", Some("B"))` or something like that (basically those defined in the build script itself). If `-D` flags are injected via env vars then this PR isn't excluding those from the flags passed to the assembler.

Stammark:
Ah yes, having actually looked at this properly now, that's correct, my bad!

The env var side would be a different topic entirely (and would be a niche case of a niche case), so ignore most of the above, sorry haha :')

In light of that, I don't have any preference for how to handle this.

All I can think of adding rn is: if we do choose to error, then we could do it gracefully rather than letting the tool derp out on us (but I wouldn't say this is a big deal either :D )

alexcrichton:
I'm personally ok with ignoring the `-D` flags here, but @tleasor to confirm, you're sure this would fix your use case too? Which is to say that you've got `.define(...)` calls but they're intended for other platforms, and it's just convenient to have them on all platforms rather than "everything but aarch64-msvc"?

tleasor:
@alexcrichton Yes, in the case that led me here assembly files are split by architecture and OS so this fixes that for MSVC ARM targets (the MSVC ARM assembler also seems to be in a superseded format, so is unlikely to ever be mixed with ARM assembly from other platforms).

The solution I'm leaning towards at the moment is to ignore -D flags for these targets as this PR currently does, but also output a message to stderr warning that this is the case. That way projects will still build but also be made aware of potential issues. Would that be a satisfactory way forward?

alexcrichton:
Seems reasonable to me!

tleasor:
@alexcrichton Thanks, this should be good to go now.

alexcrichton:
👍

PR #520, opened 2020-06-24
Use xcrun-rs for iOS SDK detection.

luxxxxy:
Use the [xcrun](https://crates.io/crates/xcrun) crate for detecting the iOS SDK.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ edition = "2018"
 
 [dependencies]
 jobserver = { version = "0.1.16", optional = true }
+xcrun = "1.0.4"
 
 [features]
 parallel = ["jobserver"]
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! Could you elaborate a bit on the rationale for this change though? I see that the crate was created a few hours before this PR, so I'm not sure what there is to gain by using this as a dependency vs what's there today

luxxxxy:
I ended up making the crate because I thought it'd be useful for my project ([Crabapple](https://github.com/Crabapple-iOS/crabapple)) and remembered that this crate also used xcrun.

I simply thought it'd lead to neater code, and could be used for other iOS/macOS related stuff in the future (such as automatically finding the tools such as `clang` that are included with the SDK)

alexcrichton:
Ah ok, in that case I think I'd prefer to leave this as-is for now. If the need arises I think it may make sense to look into an external crate but for now this is simple enough I think it's ok to have inline.

luxxxxy:
Understandable.

PR #521, opened 2020-06-25
Use a search list to find a compatible toolchain

xobs:
This adds toolchain detection support and activates it for riscv embedded targets. This enables `cc` to search for compatible toolchains using a whitelist of targets.

For example, the following builds a test project using `riscv64-unknown-elf-gcc` when targeting `riscv32imac-unknown-none-elf`:

```
[7:40:55 PM] D:/Code/Betrusted/cctest> riscv32-unknown-elf-gcc
riscv32-unknown-elf-gcc: The term 'riscv32-unknown-elf-gcc' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
[7:40:58 PM] D:/Code/Betrusted/cctest> riscv64-unknown-elf-gcc
riscv64-unknown-elf-gcc.exe: fatal error: no input files
compilation terminated.
[7:41:01 PM] D:/Code/Betrusted/cctest> cargo build --target riscv32imac-unknown-none-elf
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
[7:41:05 PM] D:/Code/Betrusted/cctest>
```

This closes https://github.com/alexcrichton/cc-rs/issues/519


xobs:
Also, I have a small test project that I put together. Nothing fancy, but it does show that it works. Is this the sort of thing that would be useful to have in CI, or not (because it's a tier 3 target)?

alexcrichton:
Nah I'm always happy to try to add more CI here so long as it doesn't put much of a burden on maintenance!

src/lib.rs
```diff
@@ -2291,11 +2291,31 @@ impl Build {
...
             "powerpc64-unknown-linux-gnu" => Some("powerpc-linux-gnu"),
             "powerpc64le-unknown-linux-gnu" => Some("powerpc64le-linux-gnu"),
-            "riscv32i-unknown-none-elf" => Some("riscv32-unknown-elf"),
-            "riscv32imac-unknown-none-elf" => Some("riscv32-unknown-elf"),
-            "riscv32imc-unknown-none-elf" => Some("riscv32-unknown-elf"),
-            "riscv64gc-unknown-none-elf" => Some("riscv64-unknown-elf"),
-            "riscv64imac-unknown-none-elf" => Some("riscv64-unknown-elf"),
+            "riscv32i-unknown-none-elf" => self.find_working_gnu_prefix(&[
+                "riscv32-unknown-elf",
+                "riscv64-unknown-elf",
```
alexcrichton, line 2296: Out of curiosity, can this 64-bit compiler be used for a 32-bit target?

src/lib.rs
```diff
@@ -2325,6 +2345,33 @@ impl Build {
...
     }
 
+    /// Some platforms have multiple, compatible, canonical prefixes. Look through
+    /// each possible prefix for a compiler that exists and return it. The prefixes
+    /// should be ordered from most-likely to least-likely.
+    fn find_working_gnu_prefix(&self, prefixes: &[&'static str]) -> Option<&'static str> {
+        let suffix = if self.cpp { "-g++" } else { "-gcc" };
+
+        for prefix in prefixes {
+            let mut cmd = Command::new(&format!("{}{}", prefix, suffix));
```
alexcrichton, line 2355: Instead of spawning a process could this perhaps probe `PATH` which should be a bit speedier?

src/lib.rs
xobs: It can be, and very frequently is. The sifive toolchains, for example, advertise "Our toolchain and emulator distributions have been carefully packaged to support both 32-bit & 64-bit ISAs." You switch what you're targeting with `-march=`, which `cc` already does.

src/lib.rs
xobs: I considered that, but then I'd have to check all `PATHEXT` on Windows for each possible executable extension, and check for the `+x` bit on Posix systems.

One concern I had here was what happens if the compiler expects something on `stdin` and never actually finishes, but I imagine that's a largely theoretical problem.

I'm also not sure how well case sensitivity works here. But I'm sure Rust's `Path::exists()` does the right thing there.

If you prefer, I can rework it.

alexcrichton: My main worry is that process spawning is pretty slow compared to checking for files. There's not need to really be 100% correct here since it's pretty unlikely you'd have multiple of these compilers in `PATH` but only one is actually executable. And AFAIK but most of these compilers don't even exist on Windows, right?

xobs: At least two of these compilers exist on Windows (`riscv-embed-none` and `riscv64-unknown-none-elf`) in compiled form, so I could imagine users running into both of these.

I've rewritten it so it now iterates through the `PATH` variable rather than forking off commands. Furthermore, it seems like `Path::exists()` is case-insensitive on Windows, so it continues to work even though I tried renaming my compiler to `riscv64-unknown-elf-gcc.EXE`.

src/lib.rs
```diff
@@ -2325,6 +2345,39 @@ impl Build {
...
+    /// Some platforms have multiple, compatible, canonical prefixes. Look through
+    /// each possible prefix for a compiler that exists and return it. The prefixes
+    /// should be ordered from most-likely to least-likely.
+    fn find_working_gnu_prefix(&self, prefixes: &[&'static str]) -> Option<&'static str> {
+        let windows_host = self
+            .get_host()
+            .map(|host| host.contains("windows"))
+            .unwrap_or(false);
+        let suffix = if self.cpp { "-g++" } else { "-gcc" };
+        let extension = if windows_host { ".exe" } else { "" };
```
alexcrichton, line 2357: I think this can be replaced with `std::env::consts::EXE_SUFFIX` perhaps?

src/lib.rs
```diff
@@ -2325,6 +2345,39 @@ impl Build {
...
+                .as_ref()
+                .and_then(|path_entries| {
+                    env::split_paths(path_entries)
+                        .find(|path_entry| path_entry.join(&target_compiler).exists())
+                })
+                .is_some()
+            {
+                return Some(*prefix);
+            }
+        }
```
alexcrichton, line 2372: Could the order of the loops be inverted here? That way if `/usr/bin` is near the front of `PATH` we won't have to search for everything everywhere that doesn't exist.

src/lib.rs
```diff
@@ -2325,6 +2345,39 @@ impl Build {
...
+        let extension = if windows_host { ".exe" } else { "" };
+
+        for prefix in prefixes {
+            let target_compiler = format!("{}{}{}", prefix, suffix, extension);
+
+            if env::var_os("PATH")
+                .as_ref()
+                .and_then(|path_entries| {
+                    env::split_paths(path_entries)
+                        .find(|path_entry| path_entry.join(&target_compiler).exists())
```
alexcrichton, line 2366: `.find(..).is_some()` can be replaced with `.any(...)`

src/lib.rs
xobs: I had no idea this was a thing. Thanks for pointing it out.

src/lib.rs
xobs: I've made this change as requested.

src/lib.rs
xobs: Thanks, I also didn't know about this construct. However, as part of inverting the two loops, this ended up being unnecessary. And in fact the whole thing got collapsed down to a single statement.

alexcrichton:
👍

PR #522, opened 2020-06-25
Use x86 hosted msvc on arm64 windows

danielframpton:
There is currently no arm64 hosted msvc toolchain, but the x86 toolchain can be used.

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -375,6 +375,7 @@ mod impl_ {
         let host = match host_arch() {
             X86 => "X86",
             X86_64 => "X64",
+            AARCH64 => "X86",
             _ => return None,
         };
         let target = lib_subdir(target)?;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! I've got a few quick questions before merging:

* Do you know if there's future plans to have a native toolchain? One where perhaps we could test if a native toolchain exist, and fall back to the old x86 ones if it doesn't?
* Could you add a comment about why aarch64 is using the x86 toolchain? 
* Is there a particular reason to avoid the x64 toolchain? Is the x86 one perhaps faster or something like that? (in emulation)

danielframpton:
I added a comment. The reason to use x86 is because that is all that is available.

I would prefer not to add the check without there being a native toolchain to test with.

alexcrichton:
Yeah makes sense for sure, was just wondering if there was already something to test with, but apparently there isn't! In any case looks great in that case, thanks again!

PR #524, opened 2020-07-01
Add the host architecture Windows 10 SDK bin directory to the PATH, not the target

arlosi:
Fixes failures during cross-compilation when the host cannot execute the target architecture Windows 10 SDK tools.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,6 +34,11 @@ jobs:
             rust: stable
             target: aarch64-apple-ios
             no_run: --no-run
+          - build: windows-aarch64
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for this! Would it be possible to add a test for this on CI?

arlosi:
I added a CI task targeting aarch64 -- hopefully did that right.

The tests already include running `rc` (which ships in the Windows Kits). Without this change, it will attempt to run the aarch64 version of the tool on the x86_64 host.

I also manually ran the aarch64 tests locally.

alexcrichton:
👍 Thanks!

PR #525, opened 2020-07-03
Add -nologo command-line option to Microsoft assembler.

dot-asm:
Hi,

I'd like to make a case for adding -nologo to Microsoft assembler command line. Trouble is that cargo treats the logo as warning, and worst part is that it's "sticky." "Sticky" in sense that once object files are generated and you re-run 'cargo build' for whatever different reason, it will keep showing the logo (multiple if assembler was invoked several times) tagged with "warning" without actually executing the assembler.

Cheers.


alexcrichton:
Seems reasonable to me!

dot-asm:
Thanks!

PR #528, opened 2020-07-08
use push_cc_arg for options add

slyshykO:
This commit forms proper command line for nvcc on windows. At least it now can compile.


alexcrichton:
Thanks!

PR #533, opened 2020-07-21
Add support for hexagon-unknown-linux-musl-clang

androm3da:


androm3da:
I can add a `cc-test/src/hexagon.S` test case too, if that makes sense.

alexcrichton:
Nah seems reasonable, thanks!

PR #534, opened 2020-07-22
Add arm64e ABI support.

luxxxxy:
Currently, arm64e support can only be found in the Apple LLVM, but it is actively attempting to be upstreamed.

Also useful for iOS/Apple Silicon-related compilation.


alexcrichton:
Thanks!

PR #536, opened 2020-07-31
Initial pass at supporting nasm

Jake-Shadle:
When targetting the Windows msvc environment, cc currently assumes you are compiling on a Windows host and attempts to find specifically named assemblers based on the target architecture when compiling `.asm` files, which obviously doesn't work great when you aren't on a windows host. This adds initial support for having a `AS` environment variable where the user can specify the assembler they actually want to use, with explicit support for `nasm`, which can assemble x86 and x86_64 windows targets from windows and non-windows hosts.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -448,21 +448,13 @@ impl Build {
         }
 
         let mut cmd = compiler.to_command();
-        let is_arm = target.contains("aarch64") || target.contains("arm");
+        let is_msvc = target.contains("msvc");
         let clang = compiler.family == ToolFamily::Clang;
-        command_add_output_file(
-            &mut cmd,
-            &obj,
-            self.cuda,
-            target.contains("msvc"),
-            clang,
-            false,
-            is_arm,
-        );
+        command_add_output_file(&mut cmd, &obj, !is_msvc && !clang && !self.cuda);
 
         // We need to explicitly tell msvc not to link and create an exe
         // in the root directory of the crate
-        if target.contains("msvc") && !self.cuda {
+        if is_msvc && !self.cuda {
             cmd.arg("-c");
         }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for this! I presume though that the syntaxes for nasm/windows are probably incompatible? If that's the case does an env var make sense for this if it's something that always has to be configured?

Additionally would it be possible to add a test or two to CI for this?

Jake-Shadle:
Yes, unfortunately the syntaxes are incompatible so it's not a drop in replacement, which is why I added this as a draft at first.

Thinking a bit more on it, maybe it would be better to just have a convenient mechanism off of `Build` or possibly `Tool` that can tell the user "I want Microsoft assembly" vs eg GAS, as I've fixed up several crates that have both Microsoft and GAS for targeting windows with either MSVC or GNU environments (though definitely not all), but they end up writing incorrect build scripts that only take TARGET and not HOST into account, as when cross compiling with clang-cl, it's the MSVC compiler from the perspective of cc-rs, but they need to use the GAS files instead of the MASM ones since the Microsoft assemblers aren't available.

Ultimately, it would be nice if clang/LLVM added support for MASM, I'm just not going to hold my breath. https://bugs.chromium.org/p/chromium/issues/detail?id=762167

alexcrichton:
Hm ok, well if it's a syntactical thing it seems like that should be an API on `Build` rather than an env var, but otherwise w/e is needed to compile nasm seems ok by me.

ChrisDenton:
@Jake-Shadle are you still wanting to work on this or is the new `llvm-ml` a solution?

Jake-Shadle:
Is that actually official now? I should have closed this regardless, most projects that have asm also have a GAS version for `windows-pc-gnu` so needing to cross compile masm is rarely an issue.

PR #537, opened 2020-08-03
filter SDKROOT for macos->macos build

### LINKED ISSUE #530: crosscompiling macos -> ios + xcodebuild + build-script problem
I call `cargo build` from my XCode project and build project for iOS.
One of my crates depend on bindgen, which depend on clang-sys,
which depend on libloading.

And build failed, with such error:

cargo:warning=In file included from src/os/unix/global_static.c:1:
cargo:warning=In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS13.5.sdk/usr/include/pthread.h:60:
cargo:warning=In file included from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS13.5.sdk/usr/include/time.h:69:
cargo:warning=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS13.5.sdk/usr/include/sys/_types/_clock_t.h:31:9: error: unknown type name '__darwin_clock_t'
cargo:warning=typedef __darwin_clock_t        clock_t;

libloading is part of build-script, so it should not be compiled for iOS,
it should be compiled for macOS, but for some reason it compiled for iOS instead of macOS.

Interesting thing that if I run `cargo build --target=aarch64-apple-ios` manually all works fine.

After analysing what environment variable set `xcodebuild` I thing culprit is 

SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS13.5.sdk

because of with it 

/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-Wall" "-Wextra" "-o" "/tmp/global_static.o" "-c" /Users/evgeniy/.cargo/registry/src/github.com-1ecc6299db9ec823/libloading-0.5.2/src/os/unix/global_static.c

report the same problem, and if I change it into

export SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk

all works fine if I call `cc` manually.

Dushistov:
fixes #530

The code mainly based on rustc source code,
but instead of removing it replace SDKROOT env variable,
because of it is important to find libc headers, like `pthread.h`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -120,6 +120,7 @@ pub struct Build {
     warnings: Option<bool>,
     extra_warnings: Option<bool>,
     env_cache: Arc<Mutex<HashMap<String, Option<String>>>>,
+    apple_sdk_root_cache: Arc<Mutex<HashMap<String, OsString>>>,
 }
 
 /// Represents the types of errors that may occur while using cc-rs.
@@ -312,6 +313,7 @@ impl Build {
             extra_warnings: None,
             warnings_into_errors: false,
             env_cache: Arc::new(Mutex::new(HashMap::new())),
+            apple_sdk_root_cache: Arc::new(Mutex::new(HashMap::new())),
         }
     }
 
@@ -1172,6 +1174,9 @@ impl Build {
             cmd.arg("-c");
         }
         cmd.arg(&obj.src);
+        if cfg!(target_os = "macos") {
+            self.fix_env_for_apple_os(&mut cmd)?;
+        }
 
         run(&mut cmd, &name)?;
         Ok(())
@@ -1887,27 +1892,9 @@ impl Build {
         };
 
         self.print(&format!("Detecting iOS SDK path for {}", sdk));
-        let sdk_path = self
-            .cmd("xcrun")
-            .arg("--show-sdk-path")
-            .arg("--sdk")
-            .arg(sdk)
-            .stderr(Stdio::inherit())
-            .output()?
-            .stdout;
-
-        let sdk_path = match String::from_utf8(sdk_path) {
-            Ok(p) => p,
-            Err(_) => {
-                return Err(Error::new(
-                    ErrorKind::IOError,
-                    "Unable to determine iOS SDK path.",
-                ));
-            }
-        };
-
+        let sdk_path = self.apple_sdk_root(sdk)?;
         cmd.args.push("-isysroot".into());
-        cmd.args.push(sdk_path.trim().into());
+        cmd.args.push(sdk_path);
         cmd.args.push("-fembed-bitcode".into());
         /*
          * TODO we probably ultimately want the -fembed-bitcode-marker flag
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I would prefer to avoid modifying the environment of the calling process if possible, only updating the environment of spawned processes. Additionally this seems not super great to hardcode a path into this crate which may not be the same for all users? Would it be possible to somehow infer that? Are you sure that if the env var isn't set it won't work?

Dushistov:
> I would prefer to avoid modifying the environment of the calling process if possible, only updating the environment of spawned processes. 

Done

> Additionally this seems not super great to hardcode a path into this crate which may not be the same for all users? Would it be possible to somehow infer that? 

Done

> Are you sure that if the env var isn't set it won't work?

Yes, I used such script to find out what is really problem in #530:

```sh
export PATH=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin:$PATH
#export SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
unset SDKROOT


cd /usr/local/var/cache//registry/src/github.com-1ecc6299db9ec823/libloading-0.5.2
"cc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-arch" "x86_64" "-Wall" "-Wextra" "-o" "/tmp/global_static.o" "-c" "src/os/unix/global_static.c"
```

Initially I inserted all environment variables into it and then binary search the problem env. variables.
If I run this script it will print `pthread.h` not found,
if I comment `unset SDKROOT` and uncomment `SDKROOT` all works fine.

So `cc` in `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin` works only with setted `SDKROOT`, `/usr/bin/cc` works just fine without `SDKROOT`,
but I run all build under `xcodebuild` and it changes default cc and set `SDKROOT` unfortunately.

src/lib.rs
```diff
@@ -2467,6 +2450,54 @@ impl Build {
             println!("{}", s);
         }
     }
+
+    #[cfg(target_os = "macos")]
```
alexcrichton, line 2454: Could this be done with `if cfg!(...)` rather than `#[cfg]`?

src/lib.rs
```diff
@@ -2467,6 +2450,54 @@ impl Build {
...
+        Ok(())
+    }
+    #[cfg(not(target_os = "macos"))]
+    fn fix_env_for_apple_os(&self, _cmd: &mut Command) -> Result<(), Error> {
+        Ok(())
+    }
+
+    fn apple_sdk_root(&self, sdk: &str) -> Result<OsString, Error> {
+        let sdk_path = self
+            .cmd("xcrun")
```
alexcrichton, line 2482: I think this will end up runnign `xcrun` once-per-object, which probably slows things down by a lot?

src/lib.rs
```diff
@@ -2467,6 +2450,54 @@ impl Build {
...
     }
+
+    #[cfg(target_os = "macos")]
+    fn fix_env_for_apple_os(&self, cmd: &mut Command) -> Result<(), Error> {
+        let target = self.get_target()?;
+        let host = self.get_host()?;
+        if host.contains("apple-darwin") && target.contains("apple-darwin") {
+            // Replace the `SDKROOT` environment variable if it's clearly set for the wrong platform, which
+            // may occur when we're linking a custom build script while targeting iOS for example.
+            // Removing is not enough, because of cc from XCode can not find include files in such case
```
alexcrichton, line 2461: Can this comment be clarified? The point of "removing is not enough" is only applicable when we're running `clang` directly out of the sysroot I suspect. Executing `/usr/bin/clang` doesn't require `SDKROOT` to be set, for example.

src/lib.rs
```diff
@@ -2467,6 +2450,54 @@ impl Build {
...
+            if let Ok(sdkroot) = env::var("SDKROOT") {
+                if sdkroot.contains("iPhone") {
```
alexcrichton, line 2463: Should this perhaps be inverted? It seems like Apple is inevitably going to release more sdks, so it might be good to not have to update this every time.

src/lib.rs
Dushistov: > I think this will end up runnign `xcrun` once-per-object, which probably slows things down by a lot?

It is the same for ios_flags functions, that called every time via:
compile_object->try_get_compiler->add_default_flags,
should `Build::ios_flags` also be fixed?

Dushistov: I added cache for both, target=macos and target=iphone.

src/lib.rs
Dushistov: Done

src/lib.rs
Dushistov: Done

src/lib.rs
Dushistov: Done

PR #539, opened 2020-08-03
Detect standalone android compiler with regex

gianluca-hilton:
The version pushed in the PR #495 only detects minimal versions of the compiler so other standalone compilers such as "armv7a-linux-androideabiNN-clang" with NN != 16 will not be detected.

Detecting standalone compiler with regex is more generic.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,6 @@ edition = "2018"
 
 [dependencies]
 jobserver = { version = "0.1.16", optional = true }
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the report! I would prefer, however, to not bring in `regex` as a dependency to this crate. Can this be rewritten with some inline string searching?

alexcrichton:
Thanks!

PR #547, opened 2020-08-24
Deprecated response files.

VZout:
To simplify the code I removed response files based on the suggestion of @alexcrichton. I now call the assemble commands multiple times.

This solves #496 as well.

Currently I'm using the same logic that splits the assemble command for all hosts. Let me know if you only want to split it only for Windows.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -929,7 +929,10 @@ impl Build {
             objects.push(Object::new(file.to_path_buf(), obj));
         }
         self.compile_objects(&objects)?;
-        self.assemble(lib_name, &dst.join(gnu_lib_name), &objects)?;
+
+        for chunk in objects.chunks(8191) {
+            self.assemble(lib_name, &dst.join(gnu_lib_name.clone()), &chunk)?;
+        }
 
         if self.get_target()?.contains("msvc") {
             let compiler = self.get_base_compiler()?;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for this! Have you confirmed that this works on both gnu-like platforms and MSVC? I think we need to be sure to pass various flags to `ar` and `lib.exe` to get this to work, but I'm not sure if we do that by default. (if possible adding a test would be great)

VZout:
> Thanks for this! Have you confirmed that this works on both gnu-like platforms and MSVC? I think we need to be sure to pass various flags to `ar` and `lib.exe` to get this to work, but I'm not sure if we do that by default. (if possible adding a test would be great)

I was able to compile `physx-rs` for windows and for android without problem. But I did minimal testing. I'll double check to make sure everything works fine.

alexcrichton:
To confirm, when you say you compile for Windows, do you mean MSVC or MinGW?

VZout:
MSVC. I'll try MinGW/GCC tomorrow. Compiling for android using clang worked.

VZout:
Apparently this doesn't work just work 🤦‍♂️ Must have been building against the wrong local version... My bad. 

I couldn't find any documentation on lld or lib.exe on how to do progressive linking like this. Do you have a reference for the various flags that need to be passed?

Also this solution is not 100% fool prove since in theory a single argument could exceed the command line limit.

alexcrichton:
Sorry I'm not super familiar with `lib.exe`, but with `ar` I'm pretty sure there's options to keep appending to the end of the archive but I'm unfortunately not sure what they are.

Jake-Shadle:
> By default, LIB names the output file using the base name of the first object or library file and the extension .lib. The output file is put in the current directory. If a file already exists with the same name, the output file replaces the existing file. To preserve an existing library, use the /OUT option to specify a name for the output file.

The [documentation](https://docs.microsoft.com/en-us/cpp/build/reference/managing-a-library?view=vs-2019) for lib.exe seems to indicate that using `/OUT` will append if the library already exists, should be easy enough to test though by calling it twice with different object files then running `/LIST`.

VZout:
> > By default, LIB names the output file using the base name of the first object or library file and the extension .lib. The output file is put in the current directory. If a file already exists with the same name, the output file replaces the existing file. To preserve an existing library, use the /OUT option to specify a name for the output file.
> 
> The [documentation](https://docs.microsoft.com/en-us/cpp/build/reference/managing-a-library?view=vs-2019) for lib.exe seems to indicate that using `/OUT` will append if the library already exists, should be easy enough to test though by calling it twice with different object files then running `/LIST`.

Huh I tried that a little while ago but didn't work. Will try again.

VZout:
@maxded Solved the issue I was having!

PR #549, opened 2020-09-12
Apply Darwin target modifications for Clang as well

shepmaster:


alexcrichton:
Thanks! Mind running rustfmt to make CI happy as well? (it may not be due to what you added, could be due to a stable rustfmt update

shepmaster:
Oops! Fixed now.

PR #550, opened 2020-10-08
Method to add multiple includes: build.includes(...)

### LINKED ISSUE #352: Support slice of files and includes
Something that would be nice would be to allow `includes(incs: &[PathBuf])` and something similar for `files` instead of just one include dir and one file at a time.

While it's not that hard to write

```Rust
let some_include_dirs = [
  "dir0", "dir1", "dir2"
];

for dir in &some_include_dirs {
    build.include(dir);
}
```

If the user could supply an array/slice directly it would reduce the code a bit.

dtolnay:
Closes #352.

This adds an `includes` method to call `include(...)` for each dir in an iterator, matching the `files` method from #205 which calls `file(...)` for each source file in an iterator.

My immediate use case for this involves a bunch of dirs that may or may not be selected (`Option<PathBuf>`) so that's what I put in the documentation, but any other iterator would work.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -337,6 +337,35 @@ impl Build {
         self
     }
 
+    /// Add multiple directories to the `-I` include path.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// # use std::path::Path;
+    /// # let condition = true;
+    /// #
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me, thanks!

PR #558, opened 2020-11-02
feat: Simplify code by using `String::split_ascii_whitespace`

Hywan:
Thanks for the great crate :-). I noticed this small optimisation. Hope it helps!


alexcrichton:
Thanks! It looks like this will bump the minimum rust version required, which is fine by me, but can you update CI to the new minimum?

Hywan:
I didn't notice! I'll update the CI.

PR #561, opened 2020-11-09
Document CXXSTDLIB and change the Android default

alexeyr:


alexcrichton:
Thanks! Can you describe in a bit more detail what the Android changes might impact here?

alexeyr:
The result should be more projects working on Android by default. The description of [`libstdc++` for Android](https://developer.android.com/ndk/guides/cpp-support#system) says

> Essentially, this library provides new and delete. In contrast to the other options available in the NDK, there is no support for exception handling or RTTI.
> There is no standard library support aside from the C++ wrappers for the C library headers such as <cstdio>.

So any projects which need more than this. In particular, the one where I ran into the problem was https://github.com/Narsil/esaxx-rs.

The potential drawback of using `libc++_shared` is that it needs to be included together with the main library/application, but at least in my experiments (running `cargo dinghy test`) it worked automatically. `libc++_static` doesn't have this problem, but it has other cases where it shouldn't be used, so the shared version seems to be safer by default.

alexcrichton:
Ok seems reasonable to me, and if it causes issues we can always try again!

PR #564, opened 2020-11-17
progressive linking

maxded:
Should fix both #547 and #496

* When building from `msvc`, linking is split into chunks based on the estimated command line length and the number of objects that need linking. If the estimated command line length doesn't exceed `1024 * 6` it will process all files in one go like before.
* On non `msvc` hosts, nothing has changed as the objects are not split into chunks.

All tests succeeded locally and as an additional test I've compiled the huge `physx` crate from `msvc` to both `msvc` and Android.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1774,66 +1774,39 @@ impl Build {
     }
 
     fn assemble(&self, lib_name: &str, dst: &Path, objs: &[Object]) -> Result<(), Error> {
-        // Delete the destination if it exists as the `ar` tool at least on Unix
-        // appends to it, which we don't want.
+        // Delete the destination if it exists as we want to
+        // create on the first iteration instead of appending.
         let _ = fs::remove_file(&dst);
-
-        let objects: Vec<_> = objs.iter().map(|obj| obj.dst.clone()).collect();
         let target = self.get_target()?;
-        if target.contains("msvc") {
-            let (mut cmd, program) = self.get_ar()?;
-            let mut out = OsString::from("-out:");
-            out.push(dst);
-            cmd.arg(out).arg("-nologo");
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1774,66 +1774,39 @@ impl Build {
...
             } else {
-                cmd.args(&objects).args(&self.objects);
+                objs.len()
             }
-            run(&mut cmd, &program)?;
+        } else {
+            objs.len()
+        };
 
+        for chunk in objs.chunks(chunk_size) {
```
alexcrichton, line 1801: Instead of calculating `chunk_size` could we perhaps conservatively just say the chunk size is 100? That seems like it would probably cover almost all use cases I think?

src/lib.rs
```diff
@@ -1877,10 +1873,7 @@ impl Build {
             for flag in self.ar_flags.iter() {
                 ar.arg(flag);
             }
-            run(
-                ar.arg("crs").arg(dst).args(&objects).args(&self.objects),
```
alexcrichton, line 1881: I think this may be accidentally dropping `self.object`? That may no longer get linked in anywhere?

alexcrichton (commented):
Thanks for this!

src/lib.rs
maxded: Sure thing!

src/lib.rs
maxded: `self.objects` is now cloned and chained together with the objects provided to the assembler in the new `assemble` method so should not be dropped.

alexcrichton: Ah oops I looked for it and failed to see that it was in front of my face, thanks though!

alexcrichton:
👍

PR #572, opened 2020-11-23
Call clang directly when cross-compiling from Windows to Android

maxded:
On Windows, the Android clang compiler is provided as a `.cmd` file instead of a `.exe` file. The `.cmd` file is spawned as a process and thus has a command-line limit of around 32000 characters, which you will most likely not reach. However, the `.cmd` launches the `clang.exe`, which results in a command-line limit of around 8200 characters, which larger projects with lots of include directories will easily reach.

As a fix, I've reintroduced response files when compiling from Windows to Android.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2120,7 +2120,7 @@ impl Build {
                 for version in 16..=23 {
                     if target.contains(&format!("i686-linux-android{}", version)) {
                         tool.args.push("-mstackrealign".into());
-                    }           
+                    }
                 }
 
                 Some(true)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2117,6 +2117,8 @@ impl Build {
                 tool.path.set_extension("exe");
                 tool.args.push(format!("--target={}", target).into());
 
+                // Additionally, shell scripts for target i686-linux-android versions 16 to 24
+                // pass the `mstackrealign` option so we do that here as well.
                 for version in 16..=23 {
                     if target.contains(&format!("i686-linux-android{}", version)) {
                         tool.args.push("-mstackrealign".into());
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2125,7 +2125,7 @@ impl Build {
                         if version > 15 && version < 25 {
                             tool.args.push("-mstackrealign".into());
                         }
-                    }          
+                    }
                 }
             };
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2109,9 +2109,9 @@ impl Build {
         // `--target=` ourselves.
         if host.contains("windows") && android_clang_compiler_uses_target_arg_internally(&tool.path)
         {
-            let mut f = || -> Option<bool> {
-                let file_name = tool.path.file_name()?.to_str()?.to_owned();
-                let (target, clang) = file_name.split_at(file_name.rfind("-")?);
+            if let Some(path) = tool.path.file_name() {
+                let file_name = path.to_str().unwrap().to_owned();
+                let (target, clang) = file_name.split_at(file_name.rfind("-").unwrap());
 
                 tool.path.set_file_name(clang.trim_start_matches("-"));
                 tool.path.set_extension("exe");
@@ -2122,10 +2122,7 @@ impl Build {
                         tool.args.push("-mstackrealign".into());
                     }
                 }
-
-                Some(true)
             };
-            f();
         }
 
         // If we found `cl.exe` in our environment, the tool we're returning is
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2121,11 +2121,11 @@ impl Build {
                 // pass the `mstackrealign` option so we do that here as well.
                 if target.contains("i686-linux-android") {
                     let (_, version) = target.split_at(target.rfind("d").unwrap() + 1);
-                    let version: u32 = version.parse().unwrap();
-
-                    if version > 15 && version < 25 {
-                        tool.args.push("-mstackrealign".into());
-                    }
+                    if let Ok(version) = version.parse::<u32>() {
+                        if version > 15 && version < 25 {
+                            tool.args.push("-mstackrealign".into());
+                        }
+                    }          
                 }
             };
         }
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR!

I was hoping it wouldn't come to this though since response files were just removed in https://github.com/alexcrichton/cc-rs/pull/564. Is there any way to reasonably shorten the command line for your project? Or perhaps consider this a bug for Android that we can spawn the process but it can't spawn a subprocess?

maxded:
Good point! Was unsure myself whos responsibility this was to "fix". As shortening the command line for my project is not possible I've come up with a second, better, solution. 

All the `.cmd` files do on Windows is call the main clang binary with the proper `--target` argument. As all that data is known to us we can construct this ourselves and call the main clang binary directly. :)

alexcrichton:
Seems like a nice strategy if we can get it to work! I think though that one of the purposes of https://github.com/alexcrichton/cc-rs/pull/495 is that the wrapper scripts pass a different `--target` than rustc, and that's intentional for the compilation of the C code?

maxded:
Yes they do pass a different `--target`, which instead of having the shell script pass it, I construct the proper `--target` myself and call clang directly. 

A call to the `aarch64-linux-android21-clang.cmd` shell script internally calls clang.exe with `--target=aarch64-linux-android21`. Basically, the proper _target_ and _clang_ "version" (clang or clang++) is in the `.cmd` file name. So by extracting those ourselves we can call clang.exe or clang++.exe directly.

On Windows, I've checked _all_ `.cmd` files and being able to extract the the proper `--target` is true for all of them! The i686-linux-android16 to 24 `.cmd` files also add the `-mstackrealign` argument, which I've done as well. 

So unless I'm misunderstanding or missing something I think we're good to go. :)

src/lib.rs
```diff
@@ -2096,6 +2096,38 @@ impl Build {
...
+
+                for version in 16..=23 {
+                    if target.contains(&format!("i686-linux-android{}", version)) {
+                        tool.args.push("-mstackrealign".into());
+                    }
+                }
+
+                Some(true)
+            };
+            f();
```
alexcrichton, line 2128: I think the inner closure should be ok here to remove, it's fine to just have one `if let` for the `file_name` binding and avoid the closure business.

src/lib.rs
```diff
@@ -2096,6 +2096,38 @@ impl Build {
...
+                let file_name = tool.path.file_name()?.to_str()?.to_owned();
+                let (target, clang) = file_name.split_at(file_name.rfind("-")?);
+
+                tool.path.set_file_name(clang.trim_start_matches("-"));
+                tool.path.set_extension("exe");
+                tool.args.push(format!("--target={}", target).into());
+
+                for version in 16..=23 {
+                    if target.contains(&format!("i686-linux-android{}", version)) {
+                        tool.args.push("-mstackrealign".into());
```
alexcrichton, line 2122: Is this what the current scripts do? Is there a reason to perhaps not pass this unconditionally? (I have no idea what this flag does)

maxded: Yea that's what the scripts do. Only the `i686-linux-android16-23` have the `-mstackrealign` option added in the script, which force realigns the stack at entry of every function. Don't think we want to do that unconditionally as I am not sure what the flag does as well.

alexcrichton: Hm ok, can you add a comment for why this arg is added? Also, is 23 the latest version? Or do newer versions not require this argument?

maxded: Newer versions do not require the arguments

alexcrichton: Hm ok, in that case could this perhaps be tweaked a bit differently to instead of looping over versions to instead take a look at the target and try to parse a version at the end? If this no longer happens then this is basically just handling older historical compilers then.

maxded: Done!

alexcrichton:
👍 

Thanks again!

Android-Distance:
```rust
// build.rs
extern crate cc;

fn main() {
    let tool = build.get_compiler();
    println!("{:#?}", tool);
    build.file("src/core.c")
        .compile("core");
}

```

```bash
D:\XXX\demo>cargo build --release --target i686-linux-android
   Compiling demo v0.1.0 (D:\Development\WorkSpace\clion\rust\study\demo)
error: failed to run custom build command for `demo v0.1.0 (D:XXX\demo)`

Caused by:
  process didn't exit successfully: `D:\XXX\demo\target\release\build\demo-eaac7e9af5279e16\build-s
cript-build` (exit code: 1)
  --- stdout
  OPT_LEVEL = Some("3")
  TARGET = Some("i686-linux-android")
  HOST = Some("x86_64-pc-windows-msvc")
  CC_i686-linux-android = None
  CC_i686_linux_android = None
  TARGET_CC = None
  CC = None
  CFLAGS_i686-linux-android = None
  CFLAGS_i686_linux_android = None
  TARGET_CFLAGS = None
  CFLAGS = None
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some("false")
  Tool {
      path: "clang.exe",
      cc_wrapper_path: None,
      cc_wrapper_args: [],
      args: [
          "--target=i686-linux-android",
          "-O3",
          "-DANDROID",
          "-ffunction-sections",
          "-fdata-sections",
          "-fPIC",
          "--target=i686-linux-android",
          "-Wall",
          "-Wextra",
      ],
      env: [],
      family: Clang,
      cuda: false,
      removed_args: [],
  }
  running: "clang.exe" "--target=i686-linux-android" "-O3" "-DANDROID" "-ffunction-sections" "-fdata-sections" "-fPIC" "--target=i686-linux-android" "-Wall" "-Wextra" "-o" "D:XXX\\demo\\target\\i686-linux-android\\release\\build\\demo-0d1113bc106cefdd\\out\\src/ffi/core.o" "-c" "src/ffi/core.c"

  --- stderr


  error occurred: Failed to find tool. Is `clang.exe` installed? (see https://github.com/alexcrichton/cc-rs#compile-time-requirements for help)
```

PR #577, opened 2021-01-23
Use clang when cross-compiling to iOS targets

gvcat:
This change allows the target flags to be set correctly when building for Mac Catalyst.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2044,6 +2044,8 @@ impl Build {
                     } else {
                         format!("{}.exe", gnu)
                     }
+                } else if target.contains("ios") {
+                    clang.to_string()
                 } else if target.contains("android") {
                     autodetect_android_compiler(&target, &host, gnu, clang)
                 } else if target.contains("cloudabi") {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2044,6 +2044,8 @@ impl Build {
                     } else {
                         format!("{}.exe", gnu)
                     }
+                } else if target.contains("ios") {
```
alexcrichton, line 2047: Could this perhaps use `-ios-` as a search string to be a bit more flexible? The string "ios" is a little short unfortunately...

gvcat: iOS has 7 different targets. See [platform-support](https://doc.rust-lang.org/beta/rustc/platform-support.html). 

- aarch64-apple-ios
- x86_64-apple-ios
- aarch64-apple-ios-macabi
- armv7-apple-ios
- armv7s-apple-ios
- i386-apple-ios
- x86_64-apple-ios-macabi

It looks like `-apple-ios` is the better choice.

alexcrichton: Sounds good to me!

gvcat: Done

PR #578, opened 2021-01-26
Avoid checking for overly-broad "cl" substring

### LINKED ISSUE #574: Tool family detection is too fragile
`Tool::with_features` attempts to detect the kind of toolchain it's dealing with by doing a substring match on the compiler name. Since `cl` is not a very specific substring, the function has to work around it with exceptions to the rule:

https://github.com/alexcrichton/cc-rs/blob/8a1d5c89f02ebc58480a6e4c5b5c2aa2eb6e07c1/src/lib.rs#L2642-L2662

This obviously falls over when new exceptions appear that aren't accounted for—in this case `powerpcle`, a new target being worked on in [Void Linux](https://voidlinux-ppc.org/) at the moment. `gcc` is incorrectly detected as `cl`, so the crate passes `-nologo` to it, and ends up failing (in this case while cross-compiling the Rust toolchain).

You could of course add "powerpcle" to the list of exceptions for now, but I feel like a more robust check is warranted to avoid this situation in the future (especially when many downstreams are going to pull an old version of the crate and thus require workarounds by further-downstream packagers). Perhaps match by the end of the string, or do some separation by hyphens/underscores, at least. You could also try parsing the version output of the compiler (e.g. with regex) for a more reliable determination, though with `cl` you have to [pass it no arguments](https://stackoverflow.com/questions/1233312/finding-version-of-microsoft-c-compiler-from-command-line-for-makefiles/1233332#1233332) instead of `-v|--version`, making it a bit fiddly.

linkmauve:
Microsoft’s compiler is apparently called "cl" on Linux and "cl.exe" on Windows.

This allows for instance `powerpcle` to be used in the triple, and avoid considering it’s running under a MSVC compiler.

This implements the suggestion by @alexcrichton here: https://github.com/alexcrichton/cc-rs/issues/574#issuecomment-748192661

Fixes #574.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2645,11 +2645,7 @@ impl Tool {
         let family = if let Some(fname) = path.file_name().and_then(|p| p.to_str()) {
             if fname.contains("clang-cl") {
                 ToolFamily::Msvc { clang_cl: true }
-            } else if fname.contains("cl")
-                && !fname.contains("cloudabi")
-                && !fname.contains("uclibc")
-                && !fname.contains("clang")
-            {
+            } else if fname.ends_with("cl") || fname == "cl.exe" {
                 ToolFamily::Msvc { clang_cl: false }
             } else if fname.contains("clang") {
                 match clang_driver {


alexcrichton:
Thanks!

PR #582, opened 2021-02-14
Pick version for iOS simulator target

badboy:
See https://github.com/rust-lang/rust/pull/81966 for the target implementation and my research.


alexcrichton:
👍

PR #583, opened 2021-02-17
Match abi and march settings for riscv on linux

kraj:
linux distros have agreed on using lp64d on 64bit and
ilp32d on 32bit for ABI, we also need to match the march
values which can support these ABIs and therefore use rv64gc
and rv32gc for march

Fixes errors when compiling libstd-rs e.g.

warning: Assembler messages:
warning: Error: -march=rv64: first ISA extension must be `e', `i' or `g'
warning: cc1plus: error: '-march=rv64': first ISA subset must be 'e', 'i' or 'g'


alexcrichton:
Thanks!

PR #588, opened 2021-03-10
Added a track-dependencies feature to only recompile when needed

JoNil:
My motivation for this is that at work we have a fairly large mixed rust and cpp codebase that we would like to transition into building completely with cargo.

This pull request adds a feature that will tell the c compiler to emit dependency information for each file. Then on a subsequent build the timestamp of each of the dependent files are checked to see if any of them is newer then the object file, Only then is the compile redone.

If having json as an optional dependency is completely out of the question i can write a manual parser for the msvc dependency information format.

Additionally rerun directives are printed so that cargo can know if the build script wants to rerun.

I have tested this with msvc, clang and gcc.

I also removed some prints that made it hard for me to spot errors in my cpp files.

I also fixed a bug where if parallel build is active the build script would claim the build succeeded even if a object file failed to compile.

To be conservative i made the feature off by default even if the feature flag is on.

PS: Do you squash on merge or should i squash manually?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,12 @@ edition = "2018"
 
 [dependencies]
 jobserver = { version = "0.1.16", optional = true }
+json = { version = "0.12", optional = true }
 
 [features]
+default = ["track-dependencies"]
 parallel = ["jobserver"]
+track-dependencies = ["json"]
 
 [dev-dependencies]
 tempfile = "3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ jobserver = { version = "0.1.16", optional = true }
 json = { version = "0.12", optional = true }
 
 [features]
-default = ["track-dependencies"]
 parallel = ["jobserver"]
 track-dependencies = ["json"]
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -978,14 +978,32 @@ impl Build {
         }
         self.compile_objects(&objects)?;
 
+        let msvc = self.get_target()?.contains("msvc");
+
         #[cfg(feature = "track-dependencies")]
-        for obj in &objects {
-            track_dependencies::emit_rerun_directives(obj);
+        if self.track_dependencies {
+            let mut unique_deps = std::collections::HashSet::new();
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks for the PR! Unfortuantely though I'm pretty hesitant to tackle this. Managing incremental build is generally quite complex and very error-prone, and it's notoriously hard to get right. There's a number of downsides with this current implementation, for example:

* I don't think this takes into account changing flags to the C compiler
* I don't think this accounts for changing C compilers
* There currently aren't any tests
* The dep-info parser seems a bit too simple to be correct perhaps?

I realize that this is a major missing feature from this crate, but I'm honestly not sure that I have the steam as a maintainer to take on this new feature. I think my preferred route to have this implemented would be to either:

1. Recommend ccache
2. Ensure that a caching crate can be built on top of `cc` and hand off maintainership of that to someone else.

I'm unfortunately just not sure that I would myself be willing to take this on and maintain it.

JoNil:
I can definitely understand being hesitant to maintain this. And yes incremental builds are hard to get right.

- I don't think this takes into account changing flags to the C compiler
This is handled by writing the command line used to build the object to a file and on build i compare the contents of that file to the current command line.

- I don't think this accounts for changing C compilers
Changing compiler from clang to gcc would be handled by the above solution, changing the link for clang from clang 10 to 11 for example is not handled.

- There currently aren't any tests
If you decide that this feature in some form is something that you want then i can definitely add some :)

- The dep-info parser seems a bit too simple to be correct perhaps?
Maybe, the format is quite simple :) Admittedly i have not tested it as thoroughly on linux as on windows.

We will start using it to build our project soon then hopefully we will find out if there are any edge cases that are currently not handled.

Let me know if you would like to go forward with this, then i will add tests and any other changes that you would want.

alexcrichton:
Sorry but I personally do not want to land this in this repository. I'd hope that this could be built on top of `cc` as-is, and if you need anything else exposed by `cc` I think it's fine to add that, but I am not up for wading into the world of incremental builds for C/C++

JoNil:
ok!

PR #592, opened 2021-03-23
Fix typo in `atl_paths` ("atlfmc" --> "atlmfc")

ColinFinck:
I haven't checked what uses this code path, but this is an obvious typo.


alexcrichton:
Thanks!

PR #599, opened 2021-05-04
allow cachepot als alternative to sccache

drahnr:
`cachepot` will be a maintained fork with a different security paradigma than the original `sccache`, it will retain the ability to build C/C++ and other common toolchains, yet will use a different threat model and will rejuvenate a lot of the internals.

As such we would like to allow-list the name `cachepot` in the `cc` crate to retain the ability to build c dependencies.

https://github.com/paritytech/sccache/pull/50


alexcrichton:
Seems reasonable to me!

drahnr:
It took a bit longer than anticipated, now we went ahead and did the rename to `cachepot`.

It'd be awesome if this could be merged!

PR #600, opened 2021-05-14
Improve docs for the C++ usage

SallySoul:
I was having some issues with the c++ standard library on different platforms (since resolved), but that lead me to notice that there was a broken doc link for get_cpp_link_stdlib, which wasn't public, so its documentation is not in the html docs.

Making get_cpp_link_stdlib public provides a good place for the missing information in the html docs, so if only for that reason I think it should be public. Of course, I understand if y'all disagree.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -592,6 +592,12 @@ impl Build {
     ///
     /// The other `cpp_*` options will only become active if this is set to
     /// `true`.
+    ///
+    /// The name of the C++ standard library to link is decided by:
+    /// 1. If [cpp_link_stdlib](Build::cpp_link_stdlib) is set, use its value.
+    /// 2. Else if the `CXXSTDLIB` environment variable is set, use its value.
+    /// 3. Else the default is `libc++` for OS X and BSDs, `libc++_shared` for Android,
+    /// `None` for MSVC and `libstdc++` for anything else.
     pub fn cpp(&mut self, cpp: bool) -> &mut Build {
         self.cpp = cpp;
         self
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR, but do you have an intended use for making this function public? It sounds like it's just here for documentation? If that's the case could this perhaps be recast differently?

SallySoul:
That's fair, I don't have a use case in mind for exposing that function. However, I found the the doc for `get_cpp_link_stdlib` useful, and I think it could be a good addition to the public docs. Maybe adding a version of the comment to the `Build::cpp` function might be useful. Something like

```
/// Set C++ support.
///
/// The other `cpp_*` options will only become active if this is set to
/// `true`.
///
/// The name of the c++ standard library to link is decided by:
/// 1. If [cpp_link_stdlib](cc::Build::cpp_link_stdlib) is set, use its value.
/// 2. Else if the `CXXSTDLIB` environment variable is set, use its value.
/// 3. Else the default is `libc++` for OS X and BSDs, `libc++_shared` for Android,
/// `None` for MSVC and `libstdc++` for anything else.
```

On the other hand, I don't want to waste your time bike shedding this, feel free to close.

alexcrichton:
I'm always fine with tweaking docs! I'm hesitant to expose a function just because it's there though, I would prefer to keep things private if possible.

SallySoul:
I appreciate that! I revised the update accordingly.

JohnTitor (approved):
Approving as this now touches the docs only. Thanks!

PR #603, opened 2021-05-22
MSVC: Don't use environment variables to find tools if the targets differ

ChrisDenton:
On Windows, when using a "Developer" powershell or command prompt, the `"VCINSTALLDIR"` environment variable is set. Currently if this is detected by `cc-rs`, then it will unconditionally assume the environment has already been correctly set up.

This PR adds a check to make sure the Visual Studio target matches the requested Rust target. If not then it falls back to using automatic detection.

Fixes: https://github.com/rust-lang/rust/issues/43468

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -47,8 +47,6 @@ pub fn find_tool(_target: &str, _tool: &str) -> Option<Tool> {
 /// Documented above.
 #[cfg(windows)]
 pub fn find_tool(target: &str, tool: &str) -> Option<Tool> {
-    use std::env;
-
     // This logic is all tailored for MSVC, if we're not that then bail out
     // early.
     if !target.contains("msvc") {
@@ -66,18 +64,6 @@ pub fn find_tool(target: &str, tool: &str) -> Option<Tool> {
         return impl_::find_devenv(target);
     }
 
-    // If VCINSTALLDIR is set, then someone's probably already run vcvars and we
-    // should just find whatever that indicates.
-    if env::var_os("VCINSTALLDIR").is_some() {
-        return env::var_os("PATH")
-            .and_then(|path| {
-                env::split_paths(&path)
-                    .map(|p| p.join(tool))
-                    .find(|p| p.exists())
-            })
-            .map(|path| Tool::with_family(path.into(), MSVC_FAMILY));
-    }
-
     // Ok, if we're here, now comes the fun part of the probing. Default shells
     // or shells like MSYS aren't really configured to execute `cl.exe` and the
     // various compiler tools shipped as part of Visual Studio. Here we try to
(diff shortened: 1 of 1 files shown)


ChrisDenton:
After feedback, I've converted this to a draft until I have a moment to update it.

There are concerns with my current approach and how it relies on an environment variable that may not be set for older versions of Visual Studio.

ChrisDenton:
This PR now adds a `find_msvc_environment` function that's intended to be used the same way as the other `find_msvc_*` probe functions. It currently only makes a difference with VS 2017 and 2019 but will fallback to the old behaviour on older versions of Visual Studio. A future PR could expand the function to better support more versions.

This PR shouldn't impact situations where `cc` was already working properly, but it will help in some cases where it used to give libs for the wrong arch.

src/windows_registry.rs
```diff
@@ -218,6 +205,48 @@ mod impl_ {
...
+        };
+        Some(vscmd_arch == arch)
+    }
+
+    /// Attempt to find the tool using environment variables set by vcvars.
+    pub fn find_msvc_environment(target: &str, tool: &str) -> Option<Tool> {
+        // Early return if the environment doesn't contain a VC install.
+        if env::var_os("VCINSTALLDIR").is_none() {
+            return None;
+        }
```
alexcrichton, line 230: I think this statement isn't necessary due to the `?` on the one below?

src/windows_registry.rs
```diff
@@ -218,6 +205,48 @@ mod impl_ {
...
+        let vs_install_dir = env::var_os("VSINSTALLDIR")?.into();
+
+        // If the vscmd target differs from the requested target then
+        // attempt to get the tool using the VS install directory.
+        if is_vscmd_target(target) == Some(false) {
+            // We will only get here with versions 15+.
+            tool_from_vs15plus_instance(tool, target, &vs_install_dir)
```
alexcrichton, line 237: I think this should perhaps return `None` to preserve what the previous code did? That way the `.or_else` chain up above will fall through to the next branch?

alexcrichton (commented):
Seems reasonable to me!

src/windows_registry.rs
ChrisDenton: The one below is slightly different in that it's `VS*` instead of `VC*`. I was trying to account for the possiblity that Visual Studio (VS) is installed but the Visual C (VC) component isn't.

src/windows_registry.rs
ChrisDenton: I wanted to try using the install directory given in the environment variable instead of probing for it. My assumption being that the user's intent was to use that instance of Visual Studio even though the given targets don't match up. But I can change it if that's not appropriate.

src/windows_registry.rs
alexcrichton: Oh oops missed that, don't mind me!

src/windows_registry.rs
alexcrichton: Ah ok makes sense! Just wanted to double check, sounds like you know a lot more about these installs than I do!

alexcrichton:
Thanks for this!

PR #607, opened 2021-06-06
Add `-arch` compiler flag on older macOS systems

MarcusCalhoun-Lopez:
I am trying to get Rust working on older macOS systems.
However, the `-arch` flag is only added on x86_64 and aarch64 systems.


alexcrichton:
Thanks!

PR #612, opened 2021-07-25
Improve CUDA support

dot-asm:
It's rather tricky to use the CUDA compiler with cc-rs, because as it is now you would customarily get a bunch of unresolved symbols linking errors. Answer is a) perform CUDA-specific `--device-link` step, b) link with CUDA run-time. ~The latter is solved as a feature.~

For reference, this is similar to #508, but things are done in a more Windows-friendly manner and a CI test is added.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -103,6 +103,7 @@ pub struct Build {
     cpp_link_stdlib: Option<Option<String>>,
     cpp_set_stdlib: Option<String>,
     cuda: bool,
+    cudart: Option<String>,
     target: Option<String>,
     host: Option<String>,
     out_dir: Option<PathBuf>,
@@ -298,6 +299,7 @@ impl Build {
             cpp_link_stdlib: None,
             cpp_set_stdlib: None,
             cuda: false,
+            cudart: None,
             target: None,
             host: None,
             out_dir: None,
@@ -612,6 +614,18 @@ impl Build {
         if cuda {
             self.cpp = true;
         }
+        self
+    }
+
+    /// Link CUDA run-time.
+    ///
+    /// This option mimics the `--cudart` nvcc command-line option. Just like
+    /// the original it accepts `{none|shared|static}`. However, in case you
+    /// don't use this method when instantiating the `Build` class, the outcome
+    /// is equivalent to `none`. This is different from nvcc's default 'static'.
+    pub fn cudart(&mut self, cudart: &str) -> &mut Build {
+        self.cuda(true);
+        self.cudart = Some(cudart.to_string());
         self
     }
 
@@ -993,6 +1007,56 @@ impl Build {
         if self.cpp {
             if let Some(stdlib) = self.get_cpp_link_stdlib()? {
                 self.print(&format!("cargo:rustc-link-lib={}", stdlib));
+            }
+        }
+
+        let cudart = match &self.cudart {
+            Some(opt) => opt.as_str(), // {none|shared|static}
+            None => "none",
+        };
+        if cudart != "none" {
+            if let Some(nvcc) = which(&self.get_compiler().path) {
+                // Try to figure out the -L search path. If it fails,
+                // it's on user to specify one by passing it through
+                // RUSTFLAGS environment variable.
+                let mut libtst = false;
+                let mut libdir = nvcc;
+                libdir.pop(); // remove 'nvcc'
+                libdir.push("..");
+                let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+                if cfg!(target_os = "linux") {
+                    libdir.push("targets");
+                    libdir.push(target_arch.to_owned() + "-linux");
+                    libdir.push("lib");
+                    libtst = true;
+                } else if cfg!(target_env = "msvc") {
+                    libdir.push("lib");
+                    match target_arch.as_str() {
+                        "x86_64" => {
+                            libdir.push("x64");
+                            libtst = true;
+                        }
+                        "x86" => {
+                            libdir.push("Win32");
+                            libtst = true;
+                        }
+                        _ => libtst = false,
+                    }
+                }
+                if libtst && libdir.is_dir() {
+                    println!(
+                        "cargo:rustc-link-search=native={}",
+                        libdir.to_str().unwrap()
+                    );
+                }
+
+                // And now the -l flag.
+                let lib = match cudart {
+                    "shared" => "cudart",
+                    "static" => "cudart_static",
+                    bad => panic!("unsupported cudart option: {}", bad),
+                };
+                println!("cargo:rustc-link-lib={}", lib);
             }
         }
 
@@ -3117,4 +3181,24 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &str) -> Option<
     } else {
         None
     }
+}
+
+fn which(tool: &Path) -> Option<PathBuf> {
+    fn check_exe(exe: &mut PathBuf) -> bool {
+        let exe_ext = std::env::consts::EXE_EXTENSION;
+        exe.exists() || (!exe_ext.is_empty() && exe.set_extension(exe_ext) && exe.exists())
+    }
+
+    // If |tool| is not just one "word," assume it's an actual path...
+    if tool.components().count() > 1 {
+        let mut exe = PathBuf::from(tool);
+        return if check_exe(&mut exe) { Some(exe) } else { None };
+    }
+
+    // Loop through PATH entries searching for the |tool|.
+    let path_entries = env::var_os("PATH")?;
+    env::split_paths(&path_entries).find_map(|path_entry| {
+        let mut exe = path_entry.join(tool);
+        return if check_exe(&mut exe) { Some(exe) } else { None };
+    })
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1205,6 +1205,9 @@ impl Build {
         if !msvc || !is_asm || !is_arm {
             cmd.arg("-c");
         }
+        if self.cuda && self.files.len() > 1 {
+            cmd.arg("--device-c");
+        }
         cmd.arg(&obj.src);
         if cfg!(target_os = "macos") {
             self.fix_env_for_apple_os(&mut cmd)?;
@@ -1809,6 +1812,21 @@ impl Build {
             .collect();
         for chunk in objs.chunks(100) {
             self.assemble_progressive(dst, chunk)?;
+        }
+
+        if self.cuda {
+            // Link the device-side code and add it to the target library,
+            // so that non-CUDA linker can link the final binary.
+
+            let out_dir = self.get_out_dir()?;
+            let dlink = out_dir.join(lib_name.to_owned() + "_dlink.o");
+            let mut nvcc = self.get_compiler().to_command();
+            nvcc.arg("--device-link")
+                .arg("-o")
+                .arg(dlink.clone())
+                .arg(dst);
+            run(&mut nvcc, "nvcc")?;
+            self.assemble_progressive(dst, &[dlink])?;
         }
 
         let target = self.get_target()?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -613,19 +613,21 @@ impl Build {
         self.cuda = cuda;
         if cuda {
             self.cpp = true;
+            self.cudart = Some("static".to_string());
         }
         self
     }
 
     /// Link CUDA run-time.
     ///
-    /// This option mimics the `--cudart` nvcc command-line option. Just like
-    /// the original it accepts `{none|shared|static}`. However, in case you
-    /// don't use this method when instantiating the `Build` class, the outcome
-    /// is equivalent to `none`. This is different from nvcc's default 'static'.
+    /// This option mimics the `--cudart` NVCC command-line option. Just like
+    /// the original it accepts `{none|shared|static}`, with default being
+    /// `static`. The method has to be invoked after `.cuda(true)`, or not
+    /// at all, if the default is right for the project.
     pub fn cudart(&mut self, cudart: &str) -> &mut Build {
-        self.cuda(true);
-        self.cudart = Some(cudart.to_string());
+        if self.cuda {
+            self.cudart = Some(cudart.to_string());
+        }
         self
     }
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -87,6 +87,25 @@ jobs:
     - run: cargo test ${{ matrix.no_run }} --manifest-path cc-test/Cargo.toml --target ${{ matrix.target }} --features parallel
     - run: cargo test ${{ matrix.no_run }} --manifest-path cc-test/Cargo.toml --target ${{ matrix.target }} --release
 
+  cuda:
+    name: Test CUDA support
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@master
+    - name: Install cuda-minimal-build-11-4
+      shell: bash
+      run: |
+        # https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=20.04&target_type=deb_network
+        wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-ubuntu2004.pin
+        sudo mv cuda-ubuntu2004.pin /etc/apt/preferences.d/cuda-repository-pin-600
+        sudo apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/7fa2af80.pub
+        sudo add-apt-repository "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /"
+        sudo apt-get update
+        sudo apt-get -y install cuda-minimal-build-11-4
+    - name: Test 'cudart' feature
+      shell: bash
+      run: env PATH=/usr/local/cuda/bin:$PATH cargo test --manifest-path cc-test/Cargo.toml --features test_cuda
+
   msrv:
     name: MSRV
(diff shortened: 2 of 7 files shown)


dot-asm:
Just in case, reference to "Windows friendliness" means that I can confirm that it works on Windows. Well, `cargo test --manifest-path cc-test/Cargo.toml --features cudart` doesn't as is, but it can be rectified by moving the new `which(nvcc)` thing up in `cc-test/build.rs`, more specifically prior to `Test that the 'windows_registry' module ...`, because it wipes %PATH%. Arranging CI test on Windows is problematic because of the lack of usable installation instructions.

Cargo.toml
```diff
@@ -19,9 +19,11 @@ edition = "2018"
 
 [dependencies]
 jobserver = { version = "0.1.16", optional = true }
+which = { version = "^4.0", optional = true }
```
alexcrichton, line 22: I'd ideally like to keep the deps of this crate pretty `slim`, and with `env::split_path` could  the `which` crate be replaced with an iteration over `PATH`?

src/lib.rs
```diff
@@ -996,6 +996,55 @@ impl Build {
             }
         }
 
+        #[cfg(feature = "cudart")]
```
alexcrichton, line 999: I think it might be best to use `if self.cuda && cfg!(feature = "cudart") { .. }` here since that means we can type-check this for other platforms instead of having cuda-only issues cropping up in the future.

alexcrichton: (same for some of the checks below)

alexcrichton (commented):
Seems reasonable to me! I don't know anything about cuda really but I'm happy to defer!

src/lib.rs
dot-asm: > (same for some of the checks below)

You mean ones for linux and msvc?

Meanwhile for reference, the original reason for `#[cfg(feature = "cudart")]` was to make dependency on `which` optional. I mean majority don't use cuda, and so they wouldn't need `which`...

Cargo.toml
dot-asm: OK, I'll see what I can do... BTW, is it OK to keep it in cc-test/Cargo.toml?

dot-asm:
I apologize for the noise from CI. Rust is not my language of 1st choice:-)

Cargo.toml
dot-asm: The trouble is that on Windows `which(x)` have to check for multiple file extensions (customarily listed in %PATHEXT% variable). So here is the question. Is it OK if the replacement subroutine won't be general-purpose `which`, but tailored for this specific use case? More specifically it would check for single `.exe` extension.

alexcrichton: oh yeah keeping it for `cc-test` is fine, but I think that this is only used for `nvcc`, right? If that's the case is `nvcc.exe` always the target? Or is it  frquently wrapped with other things like `.cmd`? (if it's always `.exe` we can just hardcode that)

src/lib.rs
alexcrichton: Ah good point about the conditional `which` dependency, but in any case the current structure of the code is what I was thinking of, thanks!

Cargo.toml
dot-asm: Checking for `.exe` alone is sufficient for the `cudart` purposes, totally. The concern was rather that if you add your own `which` subroutine, you run a risk that somebody else will want to use it, and then they are likely to have certain expectations. And the question was if it would be appropriate to curb these possible expectations. Answer appears to be "yes," and dependency on `which` crate is now removed.

Once again I want to point out that Rust is not my 1st-choice language, so that if you have any styling or any other nits, don't hold back:-)

dot-asm: > Checking for `.exe` alone is sufficient for the `cudart` purposes, totally.

Just in case, why "totally." `.exe` takes precedence in PATH evaluation, so that you can't have a wrapper script called `nvcc.cmd` at the same location as `nvcc.exe` and keep using `nvcc` as command, you would have to call it explicitly as `nvcc.cmd`. In which case you would have to set NVCC environment variable if you want to use it with `cc-rs` anyway.

src/lib.rs
```diff
@@ -3100,3 +3161,24 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &str) -> Option<
...
+    }
+
+    // If |tool| is not just one "word," assume it's an actual path...
+    if tool.components().count() > 1 {
+        let mut exe = PathBuf::from(tool);
+        return if check_exe(&mut exe) { Some(exe) } else { None };
+    }
+
+    // Loop through PATH entries searching for the |tool|.
+    env::var_os("PATH").as_ref().and_then(|path_entries| {
```
alexcrichton, line 3178: Here I think you can use `?` like so:

```rust
let path_entries = env::var_os("PATH")?;
// ...
```

src/lib.rs
```diff
@@ -3100,3 +3161,24 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &str) -> Option<
...
+        env::split_paths(path_entries).find_map(|path_entry| {
+            let mut exe = path_entry.join(tool);
+            return if check_exe(&mut exe) { Some(exe) } else { None };
```
alexcrichton, line 3181: The `return` here can be elided (the last expression in a closure is an implicit return)

src/lib.rs
```diff
@@ -3100,3 +3161,24 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &str) -> Option<
         None
     }
 }
+
+fn which(tool: &PathBuf) -> Option<PathBuf> {
```
alexcrichton, line 3165: FWIW when taking an argument by reference, `&Path` is typically preferred over `&PathBuf` (not that it'll have any real impact here, just idioms)

src/lib.rs
```diff
@@ -996,6 +996,49 @@ impl Build {
...
 
+        if self.cuda && cfg!(feature = "cudart") {
+            match which(&self.get_compiler().path) {
+                None => (),
+                Some(nvcc) => {
+                    // Try to figure out the -L search path. If it fails,
+                    // it's on user to specify one by passing it through
+                    // RUSTFLAGS environment variable.
+                    let mut libtst = false;
+                    let mut libdir = nvcc.canonicalize().unwrap();
```
alexcrichton, line 1007: I've found historically that `canonicalize()` can fail for weird reasons, but is that necesary to do here? We already find `nvcc` in `PATH` so I don't think we may need to canonicalize here?

src/lib.rs
```diff
@@ -996,6 +996,49 @@ impl Build {
             }
         }
 
+        if self.cuda && cfg!(feature = "cudart") {
+            match which(&self.get_compiler().path) {
+                None => (),
+                Some(nvcc) => {
```
alexcrichton, line 1002: With an empty `None` branch you can prpobably scope this all as:

```rust
if let Some(nvcc) = which(..) {
  // ...
}
```

src/lib.rs
```diff
@@ -996,6 +996,49 @@ impl Build {
...
+                    // Try to figure out the -L search path. If it fails,
```
alexcrichton, line 1003: Out of curiosity, do you know why `nvcc` doesn't do this logic automatically? Is there documentation we can point to somewhere recommending this `-L` path and such?

alexcrichton (commented):
Ok sounds good to me about the `which` business. I'll take on the responsibility of dealing with those bugs if they ever arise in the future :)

src/lib.rs
dot-asm: Note that `PATH` is evaluated only in the case `nvcc` here is a "single-word" command. Otherwise one can set the NVCC environment variable to explicit path to whatever you want to call `nvcc`. The idea behind `canonicalize()` is to handle the case when the environment variable is set to **relative** path. One can argue that it's an edge case that is not worth covering. If you want to say "scrap it," then just say the word:-)

src/lib.rs
dot-asm: It does do this automatically, but the problem is that `cargo` doesn't call `nvcc` to link the final binary. Instead is calls the system linker, and then you have to craft -L and -l yourself. Of course ideal would be if one could query `nvcc` and ask "what's your library path," but there is no such option:-(

Hmmm, when I think about it... **If** there is a way to dynamically override the target `linker`, and set it to `nvcc`, it might work too... In such case one wouldn't need this -L thing, but then cross-compilation might get tricky. [Just in case, reference to "cross-compilation" means that it does work, I can cross-compile CUDA program for aarch64.]

src/lib.rs
dot-asm: It says "you might have meant to ... `return Some(exe);`"...

alexcrichton: Oh that probably means you need to remove the semicolon at the end of the line as well, changing it to:

```rust
if check_exe(&mut exe) { Some(exe) } else { None }
```

src/lib.rs
alexcrichton: If `nvcc` is set to a relative path is there an issue with passing a relative path as a `-L` argument? I would naively expect that to work ok, which I think means that the `canonicalize()` probably isn't necessary here.

Cargo.toml
```diff
@@ -22,6 +22,7 @@ jobserver = { version = "0.1.16", optional = true }
 
 [features]
 parallel = ["jobserver"]
+cudart = []
```
alexcrichton, line 25: Oh one thought I just had, now that this is no longer controlling an optional dependency it's probably best to have this be a configuration method on `Build` rather than a compile-time feature

src/lib.rs
alexcrichton: The final linker is actually invoked by rustc rather than Cargo, and the target's default can be overridden with `-Clinker=/path/to/nvcc` (which can also be configured in cargo either via `RUSTFLAGS` or as a string with `CARGO_TARGET_..._LINKER=...`)

Is this a case where the Rust target should be defaulting to `nvcc` rather than `cc`?

(I should point out again I'm very naive about CUDA/nvcc/etc, I naively assume right now that there's one Rust target that everyone uses for GPU things (and that nvcc is for GPU things) and that you're compiling for that target)

src/lib.rs
dot-asm: Right, that's what I tried, and it did complain, 1.53.

src/lib.rs
dot-asm: It's surely just me, but I tend to think of unlike cases. Let's say you want to compile with NVCC set to './nvcc'. Of course nobody in their right mind would actually do it, but it's not really an excuse to disregard the possibility. Without `canonicalize()` the current code will break. Alternative can be to not rely on `libdir.pop(), libdir.pop()`, but do `libdir.pop(), libdir.push("..")`... This way most unlike case is covered without `canonicalize()`. Let me test how it works...

src/lib.rs
dot-asm: > Is this a case where the Rust target should be defaulting to `nvcc` rather than `cc`?

Yes. Or rather that's what I had in mind, it remains to be seen if it actually works all the way.

Cargo.toml
dot-asm: I don't quite follow the "on `Build`" thing. Does it refer to `[build-dependecies]`?

src/lib.rs
dot-asm: > I naively assume right now that there's one Rust target that everyone uses for GPU things (and that nvcc is for GPU things) and that you're compiling for that target.

nvcc is available for x86_64, arm64 and ppc64le, And it can do cross-compilations. It's most common scenario when you run eclipse on x86_64 and target remote arm64.

src/lib.rs
dot-asm: `canonicalize()` is gone.

src/lib.rs
dot-asm: Setting LINKER to nvcc doesn't work, because it doesn't recognize all system's linker flags.

Cargo.toml
dot-asm: For reference. Original rationale behind making `cudart` a feature was to ensure that nothing gets broken. I mean if somebody is using `cc-rs` for CUDA **now**, they ought to have some kind of solution for the linking problem. And they wouldn't directly appreciate the interference if `cc-rs` just started linking cudart. In other words it's only appropriate to make it optional and off by default. And I figured that a feature would be least invasive. But it's naturally not the only option. For example, we could instead add `.cudart()` **method**. And not as complementary, but as replacement for `.cuda()`. I mean application would have a choice to use `.cuda()` or `.cudart()`, and not `.cuda()` **and** `.cudart()`.

src/lib.rs
```diff
@@ -1205,6 +1245,9 @@ impl Build {
         if !msvc || !is_asm || !is_arm {
             cmd.arg("-c");
         }
+        if self.cuda {
+            cmd.arg("--device-c");
+        }
```
dot-asm, line 1250: Additional thought. And I suppose it's rather a question to @YoshikawaMasashi. I've effectively taken this from #508, but the flag is not **generally** required, only when you compile multiple modules. For this reason it might be more appropriate to shift the burden to application and suggest to add `.flag("--devce-c")` when constructing the compilation. Alternatively one can make the addition contingent on amount of files being compiled.

Cargo.toml
dot-asm: > I don't quite follow the "on `Build`" thing.

Ah! I think I got it now. And the above remark about `.cudart()` is effectively a reply to the suggestions. I suppose remaining question is if `.cudart()` would be considered suitable. Or maybe `.link_with_cudart()` would be considered more appropriate?

dot-asm: A method that starts with "cuda" can be considered preferred, because it would appear next to `.cuda()` method in documentation. There already is similar precedent, `.cpp_link_stdlib()`. So, `.cudart()` or `.cuda_link_cudart()`? Or maybe `.cuda_link_rt()`?

dot-asm: I've decided to go for `.cudart()`, because corresponding nvcc command-line option is called `--cudart`. And to mimick it even closer it accepts same options `{none|shared|static}`.

src/lib.rs
```diff
@@ -996,6 +1010,56 @@ impl Build {
             }
         }
 
+        let cudart = match &self.cudart {
+            Some(opt) => opt.as_str(),
+            None => "none",
+        };
+        if self.cuda && cudart != "none" {
```
alexcrichton, line 1017: I think here you can probably just do `if let Some(cudart) = &self.cudart { ... }` because if someone specifies cudart then `self.cuda` is automatically set.

Cargo.toml
alexcrichton: Ah sorry, but what you ended up on is what I was thinking!

src/lib.rs
alexcrichton: I'll have to defer to your judgement on this, I don't think I know enough about cuda to know what the best option is here.

src/lib.rs
alexcrichton: Oh it might be the case that rustc needs more invasive knowledge about nvcc if it runs nvcc, there's a few flavors of linkers implemented in rustc and nvcc may need its own flavor (unless it looks exactly like gcc, which sounds like that's not the case)

src/lib.rs
dot-asm: If I `let Some(cudart) = &self.cudart`, I'll still have to check for "none". Because cudart being `None` and `Some("none")` are equivalent cases. But yes, I can omit `self.cuda`...

alexcrichton: Oh how come `Some("none")` is equivalent to `None`? (I thought that was accidental, didn't realize it was intentional)

dot-asm: `self.cuda` is removed, added a commentary to remind that `Some("none")` is an option. I've also spotted error, I recognized "dynamic" instead of "shared."

dot-asm: > I think here you can probably just do `if let Some(cudart) = &self.cudart { ... }` because if someone specifies cudart then `self.cuda` is automatically set.

Since `.cudart()` aims to mimic `--cudart`, it recognizes all `--cudart`'s options, {none|shared|static}. So that nvcc's users feel "home."

src/lib.rs
dot-asm: I went for additional `self.files.len() > 1`.

dot-asm:
As for mimicking the `--cudart` compiler option. Currently not calling `.cudart()` method is equivalent to `--cudart none`, which is **not** nvcc's default. The rationale behind this suggestion is following. I would assume that if somebody is using the cc-rs with CUDA now, they have to do something about linking. And making sudden moves like forcing equivalent of `--cudart static` can interfere with whatever current users do. Now, one can make case for **not** worrying about the said interference, enforce the nvcc's default, and tell users to adjust. I personally would prefer the nvcc's default, but I'm merely starting with Rust-CUDA combo, so my opinion doesn't weigh much:-)

dot-asm:
> one can make case for not worrying about the said interference, enforce the nvcc's default, and tell users to adjust.

The top-most commit, "Harmonize CUDA support with NVCC default --cudart static," is to show how enforcing the nvcc's default looks like in practice. Whether or not it's accepted is up to the maintainer. I for one am obviously in favour:-) I'm pinging earlier CUDA support contributors, @peterhj, @trolleyman, @YoshikawaMasashi, for possible feedback.

alexcrichton:
Ok that all sounds good to me, and I think that this is good to go from at least my end in terms of idioms and such. Are you comfortable with the implemented semantics here? If so I can hit the button

dot-asm:
It might be appropriate to give some time for other contributors to weigh in. Naturally if they choose to... To recap, the question is if it would be appropriate to start linking with cudart by default. As opposed to having user add an additional method call to the `Builder`.

alexcrichton:
Ok well in any case this doesn't change the defaults as-is but I think it's fine to consider changing the defaults in the future as well. I'm gonna go ahead and merge this.

dot-asm:
Thank you for being patient! :-) If there are any, even remotely related problems, don't hesitate to tag my handle. I'll be happy to help. Cheers.

PR #619, opened 2021-09-03
Prefer cfg(target_feature = "crt-static") over CARGO_CFG_TARGET_FEATURE variable

wravery:
This is a fix for a bug I found using `cxx` and reported there: https://github.com/dtolnay/cxx/issues/929. It may be a workaround for a bug in `cargo` itself, I'm not sure.

If you use `.cargo/config.toml` to add the `crt-static` target feature like so, `cargo` applies it to the `rustc` command line and it applies to the `cfg`, but it is not updating the `CARGO_CFG_TARGET_FEATURE` environment variable to reflect that:
```toml
[target.'cfg(target_env = "msvc")']
rustflags = ["-C", "target-feature=+crt-static"]
```

As a result, since this crate only checks the environment variable, the default value of `Build::static_crt` is wrong if you add the target feature this way. I think the `cfg` option should always take precedence since that's what determines how `cargo` will invoke `rustc` for any files that would be linked with files compiled using `cc`. If it's not specified, I left in conditionally compiled tests for the environment variable, so if there are packages/builds depending on that they should continue to work.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1447,10 +1447,7 @@ impl Build {
                     Some(true) => "-MT",
                     Some(false) => "-MD",
                     None => {
-                        let features = self
-                            .getenv("CARGO_CFG_TARGET_FEATURE")
-                            .unwrap_or(String::new());
-                        if features.contains("crt-static") {
+                        if self.get_crt_static_feature() {
                             "-MT"
                         } else {
                             "-MD"
@@ -1613,13 +1610,8 @@ impl Build {
                     }
                 }
 
-                if self.static_flag.is_none() {
-                    let features = self
-                        .getenv("CARGO_CFG_TARGET_FEATURE")
-                        .unwrap_or(String::new());
-                    if features.contains("crt-static") {
-                        cmd.args.push("-static".into());
-                    }
+                if self.static_flag.is_none() && self.get_crt_static_feature() {
+                    cmd.args.push("-static".into());
                 }
 
                 // armv7 targets get to use armv7 instructions
@@ -2673,6 +2665,18 @@ impl Build {
                 &format!("Environment variable {} not defined.", v.to_string()),
             )),
         }
+    }
+
+    #[cfg(target_feature = "crt-static")]
+    fn get_crt_static_feature(&self) -> bool {
+        true
+    }
+
+    #[cfg(not(target_feature = "crt-static"))]
+    fn get_crt_static_feature(&self) -> bool {
+        self.getenv("CARGO_CFG_TARGET_FEATURE")
+            .unwrap_or(String::new())
+            .contains("crt-static")
     }
 
     fn print(&self, s: &str) {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! Technically though this is not a correct fix. The unfortunate part is that to "fix" this issue I'd have to redirect you to Cargo itself. Let me give you some background as to why though.

The main reason that this fix is not technically correct is that what you're doing here is inspecting the flags that `cc` itself is built with, which may not be the flags that the final crate itself is being built with. The purpose of the environment variables is to inform build scripts how their output is going to get used, and this allows for different compilation settings between build scripts themselves and the final artifact. This means that configuration of crt-static that differs between build scripts and the final output, when desired, would no longer work with `cc`.

Now that all being said you've still got an issue on your hands you'd like to fix. The problem here is how you've specified `rustflags`. By doing this through Cargo's configuration which uses `cfg` expressions you're hitting a bug in Cargo. At some point Cargo executes `rustc --print cfg` to learn what to pass to a build script. At this time Cargo doesn't even know your target is `target_env = "msvc"` so it does not pass the `rustflags` you've configured to rustc when it prints `rustc --print cfg`. This means that rustc doesn't print the `crt-static` feature, which means Cargo doesn't inform the build script that the `crt-static` feature.

Afterwards, though, when Cargo actually reads `rustflags` it has the information that `target_env = "msvc"` so it actually loads up `rustflags` you've specified and passes it to the final crate. This causes Cargo to not pass `crt-static` as an enabled feature in the env vars to the build script, but it is indeed enabled in the final compilation artifact.

I believe possible workarounds are to use `[target.x86_64-pc-windows-msvc]` (explicitly listing the target) or the `RUSTFLAGS` environment variable. Would either of those work for you?

wravery:
> I believe possible workarounds are to use `[target.x86_64-pc-windows-msvc]` (explicitly listing the target) or the `RUSTFLAGS` environment variable. Would either of those work for you?

Yes, switching to an explicit target worked, thanks! `RUSTFLAGS` also works, but I didn't want to require any global/user-wide configuration for a downstream package. I was looking for a way to keep this target-feature in sync between, `cc`, a lib crate I'm writing, and bin crates that can consume it. It seems like the best way to do that is to suggest adding this to the `.cargo/config.toml` of the bin crate, which I just verified works:

```toml
[target.x86_64-pc-windows-msvc]
rustflags = ["-C", "target-feature=+crt-static"]
```

Thanks for the detailed explanation. Would you recommend submitting an issue for `cargo` to try and resolve that `cfg` difference? It sounds like you agree that's a bug and not expected behavior.

alexcrichton:
Ah ok great! If you're able to get things working locally then that relieves the pressure to fix this more urgently at least :)

In that case then yeah I think Cargo is the best place to open a bug about this. This'll be difficult to fix in Cargo since it sort of requires running `rustc --print cfg` to a fixed point as new `#[cfg]` directives are discovered, but nonetheless it's still a bummer that this happens. I don't think you're the first one to discover this about Cargo either.

If you're up for it you may want to browse the issue tracker since there may already be an issue open about this, but otherwise yeah I think an issue on Cargo would go well.

wravery:
Sounds good, let's close this PR then. 👍🏼

PR #621, opened 2021-09-18
add prefixes for x86 to support cross-compiling from non-x86 targets

infinity0:
otherwise cross-compiling from arm64 -> i686 fails, e.g. http://crossqa.debian.net/build/rustc_1.49.0+dfsg1-1_i386_20210907094750.log


infinity0:
The linux32 build is failing I think because it is relying on a x86_64-*-gcc to have multilib and using that instead of i686-linux-gnu-gcc (which is what one would use in a general cross-compiling $host-to-$target scenario). I'll try to think of an approach that can cater to both use-cases.

infinity0:
I've used the existing approach with `find_working_gnu_prefix` - all builds green now

alexcrichton:
Thanks for this! This makes sense to me, but I'm hesitant to do so because it's such a commonly used target this seems likely to regress someone somewhere by accident. Could this implement some form of fallback where if these compilers aren't found it still stick with `cc` as the default?

infinity0:
@alexcrichton that makes sense. I think this new version should do that? The only change since previously was to explicitly wrap the x86-64 case in `self.find_working_gnu_prefix()` as well. That way, if those compilers don't exist then the function will return None as it previously did. (This causes the caller to fall back to default "cc" as you want.)

Without the "// rustfmt wrap" workaround it wants to put the whole thing including comment on the same line which I thought was uglier, but happy to change this to whatever you prefer.

PR #627, opened 2021-10-13
Fix armhf build for GCC-11

infinity0:
GCC 11 changed the default flag to put the FP option in the -march flag (arm7-a+fp) rather than a specific FPU flag. Therefore it needs to be given separately now.

See https://bugs.launchpad.net/ubuntu/+source/gcc-defaults/+bug/1939379 for some more details.

Example build failure is here: https://buildd.debian.org/status/fetch.php?pkg=rustc&arch=armhf&ver=1.55.0%2Bdfsg1-1&stamp=1634161880&raw=0

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1638,6 +1638,11 @@ impl Build {
                     && (target.contains("-linux-") || target.contains("-kmc-solid_"))
                 {
                     cmd.args.push("-march=armv7-a".into());
+
+                    if target.ends_with("eabihf") {
+                        // lowest common denominator FPU
+                        cmd.args.push("-mfpu=vfpv3-d16".into());
+                    }
                 }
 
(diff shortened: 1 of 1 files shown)


infinity0:
Note: the snippet I added is already part of another armv7-a case further below in the file. I'm actually not entirely clear why there are 2 cases to handle these similar situations; both add armv7-a as a flag, but are based on the slightly different conditions `target.starts_with("armv7a")` vs `target.starts_with("armv7")`. Anyway, it works to fix the Debian build failure.

infinity0:
Working build with this patch applied: https://buildd.debian.org/status/fetch.php?pkg=rustc&arch=armhf&ver=1.55.0%2Bdfsg1-2&stamp=1634177225&raw=0

PR #628, opened 2021-11-02
fix: find_msvc_environment argument order

CurryPseudo:
According to the call from https://github.com/alexcrichton/cc-rs/blob/b2f6b146b75299c444e05bbde50d03705c7c4b6e/src/windows_registry.rs#L74-L79, the argument order is wrong


alexcrichton:
Thanks!

PR #629, opened 2021-11-03
Add armv7 assembly for cc-test

glandium:


alexcrichton:
Thanks! Is there any particular reason for adding this though?

glandium:
I independently rediscovered https://github.com/alexcrichton/cc-rs/pull/627 (since it's not in a release), and was trying to get a simple STR using cc-test, which was not building at all because of the lack of `armv7.S`.

alexcrichton:
Ah ok makes sense! FWIW I think the `_asm` symbol can be removed since that's just me being too lazy to use preprocessor differences to deal with macos/linux differences in the x86_64.S file.

PR #634, opened 2021-11-15
Fix cross-compilation from macOS using x86_64-elf-gcc

### LINKED ISSUE #633: x86_64-elf-gcc: error: unrecognized command-line option '-arch'
`-arch` flag is an extension in Apple's compilers, but not every compiler on the Apple platform supports it. Specifically, Homebrew GCC cross-compiler rejects this flag when cross-compiling from aarch64 (M1) to x64-64.

kornelski:
This avoids setting `-arch` flag if the compiler already has name of the architecture in the file name, assuming it's already a compiler for only the desired architecture.

Fixes #633

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1619,8 +1619,15 @@ impl Build {
                 if target.contains("darwin") {
                     if let Some(arch) = map_darwin_target_from_rust_to_compiler_architecture(target)
                     {
-                        cmd.args.push("-arch".into());
-                        cmd.args.push(arch.into());
+                        let is_target_arch_specific_compiler = cmd
+                            .path
+                            .file_name()
+                            .and_then(|f| f.to_str())
+                            .map_or(false, |f| f.starts_with(arch));
+                        if !is_target_arch_specific_compiler {
+                            cmd.args.push("-arch".into());
+                            cmd.args.push(arch.into());
+                        }
                     }
                 }
 


alexcrichton:
Can you describe more of the setup of why this error comes up? This sounds like a standard gcc flag isn't being accepted by gcc which sounds... fishy?

kornelski:
I'm using a cross-compiler from Homebrew to build x86-64 executables from ARM M1 Mac host: 

```bash
uname -m
# arm64
brew install x86_64-elf-gcc 
CC=x86_64-elf-gcc cargo build --target=x86_64-apple-darwin
```

I need to use the special compiler, because I need real GCC, not Apple-wrapped-clang, and the Homebrew `gcc-11` doesn't support cross-compilation to Intel:

```bash
gcc-11 -arch x86_64
gcc-11: warning: this compiler does not support X86 (arch flags ignored)
```

The `x86_64-elf-gcc` compiler build from Homebrew is very minimal, and has lots of things disabled/missing, including support for `-arch`.

```bash
x86_64-elf-gcc -arch x86_64
x86_64-elf-gcc: error: unrecognized command-line option '-arch'
```

kornelski:
It's not some bizarre broken compiler. **It is GCC**, just configured to support a single architecture. 

This compiler is part of Homebrew, which is the de-facto standard "distro" for macOS. 

Besides, AFAIK compiler flags are not part of any official C/POSIX/ISO standard, so there's nothing that says every Mac compiler must support `-arch` flag. 

I understand the reluctance to accumulate endless workarounds in the `cc` crate. However, this crate is now so common in the Rust ecosystem that it's the gatekeeper of what compilers can be used with Rust. 

Because the `cc` crate sets the `-arch` flag on macOS unconditionally, with no configuration or opt-out, it removes possibility of using a compiler that doesn't support the `-arch` flag, which makes it practically impossible to cross-compile with GCC on macOS. 

With the rise of M1 ARM Macs, support for cross-compilation is increasingly important. 

This isn't a very niche workaround. It's required to support GCC for x86-64 from the most popular package manager on macOS on the latest Apple CPU.

alexcrichton:
If the target of your compile is `x86_64-apple-darwin` why does the compiler `x86_64-elf-gcc`, which presumably is for ELF executables for x86_64, work? The macOS x86_64 target uses Mach-O for it's object file format, and this is another reason I'm wary of this change.

kornelski:
You're right. That compiler is weirder than I thought. I'll take the issue with Homebrew.

PR #637, opened 2021-11-18
Add Loongarch64 support

zhaixiaojuan:
Loongarch is a new RISC architecture.
Loongarch doc: https://github.com/loongson-community/LoongArch-Documentation/

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2945,6 +2945,7 @@ impl Build {
             ]), // explicit None if not found, so caller knows to fall back
             "i686-unknown-linux-musl" => Some("musl"),
             "i686-unknown-netbsd" => Some("i486--netbsdelf"),
+            "loongarch64-unknown-linux-gnu" => Some("loongarch64-linux-gnu"),
             "mips-unknown-linux-gnu" => Some("mips-linux-gnu"),
             "mips-unknown-linux-musl" => Some("mips-linux-musl"),
             "mipsel-unknown-linux-gnu" => Some("mipsel-linux-gnu"),


alexcrichton:
Thanks! Is there an upstream rust target specification for this target?

zhaixiaojuan:
Yes, the definition of this target is based on the rust target specification.
And regarding rust's support for loongarch64 architecture, we are under development.

alexcrichton:
Hm I don't see the target here? https://github.com/rust-lang/rust/tree/master/compiler/rustc_target/src/spec

zhaixiaojuan:
Sorry, we are still  testing  and  developing rust-lang/rust locally. After the testing is completed, we will submit it to the upstream community.

alexcrichton:
Ah ok sounds good, would you be ok if I hold off on merging this until your target is closer to being merged into rust-lang/rust?

zhaixiaojuan:
Of course. 
However, the integration of rust-lang/rust and its dependent library libc into the community may be a long-term work, and it will take a while.

zhaixiaojuan:
If cc-rs can be incorporated into the community, it will be friendly to users of the new architecture, because they do not need to manually adapt it locally every time they use it.

alexcrichton:
Yes that makes sense @zhaixiaojuan, but I prefer to have targets landed in rust-lang/rust before they're added here.

heiher:
@alexcrichton The [loongarch64-unknown-linux-gnu](https://github.com/rust-lang/rust/pull/96971) was added as a tier 3 target, could you review this patch again please?

JohnTitor (approved):
Thank you!

PR #638, opened 2021-11-21
Default to emcc for wasm32-unknown-unknown if it's available on the $PATH.

dot-asm:
On related note one can make a case for forcing `-ffreestanding` on the unknown target to ensure that emcc doesn't have ideas about using emscripten run-time...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2467,15 +2467,25 @@ impl Build {
         let default_ar = "ar".to_string();
         let program = if target.contains("android") {
             format!("{}-ar", target.replace("armv7", "arm"))
-        } else if target.contains("emscripten") {
+        } else if target.starts_with("wasm32") && !target.contains("wasi") {
             // Windows use bat files so we have to be a bit more specific
             if cfg!(windows) {
-                let mut cmd = self.cmd("cmd");
-                cmd.arg("/c").arg("emar.bat");
-                return Ok((cmd, "emar.bat".to_string()));
+                let emar = "emar.bat";
+                if target.contains("emscripten") || which(&PathBuf::from(&emar)).is_some() {
+                    let mut cmd = self.cmd("cmd");
+                    cmd.arg("/c").arg(emar);
+                    return Ok((cmd, emar.to_string()));
+                } else {
+                    default_ar
+                }
+            } else {
+                let emar = "emar";
+                if target.contains("emscripten") || which(&PathBuf::from(&emar)).is_some() {
+                    emar.to_string()
+                } else {
+                    default_ar
+                }
             }
-
-            "emar".to_string()
         } else if target.contains("msvc") {
             match windows_registry::find(&target, "lib.exe") {
                 Some(t) => return Ok((t, "lib.exe".to_string())),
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR here, but AFAIK emscripten isn't compatible with wasm32-unknown-unknown, so I'm not sure if this would work. Does this work for a project for you locally?

dot-asm:
> Does this work for a project for you locally?

Yes. For now I do it by setting `CC` environment variable to `emcc` and `AR` to `emar`. It might happen that it depends on the fact that our C library doesn't have any dependencies on the C run-time, by its design. I even compile with `-fno-builtin` to ensure that compiler doesn't slip in C run-time calls behind the curtains. For reference, `-ffreestanding` is kind of a "stronger" version of `-fno-builtin`. Well, standalone C code is probably not as common place as this PR implies, which means that it would probably be more appropriate to stick to setting `CC` and `AR`...

dot-asm:
> Well, standalone C code is probably not as common place as this PR implies

On the other hand, I wonder what are the real-life expectations? Is it actually so that most of the stock C libraries can be compiled for wasm32-unknown-unknown with stock clang? What I'm trying to say is that if there is **working** C code that works with wasm32-unknown-unknown Rust target, is it possible that it can't have dependencies on C run-time?

Just in case for reference. The reason for all this it that we do have some `#include <std*.h>` for type definitions, which is problematic with stock `clang --target=wasm32`. ..

dot-asm:
> can be compiled ... with stock clang?

It's kind of a slippery question, isn't it? I mean it's not necessarily exclusively about "stock clang", but about what you call and link with... Either way, the point I'm trying to make is that if the only thing that actually works is standalone C, then it doesn't matter which compiler do you use...

dot-asm:
OK, after further consideration I would agree that it might be less appropriate to default to `emcc`... It still might be helpful to fall back to `emcc` **if** clang doesn't recognize `--target=wasm32`. Or rather try to fall back, because it will still be tested for availability on the `$PATH`... Should I pursue this? Or should we just close this? :-)

dot-asm:
I consider the lack of response as "close." But before I do, I wonder following. Formally speaking using system `ar` is not guaranteed to work with arbitrary binary formats(*), such as binary wasm32. One can make a case for querying `clang -print-search-dirs` and checking for availability of `llvm-ar` on the programs path. Actually, one can make the case for using llvm-ar in **all** clang cases, which should ensure that cross-compilations work smoothly...

(*) Well, `ar` can archive arbitrary files, but it can be ar's responsibility to maintain common name lookup table, which is binary-executable-format-specific.

alexcrichton:
AFAIK the `wasm32-unknown-unknown` Rust target has no equivalent with a C toolchain right now. It's akin to a "kernel" target on native platorms where there is nothing to import from the environment so the C headers are theoretically possible but must bring everything, including the allocator. This doesn't really work out well because there's no established libc, so I'm hesitant to support this because while it may work for your specific use case I feel that C support for `wasm32-unknown-unknown` is never going to work out "in the large"

dot-asm:
> It's akin to a "kernel" target on native platorms

Right, hence the suggestion to forcing `-ffreestanding`.

PR #640, opened 2021-11-24
Added arbitrary flag support to try_expand()

x87-va:
It's needed to pass macOS specific flags. For example: -F/Library/Frameworks.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -408,7 +408,7 @@ impl Build {
         self
     }
 
-    /// Add an arbitrary flag to the invocation of the compiler
+    /// Add a flag to the invocation of the ar
     ///
     /// # Example
     ///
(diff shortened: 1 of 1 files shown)


alexcrichton:
As-is this is passing flags for `ar`, not `cc`, so I don't think this may be what you're intending?

x87-va:
From documentation for ar_flag: "Add an arbitrary flag to the invocation of the compiler". For example I need to expand some macros for framework located in /Library/Frameworks:

```rust
    let src = "expando.c";
    let mut gcc = cc::Build::new();
    gcc.ar_flag("-F/Library/Frameworks");
    let expanded = gcc.file(src).try_expand().expect("Expand");
```

alexcrichton:
I believe that's a copy/paste error on the `ar_flag` method, that only goes to the `ar` invocation, not the compiler itself.

x87-va:
Ok, I've fixed it.

x87-va:
Ok, I checked this. It's really unnecessary changes.

alexcrichton:
Thanks!

PR #646, opened 2021-12-25
Support getting Windows SDK from the environment

### LINKED ISSUE #625: Support getting Windows SDK from the environment
One thing that was brought up recently with the Windows CI's update was a desire to stay with a particular version of the Windows SDK even when a newer version is avaliable. Currently `cc` always gets the latest with no easy way (as far as I'm aware) to override it. Usually this works out fine but not always.

It would be great if it could default to using the environment before falling back to finding the latest. Possibly relevant environment variables (set by `vcvarsall.bat` or developer powershell):

* `WindowsSDK_ExecutablePath_x64`
* `WindowsSDK_ExecutablePath_x86`
* `WindowsSdkBinPath`
* `WindowsSdkDir`
* `WindowsSDKLibVersion`
* `WindowsSdkVerBinPath`
* `WindowsSDKVersion`

I don't have the bandwidth to test a fix for this atm, so if anyone else wants to look into it I'd appreciate it.

nico-abram:
Should fix #625 

This uses the "WindowsSdkDir" and "WindowsSDKVersion" environment variables set by vcvars.

I only ran some simple tests, checked that it worked when running under a shell configured with vcvars and under one without (And that it correctly found the env var and paths), and I also manually tested by setting the two env vars by hand to use a different SDK version and that also worked fine.

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -663,7 +663,15 @@ mod impl_ {
     // only need to bother checking x64, making this code a tiny bit simpler.
     // Like we do for the Universal CRT, we sort the possibilities
     // asciibetically to find the newest one as that is what vcvars does.
+    // Before doing that, we check the "WindowsSdkDir" and "WindowsSDKVersion"
+    // environment variables set by vcvars to use the environment sdk version
+    // if one is already configured.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #652, opened 2022-01-30
feat: add opt_fast_release method

aminya:


aminya:
It could be marked as `unsafe` to communicate the unsafety.

thomcc:
What would the safety contract be? I'm not sure this is the typical meaning of unsafe.

ChrisDenton:
Hm, I'm going to close this as not planned at this time. I think at the very least this needs more discussion and more design work. Some of the flags especially, such as `-fp:fast`, would be concerning to me. And things like `-Ob3` (inlining more aggressively) may not necessarily lead to better performance (it depends).

Issue #105, opened 2016-10-03
Default musl compiler binary names are dodgy


raphaelcohn:
An installation of musl does not supply a `musl-g++`, but this crate assumes that it exists for C++ code. This is a bit a of a quandry to solve. It also assumes things like `musl-gcc` exist, which isn't really true unless musl has been installed alongside glibc on an existing system as a system compiler - not a cross compiler. Since most `build.rs` scripts are often targeting a cross-compilation, rather than deployment on the host (especially so if `--target` has been passed to cargo), using a reference to the system compiler seems wrong.

A slightly better approach might be to support the names used by musl-cross-make, eg `x86_64-linux-musl-cc`. This allows installation of multiple copies of musl targeting gcc (or clang) at different architectures to co-exist. These names are also the ones available by default if installing musl-cross-make on Mac OS X using Homebrew (https://github.com/FiloSottile/homebrew-musl-cross/)

Sadly, it's very difficult trying to overcome the mess of C compilation, and particularly C cross compilation using gcc, which doesn't understand how to compile for more than target. We're also trying to accommodate scenarios ranging from install and run locally to deployment. Personally, I'd make cross compilation my primary focus, and make local install and run secondary; that way, when cross compilation's sorted, local install should just work. This is the approach taken by most from-scratch toolchains and the like (eg Linux from Scratch).

raphaelcohn:
Actually, could the name of the compiler used default to that in Cargo's `--target` eg via `TARGET`?

alexcrichton:
Thanks for the report! In general *all* the compiler names used in this crate are dodgy as none of them are verified to exist, they're just hopefully a reasonable set of defaults. I've seen `musl-gcc` as being more common than other musl compilers, but that's just my own personal experience! Where does the `x86_64-linux-musl-cc` compiler come from?

Also, it should be noted that these are just attempts at being reasonable defaults. That is, it's expected they're ~60% accurate which is better than the 0% accurate `cc` at least! In other words, the compiler should always be very easily overridable via a number of means.

pmarks:
I'm hitting a similar issue. My project depends on flate2. Compiling with `cargo build --release --target x86_64-unknown-linux-musl`

yields:
failed to execute command: No such file or directory (os error 2)
Is `musl-gcc` not installed?`

Should a musl installation always be required if any dependent projects does a gcc compile? Or should the gcc crate & musl toolchain should handle this automatically?

alexcrichton:
[...]

PR #654, opened 2022-02-07
Change prefix for musl targets to not use wrapper

Timmmm:
This should fix cross-compilation to musl when you haven't manually set the compiler name. It's also more consistent with the existing musl targets (e.g. mips).

See #105 and https://github.com/FiloSottile/homebrew-musl-cross#:~:text=If%20you%20encounter%20issues

Tested with `brew install FiloSottile/musl-cross/musl-cross` on Mac.


Timmmm:
Uhm not sure why the tests are failing and there's no error messages :-/

m-ou-se:
At least on Arch Linux, the `musl` package for x86-64 contains `musl-gcc`, but not `x86_64-linux-musl-cc`.

PR #657, opened 2022-02-14
Default to llvm-ar when compiling for wasm.

dot-asm:
Defaulting to llvm-ar facilitates cases when system ar(1) doesn't
support any given cross-compile target. This is because accompanying
llvm-ar is known to match targets supported by the clang installation
at hand.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1119,7 +1119,7 @@ impl Build {
             None => "none",
         };
         if cudart != "none" {
-            if let Some(nvcc) = which(&self.get_compiler().path) {
+            if let Some(nvcc) = which(&self.get_compiler().path, None) {
                 // Try to figure out the -L search path. If it fails,
                 // it's on user to specify one by passing it through
                 // RUSTFLAGS environment variable.
@@ -2775,6 +2775,20 @@ impl Build {
                         name = format!("em{}", tool);
                         Some(self.cmd(&name))
                     }
+                } else if target.starts_with("wasm32") {
+                    // Formally speaking one should be able to use this approach,
+                    // parsing -print-search-dirs output, to cover all clang targets,
+                    // including Android SDKs and other cross-compilation scenarios...
+                    // And even extend it to gcc targets by seaching for "ar" instead
+                    // of "llvm-ar"...
+                    let compiler = self.get_base_compiler().ok()?;
+                    if compiler.family == ToolFamily::Clang {
+                        name = format!("llvm-{}", tool);
+                        search_programs(&mut self.cmd(&compiler.path), &name)
+                            .map(|name| self.cmd(&name))
+                    } else {
+                        None
+                    }
                 } else {
                     None
                 }
@@ -2801,10 +2815,10 @@ impl Build {
                         // next to 'clang-cl' and use 'search_programs()' to locate
                         // 'llvm-lib'. This is because 'clang-cl' doesn't support
                         // the -print-search-dirs option.
-                        if let Some(mut cmd) = which(&compiler.path) {
+                        if let Some(mut cmd) = which(&compiler.path, None) {
                             cmd.pop();
                             cmd.push("llvm-lib.exe");
-                            if let Some(llvm_lib) = which(&cmd) {
+                            if let Some(llvm_lib) = which(&cmd, None) {
                                 lib = llvm_lib.to_str().unwrap().to_owned();
                             }
                         }
@@ -3657,7 +3671,7 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &str) -> Option<
     }
 }
 
-fn which(tool: &Path) -> Option<PathBuf> {
+fn which(tool: &Path, path_entries: Option<OsString>) -> Option<PathBuf> {
     fn check_exe(exe: &mut PathBuf) -> bool {
         let exe_ext = std::env::consts::EXE_EXTENSION;
         exe.exists() || (!exe_ext.is_empty() && exe.set_extension(exe_ext) && exe.exists())
@@ -3670,11 +3684,25 @@ fn which(tool: &Path) -> Option<PathBuf> {
     }
 
     // Loop through PATH entries searching for the |tool|.
-    let path_entries = env::var_os("PATH")?;
+    let path_entries = path_entries.or(env::var_os("PATH"))?;
     env::split_paths(&path_entries).find_map(|path_entry| {
         let mut exe = path_entry.join(tool);
         return if check_exe(&mut exe) { Some(exe) } else { None };
     })
+}
+
+// search for |prog| on 'programs' path in '|cc| -print-search-dirs' output
+fn search_programs(cc: &mut Command, prog: &str) -> Option<PathBuf> {
+    let search_dirs = run_output(cc.arg("-print-search-dirs"), "cc").ok()?;
+    // clang driver appears to be forcing UTF-8 output even on Windows,
+    // hence from_utf8 is assumed to be usable in all cases.
+    let search_dirs = std::str::from_utf8(&search_dirs).ok()?;
+    for dirs in search_dirs.split(|c| c == '\r' || c == '\n') {
+        if let Some(path) = dirs.strip_prefix("programs: =") {
+            return which(Path::new(prog), Some(OsString::from(path)));
+        }
+    }
+    None
 }
 
 #[derive(Clone, Copy, PartialEq)]


dot-asm:
CI failures appear to have everything to do with the fact that `cargo test` fails in Windows with Rust 1.58, but not with Rust 1.57. I mean it's something in 1.58 that makes it fail. This naturally falls beyond the scope of this PR. [Even though it doesn't mean that cc-rs doesn't need an adjustment.]

dot-asm:
On a related note one can wonder if it would be appropriate to extend this approach beyond a clang-specific default. Most notably gcc also recognizes `-print-search-dirs` option and cross-gcc-compiler can steer you toward target-specific ar. I mean if I traverse `programs` from `arm-linux-gnueabi-gcc` I do find arm-specific `ar`. To be more specific, it looks like it's possible to replace the current if-else with `match compiler.family` and use `-print-search-dirs` to locate `ar` for ToolFamily::Gnu, `llvm-ar` for ToolFamily::Clang, and just pick`lib.exe` for ToolFamily::Msvc... Or at least prioritize such `match` and preserve the old code as a legacy fallback...

dot-asm:
That's unfortunate:-(

For reference, the suggestion is a generalization of a problem with wasm target. While `ar(1)` can collect literally any type of files in the .a-archive, it has a responsibility to maintain a symbol table for the binary objects in the .a-archive. The table in question can be examined with `nm -s`. Customarily `/usr/bin/ar` is aware only of native execution format and attempt to use it with foreign one makes it treat the binary objects effectively as non-binary and not generate the table. And rationale is that if `clang --target=xyz` works, accompanying `llvm-ar` would be effectively obliged to manage the `xyz` target correctly.

Before I close this, a question. Would it be helpful to open an issue and challenge clang users to provide feedback if builds with `AR` environment variable set to `llvm-ar` work reliably? Or maybe if they even have to set it in order to get it working? Especially in cross-compile scenarios...

Cheers.

alexcrichton:
Unfortunately, at least in my experience, I don't know how to get feedback on a change like this. There's so many ways that C build systems are configured in my experience all you can really do is push the change, see what breaks, and evaulate it if it's worth it to back out or not. That's not a great way of managing this however.

If you wanted to make this specific to the wasm target though that seems reasonable to me. Making this more targeted that can theoretically get generalized in the future but not necessarily immediately seems plausible. I don't think this crate is as heavily used with wasm and AFAIK all wasm C compilation goes through clang right now with a pretty well-defined set of compilers.

dot-asm:
> If you wanted to make this specific to the wasm target though that seems reasonable to me.

Adjusted accordingly.

> Making this more targeted that can theoretically get generalized in the future but not necessarily immediately seems plausible.

Like for the next major release when possible breakage can be tolerated. There are more cleanups that one can implement...

src/lib.rs
```diff
@@ -3293,9 +3308,23 @@ fn which(tool: &Path) -> Option<PathBuf> {
...
 }
+
+// search for |prog| on 'programs' path in '|cc| -print-search-dirs' output
+fn search_programs(cc: &mut Command, prog: &str) -> Option<PathBuf> {
+    let search_dirs = run_output(cc.arg("-print-search-dirs"), "cc").ok()?;
+    // clang driver appears to be forcing UTF-8 output even on Windows,
+    // hence from_utf8 is assumed to be usable in all cases.
+    let search_dirs = std::str::from_utf8(&search_dirs).ok()?;
+    for dirs in search_dirs.split(|c| c == '\r' || c == '\n') {
+        if dirs.starts_with("programs: =") {
```
alexcrichton, line 3325: Can this use `.strip_prefix` to avoid the indexing and hardcoding 11 below?

src/lib.rs
```diff
@@ -2533,6 +2533,21 @@ impl Build {
...
                 None => default_ar,
             }
+        } else if target.starts_with("wasm32") {
+            // Formally speaking one should be able to use this approach,
+            // parsing -print-search-dirs output, to cover all clang targets,
+            // including Android SDKs and other cross-compilation scenarios...
+            // And even extend it to gcc targets by seaching for "ar" instead
+            // of "llvm-ar"...
+            let compiler = self.get_base_compiler()?;
+            if compiler.family == ToolFamily::Clang {
```
alexcrichton, line 2543: Coudl this condition be moved up into the `if target.starts_with("wasm32")` condition?

src/lib.rs
```diff
@@ -2533,6 +2533,21 @@ impl Build {
...
+                match search_programs(&mut self.cmd(&compiler.path), "llvm-ar") {
+                    Some(ar) => ar.to_str().unwrap().to_owned(),
+                    None => default_ar,
```
alexcrichton, line 2546: To avoid hardcoding `default_ar` return values could this be structured with a fall-through to the case below?

src/lib.rs
```diff
@@ -2533,6 +2533,21 @@ impl Build {
                 }
                 None => default_ar,
             }
+        } else if target.starts_with("wasm32") {
```
alexcrichton, line 2536: Could this be `wasm` instead of `wasm32` to cover `wasm64` as well?

src/lib.rs
```diff
@@ -3280,7 +3295,7 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &str) -> Option<
     }
 }
 
-fn which(tool: &Path) -> Option<PathBuf> {
+fn which(tool: &Path, path_entries: Option<OsString>) -> Option<PathBuf> {
```
alexcrichton, line 3298: Could this be split into a second function instead of taking an optional `path_entries` argument? It's non-obvious to me at least that given this signature what `path_entries` is and its behavior relative to `PATH`

src/lib.rs
dot-asm: doc.rust-lang.org says it's available in 1.45.0, while your MSRV is 1.34.0. Or do I interpret 1.45.0 at [strip_prefix](https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix) incorrectly?

src/lib.rs
dot-asm: Well, yes. But first let's consider rationale behind this specific choice. It was that `rustup target list` doesn't offer any wasm64 options. Nor is `wasm` used by itself anywhere in cc-rs, always as `wasm32`.

src/lib.rs
dot-asm: Would `override_path` be better? Or `override_PATH`? I see less point in duplicating this function...

src/lib.rs
dot-asm: Well, I seem to fail to figure it out. In order to avoid `default_ar` here one has to do the whole thing in `else if` above. And here is the conundrum. It doesn't let me use `if target.starts_with("wasm32") && let Some(ar) = self.find_ar()`, while `if let Some(ar) = if target.starts_with("wasm32") { self.find_ar() } else { None }` get cargo-fmt-ed in a way that makes it really hard to follow:
```
        } else if let Some(ar) = if target.starts_with("wasm32") {
            self.find_ar()
        } else {
            None
        } {
            ar.to_str().unwrap().to_owned()
        } else {
            default_ar
        };

```

alexcrichton:

dot-asm:
Just in case, as for o-wasm label. It's actually two-part now, wasm and Windows. And as discussed in the commentary, it's possible to make a case for extending the approach, defaulting to llvm-ar, to all clang compilations. And even to gcc (arguably less error-prone than recent gcc-ar changes:-). There is only one potential problem with the latter. gcc for Windows doesn't seem to enforce UTF-8 in `-print-search-dirs` output, which makes it uncertain if there are non-ASCII characters in the paths...

ChrisDenton:
Could you separate the Windows `llvm-lib` changes in to another PR? I'd feel confident in merging that asap and I don't think it needs to be tied to the wasm changes.

dot-asm:
I've opened #758 and marked this one draft, as it would require a re-adjustment once #758 is merged.

dot-asm:
Back from draft, now as wasm-only thing.

Just in case for reference. As it stands now, the linking procedure (on non-emscripten wasm targets) fails with "archive has no index" in .a file generated by cc-rs. This modification resolves the problem.

dot-asm:
Re-based just in case.

dot-asm:
Re-based.

To recap. The problem is that the linking procedure (on non-emscripten wasm targets) fails with "archive has no index" in .a file generated by cc-rs. For completeness sake one can note that Rust 1.67 doesn't seem to require the index, presumably at the cost of longer linking times.

Just in case. One can dispute the placement of the suggested `else if` segment. In sense that one could have placed it the `match tool_opt` below. Such placement is arguably less readable. Indeed, the name of the emscripten target is `wasm32-unknown-emscripten`, so that moving the `else if target.starts_with("wasm32")` below raises the question if it will override the `if target.contains("emscripten")`. Well, it won't because of the initial `Some(t) => t`, but it's not as apparent as reading the sequence `if target.contains("emscripten") ... else if target.starts_with("wasm32")`. But if the alternative placement is considered preferred nevertheless, just say so. And of course if one chooses to extend the suggested method (of examining -print-search-dirs output) to all clang targets, then it would have to be moved below, presumably as as first `if` case...

dot-asm:
> But if the alternative placement is considered preferred nevertheless, just say so.

Though moving would entail restructuring of the `match tool_opt`, because it would be appropriate to have an option to return `None` and fall back on some default...

Mrmaxmeier:
This fixes the `archive has no index; run ranlib to add one` issue from #531 and feels minimally invasive.
(I'm posting this as a well-intentioned bump and for visibility, since searching for the `run ranlib` error only returns the closed PR for some reason :slightly_smiling_face:)

thomcc (approved):
This seems fine to me, especially since it's localized to wasm.

PR #662, opened 2022-03-08
Add support for Apple WatchOS

vladimir-ea:
Hello, I am hoping to add Apple WatchOS targets and the changes here are prerequisites for that. 
Thanks! Vlad.

### DIFF
--- a/cc-test/src/riscv64gc.S
+++ b/cc-test/src/riscv64gc.S
@@ -1,0 +1,9 @@
+.globl asm
+asm:
+    li a0, 7
+    ret
+
+.globl _asm
+_asm:
+    li a0, 7
+    ret
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -168,7 +168,7 @@ impl From<io::Error> for Error {
 }
 
 impl Display for Error {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{:?}: {}", self.kind, self.message)
     }
 }
@@ -2207,7 +2207,7 @@ impl Build {
             ArchSpec::Simulator(arch) => {
                 cmd.args.push(arch.into());
                 cmd.args
-                    .push(format!("-mwatch-simulator-version-min={}", min_version).into());
+                    .push(format!("-mwatchsimulator-version-min={}", min_version).into());
                 "watchsimulator"
             }
         };
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1416,7 +1416,7 @@ impl Build {
                     cmd.push_opt_unless_duplicate("-DANDROID".into());
                 }
 
-                if !target.contains("apple-ios") {
+                if !target.contains("apple-ios") && !target.contains("apple-watchos") {
                     cmd.push_cc_arg("-ffunction-sections".into());
                     cmd.push_cc_arg("-fdata-sections".into());
                 }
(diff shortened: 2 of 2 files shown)


alexcrichton:
Sorry but I'll be honest in that I have no idea how ios/watchos/etc all work. I'm not sure I can in good conscience continue to just slurp up everyone's patches for all sorts of esoteric platforms, especially when they're doing nontrivial things like this PR. If this can be unified with something else in this crate (I don't know what) that would be great, otherwise I'm not sure I can keep taking these patches for all these targets.

vladimir-ea:
Thanks for taking a look. This is the first platform I have tried to add ... 

WatchOS is similar to iOS, I will see if I can do something to commonise the code for those 2 platforms.

vladimir-ea:
@alexcrichton I have updated so that the WatchOS changes are consolidated with the existing iOS specific parts - hopefully this is better / clearer?

alexcrichton:

vladimir-ea:
I see. If I understand correctly, this crate is a dependency of rustc. So adding targets for this platform to Rust is gated on this crate acquiring new maintainers, is that correct?

dot-asm:
> I see. If I understand correctly, this crate is a dependency of rustc. So adding targets for this platform to Rust is gated on this crate acquiring new maintainers, is that correct?

No. Rust is self-sufficient and this crate is required only when you need to compile an external C/C++ code and call it from Rust.

vladimir-ea:
> > I see. If I understand correctly, this crate is a dependency of rustc. So adding targets for this platform to Rust is gated on this crate acquiring new maintainers, is that correct?

Thank you for clarifying. I was just observing that it is a dependency of `rustc_codegen_ssa` and `rustc_llvm`. Also, these changes do appear to be required to build stdlib for the WatchOS platform correctly. Without these changes, I can compile stdlib for the WatchOS targets but the binaries produced cannot combined in a single `lipo` archive (fails with 'cputype and cpusubtype do not match'). Whereas with these changes this succeeds and the resulting archive can be successfully deplyed to the 3 target platforms.

vladimir-ea:
> No. Rust is self-sufficient and this crate is required only when you need to compile an external C/C++ code and call it from Rust.

Thank you for clarifying.  This crate appears to be a dependency of `rustc_codegen_ssa` and `rustc_llvm`, but I don't follow exactly how it is used.

vladimir-ea:
> Those dependencies are required by the Rust compiler itself. In other words you'd need `cc-rs` if you were to build the Rust compiler for executing it on your watch. I assumed it was a safe assumption that it's not the actual goal of your endeavour.

Ok, makes sense. I do not think there will ever be a requirement to compile *on* WatchOS. 

I have experimented with building without the changes to `cc-rs` made in this diff, and it does not work. I am also building the standard library. 

What I am doing is:
- building rust compiler with WatchOS changes (see https://github.com/rust-lang/rust/pull/94736)
- building the 3 WatchOS targets (`armv7k-apple-watchos`, `arm64_32-apple-watchos`, and `x86_64-apple-watchos-sim`)
- cross-compiling an application for those 3 targets (with dependency on `cc-rs` as modified by this PR)

This all works, in as much as everything compiles. To deploy to WatchOS you create multi-architecture bundle using `lipo`. This fails with errors about mismatched cpu types. 

If I update the rust project dependency on `cc-rs` to point at this PR, then this all works, I can deploy and run the application on all 3 architectures.

dot-asm:
I [for one] find all this environment variable chasing tedious and even counterproductive, so I what you would think of #664.

dot-asm:
> To find the SDK?

No, to start the compiler. Instead of looking at environment variables like THIS_N_THAT_DEPLOYMENT_TARGET and map it to --target, -mx-version-min, you just let xcrun handle all this. Taking this PR as example. You suggest to set specific version-min if WATCHOS_DEPLOYMENT_TARGET is not set. Why **this** specific version? Will it still work with next after next xcode version? Would it still match the average expectations? Aren't the said expectations "whatever xcode says" anyway? So why not simply let xcode do its thing?

vladimir-ea:
>  So why not simply let xcode do its thing?
I don't disagree - this seems like a good idea. Would like to make these changes? As part of this PR? I am not super familiar with xcode myself, but happy to test any changes.

m-ou-se:
> does this repo now have maintainers and / or reviewers? does anyone know?

Yes, but we're short on bandwidth. I'll try to look at this tomorrow, but I currently have 302 GitHub notifications that I need to check, so I can't make any promises.

Issue #644, opened 2021-12-18
Cannot use clang-cl with LTO because MSVC's lib.exe crashes


nico-abram:
I'm trying to compile a windows project using the msvc toolchain and cross language LTO. To do this, I'm trying to use clang-cl instead of the MSVC cl.exe to compile the C/C++ code. I'm running into a problem where cc-rs tries to use lib.exe to archive the generated object files, but MSVC's lib.exe does not understand the clang object files with LLVM bitcode for LTO, so it crashes. I'm opening this issue for discussion, ideally I would like a solution like an env var that lets me override the archiver, or identifying clang-cl with LTO and trying to find LLVM-lib (I don't know if rust ships with one, but if not maybe there could be an env var with a path to an installer LLVM/bin).

I also tried to temporarily make a copy of LLVM-lib called lib.exe and put it before lib.exe in my path but it didn't work because cc-rs is finding the specific VS path to lib.exe from the registry.

nico-abram:
I think I just wanted the "AR" env var. My apologies for the useless issue.

PR #668, opened 2022-03-23
Choose AR args by program name instead of target

andylizi:
Right now `cc` will always use `-out:.. -nologo ...` for ar args on MSVC targets, regardless of what the archiver actually is. This makes it impossible to use something other than `lib.exe`.

My use case is the same as #644 — I'm not sure how they managed without this, actually :sweat_smile:.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1991,10 +1991,10 @@ impl Build {
     }
 
     fn assemble_progressive(&self, dst: &Path, objs: &[PathBuf]) -> Result<(), Error> {
-        let target = self.get_target()?;
+        let (mut cmd, program) = self.get_ar()?;
 
-        if target.contains("msvc") {
-            let (mut cmd, program) = self.get_ar()?;
+        // `get_ar()` will never return "lib" or such
+        if program.eq_ignore_ascii_case("lib.exe") {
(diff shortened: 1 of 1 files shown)


dot-asm:
> I'm not sure how they managed without this

By setting `AR` environment variable to `llvm-lib`, a lib- and llvm-compatible librarian. Arguably it's a more natural choice and if anything, it would be more appropriate to advocate for having cc-rs use it, llvm-lib, automatically.

andylizi:
> By setting `AR` environment variable to `llvm-lib`

`llvm-lib` is a thing?!

Wow :joy: I remember searching around for keywords like "llvm-ar msvc" and "llvm-ar lib", but somehow failed to find any mention of it. Today I learned...

> it would be more appropriate to advocate for having cc-rs use it, llvm-lib, automatically.

Yeah that's for sure. In the mean time, do you think I should change this PR to `.ends_with("lib.exe")` to cover for `llvm-lib` usage, or just close it and wait for #657?

dot-asm:
> In the mean time, do you think

I'm just another cc-rs user, I can't tell you what to do :-) Yes, #657 can be extended to use llvm-lib, yes, I can do that... A little bit later...

PR #669, opened 2022-03-28
Add LLVM based MinGW targets

mati865:
Rust PR: https://github.com/rust-lang/rust/pull/94872

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2226,7 +2226,8 @@ impl Build {
                     if target.contains("msvc") {
                         msvc.to_string()
                     } else {
-                        format!("{}.exe", gnu)
+                        let cc = if target.contains("llvm") { clang } else { gnu };
+                        format!("{}.exe", cc)
                     }
                 } else if target.contains("apple-ios") {
                     clang.to_string()
@@ -2252,7 +2253,10 @@ impl Build {
                 } else if self.get_host()? != target {
                     let prefix = self.prefix_for_target(&target);
                     match prefix {
-                        Some(prefix) => format!("{}-{}", prefix, gnu),
+                        Some(prefix) => {
+                            let cc = if target.contains("llvm") { clang } else { gnu };
+                            format!("{}-{}", prefix, cc)
+                        }
                         None => default.to_string(),
                     }
                 } else {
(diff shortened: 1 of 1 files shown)


dot-asm:
This is duplicate of [comment at rust-lang](https://github.com/rust-lang/rust/pull/94872#issuecomment-1083683807)

Not that my word weighs a lot here and I apologize, but I fail to see the need for an additional target. Is there a programming interface that `clang` uses/requires that `gcc` and/or Microsoft `cl` don't? What I'm trying to say is that it's rather the programming interface that defines any particular target platform, not the compiler. Compilers are supposed to be and actually are interchangeable. For example, on Windows you can use `clang-cl` instead of vendor `cl` with `*-msvc` targets. On Linux you can freely switch between `clang` and `gcc`. And I see no reason for why it would be impossible to have the similar arrangement for `*-windows-gnu` targets.

dot-asm:
Just in case for reference, my remark was refuted in the referred thread at rust-lang. To summarize, it boils down to irreconcilable differences in exception handling mechanisms.

src/lib.rs
```diff
@@ -2546,7 +2550,7 @@ impl Build {
             .as_ref()
             .map(|s| s.trim_right_matches('-').to_owned());
         cross_compile.or(match &target[..] {
-            "aarch64-pc-windows-gnu" => Some("aarch64-w64-mingw32"),
+            "aarch64-pc-windows-gnullvm" => Some("aarch64-w64-mingw32"),
```
thomcc, line 2553: Does this mean to remove the `aarch64-pc-windows-gnu` target?

thomcc (changes requested):
I'm in favor of supporting the `-gnullvm` targets, but this seems to remove one of the `-gnu` targets?

src/lib.rs
mati865: Yes, there is no AArch64 support for Windows in both Binutils and GCC. If you insist I can restore it but it's dead code.

thomcc: Can you add a comment to this effect, then?

mati865: Comment in commit message/PR description or something else?
There was no `aarch64-pc-windows-gnu` target ever available in Rust.

thomcc: Before that arm, something like this maybe?
```rs
// Note: there is no `aarch64-pc-windows-gnu` target, only `-gnullvm`
```

mati865:
Possibly due to repo move GitHub ignores pushes to this branch, moving to new PR: https://github.com/rust-lang/cc-rs/pull/734

PR #670, opened 2022-03-31
Fix clang-cl treating file as a /U arg.

roblabla:
On MacOS, paths will likely start with /Users, but clang-cl treats /U
as an argument, which leads to the source being mistaken for a flag. To
avoid this, we can pass -- before the source to ensure argument parsing
is disabled.


roblabla:
This was already previously proposed in https://github.com/alexcrichton/cc-rs/pull/670

I can confirm this works as expected.

Issue #663, opened 2022-03-08
New Maintainers


alexcrichton:
I'm reaching the end of my tenure of maintaining this crate. I can't keep up with the constant inundation of platforms into this crate and this has outgrown my own personal ability to manage. I feel nowadays like all I can do is rubber stamp big changes since I have no idea how the targets work, but at the same time it seems to be the case that code is just dumped in this crate and no one ever looks at it again until there's a bug. Personally, I can't keep up with that any more.

If someone would like to take over this crate please let me know. I realize though that this crate is very heavily depended on in the ecosystem, so I won't necessarily give this to the first person that shows up. I'll need to recognize you and you'll also need to understand the importance of this crate and its stability.

sunface:
I have a idea, maybe you should create an organization like `cc-rs`, move the repo there ,and let other maintainers join in, each of them will account for certain targets.

This project is very great and important, ,if it is just moved to another github user, probably nothing will changed and things won't getting better.

Dylan-DPC:
I can help with the organisation part. Have done so for a few crates already and being a rust team member, it might be easier

Be-ing:
Thanks for recognizing your limits and asking for help @alexcrichton. :heart:

spl:
I worked on `cc-rs` [a while back](https://github.com/alexcrichton/cc-rs/commits?author=spl). At the time, I did have some ideas for improving it, but I didn't get around to finishing them. I wouldn't mind giving a shot at maintenance, shared or otherwise.

aloucks:
@m-ou-se I think it's a great idea to incorporate this crate into the `rust-lang` organization. If for whatever reason that doesn't happen, it might make sense to create a new org to host both this crate and [cmake-rs](https://github.com/alexcrichton/cmake-rs). It might also be worthwhile to host cmake-rs in the `rust-lang` org!

dvc94ch:
any updates on this?

m-ou-se:
We've discussed this in the Rust library team meeting this week, and we think it's best if `cc` and `cmake` are both moved to the `rust-lang` organisation. Currently, we won't have the bandwith to work on new features to these crates, but we will be able to review incoming PRs and make sure these crates stay maintained.

@alexcrichton If you're on board with this, can you transfer the repos to the `rust-lang` GitHub organisation, and invite `@rust-lang-owner` as owner to the crates on crates.io?

alexcrichton:
[...]

PR #671, opened 2022-04-02
Support native library modifiers (RFC 2951)

petrochenkov:
cc https://github.com/rust-lang/rust/issues/81490

I haven't added a test because stabilization of the feature has just landed on nightly and is not available on stable yet, but I've tested it manually.


dot-asm:
I for one would argue that `link_modifies` would be more appropriate name for suggested method. Because `cc.modifiers("x")` would appear rather as compiler modifier.

petrochenkov:
I'd say the full name is "native library modifiers" because they are not necessarily about linking in the narrow sense, since they can also control bundling of native libraries into rlibs/staticlibs or potentially [deduplication](https://github.com/rust-lang/rust/issues/73319) before linking.
Not sure how to better name it in this specific interface.

dot-asm:
In the context "native" is effectively synonymous with  "ffi" and `cc-rs` is all about "ffi." In other words "native" feels redundant in the context.

Just a reminder, I'm just a cc-rs user weathering #663.

petrochenkov:
Renamed to `link_lib_modifiers` to match the modified cargo directive.

petrochenkov:
Ping @m-ou-se.
Can this be merged in some form and released before the Rust 1.61 release?
It helps to address the first compatibility note from https://github.com/rust-lang/rust/pull/96539/files.

m-ou-se:
Hi!

Is this cargo feature documented anywhere? The [docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-lib) only shows the `=[KIND=]NAME` syntax, but say nothing about any modifiers.

petrochenkov:
This is a rustc syntax, cargo just passes the `STRING` in `cargo:rustc-link-lib=STRING` directly to rustc.

But I agree that the cargo docs also need to be updated, I'll send a PR.

m-ou-se:
> This is a rustc syntax (https://doc.rust-lang.org/nightly/rustc/command-line-arguments.html#-l-link-the-generated-crate-to-a-native-library), cargo just passes the `STRING` in `cargo:rustc-link-lib=STRING` directly to rustc.

Ah great. Can you add a link to that in the doc comment for `link_lib_modifiers`?

Then one more question:

The `flag()` method takes a single flag and adds it to a list of flags, similar for `include()`, but this new `link_lib_modifiers()` method takes a string with multiple modifiers and overrides any previously set value. Do you think it should be `link_lib_modifier()` that adds a single modifier, instead?

petrochenkov:
>Can you add a link to that in the doc comment for link_lib_modifiers?

Done.

>The flag() method takes a single flag and adds it to a list of flags, similar for include(), but this new link_lib_modifiers() method takes a string with multiple modifiers and overrides any previously set value. Do you think it should be link_lib_modifier() that adds a single modifier, instead?

Both resetting and appending method could potentially make sense in both cases.
I added only a resetting method for the modifiers just because it's simpler, more general, and sufficient for now.

(I've changed the parameter from `&str` to `Option<&str>` though, so you can reset the modifiers back to none if necessary.)

helgoboss:
Thanks @petrochenkov. I'm also very interested in your addition because the Rust 1.61 change https://github.com/rust-lang/rust/pull/93901/ breaks my application at runtime (although it compiles and links successfully). Using `.link_lib_modifiers(Some("+whole-archive"))` for one specific cc build invocation fixes that. For the record, here's the commit which fixes it: https://github.com/helgoboss/realearn/commit/0055d6c96a5ce82f98bf056f48a2765cd25390ba.

petrochenkov:
@helgoboss 
FWIW, `+whole-archive` can be passed even without this API, by using [`build.cargo_metadata(false)`](https://docs.rs/cc/1.0.73/cc/struct.Build.html#method.cargo_metadata) first, and then emitting the `cargo:rust-ling-*` directives manually:
```rust
println!("cargo:rustc-link-lib=static:+whole-archive={}", lib_name);
println!("cargo:rustc-link-search=native={}", env::var("OUT_DIR"));
```

Ping @m-ou-se again.

helgoboss:
@petrochenkov Thanks, good to know. Your solution is much more convenient though.

m-ou-se:
> I added only a resetting method for the modifiers just because it's simpler, more general, and sufficient for now.
> 
> (I've changed the parameter from `&str` to `Option<&str>` though, so you can reset the modifiers back to none if necessary.)

The other methods like `flags()`, `include()`, `file()`, `define()`, are all append-only and provide no way of resetting. I think we should keep this consistent with that.

Right now it doesn't matter much with only `"+whole-archive"`, but it'd be nice if we can keep things unchanged when more useful modifiers are added in the furture.

petrochenkov:
Updated in accordance with https://github.com/rust-lang/cc-rs/pull/671#issuecomment-1176260001.

petrochenkov:
ping @m-ou-se

thomcc (approved):
This looks good to me, thanks.

>  Can this be merged in some form and released before the Rust 1.61 release?

Obviously we missed the boat here, but I'll try to cut a release after I finish going through the pending PRs and merging the ones that are viable.

PR #676, opened 2022-04-26
Use specified compiler in is_flag_supported

Ast-x64:
Use the compiler specified instead of the default one to test whether a flag can be accepted.
Fixes https://github.com/alexcrichton/cc-rs/issues/675.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -477,6 +477,9 @@ impl Build {
             .debug(false)
             .cpp(self.cpp)
             .cuda(self.cuda);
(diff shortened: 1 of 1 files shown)


thomcc (approved):
This seems very reasonable.

Can you rebase in order to retrigger CI? I'm not sure why the windows checks are cancelled, but because it's old the logs are gone, so it's not clear what happened.

(P.S. sorry for the delay on this review, should be better moving forward)

thomcc:
Thanks!

PR #677, opened 2022-04-30
Update CUDA Linux GPG repository key [and compiler version].

dot-asm:
Nvidia rotated their repository signing key on April 27, 2022.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,16 +92,16 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@master
-    - name: Install cuda-minimal-build-11-4
+    - name: Install cuda-minimal-build-11-6
       shell: bash
       run: |
         # https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=20.04&target_type=deb_network
         wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-ubuntu2004.pin
         sudo mv cuda-ubuntu2004.pin /etc/apt/preferences.d/cuda-repository-pin-600
-        sudo apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/7fa2af80.pub
+        sudo apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/3bf863cc.pub
         sudo add-apt-repository "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /"
         sudo apt-get update
-        sudo apt-get -y install cuda-minimal-build-11-4
+        sudo apt-get -y install cuda-minimal-build-11-6
     - name: Test 'cudart' feature
       shell: bash
       run: env PATH=/usr/local/cuda/bin:$PATH cargo test --manifest-path cc-test/Cargo.toml --features test_cuda
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        build: [stable, beta, nightly, linux32, macos, aarch64-ios, win32, win64, mingw32, mingw64, windows-2016]
+        build: [stable, beta, nightly, linux32, macos, aarch64-ios, win32, win64, mingw32, mingw64, windows-2019]
         include:
           - build: stable
             os: ubuntu-latest
@@ -40,7 +40,7 @@ jobs:
             target: aarch64-pc-windows-msvc
             no_run: --no-run
           - build: win32
-            os: windows-2016
+            os: windows-2019
             rust: stable-i686-msvc
             target: i686-pc-windows-msvc
           - build: win64
@@ -55,8 +55,8 @@ jobs:
             os: windows-latest
             rust: stable-x86_64-gnu
             target: x86_64-pc-windows-gnu
-          - build: windows-2016
-            os: windows-2016
+          - build: windows-2019
+            os: windows-2019
             rust: stable-x86_64
             target: x86_64-pc-windows-msvc
     steps:


dot-asm:
Additional commit replaces windows-2016 with windows-2019, because former was removed from Github Actions.

dot-asm:
The mingw error after merge is new. Should I look into it? I mean maybe how to fix it is already common knowledge...

dot-asm:
It appears to be Windows 2022-specific. Because if I force `os` to `windows-2019` in mingw32 section of the .github/workflows/main.yml, the passes... [Off for today.]

ChrisDenton:
It looks like a miscompile. Maybe a rustc issue? If I try to run the `i686-pc-windows-gnu` test binary on my machine, this the what the debugger says:

```
ModLoad: 00400000 009f8000   Z:\target\debug\deps\all-aa90ab550493a218.exe
WARNING: Image Z:\target\debug\deps\all-aa90ab550493a218.exe section 0 extends outside of image bounds
WARNING: Image Z:\target\debug\deps\all-aa90ab550493a218.exe section 1 extends outside of image bounds
DbsSplayTreeRangeMap::Add: Conflicting region 00000000 - 00000407
Unable to map Z:\target\debug\deps\all-aa90ab550493a218.exe region at 00000000, NTSTATUS 0xC0000018
*** WARNING: Unable to verify timestamp for all-aa90ab550493a218.exe
eax=00000000 ebx=00000000 ecx=00000000 edx=00000000 esi=00000000 edi=00000000
eip=004014e0 esp=00000000 ebp=00000000 iopl=0         nv up di pl nz na po nc
cs=0000  ss=0000  ds=0000  es=0000  fs=0000  gs=0000             efl=00000000
all_aa90ab550493a218+0x14e0:
004014e0 ??              ???
```

That NTSTATUS code is `STATUS_CONFLICTING_ADDRESSES` ("The specified address range conflicts with the address space").

I haven't investigated further.

dot-asm:
Just in case, it's not like it fails to run **any** mingw32 executable. The prior tests do pass and **then** it fails `--manifest-path cc-test/Cargo.toml`. As for which component is to blame. Recall that rustc is calling mingw gcc to link the binary. So it's either rustc driving GNU ld crazy, or GNU ld being crazy by itself. As for 2019 vs. 2022. Corresponding runners are documented to run different mingw versions, 8.1.0 vs. 11.20, so that it's more likely to be specific to mingw version than to Windows.

dot-asm:
> The prior tests do pass and **then** it fails `--manifest-path cc-test/Cargo.toml`

Though you're showing `Z:\target\debug\deps\all-xyzzy.exe`. More specifically there is no cc-test in the file name. On the other hand it has "all" in it... So it ought to be cc-test/test/all.rs, right?

ChrisDenton:
I am replicating the [failed CI test](https://github.com/rust-lang/cc-rs/runs/7266232308?check_suite_focus=true):

```
  cargo test  --manifest-path cc-test/Cargo.toml --target i686-pc-windows-gnu
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
    Updating crates.io index
 Downloading crates ...
  Downloaded either v1.7.0
  Downloaded lazy_static v1.4.0
  Downloaded which v4.2.5
  Downloaded libc v0.2.126
   Compiling libc v0.2.126
   Compiling either v1.7.0
   Compiling lazy_static v1.4.0
   Compiling cc v1.0.73 (D:\a\cc-rs\cc-rs)
   Compiling which v4.2.5
   Compiling cc-test v0.1.0 (D:\a\cc-rs\cc-rs\cc-test)
    Finished test [unoptimized + debuginfo] target(s) in 21.40s
     Running tests\all.rs (cc-test\target\i686-pc-windows-gnu\debug\deps\all-8b31d61675f10411.exe)
error: test failed, to rerun pass '--test all'
Caused by:
  could not execute process `D:\a\cc-rs\cc-rs\cc-test\target\i686-pc-windows-gnu\debug\deps\all-8b31d61675f10411.exe` (never executed)
```

I don't know where it "ought" to be. I can only say where it is and that the binary is broken as it won't even load.

dot-asm:
Got it! Thanks! It ought to be one of the "unusual" tests in all.rs, "OptLink" or "asm". I'll try to triage...

dot-asm:
OK, the `--release` builds actually work. Is it possible that it's mingw32 equivalent of #694? I write "equivalent" because DWARF is not used in mingw... Right? Either way, it's not impossible that it's a debugging information mismatch that is driving GNU ld crazy...

dot-asm:
> the `--release` builds actually work.

Just in case, **all** tests work with `--release`. That is even the "OptLink", "asm" and "cpp".

ChrisDenton:
Ah, yeah, I do think mingw uses DWARF for exception handling. Only msvc targets use SEH currently.

ChrisDenton:
So the workaround for now would be to set the job to `windows-2019`?

dot-asm:
Well, whatever information is there to assist the exception handling is unlikely to be a problem. Because if it was, then even `--release` build would be failing...

dot-asm:
> So the workaround for now would be to set the job to `windows-2019`?

This was tested and proven to work, yes.

ChrisDenton:
> Well, whatever information is there to assist the exception handling is unlikely to be a problem. Because if it was, then even --release build would be failing...

Sure but I mean if it's using DWARF for exception handling then I would be very surprised if it wasn't also using DWARF for debugging symbols.

dot-asm:
Opened #697 as temporary workaround.

dot-asm:
BTW. Chris! Since you can reproduce it on your computer, could you check if setting CFLAGS and CXXFLAGS environment variables to `-gdwarf-2` makes `cargo test  --manifest-path cc-test/Cargo.toml --target i686-pc-windows-gnu` work? This would allow us to figure out if it's the debugging information version mismatch that is the actual problem...

ChrisDenton:
Yep, those flags seem to fix it. 

```
$ CFLAGS=-gdwarf-2 CXXFLAGS=-gdwarf-2 cargo +stable-i686-pc-windows-gnu test --manifest-path cc-test/Cargo.toml --target-dir Z:\\cc-test\\target
   Compiling libc v0.2.126
   Compiling lazy_static v1.4.0
   Compiling either v1.7.0
   Compiling cc v1.0.73 (D:\Projects\cc-rs)
   Compiling which v4.2.5
   Compiling cc-test v0.1.0 (D:\Projects\cc-rs\cc-test)
    Finished test [unoptimized + debuginfo] target(s) in 6.00s
     Running tests\all.rs (Z:\cc-test\target\debug\deps\all-aa90ab550493a218.exe)

running 6 tests
test asm_here ... ok
test bar_here ... ok
test baz_here ... ok
test foo_here ... ok
test opt_linkage ... ok
test windows_here ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

dot-asm:
Yeah, it looks very much like mingw32 equivalent of #694, I'll leave a note there. Thanks!

PR #684, opened 2022-06-01
Map source absolute paths to OUT_DIR as relative.

### LINKED ISSUE #683: Bug: Duplicate symbols if there multiple source files have the same name
The `compile()` complains about duplicate symbols when there are multiple input files with the same name but in different directories. The `cc` names the object file by its source file and places them under `OUT_DIR`. It causes one object file to replace the other file, and pack the last object file twice in the .a static library.

This is an example of the occurrence. The `target/debug/build/xxx/output` log shows that two `empty__functions.c` files from different dirs are compiled into `empty_functions.o` in the same directory.

```
running: "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-I" "/xxx/include" "-Wall" "-Wextra" "-o" "/my-crate/target/debug/build/rclrust-msg-ac6046ca5e70df07/out/empty__functions.o" "-c" "/xxx/include/std_msgs/msg/detail/empty__functions.c"
...
running: "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-I" "/xxx/include" "-Wall" "-Wextra" "-o" "/my-crate/target/debug/build/rclrust-msg-ac6046ca5e70df07/out/empty__functions.o" "-c" "/xxx/include/std_srvs/srv/detail/empty__functions.c"
...
```

dot-asm:
If a source file was specified by absolute path, then the corresponding
object file was placed into OUT_DIR. This posed a problem if multiple
files with the same base name were specified by absolute paths.

Fixes #683.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1023,7 +1023,26 @@ impl Build {
 
         let mut objects = Vec::new();
         for file in self.files.iter() {
-            let obj = dst.join(file).with_extension("o");
+            let obj = if file.has_root() {
+                // If `file` is an absolute path, try to remove the part
+                // common with the destination directory, and graft the
+                // remaining part. Most common outcome would be removal
+                // of the home directory, next common - removal of the root
+                // directory.
+                let mut pre = dst.components();
+                let mut dst = dst.clone();
+                for comp in file.components() {
+                    if comp != pre.next().unwrap_or(Component::CurDir) {
+                        match comp {
+                            Component::Normal(c) => dst.push(c),
+                            _ => (),
+                        };
+                    }
+                }
+                dst.with_extension("o")
+            } else {
+                dst.join(file).with_extension("o")
+            };
             let obj = if !obj.starts_with(&dst) {
                 dst.join(obj.file_name().ok_or_else(|| {
                     Error::new(ErrorKind::IOError, "Getting object file details failed.")
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,11 +56,12 @@
 #![allow(deprecated)]
 #![deny(missing_docs)]
 
-use std::collections::HashMap;
+use std::collections::{hash_map, HashMap};
 use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fmt::{self, Display, Formatter};
 use std::fs;
+use std::hash::Hasher;
 use std::io::{self, BufRead, BufReader, Read, Write};
 use std::path::{Component, Path, PathBuf};
 use std::process::{Child, Command, Stdio};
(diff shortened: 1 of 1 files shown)


de-vri-es:
This method can still cause multiple distinct paths to be mapped to the same object file.

Suppose that you're compiling `/home/me/foo/bar.c` and `foo/bar.c` in `/home/me/my-crate`, this scheme will map both to the same output path. Any scheme that strips a variable prefix will have this problem. You can also get it with two absolute paths:

/home/me/foo/bar.c
/home/foo/bar.c

A different option could be to include a hash of the paths in the object name for all objects. That avoids clashes without producing very long file paths.

dot-asm:
Well, arguably suggested scenario is unlikely to be a common case, but I've added unique prefix based on absolute path for maintainers to consider :-)

de-vri-es:
I'd remove the path grafting entirely now. I'd even consider doing the same for relative paths and dump all object files in one directory. Although I guess it's nice to mimic the directory structure for relative files.

dot-asm:
> I'd remove the path grafting entirely now.

Even for relative paths? The original idea was to be as conservative as possible. I mean **not** changing the way it currently works in the most common case of relative paths. As for absolute paths, yeah, I've thought of "flattening" and reckoned that keeping traces of the directory structure could be helpful for trouble-shooting purposes. But sure, it's definitely a viable option. For both relative and absolute paths. Nobody traverses the object files' destination directories anyway...

dot-asm:
> I've thought of "flattening" and reckoned that keeping traces of the directory structure could be helpful for trouble-shooting purposes.

On the other hand, one can totally say that "flattening" would actually be more a conservative approach, because that's how the absolute paths are treated now. Fair enough, I'll remove the grafting for absolute paths and let maintainers choose...

dot-asm:
Note for maintainers. There are three suggestions to consider, one per commit. First one attempts to graft subdirectories of absolute source files' paths to output directory. The second one prefixes object files with source directories' hashes. And the third one removes grafting. There is also suggestion to use hashes as prefixes even with relative paths.

thomcc (approved):
This seems fine -- I prefer this approach to flattening.

One concern might be if the hashing in the filename defeats the ability of tools like `sccache` to cache, but I think they should only care about input filenames.

dot-asm:
> This seems fine -- I prefer this approach to flattening.

Could you clarify the course of further actions for me here. I mean should I remove all commits but the fist here, or do you simply cherry-pick it yourself?

thomcc:
Sorry, that wasn't about the commit structure. I just forgot to merge

PR #685, opened 2022-06-05
Use RUSTC_LINKER's prefix as last resort for prefix_for_target().

dot-asm:
User-provided cross-linker is a working cross-compiler with a perfectly usable prefix. Usable in the prefix_for_target() context that is. One can wonder if it's possible to even remove the majority of the mapping table... (I write "majority" because \*-windows-gnu [and no-std?] users are excused from configuring cross-linker.) Potentially fixes #666 and #637. [~Possibly even #654.~ To help #654 one would have to remove the musl mappings. So that the fallback is actually fallen on.]

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2595,11 +2595,22 @@ impl Build {
     }
 
     fn prefix_for_target(&self, target: &str) -> Option<String> {
+        // Put aside RUSTC_LINKER's prefix to be used as last resort
+        let rustc_linker = self.getenv("RUSTC_LINKER").unwrap_or("".to_string());
+        // let linker_prefix = rustc_linker.strip_suffix("-gcc"); // >=1.45.0
+        let linker_prefix = if rustc_linker.len() > 4 {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2595,11 +2595,22 @@ impl Build {
     }
 
     fn prefix_for_target(&self, target: &str) -> Option<String> {
+        // Put aside RUSTC_LINKER's prefix to be used as last resort
+        let rustc_linker = self.getenv("RUSTC_LINKER").unwrap_or("".to_string());
+        // let linker_prefix = rustc_linker.strip_suffix("-gcc"); // >=1.45.0
+        let linker_prefix = if rustc_linker.len() > 4 {
```
thomcc, line 2601: I'd probably implement this as `if rustc_linker.ends_with("-gcc") { ... }`, but I don't really think this needs to be done here (I probably will be making a case for bumping `cc-rs`'s MSRV in coherence with whatever we decide in https://github.com/rust-lang/libs-team/issues/72, so I don't think this needs to be bad for that long).

thomcc (approved):
This seems very reasonable to me.

PR #688, opened 2022-06-13
Add `whole_archive` option to `cc::Build`

onalante-msft:
Rust 1.61 no longer links with `whole-archive` in some cases, instead linking in this manner only if requested[^1]. This means that the emitted Cargo metadata can sometimes be insufficient to produce usable artifacts from `cc::Build` outputs[^2]. A workaround is to set `cargo_metadata` to `false` and print the appropriate `rustc-link-{lib,search}` arguments, but having this as a built-in option would be helpful since the default linking mode (with respect to the Cargo metadata) is static linking.

[^1]: https://github.com/rust-lang/rust/blob/1.61.0/RELEASES.md#compatibility-notes
[^2]: Concrete example: `bind_engine` and `v_check` from [here](https://github.com/Azure/iot-identity-service/blob/b8491c6b847975f62044d9b5e62e9c9800176657/key/aziot-key-openssl-engine-shared/build/engine.c#L22-L25) are omitted when linking into a `cdylib`: [build script](https://github.com/Azure/iot-identity-service/blob/b8491c6b847975f62044d9b5e62e9c9800176657/key/aziot-key-openssl-engine-shared/build/main.rs).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,6 +114,7 @@ pub struct Build {
     compiler: Option<PathBuf>,
     archiver: Option<PathBuf>,
     cargo_metadata: bool,
+    whole_archive: Option<bool>,
     pic: Option<bool>,
     use_plt: Option<bool>,
     static_crt: Option<bool>,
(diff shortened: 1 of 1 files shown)


petrochenkov:
Previous PR about this - https://github.com/rust-lang/cc-rs/pull/671.

PR #690, opened 2022-06-24
The scope of the unsafe block can be appropriately reduced

peamaeq:


thomcc (changes requested):
Mind rebasing and reformatting? I think this was probably fine before, but I guess this makes it more clear where unsafe things are happening.

dtolnay:
FYI I think this was a misguided bot. They opened 50+ similar PRs on the same day, largely broken in a variety of ways, and have not engaged with review feedback on a single one of them.

thomcc:
Then I'll close.

PR #694, opened 2022-07-07
Use a DWARF version consistent with rustc

glandium:
Rustc defaults to DWARF-2 on some targets, and DWARF-4 on others.
However using -g with the C compiler yields whatever default version the
C compiler prefers.

One side effect is that the DWARF debug info shipped in some libraries
with rustc itself (e.g. libcompiler_builtins and others) have recently
switched to DWARF-5 as a side effect of upgrading the clang version
used on rustc CI. (https://github.com/rust-lang/rust/issues/98746)

Ideally, the preferred DWARF version would be given by the rust compiler
and/or cargo, but that's not the case at the moment, so the next best
thing is something that aligns with the current defaults, although
work in under way to add a rustc flag that would allow to pick the
preferred DWARF version (https://github.com/rust-lang/rust/pull/98350)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -214,13 +214,17 @@ enum ToolFamily {
 
 impl ToolFamily {
     /// What the flag to request debug info for this family of tools look like
-    fn add_debug_flags(&self, cmd: &mut Tool) {
+    fn add_debug_flags(&self, cmd: &mut Tool, dwarf_version: Option<u32>) {
         match *self {
             ToolFamily::Msvc { .. } => {
                 cmd.push_cc_arg("-Z7".into());
             }
             ToolFamily::Gnu | ToolFamily::Clang => {
-                cmd.push_cc_arg("-g".into());
+                cmd.push_cc_arg(
+                    dwarf_version
+                        .map_or_else(|| "-g".into(), |v| format!("-gdwarf-{}", v))
+                        .into(),
+                );
             }
         }
     }
@@ -1589,7 +1593,7 @@ impl Build {
                 cmd.args.push("-G".into());
             }
             let family = cmd.family;
-            family.add_debug_flags(cmd);
+            family.add_debug_flags(cmd, self.get_dwarf_version());
         }
 
         if self.get_force_frame_pointer() {
@@ -2846,6 +2850,25 @@ impl Build {
             Some(s) => s != "false",
             None => false,
         })
+    }
+
+    fn get_dwarf_version(&self) -> Option<u32> {
+        // Tentatively matches the DWARF version defaults as of rustc 1.62.
+        let target = self.get_target().ok()?;
+        if target.contains("android")
+            || target.contains("apple")
+            || target.contains("dragonfly")
+            || target.contains("freebsd")
+            || target.contains("netbsd")
+            || target.contains("openbsd")
+            || target.contains("windows-gnu")
+        {
+            Some(2)
+        } else if target.contains("linux") {
+            Some(4)
+        } else {
+            None
+        }
     }
 
     fn get_force_frame_pointer(&self) -> bool {
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -20,7 +20,7 @@ fn gnu_smoke() {
     test.cmd(0)
         .must_have("-O2")
         .must_have("foo.c")
-        .must_not_have("-g")
+        .must_not_have("-gdwarf-4")
         .must_have("-c")
         .must_have("-ffunction-sections")
         .must_have("-fdata-sections");
@@ -53,10 +53,25 @@ fn gnu_opt_level_s() {
 }
 
 #[test]
+fn gnu_debug() {
+    let test = Test::gnu();
+    test.gcc().debug(true).file("foo.c").compile("foo");
+    test.cmd(0).must_have("-gdwarf-4");
+
+    let test = Test::gnu();
+    test.gcc()
+        .target("x86_64-apple-darwin")
+        .debug(true)
+        .file("foo.c")
+        .compile("foo");
+    test.cmd(0).must_have("-gdwarf-2");
+}
+
+#[test]
 fn gnu_debug_fp_auto() {
     let test = Test::gnu();
     test.gcc().debug(true).file("foo.c").compile("foo");
-    test.cmd(0).must_have("-g");
+    test.cmd(0).must_have("-gdwarf-4");
     test.cmd(0).must_have("-fno-omit-frame-pointer");
 }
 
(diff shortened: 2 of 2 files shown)


glandium:
The CI failures don't seem to be my doing.

dot-asm:
> The CI failures don't seem to be my doing.

Right, see #677.

src/lib.rs
```diff
@@ -2771,6 +2775,23 @@ impl Build {
         })
     }
 
+    fn get_dwarf_version(&self) -> Option<u32> {
```
dot-asm, line 2778: I would argue that it would be more than appropriate to add a commentary stating that the version choice reflects rustc defaults as of version X through Y. Idea is that whenever defaults change, there would be a hint that corresponding adjustment would be needed.

dot-asm:
On a related note, mingw32 seems to suffer from the same problem. Or at least it appears to be a reasonable conclusion to reach based on the facts that a) attempt to use latest mingw toolchain renders debug-build executables non-executable (see https://github.com/rust-lang/cc-rs/pull/677#issuecomment-1179687775), and b) imposing `-gdwarf-2` flags on C and C++ compilers mitigates the problem (see https://github.com/rust-lang/cc-rs/pull/677#issuecomment-1179794363).

mati865:
Depending on toolchain version you will end up with one of these Dwarf versions: 2, 4, 5 when targeting 32-bit mingw-w64.
Toolchain used to build Rust is rather ancient one so I'm fairly sure it uses Dwarf-2 by default. How would potential upgrade path for it look like if we choose one version? I.e. should upgrade of mingw-w64 toolchain also change Dwarf version in the spec?

src/lib.rs
```diff
@@ -2771,6 +2775,25 @@ impl Build {
...
+        // Tentatively matches the DWARF version defaults as of rustc 1.62.
+        let target = self.get_target().ok()?;
+        if target.contains("android")
+            || target.contains("apple")
+            || target.contains("dragonfly")
+            || target.contains("freebsd")
+            || target.contains("netbsd")
+            || target.contains("openbsd")
+            || target.contains("windows-gnu")
```
dot-asm, line 2787: Just in case. The original remark was specifically about mingw32, the 32-bit target. The 64-bit one was **not** observed to fail. And the thing is that "windows-gnu" would cover both 32- and 64-bit variants. At the same time the flag appears to be recognized by the 64-bit toolchain and causes no run-time problems. In other words, even though this line doesn't pinpoint the specific problem, it doesn't seem to incur any collateral damage. If it's acceptable is up to the actual cc-rs maintainers.

dot-asm:
For reference, CUDA failure appears to be transient, it failed to download package index files. They suggest it can be side effect of "Mirror sync in progress".

dot-asm:
> How would potential upgrade path for it look like if we choose one version?

Not that I can offer any real answers here, but what would be definitely helpful if linker was producing working executable files even when faced with mismatching DWARF versions. One can even wonder if it would be possible to classify it as a GNU ld bug. More specifically such an **ungraceful** failure mentioned above...

src/lib.rs
mati865: I have no idea what will happen on 64-bit targets because they typically use SEH instead of DWARF.
I worry it can work with GNU tools but introduce bugs like https://github.com/rust-lang/rust/issues/99143 with LLVM.

mati865:
> One can even wonder if it would be possible to classify it as a GNU ld bug.

IMO it's surely a bug in GNU ld but it's not clear if we can do anything other than workarounds here.

src/lib.rs
dot-asm: > 64-bit targets because they typically use SEH instead of DWARF.

Note that the 32-bit problem is not really/necessarily about exception handling. Because it was specifically debug-build binaries, ones packed with debugging information, that were rendered unexecutable. This is not to say that I'm saying that the 64-bit toolchain uses DWARF for debugging symbols. Because I'm not, as I don't really know. But I can confirm that -gdwarf appears to be an acceptable option. It can just as well be ignored, but the binaries are executable, and that's what counts for the moment.

> it can work with GNU tools but introduce bugs ... with LLVM.

But it won't be affected by this target.contains. Right? So that if anything, you should double-check and report back. Or keep this in mind and suggest a workaround later on.

glandium: > I have no idea what will happen on 64-bit targets because they typically use SEH instead of DWARF. I worry it can work with GNU tools but introduce bugs like [rust-lang/rust#99143](https://github.com/rust-lang/rust/issues/99143) with LLVM.

Exceptions handling and Debug info are more or less orthogonal. `-g` defaults to DWARF on mingw targets for both GCC and clang. I'm not sure what rust was doing for those targets before #99143, though.

glandium:
ping?

thomcc:
Can you rebase? The CI failures should be fixed now.

thomcc (approved):
This largely looks fine. I'd *really* prefer long-term this information come from cargo, via an env var or something. That seems necessary to do the right thing here if rustc allows setting this via a flag, for example (Well, I suppose we could parse rustflags...)

I won't block on this, though.

My one concern is if folks are using compilers that don't support this argument, then this might cause issues. I'm not sure if such compilers exist, but I suppose we'll find out...

glandium:
Rebased.

thomcc:
Thanks

thomcc:
This is in [1.0.77](https://github.com/rust-lang/cc-rs/releases/tag/1.0.77).

One concern I have here is that in https://github.com/rust-lang/rust/pull/104385 we'll be updating the DWARF version to DWARF-4 on Apple targets.

@glandium, are you familiar with how this will impact this patch? Is there an issue with `cc` using DWARF-2 when the rust compiler uses DWARF-4? I'd assume there isn't, but figured you'd know.

If there is, I suppose `cc` will need do version-detection on rustc to handle this case... Which is gross, but is not exactly technically challenging.

glandium:
Mixing versions shouldn't cause problems. We've been mixing DWARF-2 and DWARF-4 on mac builds of Firefox for a very long time (DWARF-2 from rust, DWARF-4 from clang). DWARF-5 was only causing problems because not all tooling supports it.
Ideally, though, cc-rs would get the information from rustc somehow.

thomcc:
> Mixing versions shouldn't cause problems. We've been mixing DWARF-2 and DWARF-4 on mac builds of Firefox for a very long time (DWARF-2 from rust, DWARF-4 from clang). DWARF-5 was only causing problems because not all tooling supports it.

That's what I thought, but wanted to confirm, thanks.

> Ideally, though, cc-rs would get the information from rustc somehow.

Yeah, I agree. Probably through cargo, although that's a bit of a headache for cargo.

PR #696, opened 2022-07-09
Detect msvc version if there is no default

### LINKED ISSUE #695: Cannot find MSVC 2019 build tools if installed with Visual Studio 2022
Reported here: https://github.com/rust-lang/rustup/issues/3024

A brief investigation shows this line to be the problem: https://github.com/rust-lang/cc-rs/blob/53272c5953cf0d8d6bfe7252935f9ff6fca56ce2/src/windows_registry.rs#L461-L462

If the Visual Studio 2022 does not have the MSVC 2022 build tools then `Microsoft.VCToolsVersion.default.txt` may be missing. However, if MSVC 2019 is installed `Microsoft.VCToolsVersion.v142.default.txt` is present. Therefore a possible fix would be to fallback to searching for this and other variations on this file name.

**EDIT:** This appears to also be broadly true for any version other than latest.

ChrisDenton:
Fixes #695

If there is no default msvc version file, then this falls back to searching for other version files. This fixes an issue discovered in https://github.com/rust-lang/rustup/issues/3024 where installing anything but the latest msvc version meant that rustup couldn't find it.

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -450,14 +450,10 @@ mod impl_ {
 
     fn vs15plus_vc_paths(
         target: &str,
-        instance_path: &PathBuf,
+        instance_path: &Path,
     ) -> Option<(PathBuf, PathBuf, PathBuf, PathBuf, PathBuf)> {
-        let version_path =
-            instance_path.join(r"VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt");
-        let mut version_file = File::open(version_path).ok()?;
-        let mut version = String::new();
-        version_file.read_to_string(&mut version).ok()?;
-        let version = version.trim();
+        let version = vs15plus_vc_read_version(instance_path)?;
+
         let host = match host_arch() {
             X86 => "X86",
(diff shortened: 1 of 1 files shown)


ChrisDenton:
Hm, these failures seem unrelated to my PR? Nothing should prevent even running tests. Looking back at other PRs it seems CI has been flaky as of late.

dot-asm:
> Hm, these failures seem unrelated to my PR?

Right, see #677.

youknowone:
@dot-asm @thomcc Does this PR requires more actions?

dot-asm:
> @dot-asm

My remark was purely on the GitHub Actions **then** failure, which is unrelated to the suggested modifications. In other words I had no comment on the actual substance, and still don't.

thomcc (approved):
Looks good to me. I'd merge but I'll leave it to @ChrisDenton to do so.

ChrisDenton:
I'd honestly forgotten about this PR. I'll rebase and check my working...

ChrisDenton:
lgtm, thanks Chris!

PR #697, opened 2022-07-10
.github/workflows/main.yml: force mingw32 test to Windows 2019.

dot-asm:
Actions' windows-latest is equipped with newer mingw, which causes
linking problems in mingw32 debug builds. See #677 for details.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,7 +48,11 @@ jobs:
             rust: stable
             target: x86_64-pc-windows-msvc
           - build: mingw32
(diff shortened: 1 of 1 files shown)


ChrisDenton:
You might also want to add a comment in the yaml calling out that this is a temporary workaround.

dot-asm:
Will do later today, as I have to run now. (I'd have to figure out if the usual way to comment works in that spot first...)

dot-asm:
Added comment.

Mark-Simulacrum (approved):

PR #701, opened 2022-07-19
Add option to emit `rerun-if-env-changed` metadata

### LINKED ISSUE #230: Use `rerun-if` features to detect when code needs rebuilding

nathaniel-daniel:
Partially fixes #230 by emitting directives for environment variables. Looking at #443, it seems uncontroversial to add an option to emit this metadata for environment variables. 

This PR adds an option called `emit_rerun_if_env_changed` to control this behavior, defaulted to false. I added a check in `getenv` to print the metadata if this flag is true. Since I use the `print` function, `cargo_metadata` must also be true for this flag to have any effect.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2802,12 +2802,12 @@ impl Build {
     }
 
     fn getenv(&self, v: &str) -> Option<String> {
-        if self.emit_rerun_if_env_changed {
-            self.print(&format!("cargo:rerun-if-env-changed={}", v));
-        }
         let mut cache = self.env_cache.lock().unwrap();
         if let Some(val) = cache.get(v) {
             return val.clone();
+        }
+        if self.emit_rerun_if_env_changed {
+            self.print(&format!("cargo:rerun-if-env-changed={}", v));
         }
         let r = env::var(v).ok();
         self.print(&format!("{} = {:?}", v, r));
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -924,8 +924,8 @@ impl Build {
         self.use_plt = Some(use_plt);
         self
     }
-    
-    /// Define whether metadata should be emitted for cargo to detect environment 
+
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2788,6 +2802,9 @@ impl Build {
     }
 
     fn getenv(&self, v: &str) -> Option<String> {
+        if self.emit_rerun_if_env_changed {
```
thomcc, line 2805: We only need to do this if the var was not cached, so this ideally would be only in the branch that fails to read from the cache.

thomcc (changes requested):
This is mostly good, but I'd like to avoid emitting this for the same var multiple times.

thomcc (approved):
This looks better. In the future it would be nice to avoid doing this for vars cargo sets on its own, but I don't know how much that matters.

thomcc:
Thinking more, I'm really unsure why we wouldn't default this to true.

nathaniel-daniel:
Thanks for the merge/fixes.

>This looks better. In the future it would be nice to avoid doing this for vars cargo sets on its own, but I don't know how much that matters.

I figured it didn't matter too much, and wasn't sure how much overhead checking a list would be. 

> Thinking more, I'm really unsure why we wouldn't default this to true.

I was going off of https://github.com/rust-lang/cc-rs/pull/443#issuecomment-535144089 for this PR. I guess its conceivable that a build script may not support being rerun, so emitting rerun metadata might break those since I think cargo metadata is emitted by default. However, I couldn't find the exact rationale behind defaulting to false. 

I wanted to create an uncontroversial PR to improve the situation around rerunning, since progress there seemed to have stalled for half a decade and I run into this issue a lot. 

Also, is #230 considered fixed by this? Is there nothing that can be done for files aside from expecting the user to emit metadata themselves?

PR #703, opened 2022-07-26
Fix xctoolchain AppleClang to corrrectly use -isysroot

youknowone:
The first commit is #708 but unseparatable.

Fix AppleClang build for -apple-darwin to add -isysroot for MacOSX.sdk correctly when cc is under xctoolchain.
Becasue `-isysroot` priors to `$SDKROOT` env variable, the original ad-hoc fix is not required anymore.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2515,7 +2515,8 @@ impl Build {
             ArchSpec::Catalyst(_) => "macosx".to_owned(),
         };
 
-        if !is_mac {
+        // AppleClang sometimes requires sysroot even for darwin
+        if cmd.is_xctoolchain_clang() || !target.ends_with("-darwin") {
             self.print(&format_args!("Detecting {:?} SDK path for {}", os, sdk));
             let sdk_path = if let Some(sdkroot) = env::var_os("SDKROOT") {
                 sdkroot
@@ -2525,7 +2526,10 @@ impl Build {
 
             cmd.args.push("-isysroot".into());
             cmd.args.push(sdk_path);
-            // TODO: Remove this once Apple stops accepting apps built with Xcode 13
+        }
+
+        // TODO: Remove this once Apple stops accepting apps built with Xcode 13
+        if !is_mac {
             cmd.args.push("-fembed-bitcode".into());
         }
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -2151,6 +2170,12 @@ impl Build {
         };
 
         let (sdk_prefix, sim_prefix, min_version) = match os {
+            Os::MacOs => (
+                "macosx",
+                "",
+                std::env::var("MACOSX_DEPLOYMENT_TARGET")
+                    .unwrap_or_else(|_| (if self.cpp { "10.9" } else { "10.4" }).into()),
```
thomcc, line 2177: As I mentioned in the other PR, I'd prefer this match the logic in https://github.com/rust-lang/rust/blob/6365e5ad9fa9e2ec867a67aeeae414e7c62d8354/compiler/rustc_target/src/spec/apple_base.rs#L105-L114

src/lib.rs
```diff
@@ -2180,11 +2210,17 @@ impl Build {
...
-        cmd.args.push("-fembed-bitcode".into());
+        // AppleClang sometimes needs sysroot even for darwin
+        if cmd.is_xctoolchain_clang() || !target.ends_with("-darwin") {
+            self.print(&format!("Detecting {} SDK path for {}", os, sdk));
+            let sdk_path = self.apple_sdk_root(sdk.as_str())?;
+            cmd.args.push("-isysroot".into());
+            cmd.args.push(sdk_path);
+        }
+
+        if !is_mac {
```
thomcc, line 2221: Can you explain why this is conditional now?

src/lib.rs
```diff
@@ -2103,12 +2111,23 @@ impl Build {
             None => false,
         };
 
-        let is_sim = match target.split('-').nth(3) {
+        let is_arm_sim = match target.split('-').nth(3) {
```
thomcc, line 2114: This should still be named `is_sim` -- As it is, further down we have x86-specific logic in an `if is_arm_sim` branch.

thomcc (changes requested):
This looks like a good idea to me, but I'm not sure this is the way to go about it. I've left some comments but will need to do further investigation of what's expected here (and review some of the work by @indygreg in `apple-sdk`).

src/lib.rs
youknowone: The whole function `apple_flags` - which was previously `ios_watchos_flags` was not called from macos target before. I'd like to keep the previous behavior for macos target in this PR. If you'd like to add this to macos too, I will remove it.

thomcc: Thanks, good catch.

youknowone:
The first commit is #708

youknowone:
@thomcc  rebase done, could you review it again?

src/lib.rs
```diff
@@ -3684,6 +3675,17 @@ impl Tool {
         self.family == ToolFamily::Clang
     }
 
+    /// Whether the tool is AppleClang under .xctoolchain
+    #[cfg(target_vendor = "apple")]
+    fn is_xctoolchain_clang(&self) -> bool {
+        let path = self.path.to_str().unwrap();
```
thomcc, line 3681: This has bitten us before, it should probably be to_string_lossy.

youknowone: Thank you, I fixed it.

thomcc (approved):
Thanks.

PR #705, opened 2022-07-27
Fix `--target` getting passed twice to the Android NDK clang on Windows

Arc-blroth:
When using the Android NDK toolchain on Windows, `cc-rs` currently attempts to pass `--target` twice: 
```
running: "C:\\Users\\arcdev\\AppData\\Local\\Android\\Sdk\\ndk\\25.0.8775105\\toolchains\\llvm\\prebuilt\\windows-x86_64\\bin\\clang.exe" "--target=armv7a-linux-androideabi21" "-O0" "-DANDROID" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "--target=armv7-linux-androideabi" "-Wall" "-Wextra" "-E" "src/android-api.c"
```
Since the second `--target` doesn't have an Android ABI version attached, this causes the NDK to not define `__ANDROID_API__`, among other things.

This seems to be a regression introduced by #572, which replaces the clang path in `get_base_compiler` and thus accidently breaks the `android_clang_compiler_uses_target_arg_internally` check later in `add_default_flags`.
This PR adds a field `has_internal_target_arg` to the `Tool` struct, which is set inside `get_base_compiler` and replaces the check in `add_default_flags`, so everything works as originally intended.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -200,6 +200,7 @@ pub struct Tool {
     family: ToolFamily,
     cuda: bool,
     removed_args: Vec<OsString>,
+    has_internal_target_arg: bool,
 }
 
 /// Represents the family of tools this tool belongs to.
@@ -1824,9 +1825,7 @@ impl Build {
         // Target flags
         match cmd.family {
             ToolFamily::Clang => {
-                if !(target.contains("android")
-                    && android_clang_compiler_uses_target_arg_internally(&cmd.path))
-                {
+                if !(target.contains("android") && cmd.has_internal_target_arg) {
                     if target.contains("darwin") {
                         if let Some(arch) =
                             map_darwin_target_from_rust_to_compiler_architecture(target)
@@ -2696,6 +2695,7 @@ impl Build {
                 let file_name = path.to_str().unwrap().to_owned();
                 let (target, clang) = file_name.split_at(file_name.rfind("-").unwrap());
 
+                tool.has_internal_target_arg = true;
                 tool.path.set_file_name(clang.trim_start_matches("-"));
                 tool.path.set_extension("exe");
                 tool.args.push(format!("--target={}", target).into());
@@ -3522,6 +3522,7 @@ impl Tool {
             family: family,
             cuda: false,
             removed_args: Vec::new(),
+            has_internal_target_arg: false,
         }
     }
 
@@ -3582,6 +3583,7 @@ impl Tool {
             family: family,
             cuda: cuda,
             removed_args: Vec::new(),
+            has_internal_target_arg: false,
         }
     }
 


ChrisDenton:
Ok, I've been looking at this and it seems reasonable to me. Since it's been awhile (sorry for the long delay) can you confirm that this is still the approach you want and that it's working for you, etc?

pitaj:
@Arc-blroth here from triage. Checking to see if you're still active.

Please respond, thanks

Arc-blroth:
@pitaj I'm still here! I've been extremely busy with other work these past few days, but I'll try to get around to looking at this within the next week.

Arc-blroth:
I kinda forgot about this PR again (whoops) I'll actually take a look sometime tomorrow morning

ChrisDenton:
Thanks! When you have a moment could you also squish the commits into one?

Arc-blroth:
(update) I'm getting a weird error in cc's inclusion of backtrace's build script when trying to build a toolchain with this and rust-lang/rust#116318 - debugging that right now...
backtrace-sus

Arc-blroth:
in the backtrace-rs buildscript `std::env::current_dir()` is `rust\library\std` and `file!()` returns `library\std\..\backtrace\build.rs`, so backtrace-rs is trying to resolve `android-api.c` at `rust\library\std\library\backtrace\src\android-api.c`, which definitely doesn't exist

pitaj:
Not exactly sure how to resolve that yet. In the meantime, we should definitely change the CI tests so this is caught by the android Docker run.
[It currently shows up but only as a warning](https://github.com/rust-lang/backtrace-rs/actions/runs/5948900382/job/16133608934#step:7:1514) which is why we didn't catch it before.

Arc-blroth:
adding
```diff
+ assert!(android_api_c.try_exists().unwrap(), "android-api.c could not be found for compilation");
```
to backtrace-rs' buildscript causes the entire build to fail (right now), so we should leave that in as a future check

pitaj:
An alternative is to just use `.expand()` instead of matching on `.try_expand()`:
```rust
let expansion = cc::Build::new().file(android_api_c).expand();
```
Which will panic on failure, which seems like the right thing to do anyways.

I feel like some of the other error cases should probably also not fail silently:
```rust
    // When will this not be utf-8? Can probably just unwrap
    let expansion = match std::str::from_utf8(&expansion) {
        Ok(s) => s,
        Err(_) => return,
    };
    println!("expanded android version detection:\n{}", expansion);
    let marker = "APIVERSION";
    // If marker not found, probably a big problem?
    let i = match expansion.find(marker) {
        Some(i) => i,
        None => return,
    };
    // Lack of the __ANDROID_API__ expansion is probably okay
    let version = match expansion[i + marker.len() + 1..].split_whitespace().next() {
        Some(s) => s,
        None => return,
    };
    // But if it exists but isn't a number? Probably sign of a problem that should stop the build
    let version = match version.parse::<u32>() {
        Ok(n) => n,
        Err(_) => return,
    };
    if version >= 21 {
        println!("cargo:rustc-cfg=feature=\"dl_iterate_phdr\"");
    }
```

Arc-blroth:
rust-lang/cargo#3946 would be really nice to have here for figuring out build.rs's absolute location, but no PR has been merged for it yet :(

Arc-blroth:
A more robust solution that could work in backtrace-rs right now would be to `include_str!("src/android-api.c`, and then throw that string into a temp folder for cc consumption, but it definitely feels incredibly hacky.

pitaj:
The file is pretty short, I would probably just inline the string altogether. I'll try writing it to OUT_DIR or something

Arc-blroth:
(update 2) Trying to cross-compile the stage1 library artifacts from Windows to Android, which is currently failing because rustc is linking the artifacts as if they were Windows binaries instead of Android binaries (???)

Arc-blroth:
It seems that this (year old) code still works!

```
running: "C:\\Users\\arc\\AppData\\Local\\Android\\Sdk\\ndk\\26.0.10792818\\toolchains\\llvm\\prebuilt\\windows-x86_64\\bin\\clang.exe" "--target=aarch64-linux-android21" "-O3" "-DANDROID" "-ffunction-sections" "-fdata-sections" "-fPIC" "-DANDROID" "-ffunction-sections" "-fdata-sections" "-fPIC" "-E" "C:\\Users\\arc\\git3\\rust\\build\\x86_64-pc-windows-msvc\\stage1-std\\aarch64-linux-android\\release\\build\\std-5c28ba14c53f822b\\out\\android-api.c"
```

Along with rust-lang/rust#116318 and rust-lang/backtrace-rs#556, the toolchain built by this trio of PR produces backtraces that Just Work:tm: on Android:

```kt
package ai.arcblroth.backtracefix

import android.os.Bundle
import androidx.activity.ComponentActivity

class MainActivity : ComponentActivity() {
    companion object {
        init {
            System.loadLibrary("backtracefix")
        }
    }

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        rust()
    }

    private external fun rust()
}
```

```rs
use std::ffi::CString;
use std::os::raw::{c_char, c_int, c_void};

fn android_log_error(text: String) {
    extern "system" {
        fn __android_log_write(prio: c_int, tag: *mut c_char, text: *mut c_char);
    }

    let tag = CString::new("backtracefix").unwrap();
    let c_text = CString::new(text.replace("\x00", " ")).unwrap();
    unsafe {
        __android_log_write(6, tag.into_raw(), c_text.into_raw());
    }
}

#[no_mangle]
pub extern "C" fn Java_ai_arcblroth_backtracefix_MainActivity_rust(_env: *mut c_void, _this: *mut c_void) {
    android_log_error(format!("=== Backtrace ===\n{:?}", std::backtrace::Backtrace::force_capture()));
}
```

Logcat output:
```
=== Backtrace ===
Backtrace [{ fn: "Java_ai_arcblroth_backtracefix_MainActivity_rust", file: "C:\Users\arc\AndroidStudioProjects\BacktraceFix\backtracefix-rs\src\lib.rs", line: 18 }, { fn: "art_quick_generic_jni_trampoline" }, { fn: "art_quick_invoke_stub" }, ...
```

workingjubilee:
Nice!

ChrisDenton (approved):
Looks good to me! Unless anyone has any objections, I'll merge this soon.

PR #706, opened 2022-07-27
README.md: Add note that rerun-if-changed is not emitted.

EvanTheB:
Issue #230 is the issue for this, this patch just documents the
current behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ fn main() {
         .file("foo.c")
         .file("bar.c")
         .compile("foo");
+    println!("cargo:rerun-if-changed=foo.c");
+    println!("cargo:rerun-if-changed=bar.c");
 }
 ```
 
@@ -100,6 +102,10 @@ In addition to the above optional environment variables, `cc-rs` has some
 functions with hard requirements on some variables supplied by [cargo's
 build-script driver][cargo] that it has the `TARGET`, `OUT_DIR`, `OPT_LEVEL`,
 and `HOST` variables.
+
+Note that `cc-rs` does not automatically tell cargo to rebuild when the c files
+are changed, nor the environment variables such as CFLAGS. See the example to
+manually instruct cargo to rebuild when the source files change.
 
 [cargo]: https://doc.rust-lang.org/cargo/reference/build-scripts.html#inputs-to-the-build-script
 


dot-asm:
Customarily the decision whether or not a package gets rebuilt depends on whether or not **any** of the files in corresponding `cargo package --list [--allow-dirty]` output are modified. Customarily you would make your .c sources to be a part of the package, so that emitting explicit `cargo:rerun-if-changed` directives wouldn't be normally required. I suppose question is if it's some unusual case, then it might be more appropriate to discuss the specifics of when it would be actually required to emit the directives.

dot-asm:
There might be some misunderstanding on my[?] side. You'd use `cargo:rerun-if-changed` in case you find that rebuilding upon **any** change in `cargo package --list` excessive. But either way, the suggested "Note that `cc-rs` does not automatically tell cargo to rebuild when the c files are changed" still doesn't adequately reflect the current state of affairs. I mean in most common situation cargo will follow modification timestamps without explicit if-changed directives.

EvanTheB:
I need to recheck as I dont understand this very well. I had a c file in my src/ folder, which does show up in `cargo package --list`. However modifying that file did not cause a rebuild. Searching the issue brought me to the mentioned issue, and adding rerun-if-changed made the behaviour what I expect (rebuild when I modify the c file). There might be something else at play.

Because pkg_config is emitting `rerun-if` commands, the default behaviour you mention may not be as expected. Its not clear what is the 'right' way to solve this. But it would have saved me half a day of debugging to have this mentioned in the cc docs somehow...

```
extern crate cc;

fn main() {
    // pkg_config is needed only to pick up the include path for log.c to use.
    // libflashrom-sys tells cargo how to link to libflashrom.
    let flashrom = pkg_config::Config::new()
        .cargo_metadata(false)
        .probe("flashrom")
        .unwrap();
    let mut log_c = cc::Build::new();
    log_c.file("src/log.c");
    for p in flashrom.include_paths {
        log_c.include(p);
    }
    log_c.compile("log.o");
    println!("cargo:rerun-if-changed=src/log.c");
}
```

dot-asm:
> Because pkg_config is emitting `rerun-if` commands, the default behaviour you mention may not be as expected.

Bingo! If you have one `rerun-if`, you need all of them. And that's what the note should say instead of "cargo never tells." I suppose it would be appropriate to mention the possibility of build-dependency crates adding `rerun-if` without you realizing it in advance...

thomcc (changes requested):
As discussed in this issue, this isn't quite right (or is at least misleading). I think it *would* be good to document the behavior, but we should be somewhat more precise about how this works. Specifically, in the common case you don't need to provide these.

PR #708, opened 2022-07-30
use MACOSX_DEPLOYMENT_TARGET when set

youknowone:
~10.4 for c and 10.9 for c++~

10.7 for x86 and 11.0 for aarch64.

ref: https://github.com/rust-lang/rust/blob/6365e5ad9fa9e2ec867a67aeeae414e7c62d8354/compiler/rustc_target/src/spec/apple_base.rs#L105-L114
which referred in https://github.com/rust-lang/cc-rs/pull/703#discussion_r1005188342

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2141,8 +2141,8 @@ impl Build {
             }
         }
 
-        if target.contains("apple-ios") || target.contains("apple-watchos") {
-            self.ios_watchos_flags(cmd)?;
+        if target.contains("-apple-") {
+            self.apple_flags(cmd)?;
         }
 
         if self.static_flag.unwrap_or(false) {
@@ -2360,7 +2360,7 @@ impl Build {
         Ok(())
     }
 
-    fn ios_watchos_flags(&self, cmd: &mut Tool) -> Result<(), Error> {
+    fn apple_flags(&self, cmd: &mut Tool) -> Result<(), Error> {
         enum ArchSpec {
             Device(&'static str),
             Simulator(&'static str),
@@ -2368,12 +2368,14 @@ impl Build {
         }
 
         enum Os {
+            MacOs,
             Ios,
             WatchOs,
         }
-        impl Display for Os {
+        impl std::fmt::Debug for Os {
             fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
                 match self {
+                    Os::MacOs => f.write_str("macOS"),
                     Os::Ios => f.write_str("iOS"),
                     Os::WatchOs => f.write_str("WatchOS"),
                 }
@@ -2381,16 +2383,22 @@ impl Build {
         }
 
         let target = self.get_target()?;
-        let os = if target.contains("-watchos") {
+        let os = if target.contains("-darwin") {
+            Os::MacOs
+        } else if target.contains("-watchos") {
             Os::WatchOs
         } else {
             Os::Ios
         };
+        let is_mac = match os {
+            Os::MacOs => true,
+            _ => false,
+        };
 
-        let arch = target.split('-').nth(0).ok_or_else(|| {
+        let arch_str = target.split('-').nth(0).ok_or_else(|| {
             Error::new(
                 ErrorKind::ArchitectureInvalid,
-                format!("Unknown architecture for {} target.", os),
+                format!("Unknown architecture for {:?} target.", os),
             )
         })?;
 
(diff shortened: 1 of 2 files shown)


youknowone:
@Mark-Simulacrum Hi, you seem to maintained this repository recently. Could you review this PR?

src/lib.rs
```diff
@@ -2151,6 +2170,12 @@ impl Build {
         };
 
         let (sdk_prefix, sim_prefix, min_version) = match os {
+            Os::MacOs => (
+                "macosx",
+                "",
+                std::env::var("MACOSX_DEPLOYMENT_TARGET")
+                    .unwrap_or_else(|_| (if self.cpp { "10.9" } else { "10.4" }).into()),
```
thomcc, line 2177: 10.4 is older than Rust's deployment target. We default to 10.7, although this may change soon.

thomcc (changes requested):
This makes a lot of other changes. Can you remove those (perhaps put them in a different PR)?

I think it's reasonable to take advantage of MACOSX_DEPLOYMENT_TARGET, but I don't think we should default to setting this to something old. I'd actually probably prefer we don't set it if not provided.

Alternatively, we could [mirror the logic used by rustc_target](https://github.com/rust-lang/rust/blob/bed4ad65bf7a1cef39e3d66b3670189581b3b073/compiler/rustc_target/src/spec/apple_base.rs#L106-L123), but I'd probably lean on not passing this if it is not set.

Also sorry for the delay here, this crate should be better maintained going forward.

thomcc:
Oh, I see, the other changes are from https://github.com/rust-lang/cc-rs/pull/703, which I hadn't seen yet.

youknowone:
I did it by respecting the policy for other platforms like IPHONEOS. if you want to make different policy for macos, I will change it. Please confirm you want it.

> I think it's reasonable to take advantage of MACOSX_DEPLOYMENT_TARGET, but I don't think we should default to setting this to something old. I'd actually probably prefer we don't set it if not provided.

youknowone:
@thomcc could you review it again?

colincornaby:
I'm still getting familiar with this project so apologies if I'm misreading something.

On line 2246, Bitcode is enabled if the platform is not Mac.
- This means Bitcode is also enabled for other places where it should not be (by my reading.) iOS Simulator and Catalyst should also not have Bitcode.
- Not sure what the stance should be on Bitcode going forward. Bitcode has been discontinued in Xcode 14.0 and many developers are stripping it out of their pipelines. However - developers that need to support Xcode 13.0 may continue to need it.
- The Bitcode generated will also not be compatible with Xcode anyway in since Rust uses a different version of Bitcode. So enabling this flag may actually break iOS builds.

src/lib.rs
```diff
@@ -2168,6 +2187,18 @@ impl Build {
...
+                "macosx",
+                "",
+                std::env::var("MACOSX_DEPLOYMENT_TARGET").unwrap_or_else(|_| {
+                    (if arch_str == "aarch64" {
+                        "11.0"
+                    } else {
+                        "10.7"
+                    })
+                    .into()
+                }),
```
BlackHoleFox, line 2200: The `rustc` PR for `--print deployment-target` just landed (https://github.com/rust-lang/rust/pull/105354), and should be available starting in 1.71 fyi. Ideally that would be queried for the version and then falling back to this hardcoded logic if the compiler is too old.

francesca64:
@colincornaby this PR is minimally invasive and doesn't actually change the existing behavior around bitcode; bitcode is already being unconditionally generated for all non-macOS Apple platforms on `main`: https://github.com/rust-lang/cc-rs/blob/57853c4bf8a89a0f4c9137eb367ac580305c6919/src/lib.rs#L2340

Deciding how `cc` should handle bitcode going forward should be left to another issue/PR, since landing this PR would be hugely beneficial to anyone developing macOS apps written in Rust.

@thomcc 
> I think it's reasonable to take advantage of MACOSX_DEPLOYMENT_TARGET, but I don't think we should default to setting this to something old. I'd actually probably prefer we don't set it if not provided.

`clang` defaults to targeting the same version of macOS currently being run, which is in contrast to Rust defaulting to a specific old version for maximum support. As it stands, it's ***impossible*** for someone developing a macOS app with any dependencies that use `cc` to target macOS versions earlier than what their development machine is running. This is nearly a non-starter for anyone who wants to make binary distributions of macOS apps, as the only workaround is to fork every such dependency to use this PR.

`cc` should follow the same conventions Rust itself does, since the current behavior is very surprising and leads to countless difficult-to-trace linker warnings about object files targeting too new of a version.

As mentioned, this PR is minimally invasive and merely extends the existing behavior around Apple platforms to account for macOS as well. I've tested this PR, and it solves all of my build problems flawlessly.

src/lib.rs
```diff
@@ -2081,34 +2081,42 @@ impl Build {
...
+            Os::MacOs
+        } else if target.contains("-watchos") {
             Os::WatchOs
         } else {
             Os::Ios
         };
+        let is_mac = match os {
+            Os::MacOs => true,
+            _ => false,
+        };
```
francesca64, line 2117:
```suggestion
        let is_mac = matches!(os, Os::MacOs);
```

[`matches!`](https://doc.rust-lang.org/std/macro.matches.html) was added in 1.42, which is below the [MSRV](https://github.com/rust-lang/cc-rs/blob/57853c4bf8a89a0f4c9137eb367ac580305c6919/.github/workflows/main.yml#L113-L126).

thomcc:
Yeah, I'm fine with this PR now. My complaints about the too old version were from an earlier version. I'd be fine with a rebased version of this.

youknowone:
I wish it can be merged at this time.

BlackHoleFox:
I opened up https://github.com/rust-lang/cc-rs/pull/848 to further this PR's changes like [said on a previous comment](https://github.com/rust-lang/cc-rs/pull/708#discussion_r1187825066), so it also contains a rebased version of the commit here. Depending on which merges first, the rebase this branch needs (a one line print conflict) might be unnecessary.

youknowone:
@BlackHoleFox Thank you!

youknowone:
I rebased this PR just in case if it need to be merged first.
But I totally agree to #848 and also prefer to merge this patch through it.

BlackHoleFox:
It looks like my forward PR merged with these changes, so this one could be closed as done @youknowone

thomcc:
Yeah this is done. Sorry for all the delays here.

PR #709, opened 2022-07-31
detect clang/gcc using --version

youknowone:
This is useful on macos because `gcc` and `cc` is actually `clang`.
`gcc` and `cc` were regarded as `clang` even when it was gcc.

`m32` and `m64` is not necessary for clang, but also harmless.
Do you want to do it as now or want to revert it and change tests run only for gcc?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3451,7 +3451,7 @@ impl Tool {
                 Some(s) => s,
                 None => {
                     // --version failed. fallback to gnu
-                    println!("cargo-warning:Running failed: {:?}", cmd);
+                    println!("cargo-warning:Failed to run: {:?}", cmd);
                     return ToolFamily::Gnu;
                 }
             };
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1808,6 +1808,16 @@ impl Build {
             family.add_force_frame_pointer(cmd);
         }
 
+        if !cmd.is_like_msvc() {
+            if target.contains("i686") || target.contains("i586") {
+                cmd.args.push("-m32".into());
+            } else if target == "x86_64-unknown-linux-gnux32" {
+                cmd.args.push("-mx32".into());
+            } else if target.contains("x86_64") || target.contains("powerpc64") {
+                cmd.args.push("-m64".into());
+            }
+        }
+
         // Target flags
         match cmd.family {
             ToolFamily::Clang => {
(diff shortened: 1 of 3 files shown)


youknowone:
@thomcc This patch is broken after rebase. Could you tell me if it is worth to fix it or not?

src/lib.rs
```diff
@@ -3270,6 +3272,35 @@ impl Tool {
...
+            cmd.arg("--version");
+
+            let stdout = match run_output(&mut cmd, &path.to_string_lossy())
+                .ok()
+                .and_then(|o| String::from_utf8(o).ok())
+            {
+                Some(s) => s,
+                None => {
+                    // --version failed. fallback to gnu
+                    println!("cargo-warning:Running failed: {:?}", cmd);
```
thomcc, line 3286:
```suggestion
                    println!("cargo-warning:Failed to run: {:?}", cmd);
```

src/lib.rs
```diff
@@ -1663,6 +1663,16 @@ impl Build {
             family.add_force_frame_pointer(cmd);
         }
 
+        if !cmd.is_like_msvc() {
```
thomcc, line 1666: This change seems unrelated?

thomcc (commented):
Yeah, I think this is fine. Sorry for the delay. Can you rebase?

src/lib.rs
youknowone: I don't remember the details anymore, but they are related to tests `gnu_i686` and `gnu_x86_64`

youknowone: Ah, I got the idea. This happened because `gcc()` now target `x86_64-apple-darwin` on macOS.
I can revert this part and add target to `gnu_i686` and `gnu_x86_64`. Is this better?

youknowone:
@thomcc Tests are fixed. Could you review it again?

youknowone:
Maybe replacement of #407

thomcc (approved):
Seems good to me.

thomcc:
It looks like this broke CI, hmmm...

src/lib.rs
dot-asm: By itself it's a formally wrong move. While the -m32/m64 are accepted by clang, the --target argument [which is always passed to clang by cc-rs] describes the intention sufficiently and more importantly more adequately. In other words the options in question should have remained in the ToolFamily::Gnu section. If they cause problems in the section in question, then it would have been more appropriate to solve it there.

On a tangential note. The powepc64 gcc defaults to -m64, so that it's actually more interesting to pass -m32 than -m64.  But of course making it symmetric would be better, i.e. pass -m32 if the target is 32-bit ppc, and -m64 otherwise...

PR #712, opened 2022-08-04
Refine CUDA support.

### LINKED ISSUE #641: Fails compiling with cuda
I tried to build cuda code using the following build script:
```rust
    cc::Build::new()
        .cuda(true)
        .cudart("static")
        .flag("arch=sm_75")
        .flag("-std=c++17")
        .include("./testing/src")
        .file("testing/src/kernel.cu")
        .compile("libtesting.a");
```
And got the following error:
`  cargo:warning=nvcc fatal   : Unknown option '--target=x86_64-unknown-linux-gnu'`

Apparently nvcc doesn't like this flag.
full error:
```
  --- stdout
  TARGET = Some("x86_64-unknown-linux-gnu")
  OPT_LEVEL = Some("0")
  HOST = Some("x86_64-unknown-linux-gnu")
  CXX_x86_64-unknown-linux-gnu = None
  CXX_x86_64_unknown_linux_gnu = None
  HOST_CXX = None
  CXX = Some("clang++ -stdlib=libc++")
  NVCC_x86_64-unknown-linux-gnu = None
  NVCC_x86_64_unknown_linux_gnu = None
  HOST_NVCC = None
  NVCC = None
  CXXFLAGS_x86_64-unknown-linux-gnu = None
  CXXFLAGS_x86_64_unknown_linux_gnu = None
  HOST_CXXFLAGS = None
  CXXFLAGS = None
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some("true")
  running: "nvcc" "-ccbin=clang++" "-Xcompiler" "-O0" "-Xcompiler" "-ffunction-sections" "-Xcompiler" "-fdata-sections" "-Xcompiler" "-fPIC" "-G" "-Xcompiler" "-g" "-Xcompiler" "-fno-omit-frame-pointer" "--target=x86_64-unknown-linux-gnu" "-I" "./testing/src" "-Xcompiler" "-Wall" "-Xcompiler" "-Wextra" "arch=sm_75" "-std=c++17" "-o" .../src/kernel.o" "-c" "testing/src/kernel.cu"
  cargo:warning=nvcc fatal   : Unknown option '--target=x86_64-unknown-linux-gnu'
  exit status: 1

  --- stderr


  error occurred: Command "nvcc" "-ccbin=clang++" "-Xcompiler" "-O0" "-Xcompiler" "-ffunction-sections" "-Xcompiler" "-fdata-sections" "-Xcompiler" "-fPIC" "-G" "-Xcompiler" "-g" "-Xcompiler" "-fno-omit-frame-pointer" "--target=x86_64-unknown-linux-gnu" "-I" "./testing/src" "-Xcompiler" "-Wall" "-Xcompiler" "-Wextra" "arch=sm_75" "-std=c++17" "-o" "/.../src/kernel.o" "-c" "testing/src/kernel.cu" with args "nvcc" did not execute successfully (status code exit status: 1).
```

nvcc version:
```
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2021 NVIDIA Corporation
Built on Mon_Sep_13_19:13:29_PDT_2021
Cuda compilation tools, release 11.5, V11.5.50
Build cuda_11.5.r11.5/compiler.30411180_0
```

dot-asm:
Instead of counting all source files we count only .cu files, ones with device code that actually require the special treatment.

[Even fixes #641.]

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1367,7 +1367,7 @@ impl Build {
         if !msvc || !is_asm || !is_arm {
             cmd.arg("-c");
         }
-        if self.cuda && self.files.len() > 1 {
+        if self.cuda && self.cuda_file_count() > 1 {
             cmd.arg("--device-c");
         }
         if is_asm {
@@ -2035,7 +2035,7 @@ impl Build {
             self.assemble_progressive(dst, chunk)?;
         }
 
-        if self.cuda {
+        if self.cuda && self.cuda_file_count() > 0 {
             // Link the device-side code and add it to the target library,
             // so that non-CUDA linker can link the final binary.
 
@@ -3009,6 +3009,13 @@ impl Build {
         let ret: OsString = sdk_path.trim().into();
         cache.insert(sdk.into(), ret.clone());
         Ok(ret)
+    }
+
+    fn cuda_file_count(&self) -> usize {
+        self.files
+            .iter()
+            .filter(|file| file.extension() == Some(OsStr::new("cu")))
+            .count()
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1690,7 +1690,7 @@ impl Build {
                             cmd.args.push("--target=aarch64-unknown-windows-gnu".into())
                         }
                     } else {
-                        cmd.args.push(format!("--target={}", target).into());
+                        cmd.push_cc_arg(format!("--target={}", target).into());
                     }
                 }
             }
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -96,16 +96,14 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@master
-    - name: Install cuda-minimal-build-11-6
+    - name: Install cuda-minimal-build-11-8
       shell: bash
       run: |
         # https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=20.04&target_type=deb_network
-        wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-ubuntu2004.pin
-        sudo mv cuda-ubuntu2004.pin /etc/apt/preferences.d/cuda-repository-pin-600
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -1367,7 +1367,14 @@ impl Build {
         if !msvc || !is_asm || !is_arm {
             cmd.arg("-c");
         }
-        if self.cuda && self.files.len() > 1 {
+        if self.cuda
+            && self
+                .files
+                .iter()
+                .filter(|file| file.extension() == Some(OsStr::new("cu")))
```
thomcc, line 1374: Nit: can this be `.iter().any(|file| ...)` or is that prevented by MSRV.

src/lib.rs
```diff
@@ -2035,7 +2042,14 @@ impl Build {
             self.assemble_progressive(dst, chunk)?;
         }
 
-        if self.cuda {
+        if self.cuda
```
thomcc, line 2045: Since you have this check (are we cuda and have \>1 .cu file), it's probably worth pulling out into a function.

thomcc (commented):
Looks good but have some minor nits.

src/lib.rs
dot-asm: Documentation says "any() is short-circuiting," so one can't `.count()` it. One could use it in the second case, but since the suggestion is to make it a function for using in both cases, then it would have to be `.filter()`...

thomcc: Oh right, yeah, `>1` rather than `>=1`, my bad.

src/lib.rs
thomcc: Nevermind, this is different.

dot-asm: As in `if self.cuda && self.cuda_files_count() > X` or as in `if self.cuda_files_count() > X`?

dot-asm: Well, a cuda_files_count() would make a sense, because the formatting is unbearable :-)

thomcc: Yeah, that would be reasonable.

dot-asm:
I used the opportunity to update the CUDA toolchain.

thomcc (approved):
This looks good to me.

PR #718, opened 2022-08-26
print cargo:rerun-if-changed inside file method

Be-ing:
so users do not need to do this in their own build.rs


dot-asm:
So far answer to this question was no, see #230.

JohnTitor:
Closing as #701 should resolve the issue. Let me know if I'm missing something.

Be-ing:
AFAICT #701 only affects environment variables? This was about compiled files.

PR #725, opened 2022-10-03
Fix : Properly detect C runtime through target_feature cfg

elast0ny:
Currently, the `cc` crate tries to auto-detect the proper C runtime based on the contents of the `CARGO_CFG_TARGET_FEATURE`. Unfortunately, this env variable does not contain an accurate list of target features and may not contain the `crt-static` feature. 

For example, if your project depends on cc and is configured to set the `crt-static` feature through a `.cargo/config.toml` file, `cc` will compile using the dynamic C runtime as it wont be able to see the `crt-static` feature through `CARGO_CFG_TARGET_FEATURE`.

See https://github.com/rust-lang/cargo/issues/6858

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1491,10 +1491,7 @@ impl Build {
                     Some(true) => "-MT",
                     Some(false) => "-MD",
                     None => {
-                        let features = self
-                            .getenv("CARGO_CFG_TARGET_FEATURE")
-                            .unwrap_or(String::new());
-                        if features.contains("crt-static") {
+                        if cfg!(target_feature = "crt-static") {
                             "-MT"
                         } else {
                             "-MD"


elast0ny:
This should be a bit more flexible than #723

src/lib.rs
```diff
@@ -1491,10 +1491,7 @@ impl Build {
                     Some(true) => "-MT",
                     Some(false) => "-MD",
                     None => {
-                        let features = self
-                            .getenv("CARGO_CFG_TARGET_FEATURE")
-                            .unwrap_or(String::new());
-                        if features.contains("crt-static") {
+                        if cfg!(target_feature = "crt-static") {
```
dot-asm, line 1494: There is a reason for why it's inappropriate to use `cfg!` here. Keyword is that it applies to the build script itself, not the compilation target.

elast0ny: My tests with these changes were able to properly infer the C runtime based on the settings in my config.toml from the parent crate. 

The cc build script gets compiled during the build invocation of the parent crate which means the cgf! directive has access to the parent crate target_feature no ?

dot-asm: The build script is compiled for the platform that executes the Rust compiler, but it doesn't necessarily mean that the compiler was asked to generate code for the same platform. Well, one can argue that `*-windows-msvc` binary code is generated only by `*-windows-msvc` toolchain, but formally speaking it's not safe assumption to make. And cc-rs doesn't make it.

ChrisDenton:
Thanks for the PR but I don't think this is the correct solution, as outlined by dot-asm, so I'm closing this.

Issue #726, opened 2022-10-10
Unable to build aarch64-apple-ios-macabi and x86_64-apple-ios-macabi targets using clang 13.1 and above


kgrech:
I am trying to compile the ring crate for the targets mentioned in the target:

```
cargo +nightly build -Z build-std --release --target aarch64-apple-ios-macabi -vv
cargo +nightly build -Z build-std --release --target x86_64-apple-ios-macabi -vv
```

Both builds fail with the same error:
```
clang: error: invalid version number in '--target=arm64-apple-ios13.0-macabi'
```

I am using M1 mac and my clang version is the following:
```
clang -v
Apple clang version 14.0.0 (clang-1400.0.29.102)
Target: arm64-apple-darwin21.6.0
```

I was trying to solve this problem by downgrading the clang. It starts working if I downgrade to 13.0. Version 13.1 and above are failing.

I see the same problem on my x64 mac as well.

thomcc:
I think the support for catalyst upstream in rustc_target is [kind of broken here](https://github.com/rust-lang/rust/blob/bed4ad65bf7a1cef39e3d66b3670189581b3b073/compiler/rustc_target/src/spec/aarch64_apple_ios_macabi.rs), we should figure out a fix there before doing *too* much here...

... That said, I've merged #727 in the meantime, since it seems like a reasonable approach that we wouldn't be upset about no matter how we resolve that.

imWildCat:
@thomcc @kgrech I created another PR to fix in the rust compiler: https://github.com/rust-lang/rust/pull/106925

madsmtm:
All of the mentioned PRs have been merged, so I believe this can be closed.

NobodyXu:
I tried `cargo +nightly build -Z build-std --release --target aarch64-apple-ios-macabi` and it still failed:

```
   Compiling ring v0.17.8 (/private/tmp/ring)
   Compiling addr2line v0.21.0
   Compiling std v0.0.0 (/Users/nobodyxu/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/std)
The following warnings were emitted during compilation:

warning: ring@0.17.8: clang: error: argument unused during compilation: '-L/usr/lib' [-Werror,-Wunused-command-line-argument]

error: failed to run custom build command for `ring v0.17.8 (/private/tmp/ring)`

Caused by:
[...]

PR #727, opened 2022-10-10
Support clang >13.1 for aarch64-apple-ios-macabi and x86_64-apple-ios-macabi targets

kgrech:
Please see the details in the issue #726 

This solves the problem for Clang 13.1 and above (including 14.0)

My main concern is that it is going to break the above targets for Clang 13.0 and below. 
Is it acceptable? Maybe we could suggest using older version of the CC crate if you use out of date compiler?

Alternatively, we could check the version of clang used, but I did not find a way to perform this check. Does anyone has any suggestions?

Overall hardcoding 13.0 is very strange idea

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1581,7 +1581,7 @@ impl Build {
                             map_darwin_target_from_rust_to_compiler_architecture(target)
                         {
                             cmd.args
-                                .push(format!("--target={}-apple-ios13.0-macabi", arch).into());
+                                .push(format!("--target={}-apple-ios-macabi", arch).into());
                         }
                     } else if target.contains("ios-sim") {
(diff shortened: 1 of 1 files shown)


thomcc:
Yeah, hardcoding 13.0 seems wrong. It looks like this isn't one of the targets where we build the string dynamically either, we just pass the platform version to the linker: https://github.com/rust-lang/rust/blob/bed4ad65bf7a1cef39e3d66b3670189581b3b073/compiler/rustc_target/src/spec/apple_base.rs#L22-L30

Perhaps we should be doing that when building a dylib (and I'd probably take a patch for that if it's needed), but for now this seems fine.

thomcc (approved):

PR #728, opened 2022-10-10
Update Android compiler detection for NDK r25

chriswailes:
This commit updates the compiler detection logic for the Android platform to work with NDK r25.  This code is largely identical to the corresponding code in rustc/src/bootstrap/cc_detect.rs.


chriswailes:
It occurs to me that if this is committed then developers using older versions of `rustc` (and targeting the corresponding Android NDK) but a new version of `cc` will encounter errors during compiler detection as the new paths will be returned.  Ideally we would check on the compiler version to determine which version of the names to return.  We could either do this using the `rustversion` crate (proc-macro) or using the `rustc_version` check (either in `build.rs` or in `lib.rs` directly as it is executed at build time as well).

Is there a preference between these options?

thomcc:
> Is there a preference between these options?

Getting the version by parsing the result of `rustc --version` is very simple (like 10-20loc tops) so I'd definitely prefer manually doing that rather than pulling in dependencies for this (one of those pulls in dependencies of its own for e.g. semver parsing, and the other is a proc macro).

You should probably get the path to rustc via the `RUSTC` env var, falling back to `rustc` if it's unset. I guess if that fails I might default to assuming a new version of rustc, to avoid a situation where the handling of old `rustc` breaks things for users of new `rustc`.

thomcc:
Note to self: this shouldn't get merged until after the discussion in https://github.com/rust-lang/rust/issues/103673 is fully resolved.

chriswailes:
Thanks for the feedback.  I'm in the middle of putting out some fires but I'll try and get to this ASAP.

NobodyXu:
This PR would need a rebase

torokati44:
Do you think this could fix https://github.com/ruffle-rs/ruffle-android/issues/62 ? 🤔

PR #732, opened 2022-10-17
Support native ARM64 MSVC toolchain

Alovchin91:
In VS 17.3 a native ARM64 MSVC toolchain has appeared. Generally it should be chosen over the emulated one.

Open questions:

- [ ] Did it actually appear in VS 17.2?
- [ ] Does it contain all the tools?
- [ ] If not, should we try several toolchains? Or is it okay if it's not complete in older, unsupported versions of VS 2022?

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -469,7 +469,11 @@ mod impl_ {
         // We use the first available host architecture that can build for the target
         let (host_path, host) = hosts.iter().find_map(|&x| {
             let candidate = path.join("bin").join(&format!("Host{}", x));
-            candidate.join(&target).exists().then_some((candidate, x))
+            if candidate.join(&target).exists() {
+                Some((candidate, x))
+            } else {
+                None
+            }
         })?;
         // This is the path to the toolchain for a particular target, running
         // on a given host
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -466,10 +466,10 @@ mod impl_ {
         let target = lib_subdir(target)?;
         // The directory layout here is MSVC/bin/Host$host/$target/
         let path = instance_path.join(r"VC\Tools\MSVC").join(version);
-        // We use the first available host architecture
+        // We use the first available host architecture that can build for the target
         let (host_path, host) = hosts.iter().find_map(|&x| {
             let candidate = path.join("bin").join(&format!("Host{}", x));
-            candidate.exists().then_some((candidate, x))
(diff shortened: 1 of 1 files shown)


ChrisDenton:
This code will need formatting and the MSRV issues fixed.

More importantly this will need some testing. Ideally we should also have Windows ARM running in CI.

briansmith:
@Alovchin91 Are you interested in rebasing this and pushing it so we can see the CI run results again?

Alovchin91:
@briansmith Done

src/windows_registry.rs
```diff
@@ -454,30 +454,34 @@ mod impl_ {
     ) -> Option<(PathBuf, PathBuf, PathBuf, PathBuf, PathBuf)> {
         let version = vs15plus_vc_read_version(instance_path)?;
 
-        let host = match host_arch() {
-            X86 => "X86",
-            X86_64 => "X64",
-            // There is no natively hosted compiler on ARM64.
-            // Instead, use the x86 toolchain under emulation (there is no x64 emulation).
-            AARCH64 => "X86",
+        let hosts = match host_arch() {
```
ChrisDenton, line 457: If you annotate `hosts` with a type (i.e. `&[&str]`) you should be able to avoid creating `Vec`s.

ChrisDenton (commented):
This looks reasonable to me, just one nit that I've commented on.

Also I accept that falling back to x86 is the reasonable solution given that's the current behaviour. However, I'd definitely be interested in a followup PR that just errors for newer versions of VS.

dpaoliello:
@Alovchin91 are you still working on this, or would you like me to take over?

To answer your questions:
* 17.4 is the first VS with support for ARM64: <https://devblogs.microsoft.com/visualstudio/arm64-visual-studio-is-officially-here/>
* The "Desktop development with C++" and ".NET Desktop development" workloads are both supported, which includes cl.exe, link.exe, msbuild.exe and the Windows SDK.
* For now, I don't believe that we should drop support for using x86 on ARM64, given that VS has no way to migrate across architectures other than re-installing.

One interesting thing we may want to do: on Windows 11 we should prefer the x64 host architecture to x86 since it should be faster.

Alovchin91:
Hi @dpaoliello, I don't currently actively work on it, so if you'd prefer to take it over, please go ahead ☺️ Otherwise I will pick it up at some point but without any concrete promises about the timeframe.

NobodyXu:
Close as #957 supersedes this and is already merged

PR #736, opened 2022-10-26
Continue to support binutils ar

jonhoo:
sfackler [observed] that the `binutils` package [provides] binaries named `$target-ar` (with no `-gcc`). This patch augments the change made in #735 to continue picking those up too.

[observed]: https://github.com/alexcrichton/openssl-src-rs/pull/163#issuecomment-1292439189
[provides]: https://packages.debian.org/bullseye/amd64/binutils-aarch64-linux-gnu/filelist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2605,12 +2605,17 @@ impl Build {
         } else if self.get_host()? != target {
             match self.prefix_for_target(&target) {
                 Some(p) => {
-                    let target_ar = format!("{}-gcc-ar", p);
-                    if Command::new(&target_ar).output().is_ok() {
-                        target_ar
-                    } else {
-                        default_ar
+                    // GCC uses $target-gcc-ar, whereas binutils uses $target-ar -- try both.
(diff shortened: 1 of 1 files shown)


jonhoo:
cc @sfackler

sfackler:
What is the rationale for this change? I think it's generally expected you have a cross binutils in addition to a cross gcc if you are cross compiling. Binutils is a dependency of gcc in Ubuntu for example.

jonhoo:
I came across this when I was in an environment where only `${target}-gcc-ar` was on `$PATH`, and not the tools provided by binutils. Whether that's a _common_ situation is hard for me to say, but it does seem like it should be a supported one given that it's a standard part of the GCC distributions.

thomcc (approved):

thomcc:
This should be pretty harmless, and avoid anything bad that would be caused by #735. I have seen the `blah-gcc-ar` before, so I'll merge it.

PR #738, opened 2022-10-29
Default `emit_rerun_if_env_changed` to true

thomcc:
As mentioned in https://github.com/rust-lang/cc-rs/pull/701#issuecomment-1295758140, I don't think there's a reason to default this to false. `rerun-if-env-changed` is purely additive (unlike `rerun-if-changed`, which has a fairly good default cleared if you emit it at all), so it should never hurt.

---

I've also added code to skip emitting it for vars that provided by cargo, otherwise every use of `cc` will end up with `cargo:rerun-if-env-changed` invocations for pointless things like `DEBUG`, `OPT_LEVEL`, and so on. I don't think these are likely to hurt anything, but they definitely don't help, and [cargo's docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html?highlight=rerun-if#cargorerun-if-env-changedname) say not to.

An alternative to this would be to add a different function for vars provided by cargo, but that sounds like a headache to maintain, as reviewers could easily not notice the wrong function getting called for a given variable.

---

I'd like to get this in before the release, but mostly for the change of the default (since changing this *after* the release would be breaking) rather than skipping the cargo vars (which is not actually necessary).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2854,11 +2854,25 @@ impl Build {
     }
 
     fn getenv(&self, v: &str) -> Option<String> {
+        // Returns true for environment variables cargo sets for build scripts:
+        // https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts
+        //
+        // This handles more of the vars than we actually use (it tries to check
+        // complete-ish set), just to avoid needing maintenance if/when new
+        // calls to `getenv`/`getenv_unwrap` are added.
+        fn provided_by_cargo(envvar: &str) -> bool {
+            match envvar {
+                v if v.starts_with("CARGO") || v.starts_with("RUSTC") => true,
+                "HOST" | "TARGET" | "RUSTDOC" | "OUT_DIR" | "OPT_LEVEL" | "DEBUG" | "PROFILE"
+                | "NUM_JOBS" | "RUSTFLAGS" => true,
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
Thanks for the explanation, makes sense to me 👍

briansmith:
I think this is a breaking change. In my CI, I was (perhaps wrongly) setting a bunch of environment variables when doing `cargo test` but then not setting them when I ran `cargo test --doc` immediately after. Previously this worked fine because it didn't re-run my build script (and so didn't re-run anything using cc-rs) to run the doctests but afterwords my CI broke. I have fixed it on my end but I want to notify you and others that this is breaking for some as I think I that might be a surprise.

dtolnay:
> `rerun-if-env-changed` is purely additive (unlike `rerun-if-changed`, which has a fairly good default cleared if you emit it at all), so it should never hurt.

I think this is a misunderstanding. Cargo's behavior is that if your build script emits _neither_ rerun-if-changed _nor_ rerun-if-env-changed, then it will default to rerunning when any file within the package is changed. This is the fairly good default that you mention. But whether you emit a rerun-if-changed or rerun-if-env-changed, either way the default behavior goes away.

NobodyXu:
Yeah we could improve our doc here to mention the default behavior of cargo

NobodyXu:
Opened #1508

PR #739, opened 2022-10-29
Be more defensive about how windows RegistryKeys are handled

thomcc:
Few things along a general theme of skepticism in this registry code.

- Fixes an issue where `HKEY_LOCAL_MACHINE` is not sign-extended into 64 bits on 64 bit targets.

	From discussion with @ChrisDenton it's unclear why this works, perhaps internal details, or odd handle behavior with `KEY_WOW64_32KEY`. Dunno, easy fix (hopefully).

- Replaces `Repr::Const(...)` with `Repr::LocalMachine`.

    According to <https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexw>, using this with some HKEY types (`HKEY_PERFORMANCE_DATA` and presumably its subkeys) requires using the API in a totally different way, and knowing exactly which type you're dealing with. If you try to use the API as you normally, it seems like you'd end up with UB.

	Since we don't care at all about `HKEY_PERFORMANCE_DATA`, this just makes it clearer that we should never have other `Repr::Const`s than `HKEY_LOCAL_MACHINE`.

- Behaves a lot more defensively in `RegistryKey::query_str`. Specifically, it adds some asserts that the values we get are as we'd expect, and defensively zeroes the memory we allocate for the wide string.
	
	This is being a bit paranoid, but most of these weird behaviors seem like they could happen if we happened to have a `HKEY_PERFORMANCE_DATA`-derived key, and there's no easy way for us to check for that.
	
	These checks try to minimize how disastrous it would be if we did end up with one (panic rather than either heap buffer overflow or returning a pile of uninit memory to the caller).

- Fixes a small oversight in a case where we'd panic the value we read happens to be the empty string. Presumably this couldn't actually happen, or we'd have gotten a bug report by now.

- Adds some comments explaining this stuff, to make it more clear which things are done or not done intentionally.
	
r? @ChrisDenton

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -30,7 +30,7 @@ type REGSAM = u32;
 
 const ERROR_SUCCESS: DWORD = 0;
 const ERROR_NO_MORE_ITEMS: DWORD = 259;
-// Sign-extend into 64 bits if needed (although it seems to work even if not).
+// Sign-extend into 64 bits if needed.
 const HKEY_LOCAL_MACHINE: HKEY = 0x80000002u32 as i32 as isize as HKEY;
 const REG_SZ: DWORD = 1;
 const KEY_READ: DWORD = 0x20019;
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -14,7 +14,8 @@ use std::ops::RangeFrom;
 use std::os::raw;
 use std::os::windows::prelude::*;
 
-pub struct RegistryKey(Repr);
+/// Must never be `HKEY_PERFORMANCE_DATA`.
+pub(crate) struct RegistryKey(Repr);
 
 type HKEY = *mut u8;
 type DWORD = u32;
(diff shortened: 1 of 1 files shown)


src/registry.rs
```diff
@@ -29,7 +30,8 @@ type REGSAM = u32;
 
 const ERROR_SUCCESS: DWORD = 0;
 const ERROR_NO_MORE_ITEMS: DWORD = 259;
-const HKEY_LOCAL_MACHINE: HKEY = 0x80000002 as HKEY;
+// Sign-extend into 64 bits if needed (although it seems to work even if not).
+const HKEY_LOCAL_MACHINE: HKEY = 0x80000002u32 as i32 as isize as HKEY;
```
ChrisDenton, line 34:
```suggestion
// Sign-extend into 64 bits if needed.
const HKEY_LOCAL_MACHINE: HKEY = 0x80000002u32 as i32 as isize as HKEY;
```
I don't think we need to mention this here. There's a lot of stuff in the Windows API that happens to work for various reasons. If people are wondering about why this changed then they can use blame to find this PR.

ChrisDenton (approved):
This is maybe a bit paranoid in places but this code isn't performance sensitive code so extra checks won't hurt. I only have one small nit but nothing I feel strongly about.

thomcc:
Yep makes sense to me.

PR #741, opened 2022-10-30
Prefer `-ar` to `-gcc-ar`

nagisa:
As observed in https://github.com/cross-rs/cross/issues/1100, in some situations `*-gcc-ar` might actually be just broken.

I believe this is most likely just a plain bug in gcc (failing to consider `--disable-lto` configuration option somewhere in their build setup,) but for the time being `*-ar` tends to avoid this problem altogether.

Code added in https://github.com/rust-lang/cc-rs/pull/736 appears to be preferring `*-gcc-ar`, but no strong rationale is given to prefer one over the other. `*-gcc-ar` being outright broken sometimes seems like a rationale strong enough to continue preferring binutils’ `ar`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2623,9 +2623,12 @@ impl Build {
             match self.prefix_for_target(&target) {
                 Some(p) => {
                     // GCC uses $target-gcc-ar, whereas binutils uses $target-ar -- try both.
-                    // Prefer -gcc-ar if it exists, since that matches what we'll use for $CC.
+                    // Prefer -ar if it exists, as builds of `-gcc-ar` have been observed to be
+                    // outright broken (such as when targetting freebsd with `--disable-lto`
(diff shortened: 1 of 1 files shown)


jonhoo (approved):
I think ideally we'd prefer whichever one is "newer", but yeah, given that you've observed `gcc-ar` being broken, seems fine to me to swap the preference 👍

jonhoo:
I've also updated https://github.com/alexcrichton/openssl-src-rs/pull/164 to have the updated preference.

thomcc (approved):
Yep, seems good to me.

Emilgardis:
What's the status on getting this released?

thomcc:
Very soon. Release notes might be delayed a little but should be published within the next day, probably (still not automated and for various reasons is a multi-person ordeal)

thomcc:
Published https://github.com/rust-lang/cc-rs/releases/tag/1.0.75

PR #742, opened 2022-10-31
Add debuginfo flags for msvc assemblers

nagisa:
When code is being built with debuginfo, we need to instruct the assemblers accordingly.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1961,8 +1961,16 @@ impl Build {
             cmd.arg("-I").arg(directory);
         }
         if target.contains("aarch64") || target.contains("arm") {
+            if self.get_debug() {
+                cmd.arg("-g");
+            }
+
             println!("cargo:warning=The MSVC ARM assemblers do not support -D flags");
         } else {
+            if self.get_debug() {
+                cmd.arg("-Zi");
+            }
+
             for &(ref key, ref value) in self.definitions.iter() {
                 if let Some(ref value) = *value {
                     cmd.arg(&format!("-D{}={}", key, value));


thomcc (commented):
This matches the documentation on the arm assemblers https://learn.microsoft.com/en-us/cpp/assembler/arm/arm-assembler-command-line-reference?view=msvc-170

The x86 arg seems a bit odd tho https://learn.microsoft.com/en-us/cpp/assembler/masm/ml-and-ml64-command-line-reference?view=msvc-170, but maybe it's just a lack of familiarity with the term "CodeView" for a debuginfo format.

I don't have a good way to test this at the moment, so I might punt to @ChrisDenton in case he's slightly more familiar than I.

nagisa:
FWIW my approach to testing this was basically just slapping a `cmd.arg("-Zi")` in my `build.rs` and running my stuffs in a debugger.

Two things to keep in mind when interpreting this change for x86: this flag is somewhat shared between the `ml` assembler and the c/c++ compiler (`cl`), though it seems to be closer to `/Z7`: https://learn.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format?view=msvc-170

And `/` is interchangeable with `-`. Since `cc-rs` appears to be using `-`, I stuck to that convention as well.

ChrisDenton:
This does look correct to me but I'll test when I have a moment.

ChrisDenton (approved):
Ok, I tested in WinDbg. With this patch I get an interactive code view when debugging, which is a very nice enhancement!

Btw, it'd be good if CI were able to test debugging info in a similar way to the main rust repo. But setting that up is way beyond the scope of this PR.

PR #747, opened 2022-11-08
Don't separate files/opt when using msvc assembler

roblabla:
The MSVC assemblers don't support using `--` to mark the end of the options and the start of the "verbatim" file list. When the compiler family is MSVC with clang-cl, the assembler used will be the standard MSVC assembler.

See https://github.com/rust-lang/rust/pull/104152#issuecomment-1307816424

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1345,7 +1345,7 @@ impl Build {
         if self.cuda && self.files.len() > 1 {
             cmd.arg("--device-c");
         }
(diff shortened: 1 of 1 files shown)


thomcc (approved):
Looks good to me. Thanks.

thomcc:
If this fixes rust-lang/rust I'll try to get a hotfix release out later today or so.

Issue #751, opened 2022-11-16
Feature request: support setting asm-specific flags


nicholasbishop:
Background: clang fails with an error if you pass asm options when compiling a non-asm input. For example:
```console
$ clang -c foo.S -Wa,-defsym,abc=0
(no error!)

$ clang -c foo.c -Wa,-defsym,abc=0
error: unknown argument: '-defsym'
```

It would be helpful to have a `Build::asm_flag` method for adding flags specific to compiling assembly. The flags passed in this way would only be added to the command for `.S`/`.asm` files.

thomcc:
Patch welcome.

dot-asm:
Formally speaking, one could have used `flag_if_supported("-no-integrated-as")` in this case. And one can wonder if it would be generally more appropriate. Indeed, since the suggestion is to pass the option to multiple files, it implies that the symbol definition is intended to be global. In which case C is formally entitled to refer to it too :-)

PR #752, opened 2022-11-17
Add `Build::asm_flag`

nicholasbishop:
The method allows flags to be specified that are only applied to assembly input files, not C files. This is helpful with clang, which only accepts certain assembly options if the input file is asm, not C.

Closes https://github.com/rust-lang/cc-rs/issues/751

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,6 +97,7 @@ pub struct Build {
     flags_supported: Vec<String>,
     known_flag_support_status: Arc<Mutex<HashMap<String, bool>>>,
     ar_flags: Vec<String>,
+    asm_flags: Vec<String>,
     no_default_flags: bool,
     files: Vec<PathBuf>,
     cpp: bool,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -3471,3 +3495,13 @@ fn which(tool: &Path) -> Option<PathBuf> {
         return if check_exe(&mut exe) { Some(exe) } else { None };
     })
 }
+
+/// Check if the file's extension is either "asm" or "s", case insensitive.
+fn is_asm(file: &Path) -> bool {
+    if let Some(ext) = file.extension() {
+        let ext = ext.to_ascii_lowercase();
+        ext == "asm" || ext == "s"
```
thomcc, line 3503: This probably will need some adjustment over time, but I don't see anything it's concretely missing now.

thomcc (approved):
Looks fine.

thomcc:
Ah, there's no to_ascii_lowercase on OsStr back then, that will need to be fixed.

nicholasbishop:
I changed the test to disable it if the `parallel` feature is enabled since the execution order seems to not be deterministic -- is solution OK?

thomcc:
Yeah that's fine.

nicholasbishop:
Great, thanks for the quick review :)

thomcc:
This is in [1.0.77](https://github.com/rust-lang/cc-rs/releases/tag/1.0.77).

Issue #754, opened 2022-11-21
Cross-compilation of `zstd-sys` for `x86_64-pc-windows-msvc` target fails starting at 1.0.77


Hainish:
I am cross-compiling `zstd-sys` to `x86_64-pc-windows-msvc` in https://github.com/EFForg/apkeep/blob/5bffbf77cba71751884a46234dab66d83d5b0dbe/build-remote.sh#L75-L101.  When `cc` is updated from `1.0.76` to `1.0.77`, the build starts failing with the following:

```
  running: "clang-cl" "-nologo" "-MD" "-O2" "-Brepro" "-m64" "-Wno-unused-command-line-argument" "-fuse-ld=lld-link" "/imsvc/home/admin/xwin/crt/include" "/imsvc/home/admin/xwin/sdk/include/ucrt" "/imsvc/home/ad
min/xwin/sdk/include/um" "/imsvc/home/admin/xwin/sdk/include/shared" "-I" "zstd/lib/" "-I" "zstd/lib/common" "-I" "zstd/lib/legacy" "-fvisibility=hidden" "-DZSTD_LIB_DEPRECATED=0" "-DXXH_PRIVATE_API=" "-DZSTDLIB
_VISIBILITY=" "-DZDICTLIB_VISIBILITY=" "-DZSTDERRORLIB_VISIBILITY=" "-DZSTD_LEGACY_SUPPORT=1" "-Fo/home/admin/apkeep/target/x86_64-pc-windows-msvc/release/build/zstd-sys-6caf8d3f6c2dc823/out/zstd/lib/legacy/zstd
_v06.o" "-c" "--" "zstd/lib/legacy/zstd_v06.c"
  cargo:warning=clang: warning: unknown argument ignored in clang-cl: '-fvisibility=hidden' [-Wunknown-argument]
  exit status: 0
  running: "clang-cl" "-nologo" "-MD" "-O2" "-Brepro" "-m64" "-Wno-unused-command-line-argument" "-fuse-ld=lld-link" "/imsvc/home/admin/xwin/crt/include" "/imsvc/home/admin/xwin/sdk/include/ucrt" "/imsvc/home/ad
[...]

PR #755, opened 2022-11-23
Only pass `.asm` files to masm

thomcc:
This changes the handling of assembly on MSVC targets so that only `.asm` files are passed to masm. This is more in line with how things were handled in cc <1.0.77, and would have avoided https://github.com/rust-lang/cc-rs/issues/754.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1339,12 +1339,14 @@ impl Build {
     }
 
     fn compile_object(&self, obj: &Object) -> Result<(), Error> {
-        let is_asm = is_asm(&obj.src);
+        let asm_ext = AsmFileExt::from_path(&obj.src);
+        let is_asm = asm_ext.is_some();
         let target = self.get_target()?;
         let msvc = target.contains("msvc");
         let compiler = self.try_get_compiler()?;
         let clang = compiler.family == ToolFamily::Clang;
-        let (mut cmd, name) = if msvc && is_asm {
+
+        let (mut cmd, name) = if msvc && asm_ext == Some(AsmFileExt::DotAsm) {
             self.msvc_macro_assembler()?
         } else {
             let mut cmd = compiler.to_command();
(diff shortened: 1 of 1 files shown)


thomcc:
I'll probably fix test failures (if any) tomorrow.

src/lib.rs
```diff
@@ -3496,14 +3512,8 @@ fn which(tool: &Path) -> Option<PathBuf> {
...
-            return ext == "asm" || ext == "s";
-        }
-    }
-
-    false
+#[derive(Clone, Copy, PartialEq)]
+enum AsmExt {
+    S,
+    Asm,
 }
```
ChrisDenton, line 3519: Hm, if I didn't already know what the `asm_ext` function did then reading `asm_ext == Some(AsmExt::Asm)` above may feel like we're using a very redundantly named enum. Maybe the variants should be a bit more descriptive of their intended use? E.g. something like `AsmExt::Msvc` or `AsmExt::Masm` would seem clearer?

I don't feel particularly strongly about this and I do get the advantage of the variant names matching the tested extension.

ChrisDenton (commented):
Effectively restoring the previous behaviour to prevent unintentional breakages looks good to me. I only have a few minor nits.

src/lib.rs
```diff
@@ -455,6 +455,20 @@ impl Build {
         self
     }
 
+    fn asm_ext(&self, file: &Path) -> Option<AsmExt> {
```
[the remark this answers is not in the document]
thomcc: Yeah, I had moved it here since my first cut used target info. I don't mind moving it to a constructor on AsmExt.

src/lib.rs
thomcc: I've definitely seen `.S` used on MSVC-heavy projects, so I think that might be more confusing.

ChrisDenton: Fair.

thomcc: I've renamed it to `AsmFileExt::DotAsm` which should be more clear.

thomcc:
The fact that `cargo +1.34.0 test` doesn't work on my machine (since 1.34 predates aarch64-apple-darwin) is very annoying.

src/lib.rs
```diff
@@ -3496,14 +3498,27 @@ fn which(tool: &Path) -> Option<PathBuf> {
...
-    if let Some(ext) = file.extension() {
-        if let Some(ext) = ext.to_str() {
-            let ext = ext.to_lowercase();
-            return ext == "asm" || ext == "s";
+#[derive(Clone, Copy, PartialEq)]
+enum AsmFileExt {
+    /// `.asm` files. On windows targets, we assume these should be passed to
+    /// MASM (`ml{,64}.exe`).
+    DotAsm,
+    /// `.s` or `.S` files, which do not have the special handling on windows.
```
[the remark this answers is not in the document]
thomcc: Ah, yeah, good point.

ChrisDenton (approved):
LGTM, thanks!

MarijnS95:
> and would have avoided #754.

Fwiw this is more-so caused by the `zstd-sys` build script looking at _host_ OS rather than _target_ OS, while cross-compiling. However, it seems that its choice to include this file comes down to compiler support, so if the cross-compiler was having `clang-cl` available that might explain why it was previously building fine.

thomcc:
> Fwiw this is more-so caused by the zstd-sys build script looking at host OS rather than target OS, while cross-compiling.

Yeah, the point is more: you should actually be allowed to compile that file with clang-cl, even if the logic you used to decide whether or not to include it was buggy.

MarijnS95:
@thomcc is this still on track to be released? We're currently failing to compile `tract-linalg` because it detects `ml64.exe` isn't available and passes `.S` files to `cc`, which ends up invoking `ml64.exe` anyway.

https://github.com/sonos/tract/blob/main/linalg/build.rs

thomcc:
Ah, thanks for the reminder, I'll do a release as soon as possible.

thomcc:
@MarijnS95 Sorry for the delay, it's in https://github.com/rust-lang/cc-rs/releases/tag/1.0.78.

thomcc:
A wrinkle here is that `meson` projects are expected to use `.asm` file as the file extension, regardless of assembler. I wonder if this will cause issues for us...

ChrisDenton:
Oh hm. Maybe! We probably should have ways to override any and all heuristics in any case.

PR #758, opened 2022-11-30
Default to llvm-lib when using clang-cl in msvc environment.

dot-asm:
The problem is that the vendor librarian can't handle object modules compiled with clang-cl -flto. llvm-lib on the other hand can handle any object modules, which makes it a preferred option.

At [@ChrisDenton's request](https://github.com/rust-lang/cc-rs/pull/657#issuecomment-1332348566).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2665,10 +2665,29 @@ impl Build {
 
             "emar".to_string()
         } else if target.contains("msvc") {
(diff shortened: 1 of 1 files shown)


ChrisDenton:
Thanks! I'd like to take a little bit of time to run some tests locally but otherwise this looks good.

ChrisDenton (approved):

PR #759, opened 2022-12-01
Fix "-arch arm64" flag for aarch64-ios-sim

gh2o:
Passing "-arch arm64" as a single string to clang causes an "argument unused during compilation" warning that becomes an error if -Werror is enabled. Fix passes it as two separate arguments (unfortunately "-arch=arm64" does not work either).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2212,7 +2212,7 @@ impl Build {
             }
         } else if is_sim {
             match arch {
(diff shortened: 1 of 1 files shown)


gh2o:
Bump :)

thomcc (approved):
Oh yeah, that seems completely wrong

Issue #565, opened 2020-11-18
Explicit method to specify C/C++ standard


alexeyr:
It would be nice to specify the standard explicitly as a method on `Build`, e.g. `Build::new().standard("c++14")`. This abstracts from difference between GCC/Clang `-std=` and MSVC `/std:`. Questions to answer:

1. What to do if you pass `gnu14` on MSVC: translate to the closest `c++14` or give an error?
2. Similarly, MSVC doesn't yet support `/std:c++20` and requires `/std:c++latest` for C++20 features. Should `standard("c++20")` test if `/std:c++20` works and use `/std:c++latest` if it isn't?

hikari-no-yume:
The simplest thing you can do right now is something like this:

```rust
if build.get_compiler().is_like_msvc() {
    build.flag("/std:c++17")
} else {
    build.flag("-std=c++17")
}
```

But it doesn't solve the issues you mention with unusual versions…

Do you think it's be useful to have this even if it only supports the simple cases? If so, I might contribute a fix for this myself, because I need this flag for one of my projects.

hikari-no-yume:
I decided to just do it :) https://github.com/rust-lang/cc-rs/pull/761

thomcc:
Fixed by #761, will be in the next release.

PR #761, opened 2022-12-05
Add method for specifying C/C++ standard version

hikari-no-yume:
As requested by https://github.com/rust-lang/cc-rs/issues/565.

Please have a look at the documentation I have included. :)

I am not doing any abstraction over special cases like the name for the work-in-progress standard version. It's just the basics.

I want to propose adding this to `cc` because:

* It should be useful fairly often, considering that using the right standard version can be critical to compiling modern code. I have a C++17 dependency in one of my projects, and I couldn't build my code interfacing with it without specifying the standard version, because even very new compilers seem to default to older versions. As far as I know, the compatibility between C++ versions isn't perfect, so that might be why.
* It can save people the effort of having to consult both GCC/Clang and MSVC documentation, and improve the portability of code from people who haven't tested on Windows:
  * The GCC/Clang and MSVC flag syntaxes are not quite the same, so without support in the `cc` crate, the developer has to write conditional, platform-specific code.
  * The common subset of supported versions between GCC/Clang and MSVC is not something you can intuitively guess. The documentation added here can help.
* While I haven't done that here, it could be enhanced in future to handle special cases like the different names for the WIP future C and C++ standards.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,6 +108,7 @@ pub struct Build {
     cpp_set_stdlib: Option<Arc<str>>,
     cuda: bool,
     cudart: Option<Arc<str>>,
+    std: Option<Arc<str>>,
     target: Option<Arc<str>>,
     host: Option<Arc<str>>,
     out_dir: Option<Arc<Path>>,
@@ -314,6 +315,7 @@ impl Build {
             cpp_set_stdlib: None,
             cuda: false,
             cudart: None,
+            std: None,
             target: None,
             host: None,
             out_dir: None,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -665,6 +667,36 @@ impl Build {
...
+    /// Specify the C or C++ language standard version.
+    ///
+    /// Common values accepted by modern versions of GCC, Clang and MSVC:
+    /// - `c11` for ISO/IEC 9899:2011
+    /// - `c17` for ISO/IEC 9899:2018
+    /// - `c++14` for ISO/IEC 14882:2014
+    /// - `c++17` for ISO/IEC 14882:2017
+    /// - `c++20` for ISO/IEC 14882:2020
+    ///
+    /// Other values like `c99`, `gnu89` and `c++11` have less broad support.
```
riidefi, line 679: It should be noted that this isn't entirely true: `c++11` has the *most* broad support of any standard listed here across the greatest variety of compilers and platforms.

hikari-no-yume: Maybe I should have been more direct here: [`c99`, `gnu89` and `c++11`](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170) are not supported by the `/std:` flag to MSVC.

hikari-no-yume: Even Visual Studio 2015 doesn't support `c++11` as a language standard flag, if the documentation is to be believed. The minimum is `c++14`.

For the same reasons, I mention `c99` as less widely supported, because MSVC has never supported C99, only C11. That's because C11 made some features optional that were mandatory in C99.

Microsoft also don't support the GNU extensions, but that's less surprising.

thomcc:
There are conflicts I guess. Can you rebase?

hikari-no-yume:
Sure, I can do that later today.

Maybe I can take that opportunity to edit the doc comment a bit to address the review comment above.

thomcc:
Sure. I think it's fine as-is but additional clarification wouldn't hurt.

hikari-no-yume:
I rebased it now. The conflict was caused by the code being changed to use `Arc<str>` instead of `String` for all the other fields. I'm not sure whether that has any benefit for this field, but I changed it to match. I also clarified the thing with MSVC and C++11.

thomcc:
Thanks.

Issue #82, opened 2016-06-22
Support a convention for specifying more paths to toolchain binaries


luser:
Currently cargo supports setting `linker` and `ar` in a config file, which is useful but is not always sufficient for cross-compiling crates that build native libraries.

While cross-compiling Servo I found that I had to specify quite a few extra tools to get everything to compile:
https://gist.github.com/luser/a33e5070d1c55a7d2c46fe763a9d1543#file-servo-cross-mac-L59

This was mostly because various crates used different build systems for building native libraries--autotools, cmake, etc. I think it'd be useful if the gcc crate supported a convention for specifying more tools in `.cargo/config` so that common crates for building native libraries (like `cmake`) could use that support to drive their builds.

Specifically, I think it would be useful if it allowed at least setting cc, cxx, ranlib, and exposed those values through its API.

luser:
Interestingly, AFAICT cargo does not expose the values of `linker` or `ar` from `.cargo/config` to build scripts even if they're set.

alexcrichton:
Yeah currently Cargo doesn't expose `linker` and/or `ar` to build scripts, but that probably wouldn't push this over the finish line either as you probably also want to at least read `cxx` (if not `cc` before `linker`). At the very least it'd be good to canonicalize this information all into one location, however, even if Cargo isn't passing the information.

It's probably worth at least doing some parsing of `.cargo/config` to learn information like this. I'd need to double check to ensure a build script can figure out what directory it was invoked from, however. I forget if that's passed through or not...

GlenDC:
> I'd need to double check to ensure a build script can figure out what directory it was invoked from

That depends on the user. I think you can specify the linker just by giving the name, in case it is in your environment PATH. Personally I don't have it in the path and just gave the full (absolute) location to the file, which would be enough to get the directory, as all the different tools can find in the same directory as the linker.

But even if the user doesn't give the full path, we should be able to get the directory of the executable on all development OS's, no? I know on Unix based systems it's possible with tools like `whereis`/`which`, and I suppose Windows has its own system call to get it as well.

jdub:
I just noticed that `gcc` doesn't take advantage of the standard-ish `TOOLCHAIN_PREFIX` or `CROSS_COMPILE` environment variables in `HOST` != `TARGET` scenarios. Would you accept a change to use these when available?

alexcrichton:
@jdub certainly! I hadn't heard of those before but if they're commonly in use I'd love to have support for them.

[...]

PR #763, opened 2022-12-12
Expose get_archiver and get_ranlib

jonhoo:
This PR exposes cc's inferred archiver through a similar mechanism as `get_compiler`: with `get_archiver`/`try_get_archiver`.

As part of this, I also realized that cc wouldn't pick up `ARFLAGS`, so I added support for that. Through that, I also realized that there was currently one place where we didn't respect `Build::ar_flags` when ar was called, so I fixed that too.

Since ranlib is exposed more or less exactly like ar, I also added a `get_ranlib`/`try_get_ranlib` combination. This is, in part, to enable `openssl-src` to move to re-use `cc`'s AR (and now RANLIB) detection rather than rolling its own. See alexcrichton/openssl-src-rs#164.

Note that I didn't re-use `Tool` for the return value for these getter functions, and instead just exposed `Command` directly. `Tool` has a number of relatively compiler-specific things, so it felt wrong to use. One important caveat to this is that `Command::get_program` was only added in Rust 1.57.0, so users on older versions of Rust won't have a way to recover the inferred `ar` program path, but that feels like it's probably okay given that we didn't even expose `get_archive` until now.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -115,6 +115,7 @@ pub struct Build {
     env: Vec<(OsString, OsString)>,
     compiler: Option<PathBuf>,
     archiver: Option<PathBuf>,
+    ranlib: Option<PathBuf>,
     cargo_metadata: bool,
     link_lib_modifiers: Vec<String>,
     pic: Option<bool>,
@@ -320,6 +321,7 @@ impl Build {
             env: Vec::new(),
             compiler: None,
             archiver: None,
+            ranlib: None,
             cargo_metadata: true,
             link_lib_modifiers: Vec::new(),
             pic: None,
@@ -916,6 +918,17 @@ impl Build {
         self.archiver = Some(archiver.as_ref().to_owned());
         self
     }
+
+    /// Configures the tool used to index archives.
+    ///
+    /// This option is automatically determined from the target platform or a
+    /// number of environment variables, so it's not required to call this
+    /// function.
+    pub fn ranlib<P: AsRef<Path>>(&mut self, ranlib: P) -> &mut Build {
+        self.ranlib = Some(ranlib.as_ref().to_owned());
+        self
+    }
+
     /// Define whether metadata should be emitted for cargo allowing it to
     /// automatically link the binary. Defaults to `true`.
     ///
@@ -2109,9 +2122,6 @@ impl Build {
             let mut out = OsString::from("-out:");
             out.push(dst);
             cmd.arg(out).arg("-nologo");
-            for flag in self.ar_flags.iter() {
-                cmd.arg(flag);
-            }
             // If the library file already exists, add the library name
             // as an argument to let lib.exe know we are appending the objs.
             if dst.exists() {
@@ -2145,9 +2155,6 @@ impl Build {
             // In any case if this doesn't end up getting read, it shouldn't
             // cause that many issues!
             ar.env("ZERO_AR_DATE", "1");
-            for flag in self.ar_flags.iter() {
-                ar.arg(flag);
-            }
             run(ar.arg("cq").arg(dst).args(objs), &cmd)?;
         }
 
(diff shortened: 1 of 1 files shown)


jonhoo:
With this, I think we're also further along on https://github.com/rust-lang/cc-rs/issues/82 ?

jonhoo:
Ah, because I now use `Command::get_program` to grab the program name to use in error messages, this also bumps MSRV. How unacceptable do we think that is? I could probably find a way to have it carry through the name all the way to bypass the need to use `Command::get_program`, though it'd be unfortunate added complexity.

jonhoo:
Nudge on this @thomcc

thomcc:
Sorry, thanks for the ping. Will get to it tomorrow.

thomcc:
> As part of this, I also realized that cc wouldn't pick up ARFLAGS, so I added support for that

A problem here is we already pass our own flags to it, so I think if the user provides ARFLAGS, it will collide with these. Usually I see these used in cases like `ARFLAGS=Drcs`, in which case <https://github.com/rust-lang/cc-rs/blob/main/src/lib.rs#L2098> would collide, which is an error (at least on some `ar`s).

(That said I am in favor of supporting `ARFLAGS`)

jonhoo:
Hmm, I'm not sure what to do about that. Are you thinking we'd specifically check if `s` is already among the flags and if so not add it in that location?

thomcc:
Presumably if they set ARFLAGS we don't want to pass anything?

jonhoo:
Ohh, I see what you mean. Sure, I can make us not set flags if `ARFLAGS` is set!

jonhoo:
Should that also apply if they've called `Build::ar_flag`?

jonhoo:
Thinking some more about it, I'm also not sure you'd actually want to skip `s` for example even if `ARFLAGS` is set. `ARFLAGS` (like `CFLAGS`) I think is "flags that should always be set", whereas the `s` in the code you reference is more like a "mode" argument that you only want to pass to accomplish a particular task (as opposed to always). At least that's how it's used there — we _specifically_ want to run `ar` _with `s`_ to accomplish something.

I'm also a bit worried about the debuggability of "when I set `ARFLAGS`, `cc` stops working on some platforms" just because it causes `s` to _not_ be passed when it was otherwise needed. But you have a better sense than I do of how likely folks are to run into this (by setting `ARFLAGS` to something that doesn't include `s`).

jonhoo:
There are also cases like

https://github.com/rust-lang/cc-rs/blob/main/src/lib.rs#L2108-L2120

where it feels like the arguments aren't something that'd generally be included in `ARFLAGS`. They're things we want to pass _in addition_.

jonhoo:
nudge @thomcc

thomcc:
Ah, sorry, thanks for the ping. Looking online for other uses of ARFLAGS (in makefiles and such) it generally looks like they include stuff like `s`, and completely replace the exiting set of flags.

That said, I guess we already support `ar_flags` and they don't have that behavior...

jonhoo:
But presumably they don't replace _all_ the flags. For example, input and output files still need to be specified.

I guess the question here is what behavior is less surprising and less difficult to work around. I worry about the case where suddenly things _stop_ working because your `ARFLAGS` don't include `s` and you hit the case where we previously would inject `s` for you. That's harder to debug (I think) than having `ar` complain because you're passing `s` twice. We have to weigh that against the likelihood that someone wants to _always_ pass `s`, even when it's not required (as per the current logic), because they won't be able to have that behavior if we explicitly add `s` (and `ar` errors on `ss`).

Basically: do we want to introduce hard-to-debug (but maybe rare?) errors, or make it impossible to *always* pass `s` (via `ARFLAGS`)?

thomcc:
@danakj in https://github.com/chromium/chromium/commit/8f52776cd0cf38eeda26839d11e989f05383eb63 you seem to be griping that we don't support ARFLAGS. Can you weigh in on the semantics you'd like them to have? At the moment we're at a bit of an impasse.

danakj:
Hi, yes! (phone typing but I’ll do my best)

In the end we changed our archive tool to one which supported -nologo on Windows (i think it was lld-llvm). However with the situation we had:
1) it was windows but we are not using msvc, we’re using clang we built ourselves
2) the default ar choice did not know “windowsisms” like -nologo
3) so archive failed

I was not looking for a way to override the mode flag of the ar tool but to (remove and) replace other flags, specifically -nologo.

tldr I think I would treat the mode flags like input files and always set them, but if ARFLAGS is set i would use that in place of any extra flags (even platform specific ones like -nologo).

thomcc:
That seems reasonable to me. @jonhoo do you mind changing this PR to do that?

jonhoo:
Yep, I'll make that change tomorrow!

Just to confirm, that would mean we _should_ set `s` even if `ARFLAGS` is passed in?

danakj:
Yep

jonhoo:
Oh, and, should the same apply if any arguments have been added using `Build::ar_flag`?

thomcc:
I think so, for consistency. But I suppose there is a compatibility argument...

danakj:
Perhaps there is another way to remove flags with Build::ar_flag as the interface is more rich than an environment variable?

jonhoo:
Done in https://github.com/rust-lang/cc-rs/pull/763/commits/7a0c2afeffa63e1679f58714f8d0e2e824687450 👍

jonhoo:
The CI failures are because of the MSRV bump as a result of using `CommandBuilder::get_program`.

thomcc:
How difficult is that to avoid?

I don't know that we have a strong MSRV policy for this crate, but I'd weakly like to keep this to what works on debian stable, at least until the `libc` crate bumps it's MSRV (at which point almost all Rust crates will break on deb stable, so there's really no point anymore).

This is mostly to reduce false bug reports and nagging more than anything else, though, so if it's impossible to avoid it's not the end of the world.

jonhoo:
Worked around it (I think) in https://github.com/rust-lang/cc-rs/pull/763/commits/de69fc7547783af3b2b3c03688c54a7d3061e7b5.

It's not _entirely_ clear what the second argument to `run` is _supposed_ to be — sometimes it's referred to as the command's "args", and sometimes as just a program name. I went with just the program name though.

thomcc (approved):
Looks great. Thanks for putting up with all the back and forth.

thomcc:
I'll try to get a release out this weekend or so.

jonhoo:
> Thanks for putting up with all the back and forth.

Not at all — it's important to weigh the trade-offs here!

> I'll try to get a release out this weekend or so.

Thank you!

Next up, https://github.com/alexcrichton/openssl-src-rs/pull/173 😅

Issue #766, opened 2022-12-20
Cross compile build failure


Yuri6037:
When attempting to cross compile from aarch64-unknown-linux-gnu to x86_64-unknown-linux-musl this crate chooses a completely wrong compiler name: `musl-gcc`. The `must-gcc` compiler cannot be used to build x86_64 under an aarch64 machine.

Why is `cc-rs` not using RUSTC_LINKER in this case which points to a valid compiler/linker for the target platform?

dot-asm:
In a way the question answers itself. I mean the answer to the "why" is obviously somebody asked the related "why" question before and came up with an answer that made sense in a specific situation:-) But on a serious note, a similar question was [asked](https://github.com/rust-lang/cc-rs/pull/685#issue-1261002642) and even with a reference to musl. The suggestion was to scrap most of the mapping table, more specifically entries for targets that rely on cross-linker having to be specified (and as a result being passed as RUSTC_LINKER). As an alternative one can flip the priorities and use the table as a fallback. As opposed to using RUSTC_LINKER as fallback that is. (I also [advocate](https://github.com/rust-lang/cc-rs/issues/693#issuecomment-1180192993) for providing an interface to interrogate rustc for linker, so that we have something to work with in case RUSTC_LINKER is not set...)

PR #767, opened 2022-12-22
Prioritize RUSTC_LINKER over target->cross-compile-prefix table.

dot-asm:
Rationale is that if rustc is explicitly instructed to use a specific linker, using it as base for cross-compile prefix would be better aligned with user expectations.

This is in response to #766. Should fix #654.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2864,23 +2864,14 @@ impl Build {
     }
 
     fn prefix_for_target(&self, target: &str) -> Option<String> {
-        // Put aside RUSTC_LINKER's prefix to be used as last resort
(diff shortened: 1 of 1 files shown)


dot-asm:
Re-based and aligned with recent MSRV adjustment.

thomcc (approved):
Hmm, sorry, I think I missed this PR when it was submitted. It seems hard to argue with this, thanks.

thomcc:
I'm going to let it sit shortly since I'd like to test it a little, but will merge it later this week.

thomcc:
Eh, I'll just test it on master when I get to it.

dot-asm:
You should be able to close #766 and even #637 (since earlier actually).

PR #768, opened 2023-01-05
Remove bitcode from iOS / watchOS

keith:
Starting with Xcode 14 Apple deprecated bitcode and stopped accepting applications built with it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2279,18 +2279,6 @@ impl Build {
 
         cmd.args.push("-isysroot".into());
         cmd.args.push(sdk_path);
-        cmd.args.push("-fembed-bitcode".into());
-        /*
-         * TODO we probably ultimately want the -fembed-bitcode-marker flag
-         * but can't have it now because of an issue in LLVM:
-         * https://github.com/rust-lang/cc-rs/issues/301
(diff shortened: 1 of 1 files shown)


keith:
realistically it might be the case that this isn't the right time to do this, since technically folks could use a newer rust + older Xcode until Apple stops accepting apps from Xcode 13 (which I think normally happens around spring). I discovered this working on https://github.com/rust-lang/rust/pull/106476 but I found another workaround for that for now

JohnTitor:
> realistically it might be the case that this isn't the right time to do this, since technically folks could use a newer rust + older Xcode until Apple stops accepting apps from Xcode 13 (which I think normally happens around spring).

Yeah, I'm concerned about that too. So, at least that flag doesn't produce an error on a newer toolchain, right? Then, maybe we should just add FIXME for now, and remove it when rustc drops the support for older Xcode, or Xcode starts to emit an error for it.

keith:
Correct it still works fine because all tooling expects the build process to do a final bitcode strip step in the case you don't actually need bitcode. It's just the case I found in the compiler that tripped it up.

Updating the comment here https://github.com/rust-lang/cc-rs/pull/769 !

PR #770, opened 2023-01-13
More env overrides

lu-zero:


lu-zero:
The CI apparently had a network glitch.

thomcc:
Interesting idea. I'll have to think about it.

thomcc:
This is implemented in an opt-in way by https://github.com/rust-lang/cc-rs/pull/818. So I don't think we want this anymore.

lu-zero:
I looked at the change, you do not want to have this feature opt-in since it would sort of defeat the purpose of having a standard way to pass optional flags to any crate that's using cc-rs.

PR #773, opened 2023-01-24
fix msbuild from vs2022 not being found

seritools:
Without this (and only VS 2022 installed), `find_tool` with `msbuild` returns `C:\Windows\Microsoft.NET\Framework64\v4.0.30319\MSBuild.exe` instead of the MSBuild.exe from the actual VS installation.

In other words:

```rs
cc::windows_registry::find_tool("x86_64-pc-windows-msvc", "msbuild.exe").unwrap().path()
```

returns `C:\Windows\Microsoft.NET\Framework64\v4.0.30319\MSBuild.exe` instead of something like `C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\MSBuild\Current\Bin\amd64\MSBuild.exe`, as `find_msbuild` is missing the check for VS17.

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -866,7 +866,9 @@ mod impl_ {
     // see http://stackoverflow.com/questions/328017/path-to-msbuild
     pub fn find_msbuild(target: &str) -> Option<Tool> {
         // VS 15 (2017) changed how to locate msbuild
-        if let Some(r) = find_msbuild_vs16(target) {
+        if let Some(r) = find_msbuild_vs17(target) {
+            return Some(r);
+        } else if let Some(r) = find_msbuild_vs16(target) {
             return Some(r);
         } else if let Some(r) = find_msbuild_vs15(target) {
             return Some(r);


ChrisDenton (approved):
Looks great, thanks!

malkia:
The issue above mentions VS2022, yet the change has vs17 in it.

Also VS2022 is no longer installed in "C:\Program Files (x86)" (by default), but in "C:\Program Files"

ChrisDenton:
The VS2017 method uses COM to find the installation, which will work for later VS versions too. It will return whatever the path is, be it in `Program Files (x86)`, `Program Files` or elsewhere. You can test this yourself. E.g.:

```rust
fn main() {
    let tool = cc::windows_registry::find_tool("x86_64-pc-windows-msvc", "msbuild.exe").unwrap();
    println!("{}", tool.path().display());
}
```

If you find there's still a problem, then please create a new issue with details so it can be investigated.

seritools:
@malkia Visual Studio 2022 has the version number 17.

malkia:
> 

oh right, feel dumb now :) It's just that I was thinking about the Visual Studio 2017 and got it mixed up! Thanks!!!

PR #774, opened 2023-01-24
Add new fn `Build::flag_if_supported_with_fallbacks`

NobodyXu:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -562,6 +562,40 @@ impl Build {
         self
     }
 
+    /// Find the first flag in `flags` that is supported
+    /// and add that to the invocation of the compiler.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     // -Oz is only supported by clang, -Os is supported by both
+    ///     // gcc and clang, while -O2 is universally supported.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -562,6 +562,40 @@ impl Build {
...
+    ///
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     // -Oz is only supported by clang, -Os is supported by both
+    ///     // gcc and clang, while -O2 is universally supported.
+    ///     .flag_if_supported_with_fallbacks(["-Oz", "-Os", "-O2"])
+    ///     .compile("foo");
+    /// ```
+    pub fn flag_if_supported_with_fallbacks<'flag, It>(&mut self, flags: It) -> &mut Build
```
NobodyXu, line 578: This function is extracted from [`zstd-sys build.rs#L77`](https://github.com/gyscos/zstd-rs/blob/1bed85bddff3f0b08e94c68aa246b716a701870d/zstd-safe/zstd-sys/build.rs#L77).

I think it will also be useful to other crates to use this.

NobodyXu:
I just realized that the implementation is buggy.

It should check the flags in `Build::try_get_compiler` instead of in `Build::flag_if_supported_with_fallbacks`, as the users could change the configuration after that.

PR #775, opened 2023-01-24
Fix clippy warnings

NobodyXu:


src/bin/gcc-shim.rs
```diff
@@ -1,48 +1,119 @@
...
 
-        // If the file exists, commands have already run. Try again.
         if candidate.exists() {
-            continue;
+            // If the file exists, commands have already run. Try again.
+            None
+        } else {
+            Some(candidate)
         }
+    }).expect_fmt(format_args!("Cannot find the first nonexistent candidate file to which the program's args can be written under out_dir '{}'", out_dir.display()));
```
NobodyXu, line 85: In the original for-loop, after iterating over `0..` and cannot find a nonexistent candidate file, it would simply continue to create the `libfoo.a` even if no candidate file is written to.

Using iterator-style loop here, we avoid the implicit `break` by having `Iterator::find_map` returns an `Option<PathBuf>` which we then calls `expect_fmt` on.

src/bin/gcc-shim.rs
```diff
@@ -1,48 +1,119 @@
 #![cfg_attr(test, allow(dead_code))]
 
 use std::env;
+use std::fmt;
 use std::fs::File;
-use std::io::prelude::*;
+use std::io::{self, prelude::*};
 use std::path::PathBuf;
 
+trait EnumUnwrapExt {
```
thomcc, line 9: I feel like this doesn't carry it's weight, if I'm being honest. Almost half the file is dedicated to this helper (56 lines out of 119, 47%), so I'd rather we just use `unwrap_or_else(|| panic!(...))` and such.

Especially given that the gcc-shim is mostly a test-only thing?

thomcc (commented):
I'm unsure here. We don't use clippy at the moment, and none of these are issues that seem that worth catching, so I'm not sure how worthwhile it is that we start. If we wanted to start using clippy, I feel like we probably want to add it to CI too.

Regarding the gcc-shim bugfix, obviously that's fine. I think the other changes in gcc-shim.rs are a bit more involved than they need to be.

src/bin/gcc-shim.rs
NobodyXu: I've replaced this with `unwrap_or_else` but I also include the `error` in `Result::unwrap_or_else`.

NobodyXu:
> I'm unsure here. We don't use clippy at the moment, and none of these are issues that seem that worth catching, so I'm not sure how worthwhile it is that we start. If we wanted to start using clippy, I feel like we probably want to add it to CI too.

I know this PR is quite large and probably would cause merge conflicts in other PR, so I'm also a bit unsure.

However, I do think some of the clippy suggestion indeed improves the codebase.

src/lib.rs
```diff
@@ -2880,123 +2859,125 @@ impl Build {
...
-            "x86_64-rumprun-netbsd" => Some("x86_64-rumprun-netbsd"),
-            "x86_64-unknown-linux-gnu" => self.find_working_gnu_prefix(&[
-                "x86_64-linux-gnu", // rustfmt wrap
-            ]), // explicit None if not found, so caller knows to fall back
-            "x86_64-unknown-linux-musl" => Some("musl"),
-            "x86_64-unknown-netbsd" => Some("x86_64--netbsd"),
-            _ => linker_prefix,
-        }
-        .map(|x| x.to_owned()))
+        cross_compile.or_else(|| {
```
NobodyXu, line 2862: This is an example where clippy suggestion makes the code more efficient by evaluating this `match` block lazily, although it also creates a lot of diff...

NobodyXu, line 2862: @thomcc This is an example where clippy suggestion makes the code more efficient by evaluating this match block lazily, although it also creates a lot of diff...

thomcc:
Can you pull the gcc-shim bugfix into its own PR? I think we want that regardless of whether or not we're going to turn on clippy (which needs some more thought -- I'd like to do a more careful look at the changes to decide if there are any lints we'd rather disable, for example).

NobodyXu:
Sure, I will do that.

NobodyXu:
Closing this since cc doesn't use clippy and there isn't anything worth fixing.
This PR also introduces a lot of change that will create merge conflicts for other PRs, so I don't think it worth continuing.

PR #777, opened 2023-01-26
Fix `gcc-shim`, apply clippy warning & optimizations to it

NobodyXu:
Extracted from #775


thomcc (approved):
This looks fine. I kind of think eventually we should remove this binary target (moving it to something more explicitly test-only) but I think that should be done later.

PR #779, opened 2023-01-28
Fix `Build::compile_objects` on error and when used in multithreading 

NobodyXu:
Before this commit, each call to `Build::compile_objects` release a raw
token regardless of whether there is a token acquired by this process.

As such, if `Build::compile_objects` is used in multithreading context,
then it would call `Client::release_raw` for each thread and thus
increasing the parallelism far beyond the limit.

This commit fixed the bug by not acquiring the implicit global token
when initializing global `jobserver::Client` and only obtain token when
compiling objects.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1221,13 +1221,8 @@ impl Build {
     fn compile_objects(&self, objs: &[Object]) -> Result<(), Error> {
         use std::sync::Once;
 
-        // Limit our parallelism globally with a jobserver. Start off by
-        // releasing our own token for this process so we can have a bit of an
-        // easier to write loop below. If this fails, though, then we're likely
-        // on Windows with the main implicit token, so we just have a bit extra
-        // parallelism for a bit and don't reacquire later.
+        // Limit our parallelism globally with a jobserver.
         let server = jobserver();
-        let reacquire = server.release_raw().is_ok();
 
         // When compiling objects in parallel we do a few dirty tricks to speed
         // things up:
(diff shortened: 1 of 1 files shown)


NobodyXu:
@thomcc Can we have this PR merged before the release #778?

thomcc:
This is complex enough that I'd like to have a bit more time to review. I just got some vaccines this weekend and am slightly under the weather, and would rather not r+ some unsafe code I haven't had time to think about (sorry!). I'll review over the week and try to cut a release next weekend if it seems good.

NobodyXu:
Thanks @thomcc

NobodyXu:
@thomcc Friendly ping as this PR and #780 has been stale for really long time.

joshtriplett:
Initial reaction: I don't think we should be using a static mut for this. Is there some structure by which we could make this a field of the relevant object instead?

NobodyXu:
> Initial reaction: I don't think we should be using a static mut for this. Is there some structure by which we could make this a field of the relevant object instead?

`once_cell::sync::OnceCell` or `std::sync::OnceLock` is ideal for this situation.

However it seems that `cc` doesn't want to introduce another dependency `once_cell` and the msrv is on 1.46

NobodyXu:
> Initial reaction: I don't think we should be using a static mut for this. Is there some structure by which we could make this a field of the relevant object instead?

I've fixed this by having an immutable global variable for token:

```rust
static TOKEN: Mutex<Option<jobserver::Acquired>> = Mutex::new(None);
```

the `jobserver` still uses mutable global variable as before, but that is due to `cc` having a msrv of 1.46 so we cannot use `std::sync::OnceLock` and `cc` also don't want to have `once_cell` as a dependency.

NobodyXu:
I've updated the PR to only obtain jobserver token when actually compiling objects, so there is no implicit global token now.

Also, I've replaced use of `Option` with `MaybeUninit` and eliminate an `unwrap()` inside `jobserver`.

src/lib.rs
```diff
@@ -1265,36 +1260,43 @@ impl Build {
...
+                match client.available() {
+                    Ok(0) | Err(_) => {
+                        // There is no available tokens.
+                        // This is probably because `cc::Build::compile_objects` is called in
+                        // `build.rs` where the `cargo` holds the token.
+                        //
+                        // In that case, we will release one raw token just to prevent deadlock,
+                        // but this will increase the allowed global concurrency by one.
+                        let _ = client.release_raw();
+                    }
```
NobodyXu, line 1297: @joshtriplett I implemented what you requested (only acquire and release token when actually compiling) but with a catch.

`cargo` would acquire one token before running the `build.rs`, so the available token could be `0` when we obtained the jobserver via `jobserver::Client::from_env` and causing deadlock.

To prevent this, I added this piece of code which release one token into the jobserver, but it cannot be re-acquired unless I added back the global state for the obtained token.

NobodyXu: Oops, the CI is still in deadlock.

NobodyXu:
Hmmm given the deadlock caused by PR and the global state required to fix it, I suppose that it's better leave it as-is for now.

PR #780, opened 2023-01-30
Optimize `Build::compile_objects`: Only spawns one thread

NobodyXu:
This PR optimizes `Build::compile_objects` so that it only spawns one thread for printing warnings for both `cfg(features = "parallel")` and `cfg(not(features = "parallel"))`.

This PR achieves this by using `os_pipe` to create a pair of pipe fds and use that through out the entire `Build::compile_objects` for `Command::stderr`.

This eliminates the thread spawns for each process just to forward warnings.

The second optimization is specific to `cfg(features = "parallel")`, which previously spawns one thread per process creation.

I replaced that by simply adopting `Command::spawn` and collects the `Command`, program_name, `process::Child` and the `jobserver::Acquired` token into a `Vec`, then iterate over them.

This removes all the `unsafe` code in there while also significantly improves performance, especially when the users has a lot of source files but each don't take long to compile (sounds like common cases for me).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ edition = "2018"
 
 [dependencies]
 jobserver = { version = "0.1.16", optional = true }
+os_pipe = "1"
 
 [features]
 parallel = ["jobserver"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1218,8 +1218,7 @@ impl Build {
     }
 
     #[cfg(feature = "parallel")]
-    fn compile_objects<'me>(&'me self, objs: &[Object]) -> Result<(), Error> {
-        use std::sync::atomic::{AtomicBool, Ordering::SeqCst};
+    fn compile_objects(&self, objs: &[Object]) -> Result<(), Error> {
         use std::sync::Once;
 
         // Limit our parallelism globally with a jobserver. Start off by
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -19,6 +19,7 @@ edition = "2018"
 
 [dependencies]
 jobserver = { version = "0.1.16", optional = true }
+os_pipe = "1"
```
NobodyXu, line 22: The msrv CI for windows failed because it automatically pulls in the latest version of `os_pipe` v1.1.2, which depends on windows-sys v0.42.

windows-sys v0.42 seems to use `if`, `match`, `||` and `&&` in `const`, hence failed in rust v1.34.

However, we only need `os_pipe` v1.0.0 to work here, I've tested this with `cargo update --Zminimal-version` and all the tests ran just fine.

So I think this does not break the msrv given that the users manage the `Cargo.lock` to not use the latest version of `os_pipe`, but not entirely sure about that.

src/lib.rs
```diff
@@ -3710,3 +3671,56 @@ impl AsmFileExt {
         None
     }
 }
+
+struct PrintThread {
+    handle: Option<JoinHandle<()>>,
+    pipe_writer: Option<os_pipe::PipeWriter>,
+}
```
NobodyXu, line 3678: I can make forwarding warnings more efficient by creating and re-use the same printing thread for all `cc::Build::compile_objects` call, but then I cannot ensure all warnings are forwarded.

This might be useful if users call `cc::Build::compile_objects multiple times, but not sure whether it's worth losing some warnings.

Cargo.toml
thomcc: I agree this doesn't practically update the MSRV. I think a reasonable solution would be to run `cargo update -p os_pipe --precise 1.0.0` (possibly after `cargo generate-lockfile` if needed) in the CI script.

That said, do you know what version these features require? I don't mind bumping it a bit.

src/lib.rs
thomcc: I think it's fine as-is for now.

Cargo.toml
NobodyXu: > That said, do you know what version these features require? I don't mind bumping it a bit.

I looked into the tracking issue https://github.com/rust-lang/rust/pull/57563 and it seems to be the first item in the list https://github.com/rust-lang/rust/issues/49146 , which is stablised in 1.46.

P.S. If we are bumping msrv to 1.46, then I can replace the `Option` used in `fn jobserver()` with `mem::MaybeUninit`.

thomcc: 1.46 should be fine. I'll ask if anybody objects on Zulip, but my main goal was to avoid going past Debian stable (which is on 1.48), since it tends to cause a lot of spurious bug reports. Even then, those would probably be directed at the crate that causes the compilation error (in this case windows-sys), rather than us.

That said, I'm slightly unsure about adding the dependency on os_pipe, since it's not nearly as widely used as cc, and is pretty lightweight (so it might be better to implement manually). That said, it might be fine.

NobodyXu: > (so it might be better to implement manually).

That's definitely doable, though I personally don't want more `unsafe` code in `cc`.

thomcc: Sure, although I think that might be easier unsafe to audit than the current code.

thomcc: It's also notable that we don't currently have any non-optional dependencies, and even among optional dependencies we don't have any that aren't maintained by rust-lang team members (this would change that). So I'm not sure, it might be worth being a bit conservative here.

I'm largely inclined to think windows-sys and libc are fine, though.

NobodyXu: > So I'm not sure, it might be worth being a bit conservative here.

Hmmm, maybe a good idea to move `os_pipe` into rust-lang considering that it is also a fundamental lib used by other crates?

CC @oconnor663 owner of `os_pipe`.

thomcc: In any case I still need to review this code and use of the dependency (sadly, it's looking like I might not get to this until the weekend, not sure).

oconnor663: Happy to help if I can. Luckily the whole crate isn't very much code :)

NobodyXu:
@thomcc Friendly ping just in case you forgot this PR

thomcc:
Ah, sorry, I've been very busy and sadly been neglecting this more than I should. Give me a week or so, I should have some time after that.

NobodyXu:
@thomcc Pinging as this has been stale for really long time.

NobodyXu:
@thomcc Pinging since this has been stale for another 2 months.

NobodyXu:
@JohnTitor Can you review this please if you have time?
It seems that @thomcc is too busy to review this PR and it has been stale for really long time.

This PR isn't very complex and it should not take you too much time to review it.

NobodyXu:
@thomcc Please review this if you have time, it has been stale for really long time.

src/lib.rs
```diff
@@ -1357,26 +1318,30 @@ impl Build {
...
 
-        impl Drop for JoinOnDrop {
+        impl Drop for KillOnDrop {
             fn drop(&mut self) {
-                if let Some(thread) = self.0.take() {
-                    drop(thread.join());
-                }
+                let child = &mut self.0;
+
+                child.kill().ok();
```
joshtriplett, line 1327: Nit, but I'm not a fan of using `.ok()` as an idiom to ignore the return value here. I'd suggest using `let _ = ` and a comment.

(Or, possibly, calling `.expect()` since it's probably worth killing the process if this fails. But that'd be a semantic change from the previous code, so *if* done at all that should be a separate PR.)

joshtriplett:
I love this PR. Thank you!

I reviewed the code, and both the intent and implementation seems reasonable to me. Given that, and that it passes CI, I'm going to go ahead and merge it.

joshtriplett (approved):

NobodyXu:
@apoelstra The error log says "linking with cc failed", meaning it isn't executing any cc commands but rather failed linking with cc.

Since you are doing a no-std-test, I think this is where the error comes from, since cc relies on linking with libc on unix to work.

apoelstra:
@NobodyXu this makes sense, but is somehow even more confusing than my original theory :(. Why would this PR cause a *link* error with the `cc` crate itself?

joshtriplett:
I am similarly not seeing how this PR causes the observed issue. Can you file a new issue so this doesn't get lost?

PR #781, opened 2023-01-31
Bump MSRV to 1.46.0

thomcc:
Discussed [here](https://rust-lang.zulipchat.com/#narrow/stream/351149-t-libs.2Fcrates/topic/Any.20objections.20to.20bumping.20.60cc.60's.20MSRV).

Motivation is #780, but better to do as a separate patch (both so that I remember to note it in relnotes and because it's not conceptually related to that change)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -117,7 +117,7 @@ jobs:
     steps:
     - uses: actions/checkout@master
     - name: Install Rust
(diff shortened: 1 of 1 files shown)


JohnTitor (approved):
I think 1.46 is still old enough and the motivation sounds sensible to me.

thomcc:
Thank you.

PR #782, opened 2023-01-31
Speedup `Build::clone`: Use `Arc` instead of `String`/`PathBuf`/`OsString`

NobodyXu:
TBF, I'm not sure how much this will help real-world usage.

I submit this PR based on the observation that the APIs of `Build` accepts `&str`, `&Path` and `&OsStr`, so converting them to an `Arc<T>` should be as cheap as `String`/`PathBuf`/`OsString` given that `build-overrides` is compiled without optimization, so the compiler cannot simply omit these heap allocation (and even with optimization enabled `-O3`, it's hard to do so).

While `Arc::drop` is indeed due to atomic operations, I think this is mostly going to be ok given that the heap deallocation is likely to be more expensive than that since `build-overrides` uses platform-default allocator, which is often not the most optimized.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,33 +91,33 @@ pub mod windows_registry;
 /// documentation on each method itself.
 #[derive(Clone, Debug)]
 pub struct Build {
-    include_directories: Vec<PathBuf>,
-    definitions: Vec<(String, Option<String>)>,
-    objects: Vec<PathBuf>,
-    flags: Vec<String>,
-    flags_supported: Vec<String>,
+    include_directories: Vec<Arc<Path>>,
+    definitions: Vec<(Arc<str>, Option<Arc<str>>)>,
+    objects: Vec<Arc<Path>>,
+    flags: Vec<Arc<str>>,
+    flags_supported: Vec<Arc<str>>,
     known_flag_support_status: Arc<Mutex<HashMap<String, bool>>>,
-    ar_flags: Vec<String>,
-    asm_flags: Vec<String>,
+    ar_flags: Vec<Arc<str>>,
+    asm_flags: Vec<Arc<str>>,
     no_default_flags: bool,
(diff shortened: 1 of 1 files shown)


thomcc:
I'm not sure this is worth the trouble, TBH... How often does it get cloned? I suppose it doesn't hurt much.

NobodyXu:
> I'm not sure this is worth the trouble, TBH... How often does it get cloned? I suppose it doesn't hurt much.

I'm also not sure about this, but the change here shall have minimal performance impact on code that does not use Build::clone, that's why I submit the PR even I'm not sure.

But feel free to close it if you feel like there's too much to review and not much gain.

thomcc:
The review is easy but it causes churn on other PRs. That's the big downside. That said I suppose it's fine.

thomcc (approved):
Looks harmless, as mentioned.

K4sum1:
As someone who doesn't really know how to actually code and has had to revert progressive linking to get a working final build, this almost feels like a fuck me in particular even though nobody could've foreseen me being here three years later trying to build Firefox 128 with cursed build system. Literally "it's not worth the trouble" yet approved and now I'm stuck because idk how to go forward without reverting this which conflicts even more with later releases and AAAAAAAAAAA.

I actually don't know how to translate this to use the new string whatever as a lot of it has no applicable replacement in this commit, so for 128 I just reverted to older cc-rs and it works I guess.

https://github.com/Eclipse-Community/r3dfox-old/commit/e42fbcd26afc150b26c8666527ae3a6bee1f0a6c

https://github.com/Eclipse-Community/r3dfox-old/commit/9509e9d29648b93fd9e87961297ad01ee9b6d21a

NobodyXu:
If progressive linking has been an issue for your project, please open a new issue with context so that we can help you with it

PR #784, opened 2023-02-01
Do not pull in env var `CXXFLAGS`/`CFLAGS` in `Build::is_flag_supported`

NobodyXu:
Also enable the warnings and extra warnings flags regardless of env var `CXXFLAGS`/`CFLAGS` and warnings_into_errors flags.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -499,18 +499,40 @@ impl Build {
         let obj = out_dir.join("flag_check");
         let target = self.get_target()?;
         let host = self.get_host()?;
+
+        // Use a new Build instance here
         let mut cfg = Build::new();
-        cfg.flag(flag)
-            .target(&target)
-            .opt_level(0)
+        cfg.target(&target)
             .host(&host)
             .debug(false)
             .cpp(self.cpp)
             .cuda(self.cuda);
-        if let Some(ref c) = self.compiler {
-            cfg.compiler(c.clone());
+
+        cfg.compiler = self.compiler.clone();
+
+        // Call Build::get_base_compiler compiler here instead of
+        // Build::try_get_compiler to avoid pulling in environment variables
+        // CXXFLAGS or CFLAGS.
+        let mut compiler = cfg.get_base_compiler()?;
+
+        // Disable default flag generation via environment variable
+        let no_defaults = cfg.getenv("CRATE_CC_NO_DEFAULTS").is_some();
+
+        if !no_defaults {
+            cfg.add_default_flags(&mut compiler, &target, "0")?;
+        } else {
+            println!("Info: default compiler flags are disabled");
         }
-        let mut compiler = cfg.try_get_compiler()?;
+
+        // Enable all warnings
+        compiler.push_cc_arg(compiler.family.warnings_flags().into());
+
+        if let Some(wflags) = compiler.family.extra_warnings_flags() {
+            compiler.push_cc_arg(wflags.into());
+        }
+
+        // And turn these warnings into errors
+        compiler.push_cc_arg(compiler.family.warnings_to_errors_flag().into());
 
         // Clang uses stderr for verbose output, which yields a false positive
         // result if the CFLAGS/CXXFLAGS include -v to aid in debugging.
@@ -519,6 +541,7 @@ impl Build {
         }
 
         let mut cmd = compiler.to_command();
+
         let is_arm = target.contains("aarch64") || target.contains("arm");
         let clang = compiler.family == ToolFamily::Clang;
         command_add_output_file(
@@ -537,7 +560,10 @@ impl Build {
             cmd.arg("-c");
         }
 
-        cmd.arg(&src);
+        cmd.arg(&src).stdout(Stdio::null());
+
+        // Add the flag to be tested
+        cmd.arg(flag);
 
         let output = cmd.output()?;
         let is_supported = output.status.success() && output.stderr.is_empty();


thomcc:
Please move the Cow changes to another PR. I think you may be overthinking the impact of these strings though.

NobodyXu:
> Please move the Cow changes to another PR.

Done by `git reset HEAD~5`.

> I think you may be overthinking the impact of these strings though.

Hmmm yeah it could be since cloning these strings are cheap anyway.

src/lib.rs
```diff
@@ -519,6 +541,7 @@ impl Build {
         }
```
mqudsi, line 541: Technically if you are no longer inheriting CFLAGS/CXXFLAGS then this is no longer needed, but this is arcane enough knowledge and sufficiently relevant that it probably deserves to be preserved as a comment.

src/lib.rs
```diff
@@ -537,7 +560,10 @@ impl Build {
...
 
-        cmd.arg(&src);
+        cmd.arg(&src).stdout(Stdio::null());
+
+        // Add the flag to be tested
+        cmd.arg(flag);
 
         let output = cmd.output()?;
         let is_supported = output.status.success() && output.stderr.is_empty();
```
mqudsi, line 569: If you're now converting warnings into errors, then the second half of this check is now unnecessary and ends up requiring additional handling like the `-v` stuff above and might yet still trigger false negatives.

mqudsi (commented):
Forgive me for asking but is this correct? With this patch you are now more correctly testing whether or not the named executable supports the flag in question - I agree. 

But if the compiler is going to (ultimately) be run w/ some or all the currently resident CFLAGS/CXXFLAGS, do you want to know whether or not the compiler supports the flag *at all* or do you want to know if the compiler supports the flag *with the flags you're going to be using it with*?

NobodyXu:
> But if the compiler is going to (ultimately) be run w/ some or all the currently resident CFLAGS/CXXFLAGS, do you want to know whether or not the compiler supports the flag _at all_ or do you want to know if the compiler supports the flag _with the flags you're going to be using it with_?

I'm not sure about this now that you mention it, but pulling in the env variables could specify flags to ignore warnings or use some other flags to change compiler behavior in some unknown way, so I think it's just better to only test the flag requested by the users.

> I think any context as to what precipitated this PR would help shed light on the answer to this question.

The initial motivation is for performance optimization, so that I can later create `Build::is_flags_supported` and then parallelize them.

But when reading the code, I notice this and I really don't think pulling in env variable when testing supported flag is a good idea, hence the change.

mqudsi:
The compiler doesn't (usually? ever?) use the environment variable directly so it's just what cc-rs does with it that matters. If you absolutely *need* warnings to be treated as errors, as long as your `-Werror` (or whatever) comes after the CFLAGS one then that's what will end being used (e.g. `clang -Wno-error -Werror foo.c` is the same as `clang -Werror foo.c`)

One reason why I see a potential problem with this code is that it's common for build scripts (in other build systems, at any rate) to temporarily push to CFLAGS/CXXFLAGS in order to influence the behavior of an executed build system function.

It's the inverse of your case but [here's a real example from the fish CMake scripts](https://github.com/fish-shell/fish-shell/blob/fb3056d7a1de73a8f03e1f53601561d1d680ffea/cmake/ConfigureChecks.cmake#L144-L150):

```cmake
# glibc 2.30 deprecated <sys/sysctl.h> because that's what glibc does.
# Checking for that here rather than hardcoding a check on the glibc
# version in the C++ sources at point of use makes more sense.
SET(OLD_CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
check_include_files("sys/types.h;sys/sysctl.h" HAVE_SYS_SYSCTL_H)
SET(CMAKE_C_FLAGS "${OLD_CMAKE_C_FLAGS}")
```

As with `cc.is_flag_supported()`, cmake's `check_include_files()` doesn't provide a way for users to customize how it handles the test. But the fact that it *doesn't* elide CFLAGS/CXXFLAGS before performing the test means the end user/dev can influence how the test is run. It *is* hacky, but it's also powerful and leaves the developer with an option besides forking cc-rs just to run their test.

I'm not stomping my foot down and saying we can't treat warnings as errors or inherit CFLAGS/CXXFLAGS when invoking a test, but I *am* saying that this isn't just a stylistic change or a refactor and needs to be given the proper thought. The change has user-visible ramifications and while it does make the behavior of the function more deterministic, more deterministic is not always better.

I don't have a perfect example for you of where testing a flag may return a warning but the user may want to ignore that warning and consider the test as passed off the top of my head, but I've dealt with various platforms, compilers, and build systems over the years that I *can* envision that such a thing is not only possible, but even likely.

You [contributed the PR](https://github.com/rust-lang/cc-rs/commit/d69bc7db4493685508a1fecab8b98c5de811bd45) that now checks for a non-zero exit code in addition to checking if stderr is empty. If *I* were to have filed that PR, it would have been to check the exit code *instead of* whether stderr is empty and left it at that (without this PR here), allowing the dev to ultimately influence what constitutes a test pass or failure instead of boxing them in.

thomcc:
Yeah I'm not sure about this. There are some flags that only work in conjunction with others, for example consider something like `-msse2` and `--target=x86_64-unknown-linux-gnu`.

NobodyXu:
> Yeah I'm not sure about this. There are some flags that only work in conjunction with others, for example consider something like `-msse2` and `--target=x86_64-unknown-linux-gnu`.

Oh that's true, I didn't consider that...

thomcc:
So I don't think we want to do this. It's also the kind of change that's very risky given how many environments `cc-rs` is used in that we don't have tons of test coverage, so we'd need a really strong reason to do it (even if we didn't know that there are cases where it's invalid).

Do you have a use-case this unblocks? If so, it *might* be worth adding as an opt-in feature.

NobodyXu:
> Do you have a use-case this unblocks?

I don't have any of them in mind.

thomcc:
Okay, then I think we should close this, sorry. Thanks for the work, though

(P.S. I should be able to get to your other PRs tomorrow evening — been busier than I had hoped this week)

PR #786, opened 2023-02-02
If a file path references a parent directory, transform it to ensure we keep inside the OUT_DIR

jmesmon:
Otherwise, source file paths like `../../../../foo.c` can result in writing `$CARGO_TARGET_DIR/foo.o` (or worse escapes into the wider filesystem).

This is a slightly more aggressive application of object path re-writing than strictly needed to ensure we don't have `OUT_DIR` escapes, but it should be perfectly effective.


thomcc (approved):
Nice catch.

PR #787, opened 2023-02-02
`#[derive(Default)]` for `Build` and use `Build::default` in `Build::new`

NobodyXu:
To achieve this, extract new type `BuildCargoMetadataConfig` contains two fields `cargo_metadata` and `emit_rerun_if_env_changed`, which is initialized to `true` in
`<BuildCargoMetadataConfig as Default>::default` impl which is implemented manually.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,12 +84,27 @@ mod vs_instances;
 
 pub mod windows_registry;
 
+#[derive(Clone, Debug)]
+struct BuildCargoMetadataConfig {
+    cargo_metadata: bool,
+    emit_rerun_if_env_changed: bool,
+}
+
+impl Default for BuildCargoMetadataConfig {
+    fn default() -> Self {
+        Self {
+            cargo_metadata: true,
+            emit_rerun_if_env_changed: true,
+        }
+    }
+}
+
 /// A builder for compilation of a native library.
 ///
 /// A `Build` is the main type of the `cc` crate and is used to control all the
 /// various configuration options and such of a compile. You'll find more
 /// documentation on each method itself.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct Build {
     include_directories: Vec<Arc<Path>>,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -84,12 +84,27 @@ mod vs_instances;
 
 pub mod windows_registry;
 
+#[derive(Clone, Debug)]
+struct BuildCargoMetadataConfig {
+    cargo_metadata: bool,
+    emit_rerun_if_env_changed: bool,
+}
```
mqudsi, line 91: This seems like an arbitrary grouping of random variables with the only combining theme being that `Default::default()` doesn't suit them, and it requires nesting field access everywhere in the code just so that you can `#[derive(Default)]` on `Build` rather than implement it manually (which is what `Build::new()` was essentially doing).

The typical reason for doing this in another programming language would be to make the codebase more future-proof so that if someone adds a property/field to the class/struct they won't forget to add the correct default to the constructor (`Build::new()`) but that's not a concern in rust since it'll refuse to compile if you're not instantiating all the struct's fields at each and every call site.

I'm not sure what other benefit this brings? If that's the only reason then it's really not a net improvement for maintainability, discoverability, or ergonomics.

NobodyXu: > This seems like an arbitrary grouping of random variables with the only combining theme being that Default::default() doesn't suit them, and it requires nesting field access everywhere in the code just so that you can #[derive(Default)] on Build rather than implement it manually (which is what Build::new() was essentially doing).

While that is my initial motivation, the variables that are put here are related to cargo metadata emit which is enabled by default, so it's not random.

> I'm not sure what other benefit this brings? If that's the only reason then it's really not a net improvement for maintainability, discoverability, or ergonomics.

It does make `Build::new` a lot simpler and easier to add new fields, considering that `Build` is actually used like a builder-pattern, any new field to be add would probably be an `Option<T>` or `bool`.

NobodyXu:
Hmmm I think the status-quo is ok.
While using derive(Default) would reduce the amount of code, it isn't something necessary and probably only has some minor effect on readability.

PR #757, opened 2022-11-26
Fix `Build::is_flag_supported`: Check exit status of the compiler

NobodyXu:
Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

PR #790, opened 2023-02-08
Check exit status only in `Build::is_flag_supported`

NobodyXu:
This commit/PR enables warnings_into_errors, which causes the compiler to fail if there is any warning unless user override them by specifying env variable `CFLAGS` or `CXXFLAGS`.

Thus we no longer need to check stderr for warnings and can now only check exit status of the compiler.

This commit also fixed the fn `reset_env` in `tests/test.rs` to use `std::env::remove_var` instead of `std::env::set_env($var, "")`.

This commit/PR is done according to the advice from @mqudsi in https://github.com/rust-lang/cc-rs/pull/784#issuecomment-1414333799

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -507,11 +507,8 @@ impl Build {
             .host(&host)
             .debug(false)
             .cpp(self.cpp)
-            .cuda(self.cuda);
-
-        if !cfg.has_flags() {
-            cfg.warnings_into_errors(true);
-        }
+            .cuda(self.cuda)
+            .warnings_into_errors(true);
 
         if let Some(ref c) = self.compiler {
             cfg.compiler(c.clone());
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -499,6 +499,7 @@ impl Build {
         let obj = out_dir.join("flag_check");
         let target = self.get_target()?;
         let host = self.get_host()?;
+
         let mut cfg = Build::new();
         cfg.flag(flag)
             .target(&target)
@@ -507,6 +508,11 @@ impl Build {
             .debug(false)
             .cpp(self.cpp)
             .cuda(self.cuda);
+
+        if !cfg.has_flags() {
+            cfg.warnings_into_errors(true);
+        }
+
         if let Some(ref c) = self.compiler {
             cfg.compiler(c.clone());
         }
@@ -537,10 +543,13 @@ impl Build {
             cmd.arg("-c");
         }
 
-        cmd.arg(&src);
+        cmd.arg(&src)
+            .stdin(Stdio::null())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null());
 
-        let output = cmd.output()?;
-        let is_supported = output.status.success() && output.stderr.is_empty();
+        let status = cmd.status()?;
+        let is_supported = status.success();
 
         known_status.insert(flag.to_owned(), is_supported);
         Ok(is_supported)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -537,10 +543,13 @@ impl Build {
             cmd.arg("-c");
         }
 
-        cmd.arg(&src);
+        cmd.arg(&src)
+            .stdin(Stdio::null())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null());
```
mqudsi, line 549: Should we inherit stdout and stderr?

NobodyXu: I believe we should not, since stdout and stderr is for messages from compiler when building the project.
`Build::is_flag_supported` does not build the project but only test the flags, so I think the stdout and stderr should be piped to null.

NobodyXu: > But we don't provide any other mechanism for the developer to check why a test failed, unlike other build systems that write the output of all these checks/tests to a log file.

Ehh that's true, I can set that to inherit but I think the stdout/stderr might be a little bit confusing to users.
If that's fine, then I can remove these lines.

mqudsi: Let’s wait for another opinion? That was just my two cents.

thomcc:
I'm not sure about this. I'm pretty sure there's a reason the current code was the way it was... I'll have to go looking.

mqudsi:
I had done some git spelunking when I was looking into #784 and found that it was just originally added as "execute and check if stderr is empty" to see if the presence of the flag caused any errors or warnings. Later @NobodyXu added the exit code check in #757 but without any rationale or motivating bug (I don't think it changed the behavior in any way since "exit code is 0" is probably a strict subset of "stderr is empty").

If you're not using `-Werror` then checking the stderr is required, but if you're using `-Werror` then checking the stderr is overkill because it necessitates workarounds like 

https://github.com/rust-lang/cc-rs/blob/0d9a0f84e43fdb3eef34ff9726f7cb915a2b619a/src/lib.rs#L515C9-L519

which is only needed because stderr may sometimes contain non-error output (because no one says otherwise).

src/lib.rs
```diff
@@ -507,6 +508,11 @@ impl Build {
             .debug(false)
             .cpp(self.cpp)
             .cuda(self.cuda);
+
+        if !cfg.has_flags() {
+            cfg.warnings_into_errors(true);
+        }
```
mqudsi, line 514: I think this is too generous, flags is most likely going to be something like `-O2` than `-Wno-error`. We think we should always apply `-Werror` but just make sure it comes before `CFLAGS`/`CXXFLAGS`.

mqudsi:
I've been rethinking what I said in #784, and I'm not sure using CFLAGS and CXXFLAGS is necessarily the right approach. If a user has an invalid flag in CFLAGS, that means every `is_flag_supported()` call will fail/return false (this is the current behavior).

src/lib.rs
NobodyXu: > I think we should always apply -Werror but just make sure it comes before CFLAGS/CXXFLAGS.

I've enabled `-Werror` in any case but I can't put it before `CFLAGS`/`CXXFLAGS` unless I change `Build::try_get_compiler` or do sth. similar to #784

NobodyXu:
Hmmm I think it's better to check stderr just to be sure that the flag is indeed supported and there might be cases where the flag is just ignored and compiler exits with success.

PR #792, opened 2023-02-09
Update some doc comments

mqudsi:
Super minor, only filing because the nvcc comment was bothering me every time I saw it.

* Clarify the nvcc behavior documented for `push_cc_arg()` and reformat its comment.
* Update the doc comments for functions that handle duplicate flags to clarify that they only check optimization flags at the moment.


JohnTitor (approved):
LGTM, thanks!

PR #793, opened 2023-02-09
Prevent cloning on every `get_xxx()` call

mqudsi:
Return a `Cow<str>` instead of `String`. We never use the allocated result and just compare against it, so there's no need to re-allocate each time.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,6 +56,7 @@
 #![allow(deprecated)]
 #![deny(missing_docs)]
 
+use std::borrow::Cow;
 use std::collections::{hash_map, HashMap};
 use std::env;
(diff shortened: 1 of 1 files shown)


thomcc (approved):
Not public API so seems fine. Any reason why this is just for get_host? There are a few nearby that seem harmless to do it for.

mqudsi:
I only changed the one I was using. Happy to update the rest 👍.

thomcc:
I don't think it's super important (these strings are all very small). That said,  you feel a desire to anyway, it seems good to be consistent to the ones which are not public API, and exist also as a local.

mqudsi:
I've changed the other functions. Same case as before with the results of these private functions never being stored.

PR #794, opened 2023-02-09
Change approach for native and cross-compilation with clang++ on FreeBSD

mqudsi:
Correctly fix C++ compilation w/ clang++ on FreeBSD

    When cross compiling, we are forced to emit the `--target=<triple>` compiler
    flag, but this can break clang++ C++ builds on FreeBSD if the triple isn't
    set to a value that causes clang++ to infer the use of libc++ instead of
    libstdc++ (the latter of which does not support C++11 and beyond on FreeBSD).

    By manually specifying the usage of `-std=libc++` we avoid the need to know
    which version of FreeBSD we are targeting (as we do not have that information
    from the rust target triple and cannot infer it from the host as we are
    cross-compiling).

    Note that we were already specifying a default stdlib value libc++ for
    `cpp_link_stdlib` under FreeBSD but that is not sufficient.

    There are other solutions that would work when not cross compiling such as
    omitting the `--target` altogether, but there's no compelling reason to add more
    code if this solution works for both native and cross-compiling cases.

    This fix supersedes #788. See relevant discussions in #785 and #788.

(Updated)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1720,26 +1720,30 @@ impl Build {
                         } else if target.contains("aarch64") {
                             cmd.args.push("--target=aarch64-unknown-windows-gnu".into())
                         }
-                    } else if target.ends_with("-freebsd") && self.get_host()?.ends_with("-freebsd")
-                    {
-                        // clang <= 13 on FreeBSD doesn't support a target triple without at least
-                        // the major os version number appended; e.g. use x86_64-unknown-freebsd13
-                        // or x86_64-unknown-freebsd13.0 instead of x86_64-unknown-freebsd.
-                        // The current version is appended. If it doesn't align with your goals, pass
-                        // .flag("--target=...") in the build script or adjust CXXFLAGS accordingly.
-                        let stdout = std::process::Command::new("freebsd-version")
-                            .output()
-                            .map_err(|e| {
-                                Error::new(
-                                    ErrorKind::ToolNotFound,
-                                    &format!("Error executing freebsd-version: {}", e),
-                                )
-                            })?
-                            .stdout;
-                        let stdout = String::from_utf8_lossy(&stdout);
-                        let os_ver = stdout.split('-').next().unwrap();
+                    } else if target.ends_with("-freebsd") {
+                        // FreeBSD only supports C++11 and above when compiling against libc++
+                        // (available from FreeBSD 10 onwards). Under FreeBSD, clang uses libc++ by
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1662,7 +1662,11 @@ impl Build {
                 if !(target.contains("android")
                     && android_clang_compiler_uses_target_arg_internally(&cmd.path))
                 {
-                    if target.contains("darwin") {
+                    if self.get_host()? == target {
+                        // We are not cross-compiling so we do not need to explicitly
```
thomcc, line 1666: I think this is likely to break too much. We shouldn't change this for no reason.

thomcc (changes requested):
The freebsd change is mostly what I had in mind. But the location of the target==host check will change the behavior for every non-cross scenario is going to break *way* too much.

I'm sympathetic to your suggestion from the other thread that it's better to error than silently be subtly broken. I don't know that it applies here though — this code isn't exactly new and people have been using it successfully for a while. I think it needs a high degree of motivation to change. That is:

1. I don't see evidence that the behavior at the moment (which essentially always passes `--target`) is in error. There are cases when we definitely need to do this even when not cross-compiling (It's possible for rustc to think it's targetting the host, and the c compiler to not).

2. This crate is pretty old, and is used in far more configurations than we can possibly have tests for.
    	
	Not to mention I think this will concretely break usage when called from xcodebuild, when using `cargo.target.<triple>.runner`+qemu (or similar) to run the build script, etc. Perhaps I don't particularly care about such configurations, but this just breaks them for no reason, which is not good.

    While I don't take take the stance that we must keep perfect 100% compatibility with every potential configuration (doing so would effectively freeze changes to `cc`)... it's better not to cause problems for no reason.

mqudsi:
That was my original thinking in 785, happy to do the same here. Thanks for reviewing.

thomcc:
Ah, apologies if I misunderstood. (I should admit I've been a bit distracted by goings on at \$dayjob)

mqudsi:
No worries. My bit about "breaking rather than being subtly wrong" was only in argument to not go with the approach in #788. I support minimizing the impact of any changes to this crate and fully agree with the points you made. I just misunderstood what your recommendation was from the other thread but I'm glad we're on the same page now.

I've updated the branch to just use the `-stdlib` approach for both native and cross-compiling with no other changes. I've tested that this continues to work for native compilation.

thomcc (approved):
Sorry for the delay here.

mqudsi:
No worries. Thanks for merging!

PR #799, opened 2023-03-22
Refine CUDA documentation.

dot-asm:
Original wording gives an impression that user-provided flags are prefixed with -Xcompiler, which is not the case. -Xcompiler is added only to internally generated flags.


dot-asm:
Just in case. One can wonder if it would be more appropriate to prefix user-provided flags with -Xcompiler on user's behalf. As opposed to documenting it as suggested. Problem is that we don't want to break build scripts that had to figure out the current state of affairs and explicitly prefix their flags.  Though it's not impossible to take a note if build script has passed -Xcompiler and add one if deemed necessary. "Deemed necessary" refers to the fact that it's not really appropriate to prefix **all** flags, only those not recognized by nvcc. For example, it would be inappropriate to prefix `-std=c++NN`, because it's meaningful to both CUDA and host compiler.

src/lib.rs
```diff
@@ -640,10 +640,11 @@ impl Build {
...
     ///
-    /// Enabling CUDA will pass the detected C/C++ toolchain as an argument to
-    /// the CUDA compiler, NVCC. NVCC itself accepts some limited GNU-like args;
-    /// any other arguments for the C/C++ toolchain will be redirected using
-    /// "-Xcompiler" flags.
+    /// Enabling CUDA will invoke the CUDA compiler, NVCC. While NVCC accepts
+    /// most common compiler flags, e.g. `-std=c++17`, some project-specific
+    /// flags might have to prefixed with "-Xcompiler" flag, for example as
+    /// `.flag("-Xcompiler").flag("-fpermissive")`. Consult Nvidia
+    /// documentation for further information.
```
JohnTitor, line 647: I think it'd be great if we could also link to it.

```suggestion
    /// `.flag("-Xcompiler").flag("-fpermissive")`. See the Nvidia
    /// documentation for more information.
```

src/lib.rs
```diff
@@ -3296,12 +3297,14 @@ impl Tool {
...
     }
 
-    /// Push a flag to the end of the compiler's arguments list.
+    /// Push an "exotic" flag to the end of the compiler's arguments list.
     ///
-    /// Currently `-Xcompiler` is inserted before the passed flag when compiling
-    /// CUDA sources since NVCC only accepts a limited set of GNU-like flags,
-    /// while the rest must be prefixed with the `-Xcompiler` flag to get passed
-    /// to the underlying C++ compiler.
+    /// Nvidia compiler accepts only most common compiler flags like `-D`,
```
JohnTitor, line 3302:
```suggestion
    /// Nvidia compiler accepts only the most common compiler flags like `-D`,
```

dot-asm:
Resolved the nits and re-based just in case.

PR #808, opened 2023-05-17
Be more stringent about handling Android NDK.

jfgoog:
Wrapper scripts for the NDK have the format `<target>-<clang>`, so when checking to see if the filename contains "android" (which will be part of the target), split off the target part of the filename and only check that, instead of checking the entire filename.

This is a very minor improvement, but it is more correct and does have a practical benefit for us on the Android toolchain team: Our build process for rustc uses wrapper scripts that include the target in their name, and the cc crate is misidentifying them as being from the NDK and causing our Windows build to fail without a workaround.


ChrisDenton (approved):
This seems reasonable to me. Though could you squish the commits?

jfgoog:
Thanks! Squashed.

PR #812, opened 2023-06-05
Remove -fembed-bitcode flag

keith:
Apple has recently stopped accepting apps built with any Xcode version below 14.1, and rejects apps built with bitcode starting with Xcode 14.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2530,11 +2530,6 @@ impl Build {
             cmd.args.push(sdk_path);
         }
 
-        // TODO: Remove this once Apple stops accepting apps built with Xcode 13
-        if !is_mac {
-            cmd.args.push("-fembed-bitcode".into());
-        }
-
         Ok(())
     }
 


keith:
the only other case that's potentially ambiguous here is using LTO, where you might want the bitcode for that. How is this supposed to be handled by this? because also in the bitcode case this could produce only bitcode files and no object files instead

keith:
I'm less sure about that now actually, I'm not 100% sure how to tell if the embedded bitcode is being used in the case of LTO builds but with a quick test generating a `-map` file from the linker it doesn't appear to be using that

thomcc:
I don't believe we need to worry about cross-lang LTO, you pretty much have to do a bunch of special stuff to set that up anyway. I think this is correct, but ideally we'd detect the xcode version and do it based on that.

I have a patch that does that somewhere, but need to dig it out (I won't have time in the next couple weeks, maybe after)

keith:
I don't think we need to detect the Xcode version for this because Apple no longer accepts App Store submissions with anything other than [Xcode 14.1+](https://developer.apple.com/news/?id=jd9wcyov&1680024622) so it's no longer possible to submit apps that include bitcode at all since Xcode 14.1 doesn't support it. Outside of the App store there shouldn't be any reason for folks to use bitcode

thomcc:
That seems reasonable. I'll merge this after the current release (which is quite large), I'm concerned it may have a large degree of fallout (if there's a mistake here), and would rather that not be confusable with all the other changes that are going to go into 1.80.

keith:
Sounds good, thanks! Feel free to cc me on any issues that arise

thomcc:
This conflicts but is worth doing. Mind rebasing?

keith:
rebased!

thomcc:
Ugh, sorry -- happened to merge changes before getting to this that broke it again. Can you rebase again?

keith:
Done!

thomcc (approved):

thomcc:
Hmmm... I'll merge this now, and look into the CI bustage tomorrow...

keith:
Thanks!

PR #817, opened 2023-07-10
Optimize `cc::Build::try_compile`: Reuse `PrintThread`

NobodyXu:
Creating a new thread is very expensive, this commit optimizes calls to `cc::Build::assemble`, `cc::Build::assemble_progressive` and `cc::Build::compile_objects` to avoid repeated creation, destruction and joining on threads.

It also eliminates a few unnecessary heap allocation.


thomcc (approved):
Creating a new thread is not that expensive, but this patch is not unreasonable either.

PR #818, opened 2023-07-15
add a function to populate flags from an arbitrary environment variable

nagisa:
My specific use-case is to add ability for the end users to specify crate-specific environment variables, but with how general this method is I could imagine numerous other applications too.

Why does this need to be in `cc-rs`, you ask? The code to split up the `CFLAGS` or `CXXFLAGS` environment variables into flags seems trivial today, but I worry that it might not remain the case indefinitely into the future. If everybody is expected to do this in their own build scripts, we’re certain to run into a very inconsistent experience as soon as parsing needs an adjustment.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -563,6 +563,35 @@ impl Build {
         self
     }
 
+    /// Add flags from the specified environment variable.
+    ///
+    /// Normally the `cc` crate will consult with the standard set of environment
+    /// variables (such as `CFLAGS` and `CXXFLAGS`) to construct the compiler invocation. Use of
+    /// this method provides additional levers for the end user to use when configuring the build
+    /// process.
+    ///
+    /// Just like the standard variables, this method will search for an environment variable with
+    /// appropriate target prefixes, when appropriate.
+    ///
+    /// # Examples
+    ///
+    /// This method is particularly beneficial in introducing the ability to specify crate-specific
+    /// flags.
+    ///
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     .try_flags_from_environment(concat!(env!("CARGO_PKG_NAME"), "_CFLAGS"))
+    ///     .expect("the environment variable must be specified and UTF-8")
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -3133,6 +3139,34 @@ impl Build {
         }
     }
 
+    fn getenv_with_target_prefixes(&self, var_base: &str) -> Result<String, Error> {
```
nagisa, line 3142: These functions have been moved and renamed to be grouped with other getenv family of functions, with no other changes intended.

src/lib.rs
```diff
@@ -3133,6 +3139,34 @@ impl Build {
...
+        match res {
+            Some(res) => Ok(res),
+            None => Err(Error::new(
+                ErrorKind::EnvVarNotFound,
+                &format!("Could not find environment variable {}.", var_base),
+            )),
+        }
+    }
+
+    fn envflags(&self, name: &str) -> Result<Vec<String>, Error> {
```
nagisa, line 3162: This function has been changed to return an `Error` when `getenv_with_target_prefixes` fails in order to allow for the possibility to handle this error in the newly added method.

thomcc:
This seems fine to me, but probably needs some documentation.

nagisa:
What sort of documentation do you think envision in addition to what is present in the doc-comment and where should it go?

thomcc:
Do you think it's worth mentioning with https://github.com/rust-lang/cc-rs#external-configuration-via-environment-variables? I could see an argument that it's not.

nagisa:
Done! I tailored that part of the documentation specifically towards the end-users (people compiling code/crates,) as that whom this documentation appeared to be aimed at.

thomcc:
This all looks good to me, I'll take another look some time it's not 2AM before I merge, though.

thomcc (approved):
LGTM

nagisa:
Do you have a time in mind for then there might be a new version? I can submit a PR with version bumps if that all it needs.

Issue #819, opened 2023-07-17
Windows - MinGW gcc: Compiled object always gets put in the root project directory


Spencer1O1:
### The issue:
I have a C file in the root of my Rust project: my_interop.c (the issue persists wherever I place it)
My build.rs script, about as simple as it can get, looks like this:
```rust
fn main() {
    cc::Build::new()
        .file("my_interop.c")
        .compile("my_interop");
    println!("cargo:rustc-link-lib=user32"); // my_interop.c includes <windows.h>
}
```
When building, the log says,
```bash
running: "C:\\MinGW\\bin\\gcc.exe" "-O0" "-ffunction-sections" "-fdata-sections" "-g" "-fno-omit-frame-pointer" "-m64" "-Wall" "-Wextra" "-FoC:\\Dev\\my-project\\target\\debug\\build\\my-project-4a742f3493d90fa9\\out\\my_interop.o" "-c" "my_interop.c"
```
which should put the object file in ``target/debug/build/my-project-.../out/get_wp_handle.o``
However, I find the object file in the root directory, outside of target.

Then, in the linking step, the log says,
```bash
running: "C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.34.31933\\bin\\HostX64\\x64\\lib.exe" "-out:C:\\Dev\\my-project\\target\\debug\\build\\my-project-4a742f3493d90fa9\\out\\libmy_interop.a" "-nologo" "C:\\Dev\\my-project\\target\\debug\\build\\my-project-4a742f3493d90fa9\\out\\my_interop.o"
  LINK : fatal error LNK1181: cannot open input file 'C:\Dev\my-project\target\debug\build\my-project-4a742f3493d90fa9\out\my_interop.o'
  exit code: 1181
```
which is obviously because the object file got put in the root directory rather than the designated output directory.

In my build script I have tried adding ``.output_dir("some/specific/dir")`` but the object file still gets put at the root.

### The temporary solution:
If I change my build script to be
```rust
fn main() {
    cc::Build::new()
        .file("my_interop.c")
        .out_dir(".")
        .compile("my_interop");
    println!("cargo:rustc-link-lib=user32"); // my_interop.c uses #include <windows.h>
}
```
the linking step will *expect* the object file to be in the root directory.
It compiles as needed, and the project runs fine.
However, now the issue is that all of the other output files get put there too.
We end up a bunch of junk like this,
```
<Dir>    .
[...]

Issue #800, opened 2023-03-29
`cc` passes `-Fo` flag to `gcc` on windows, which is ignored


divagant-martian:
This should be `-o` similar to previous issue #482

Spencer1O1:
I submitted a pull request that resolves this issue [#820](https://github.com/rust-lang/cc-rs/pull/820)
Just waiting to get it merged

Spencer1O1:
@thomcc merged the request so this issue should be closed now.

PR #820, opened 2023-07-17
Use gnu -o flag for obj out path instead of -Fo when using gcc & g++ on Windows

Spencer1O1:
Did the same thing that @datalus did for clang in commit [e0d6380ddc038920e01a4778f9e3947bf8f40c30](https://github.com/rust-lang/cc-rs/pull/483/commits/e0d6380ddc038920e01a4778f9e3947bf8f40c30)
Resolves issues [#800](https://github.com/rust-lang/cc-rs/issues/800) and [#819](https://github.com/rust-lang/cc-rs/issues/819)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -522,12 +522,14 @@ impl Build {
         let mut cmd = compiler.to_command();
         let is_arm = target.contains("aarch64") || target.contains("arm");
         let clang = compiler.family == ToolFamily::Clang;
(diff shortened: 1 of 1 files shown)


thomcc (approved):
Beginning to feel that we should just pass the compiler family in, rather than having all these checks... Oh well, for now this is fine I suppose.

Spencer1O1:
@thomcc Yes I thought the same thing, but I don't have time to go change everything right now. As long as it works...

PR #822, opened 2023-07-18
Vendor dependency `os_pipe`

NobodyXu:


nagisa:
Could you provide motivation for this change in the PR description?

NobodyXu:
> Could you provide motivation for this change in the PR description?

Its motivation was explained by @thomcc and @joshtriplett on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/351149-t-libs.2Fcrates/topic/cc.20request.20for.20reviews/near/373980518)

workingjubilee:
Jotting out a quick resummary is still good.

thomcc:
Basically the reason is that the existing crate is not particularly portable, and this fixes those issues (it's not actually a straight vendoring).

My stance is that this optimization is not really worth this complexity and non-portability (this PR even has to explicitly compile_error on non-windows/non-unix, which is something we would support before so long as std::process/std::fs worked...), but we've already landed it, and this at least reduces the number of things people have to PR in order to get their targets working.

NobodyXu:
>this PR even has to explicitly compile_error on non-windows/non-unix, which is something we would support prior to #780 so long as std::process/std::fs worked...)

@thomcc I just looked into rust's stdlib and it seems that process spawning is only supported on windows and unix.

In the stdlib's [`sys` module](https://github.com/rust-lang/rust/blob/77e24f90f599070af2d8051ef9adad7fe528dd78/library/std/src/sys/mod.rs#L28), it supports the following OSes:
 - unix
 - windows
 - solid_asp3
 - hermit
 - wasi
 - wasm
 - sgx: `cfg(all(target_vendor = "fortanix", target_env = "sgx"))`

And I found that other than Unix and Windows, none support process spawning:
 - [proof that solid_asp3 does not support process spawning](https://github.com/rust-lang/rust/blob/77e24f90f599070af2d8051ef9adad7fe528dd78/library/std/src/sys/solid/mod.rs#L40)
 - [proof that hermit does not support process spawning](https://github.com/rust-lang/rust/blob/77e24f90f599070af2d8051ef9adad7fe528dd78/library/std/src/sys/hermit/mod.rs#L40)
 - [proof that wasi does not support process spawning](https://github.com/rust-lang/rust/blob/77e24f90f599070af2d8051ef9adad7fe528dd78/library/std/src/sys/wasi/mod.rs#L43)
 - [proof that wasm does not support process spawning](https://github.com/rust-lang/rust/blob/77e24f90f599070af2d8051ef9adad7fe528dd78/library/std/src/sys/wasm/mod.rs#L39)
 - [proof that sgx does not support process spawning](https://github.com/rust-lang/rust/blob/77e24f90f599070af2d8051ef9adad7fe528dd78/library/std/src/sys/sgx/mod.rs#L32)

FYI, here is the [link to `unsupported/process.rs`](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unsupported/process.rs), it simply panics on process spawning.

@thomcc So no, my PR does not introduce any regression since `cc` only support windows and unix before it lands and process spawning on non-unix and non-windows OS has never been supported by stdlib anyway.

NobodyXu:
@thomcc I also discovered a pattern: OS that does not support pipe creation also don't support process spawning.

tests/support/mod.rs
```diff
@@ -34,16 +34,16 @@ impl Test {
         // lesser of the two evils.
         env::remove_var("RUSTC_WRAPPER");
 
-        let mut gcc = PathBuf::from(env::current_exe().unwrap());
+        let mut gcc = env::current_exe().unwrap();
```
joshtriplett, line 37: This change *seems* unrelated to this PR?

tests/support/mod.rs
```diff
@@ -34,16 +34,16 @@ impl Test {
...
         if gcc.ends_with("deps") {
             gcc.pop();
         }
         let td = Builder::new().prefix("gcc-test").tempdir_in(&gcc).unwrap();
         gcc.push(format!("gcc-shim{}", env::consts::EXE_SUFFIX));
         Test {
-            td: td,
-            gcc: gcc,
+            td,
+            gcc,
```
joshtriplett, line 46: This change seems unrelated to this PR. (I'm all for this kind of cleanup, but please do it in a separate PR.)

joshtriplett:
I'm wondering if we can simplify this substantially.

The `PipeReader`/`PipeWriter` wrappers are nice for the `os_pipe` crate to keep users from mixing up the two ends and reading/writing the wrong one, but they add substantial complexity. For the purposes of this crate, it seems sufficient to return `(File, File)`, and take care to use them correctly.

That then eliminates the need for `impl_Read_by_forward` and `impl_Write_by_forward`, as well as the `From` impls converting to `Stdio` (because `File` already supports that).

I also don't think it's worth including the test-support binaries in `src/bin`, and the tests that depend on those.

thomcc (approved):
This is a lot more reasonable. Thanks to both you and @joshtriplett for all the changes.

PR #825, opened 2023-07-18
Disambiguate Windows run-times when using clang to compile.

dot-asm:
This is with reference to https://github.com/rust-lang/cc-rs/issues/819#issuecomment-1640309753. In the comment I'm suggesting to take it as far as rejecting the use of gcc in the msvc compilation, while this PR only issues a warning.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2549,13 +2549,7 @@ impl Build {
         }
 
         if target.contains("msvc") && tool.family == ToolFamily::Gnu {
-            println!(
-                concat!(
-                    "cargo:warning=GNU compiler is not supported for this target, ",
-                    "consider --target={} instead."
-                ),
-                target.replace("msvc", "gnu")
-            );
+            println!("cargo:warning=GNU compiler is not supported for this target");
         }
 
         Ok(tool)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1619,6 +1619,13 @@ impl Build {
                     cmd.push_opt_unless_duplicate(format!("-O{}", opt_level).into());
                 }
 
+                if cmd.family == ToolFamily::Clang && target.contains("windows") {
+                    // Disambiguate mingw and msvc on Windows. Problem is that
+                    // depending on the origin clang can default to a mismatchig
+                    // run-time.
+                    cmd.push_cc_arg(format!("--target={}", target).into());
+                }
+
                 if cmd.family == ToolFamily::Clang && target.contains("android") {
                     // For compatibility with code that doesn't use pre-defined `__ANDROID__` macro.
                     // If compiler used via ndk-build or cmake (officially supported build methods)
@@ -2539,6 +2546,16 @@ impl Build {
                     tool.env.push((k.to_owned(), v.to_owned()));
                 }
             }
+        }
+
+        if target.contains("msvc") && tool.family == ToolFamily::Gnu {
+            println!(
+                concat!(
+                    "cargo:warning=GNU compiler is not supported for this target, ",
+                    "consider --target={} instead."
+                ),
+                target.replace("msvc", "gnu")
+            );
         }
 
         Ok(tool)


thomcc:
Yeah, my impression is it's pretty much always wrong to use gcc for a -msvc target. @ChrisDenton any objections to merging this?

ChrisDenton:
Yes the current gnu toolchains, unlike llvm, do not aim to be compatible with msvc.

src/lib.rs
```diff
@@ -1619,6 +1619,13 @@ impl Build {
                     cmd.push_opt_unless_duplicate(format!("-O{}", opt_level).into());
                 }
 
+                if cmd.family == ToolFamily::Clang && target.contains("windows") {
+                    // Disambiguate mingw and msvc on Windows. Problem is that
+                    // depending on the origin clang can default to a mismatchig
+                    // run-time.
+                    cmd.push_cc_arg(format!("--target={}", target).into());
```
ChrisDenton, line 1626: Does this still work if there's a mismatch between the clang target name and rust target name?

src/lib.rs
```diff
@@ -2541,6 +2548,16 @@ impl Build {
             }
         }
 
+        if target.contains("msvc") && tool.family == ToolFamily::Gnu {
+            println!(
+                concat!(
+                    "cargo:warning=GNU compiler is not supported for this target, ",
+                    "consider --target={} instead."
+                ),
+                target.replace("msvc", "gnu")
```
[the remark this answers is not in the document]
dot-asm: Literally "always"? No. I want to remind that [my actual position](https://github.com/rust-lang/cc-rs/issues/819#issuecomment-1640309753) is actually to fail the compilation or ignore the `$CC` environment value at this point. If the remark means that you agree, just say which option of the two is preferred, and I'll implement that. If not, then the question is if the suggestion is to soften the message to just "not supported" without making any suggestions?

Just in case for reference, the original rationale was that the `target.replace("msvc", "gnu")` arguably does the justice even if matching target does not exist. Again, the scenario is that the user actively sets `$CC` to gcc and asks to compile for msvc, which is error-prone effectively by definition. If the counterpart target doesn't exist, and the user feels strong about it, let him bring it up with Rust.

src/lib.rs
dot-asm: To my knowledge, there is no mismatch between Rust and clang. Which shouldn't really come as surprise, both are llvm-based.

dot-asm: Oh! Maybe you refer to aarch64-pc-windows-llvm, or what was it called? Fair enough, I'll double-check...

dot-asm: > Oh! Maybe you refer to aarch64-pc-windows-llvm,

It's `*-pc-windows-gnullvm`, i.e. not just aarch64, and there is "gnu" in the name. Surprisingly enough clang appears to accept gnullvm and even do arguably right thing. It appears to be due to a pattern matching, because it also accepts gnufoobar, gnu-foobar, and even msvcfoobar and msvc-foobar. And it generates code targeting mingw runtime if there is gnu somewhere in the target name, and vc runtime otherwise.

thomcc: Yes, clang accepts far more target triples than rustc, which only accepts a finite set (clang, due to compatibility with GCC, does quite a bit of massaging of the value).

dot-asm: To summarize, the answer to the original question is "yes, it still works."

src/lib.rs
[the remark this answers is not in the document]
ChrisDenton: Yes, I think it would be worse to suggest a target that does not exist. We either need to be sure it does exist or else just suggest a gnu target in less exact terms.

src/lib.rs
ChrisDenton: We can presumably check that it continues to work with official targets. My only concern would be third party rustc targets (which can be named anything). But we can either say they aren't supported or require setting the `target` to a llvm compatible one.

src/lib.rs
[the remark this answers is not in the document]
dot-asm: Fixed.

src/lib.rs
dot-asm: > We can presumably check that it continues to work with official targets.

That's what "yes, it still works" effectively mean. It still works with official targets.

ChrisDenton (approved):
I mean, I'm ok with this if @thomcc is. Though we might want to keep an eye out in case anyone is tripped up by the change.

thomcc:
I'm okay with it too.

dot-asm:
With cross-reference to #811, a case can be made that it would be appropriate to issue "unsupported" warning even for `clang++` [not to be confused with clang-cl] being used with msvc target.

PR #828, opened 2023-07-20
Cleanup mod com, registry, setup_config and winapi

NobodyXu:
Uses items exported from the official crate `windows_sys` instead of homebrew definition whenever possible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ jobserver = { version = "0.1.16", optional = true }
 libc = "0.2.62"
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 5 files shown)


thomcc:
Are any of the types from windows-rs this adds to public APIs more complex than simple typedefs?

E.g. would we have to cut a breaking change if we bump the windows-sys version? (That seems unacceptable given how frequently they cut breaking changes)

NobodyXu:
> Are any of the types from windows-rs this adds to public APIs more complex than simple typedefs of primitives?

Nope.

> E.g. would we have to cut a breaking change if we bump the windows-sys version? (That seems unacceptable given how frequently they cut breaking changes)

No, windows-sys is not used in public interfaces, all changes are to modules that are used internally.

thomcc (approved):
Seems fine then.

PR #829, opened 2023-07-20
Fix `Build::compile_objects` deadlock on parallel

NobodyXu:
`jobserver::Client::acquire` may block if there's not enough token.

To avoid blocking, this patch waits for child in a separate thread so that once the old compiler job is done, it may continue spawning new job instead of blocking forever.

This PR also makes sure that the released token is always re-acquired.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1258,7 +1258,7 @@ impl Build {
 
     #[cfg(feature = "parallel")]
     fn compile_objects(&self, objs: &[Object], print: &PrintThread) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


NobodyXu:
P.S. This could be much easier if stdlib provides a simple async runtime.

thomcc (approved):
Unfortunate, but seems necessary.

NobodyXu:
> Unfortunate, but seems necessary.

Thanks for the review, with this change it still spawns much less thread than before, where two threads are spawned for every object.

PR #831, opened 2023-07-21
Optimize `Build::get_out_dir`: Return `Cow<'_, Path>`

NobodyXu:
to avoid heap allocation if possible.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3121,15 +3121,18 @@ impl Build {
         self.force_frame_pointer.unwrap_or_else(|| self.get_debug())
     }
 
-    fn get_out_dir(&self) -> Result<PathBuf, Error> {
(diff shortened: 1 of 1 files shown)


NobodyXu:
`Build::get_{target, host, opt_level, debug}` also returns `Cow<str>`, so it does make sense to also change `Build::get_out_dir` to return `Cow<'_, Path>`

NobodyXu:
Closing this as I think there is a more aggressive optimization that can be done.

NobodyXu:
> Closing this as I think there is a more aggressive optimization that can be done.

Reopened this since I realize that the more aggressive optimization I have in mind does not involve optimization for `Build::get_out_dir`.

thomcc (approved):
Seems fine.

PR #832, opened 2023-07-21
Optimize `Build::env_cache`: Store `Option<Arc<str>>` as key

NobodyXu:
instead of `Option<String>` so that getting environment variable from `env_cache` does not need heap allocation at all.


thomcc (approved):
Fair enough. These strings are really small though, I don't really think it's worth all the trouble you're going through. That said, it's already done and doesn't make the code much more complex.

PR #833, opened 2023-07-21
Optimize `Build::compile_objects` on feature parallel

NobodyXu:
Fallback to serial compilation if `objs.len() <= 1`.


thomcc (approved):
Seems reasonable.

PR #834, opened 2023-07-23
Bump version to 1.0.80

thomcc:
There have been a lot of changes, so I think it's worth getting a release out

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cc"
-version = "1.0.79"
+version = "1.0.80"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/rust-lang/cc-rs"


dot-asm:
I've spotted few inaccuracies in project's README. Consider #835 prior the version bump:-)

complexspaces:
Hey there, sorry to jump in,

Would you please reconsider the changes that introduced `windows_sys` to `cc` before making this release? In general the `windows` ecosystem is extremely fractured at the moment, and most of the ecosystem is using 2-3 different versions of the crate at once (slower adopters + MSRV, hasn't been maintained in a bit, and extremely early adopters of new versions). 

For downstream users who have a larger number of third-party dependencies, managing all of the various windows ecosystem versions can be complicated. My fear is that this change will only make the situation worse by adding the dependency to one of the most commonly used crates for Rust, and bloating lockfiles further. Additionally, the current `main` introduces a fairly sizable compile-time regression. On my AMD 7950x, this increases compile times by ~3x. The average compile time on the last release (`1.0.79`) takes 1.5s while `main` is now taking ~4.0-4.5 seconds. 

Finally, even though I didn't explicitly test it, I expect this to hamper times more since it will require two compilations of `windows-sys` even if they have the same version because `cc` is used as a build dependency, and those don't share sub-dependencies with regular ones.

thomcc:
Yeah, I guess that's fair. I guess we could use the windows-sys bindings generator. Given how much effort the old version went through to keep the dependency profile light (for this reason), it's probably worth keeping that in mind...

I won't have time to work on this for a week or two, but I guess I'd take a patch in the mean time.

NobodyXu:
I looked into windows-sys CI and it seems that they use [riddle](https://github.com/microsoft/windows-rs/blob/master/crates/tools/riddle/src/main.rs) for generation with [args](https://github.com/microsoft/windows-rs/blob/master/crates/tools/sys/bindings.txt) for sys.

thomcc:
I meant the `windows_bindgen` crate. We use it already for the stdlib's binding generation: <https://github.com/rust-lang/rust/blob/master/src/tools/generate-windows-sys/src/main.rs>

NobodyXu:
Thanks, I would have a look at it.

ChrisDenton:
Riddle is the new tool that will replace `windows_bindgen` for this purpose. Migrating from `windows_bindgen` ro `riddle` should hopefully not be difficult.

NobodyXu:
@complexspaces I've "vendored" windows-sys in #837, could you test the compile-time again please and check if there's still a regression?

complexspaces:
@NobodyXu I gave the `main` branch with your changes a try and the compilation times seem to be back roughly where they were for `1.0.79`. Thank you for doing that so quickly.

NobodyXu (approved):
Can we have an actual release with #840 merged?

thomcc:
Yeah. I think I can actually do the release, I just need someone to approve this PR and #840 (due to the branch protection I can't merge without it).

workingjubilee (approved):

thomcc:
Thanks! I'll cut the release after work today.

thomcc:
https://github.com/rust-lang/cc-rs/releases/tag/1.0.80

PR #836, opened 2023-07-25
add armv8r-none-eabihf

chrisnc:
Let cc-rs recognize the armv8r-none-eabihf target.

Related to rust-lang/rust#110482.


NobodyXu (approved):
Thanks!

PR #837, opened 2023-07-26
Vendor `windows-sys` crate

NobodyXu:
to avoid additional dependencies bring in by `windows_sys`, espeically to `build-dependencies`.


thomcc (approved):
This is great, thank you.

thomcc:
I guess `const fn transmute` is 1.56? I don't care that much about MSRV so I'd be okay bumping it. That's still very conservative anyway.

gen-windows-sys-binding/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+const PRELUDE: &str = r#"// This file is autogenerated.
+//
+// To add bindings, edit windows_sys.lst then run:
+//
+// ```
+// cd generate-windows-sys/
+// cargo run
+// ```
+//
+// ignore-tidy-filelength
```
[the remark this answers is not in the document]
NobodyXu: Thanks, I've remove this.

NobodyXu:
> I guess `const fn transmute` is 1.56? I don't care that much about MSRV so I'd be okay bumping it. That's still very conservative anyway.

I've replaced use of it with `addr as *mut T`, it will still work just a bit more derived from current std implementation, though it doesn't make much of a difference until std implements `strict_provenance`.

gen-windows-sys-binding/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+// ```
+// cd generate-windows-sys/
+// cargo run
+// ```
+"#;
+
+const POSTLUDE: &str = r#"
+/// Adapted from https://doc.rust-lang.org/src/core/ptr/mod.rs.html#600-607
+const fn invalid_mut<T>(addr: usize) -> *mut T {
+    addr as *mut T
```
thomcc, line 23: A comment that this would be `transmute()` except for MSRV requirement might be good.

NobodyXu:
> It's not really vendoring so much as using their windows-bindgen tool...

Well, anyway we are able to replace the massive `windows-sys` with 188 lines of rust.

gen-windows-sys-binding/src/main.rs
NobodyXu: Thanks!
I've added comment about MSRV and the strict_provenance.

thomcc:
This looks great. Thanks for all your work.

Issue #772, opened 2023-01-20
Fails to compile with nvcc when passing RUSTFLAGS='-C target-feature=+crt-static'


solkitten:
When I'm trying to make a static build of my rust application with some cuda code doing:
`RUSTFLAGS='-C target-feature=+crt-static' cargo build --release`
I'm getting the following error under linux:

```
 cargo:warning=nvcc fatal   : Unknown option '-static'
  exit status: 1

```

Tested with CUDA 12

dot-asm:
While I know how to fix it, one can wonder what's up the `-static` flag to start with. Indeed, cc-rs produces .a files for rustc to link with, while from C compiler's view the`-static` flag is a **link** flag. In other words at no point `-static` is actually meaningful in the context, because cc-rs never links anything. In yet another words, I'd say that cc-rs shouldn't issue the flag in question. As opposed to fixing it on the nvcc side.

dot-asm:
On a tangential note. As for "cc-rs never links anything." One can also argue that cc-rs shouldn't issue `-shared` flag either. For the same reason, it's a **link** flag to C compiler. Well, documentation suggests that cc-rs can produce .so, but it never worked (see for example #594). And even if it's implemented, it's not really an excuse to pass the flag at the `-c` stage. I mean shared support would be about adding an explicit link-shared-library step. Or replacing the `ar` step with it? Either way, that's where the -shared flag would be used, not with `cc -c`...

dot-asm:
Just in case. Tracing the introduction of corresponding methods back suggests that it was triggered by a linking error suggesting to recompile with `-fPIC`. So one would expect `-static/-shared` to affect the compiler's command line, but I don't see that it does. In sense that `cc -shared -c ...` doesn't add -fPIC to compiler command line... Or vice versa... The original dispute was about musl target, but the same can be said about `musl-gcc`... As well as about `cc` on an all-musl system... Was it the case at some point in time? Well, not even gcc-2.95.3 adds `-fPIC` if I pass `-shared`... In all cases `-static/-shared` appeared as pure link flags, i.e. how to link things.

dot-asm:
[...]

PR #838, opened 2023-07-26
No-op shared_flag() and static_flag() methods.

dot-asm:
The compiler flags enabled by the methods in question are link flags and are never actually used in the cc-rs context. The shared_flag() is documented as reserved for future support for linking shared libraries, and static_flag() is effectively deprecated.

This is with reference to https://github.com/rust-lang/cc-rs/issues/772#issuecomment-1651365558 and forward.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -597,8 +597,8 @@ impl Build {
 
     /// Set the `-shared` flag.
     ///
-    /// When enabled, the compiler will produce a shared object which can
-    /// then be linked with other objects to form an executable.
+    /// The method is reserved for linking a shared library, currently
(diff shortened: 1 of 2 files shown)


NobodyXu:
@dot-asm A lot of your PRs were closed in batch, is it an accident?

dot-asm:
No, it's a conscious choice. Just in case, it's not **a** random "lot," but **all** open ones. I've also removed myself from the watchers' list.

NobodyXu:
@dot-asm Is it because members of cc-rs is too slow to review and respond to your PR and some of them has been open for a long time?

PR #839, opened 2023-08-01
fix(flag_check): never link to avoid false positives

RubixDev:
Using `flag_if_supported` when targeting `wasm32-unknown-unknown` often (or maybe always) results in false negatives, causing supported flags to not be applied. This happens, because `clang` is trying to link the binary with `wasm-ld` which must be installed, and even with `wasm-ld` installed `clang` still gives me this error:

```
wasm-ld: error: unknown file type: /lib/crt1.o
wasm-ld: error: cannot open /usr/lib/clang/15.0.7/lib/libclang_rt.builtins-wasm32.a: No such file or directory
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
```

This is easily avoided by telling clang to skip linking by using `-c`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -535,11 +535,8 @@ impl Build {
             is_arm,
         );
 
-        // We need to explicitly tell msvc not to link and create an exe
-        // in the root directory of the crate
-        if target.contains("msvc") && !self.cuda {
-            cmd.arg("-c");
-        }
+        // Checking for compiler flags does not require linking
+        cmd.arg("-c");
 
         cmd.arg(&src);
 


src/lib.rs
```diff
@@ -537,7 +537,9 @@ impl Build {
 
         // We need to explicitly tell msvc not to link and create an exe
         // in the root directory of the crate
-        if target.contains("msvc") && !self.cuda {
+        // Additionally, checks for wasm32-unknown-unknown can easily falsely
+        // fail when trying to link
+        if (target.contains("msvc") && !self.cuda) || &*target == "wasm32-unknown-unknown" {
```
dot-asm, line 542: Formally speaking one can make a case for adding `-c` **unconditionally**, i.e. remove the if altogether. This is because cc-rs doesn't do linking and always uses `-c` when invoking the compiler.

dot-asm: If removing `if` feels inappropriate, consider the following. It's commonplace that clang on people's paths accepts cross-compile-y `--target` without having matching libraries or headers. And the thing is that the detection will fail in all these cases for the exact reason in the description. In other words the problem is not actually wasm32-unknown-unknown-specific, and it's more than appropriate to generalize it. Removing `if` is the ultimate way to generalize it:-)

RubixDev: Good point, I just wasn't sure whether for some reason it might have to run without `-c` if cuda is enabled, or why else would that be checked explicitly? If that's not a problem though I would be happy to change the PR to add the `-c` flag unconditionally.

dot-asm: cc-rs **always** calls the compiler with -c when actually compiling things, cuda or not, it never does any linking(*). In other words, if there is a question whether or not a particular flag is supported, it's a question about the compile-time flag, never link-time. Hence no real need for a link attempt.

(*) Well, cuda does require a device-side linking step, but it falls beyond the context in question, whether or not a general compiler flag is supported. Cuda users are unlikely to be unsure about their compiler flags, and hence won't generally invoke the method in question. If anything the test in cuda context would look a little bit different anyway.  [For reference, yes, I am a cuda user myself.]

RubixDev: Okay good, I've changed it to always include `-c`

thomcc (approved):
Doing this unconditionally does worry me. I do see @dot-asm's arguments for why it's correct, but it is still somewhat worrisome.

I think I'll take this PR, but I wouldn't be surprised if we end up needing to make this conditional again due to issues in the release.

PR #848, opened 2023-08-04
Use rustc's Apple deployment target defaults when available

BlackHoleFox:
To follow on https://github.com/rust-lang/rust/pull/105354 being released in 1.71, and soon https://github.com/rust-lang/rust/pull/104385, the changeset here aims to make Apple deployment targets more consistent and support future minimum bumps better. To do so, the root change is to try asking `rustc` for what it thinks the default deployment target version is before using a hardcoded one after `cc` checks the environment directly for an overridden version. For any users who aren't on 1.71+, everything will work exactly the same.

While in the area, duplication of checking the iOS and watchOS default deployment target elsewhere in the library was removed and a bad watchOS version (2.0 -> 5.0) was corrected. 

Also, this branch forks from #708 and has a rebased version of its one commit. If only GitHub supported stacked PRs... Due to that though, its better to review this per commit.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2367,31 +2367,16 @@ impl Build {
             Catalyst(&'static str),
         }
 
-        enum Os {
-            MacOs,
-            Ios,
-            WatchOs,
-        }
-        impl std::fmt::Debug for Os {
-            fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-                match self {
-                    Os::MacOs => f.write_str("macOS"),
-                    Os::Ios => f.write_str("iOS"),
-                    Os::WatchOs => f.write_str("WatchOS"),
-                }
-            }
-        }
-
         let target = self.get_target()?;
         let os = if target.contains("-darwin") {
-            Os::MacOs
+            AppleOs::MacOs
         } else if target.contains("-watchos") {
-            Os::WatchOs
+            AppleOs::WatchOs
         } else {
-            Os::Ios
+            AppleOs::Ios
         };
         let is_mac = match os {
-            Os::MacOs => true,
+            AppleOs::MacOs => true,
             _ => false,
         };
 
@@ -2465,29 +2450,11 @@ impl Build {
             }
         };
 
-        let (sdk_prefix, sim_prefix, min_version) = match os {
-            Os::MacOs => (
-                "macosx",
-                "",
-                std::env::var("MACOSX_DEPLOYMENT_TARGET").unwrap_or_else(|_| {
-                    (if arch_str == "aarch64" {
-                        "11.0"
-                    } else {
-                        "10.7"
-                    })
-                    .into()
-                }),
-            ),
-            Os::Ios => (
-                "iphone",
-                "ios-",
-                std::env::var("IPHONEOS_DEPLOYMENT_TARGET").unwrap_or_else(|_| "7.0".into()),
-            ),
-            Os::WatchOs => (
-                "watch",
-                "watch",
-                std::env::var("WATCHOS_DEPLOYMENT_TARGET").unwrap_or_else(|_| "2.0".into()),
-            ),
+        let min_version = self.apple_deployment_version(os, &target, arch_str);
+        let (sdk_prefix, sim_prefix) = match os {
+            AppleOs::MacOs => ("macosx", ""),
+            AppleOs::Ios => ("iphone", "ios-"),
+            AppleOs::WatchOs => ("watch", "watch"),
         };
 
         let sdk = match arch {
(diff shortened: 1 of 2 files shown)


tests/test.rs
```diff
@@ -459,3 +459,19 @@ fn asm_flags() {
     test.cmd(1).must_have("--abc");
     test.cmd(2).must_have("--abc");
 }
+
+#[test]
+fn gnu_apple_darwin() {
+    for (arch, version) in &[("x86_64", "10.7"), ("aarch64", "11.0")] {
```
thomcc, line 465: A comment that this is expected to break after https://github.com/rust-lang/rust/pull/104385 would be appreciated.

src/lib.rs
```diff
@@ -3638,6 +3703,22 @@ fn command_add_output_file(
     }
 }
 
+#[derive(Clone, Copy)]
+enum AppleOs {
+    MacOs,
+    Ios,
+    WatchOs,
+}
+impl Display for AppleOs {
```
thomcc, line 3712: This looks like it's only used in error messages, no? Assuming that's right, can it be a `Debug` impl to hint that it's not semantically important?

src/lib.rs
```diff
@@ -2290,8 +2283,19 @@ impl Build {
             None => false,
         };
 
-        let arch = if is_catalyst {
-            match arch {
+        let arch = if is_mac {
+            match arch_str {
+                "i686" => ArchSpec::Device("-m32"),
+                "x86_64" | "aarch64" => ArchSpec::Device("-m64"),
```
thomcc, line 2289: Needs to handle `x86_64h` here (and look for other places where `arch_str` should handle it).

src/lib.rs
```diff
@@ -3285,6 +3290,66 @@ impl Build {
...
+    fn apple_deployment_version(
+        &self,
+        os: AppleOs,
+        target: &str,
+        arch_str: Option<&str>,
+    ) -> String {
+        fn rustc_provided_target(rustc: Option<&str>, target: &str) -> Option<String> {
+            let rustc = rustc?;
+            let output = Command::new(rustc)
+                .stderr(Stdio::piped())
```
thomcc, line 3302: stdout/stderr are automatically set to piped when you use `.output()` so there's no need to specify them.

src/lib.rs
```diff
@@ -3285,6 +3290,66 @@ impl Build {
...
+                .stdout(Stdio::piped())
+                .args(&["--target", target])
+                .args(&["--print", "deployment-target"])
+                .output()
+                .ok()?;
+
+            if output.status.success() {
+                std::str::from_utf8(&output.stdout)
+                    .unwrap()
+                    .split('=')
```
thomcc, line 3312: I would prefer this use `.strip_prefix("deployment_target=")`, to ensure we notice if the output format changes.

src/lib.rs
```diff
@@ -3285,6 +3290,66 @@ impl Build {
...
+        let rustc = self.getenv("RUSTC");
+        let rustc = rustc.as_deref();
+        // note the hardcoded minimums here are subject to change in a future compiler release,
+        // so the ones rustc knows about are preferred. For any compiler version that has bumped them
+        // though, `--print deployment-target` will be present and the fallbacks won't be used.
+        //
+        // the ordering of env -> rustc -> old defaults is intentional for performance when using
+        // an explicit target
+        match os {
+            AppleOs::MacOs => env::var("MACOSX_DEPLOYMENT_TARGET")
```
thomcc, line 3331: Hm... should we emit rerun-if-changed flags for these env vars in any situation? (Perhaps it would be needed in complex situations with cross-compiling?)

tests/test.rs
BlackHoleFox: I found `__set_env` and used that to override the version used, so this now should be immune to future minimum support bumps. Does that work instead?

src/lib.rs
BlackHoleFox: Swapped it out.

src/lib.rs
BlackHoleFox: Should be done now, it seems like the original PR predated that target being supported.

src/lib.rs
BlackHoleFox: Originally I opted to avoid that (a prior revision of the branch actually ran it through the caching env paths) for consistency reasons. Only `cc` would rerun, so an arbitrary number of objects not created by `cc` in the build tree would have a different deployment target and that seems worse then doing nothing and making people run `cargo clean` to switch.

I think that if this is desired behavior, `rustc` or `cargo` should handle it instead and invalidate the entire build tree at once.

youknowone (approved):
This is a really good idea.

thomcc (approved):

nox:
Prior to this change, the cc crate just didn't emit any `-mmacosx-version-min` flag on its own, right? That's a breaking change on its own, and I don't think #901 changes that.

PR #851, opened 2023-08-09
fix: Set environment variables in `Tool` from `try_get_compiler`

oblique:
[`get_compiler`](https://docs.rs/cc/latest/cc/struct.Build.html#method.get_compiler) wasn't forwarding the environment variables modifications to `Tool`, however these are very useful configurations, especially under MacOS.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -197,6 +197,7 @@ pub struct Tool {
     cc_wrapper_args: Vec<OsString>,
     args: Vec<OsString>,
     env: Vec<(OsString, OsString)>,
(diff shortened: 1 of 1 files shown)


oblique:
@Amanieu Sorry for pinging you, I didn't know who else to ping. What is you opinion/review on this?

youknowone:
Is this related to lost `-isysroot` problem with AppleClang?

oblique:
@youknowone I don't know about the problem. Do you have a link to it? If the problem is when using `Tool`, then maybe it is related.

PR #854, opened 2023-08-12
fix: fix the usage of include directories for the Rust analyzer

aminya:


ChrisDenton (approved):
Sure, this seems slightly clearer.

PR #857, opened 2023-08-17
Separate source args with -- when compiling assembly with clang-cl

roblabla:
Currently, when compiling with clang-cl, we separate the source from the arguments by inserting a `--`. This avoids compilation errors due to conflicts between source paths and arguments (especially on macos, where path will often begin with /Users, but /U is a clang-cl flag).

However, this separator is disabled when we are compiling assembly for msvc targets, as those will usually be compiled through `ml.exe`, which doesn't understand this separator. Problem is, this is not always the case: Only `.asm` files are compiled using `ml.exe`. Any asm using a different extension will still go through the normal C compiler (clang-cl), and hence we need to provide the separator in that case.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1479,7 +1479,8 @@ impl Build {
         let clang = compiler.family == ToolFamily::Clang;
         let gnu = compiler.family == ToolFamily::Gnu;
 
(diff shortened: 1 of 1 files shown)


thomcc (approved):
Yeah, I see. I wish we had tests for this but so it goes.

PR #866, opened 2023-09-20
Mirror clang's link stage on Windows.

dot-asm:
Fixes problem discussed in #811.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1171,6 +1171,11 @@ impl Build {
                     atlmfc_lib.display()
                 ));
             }
+
+            if compiler.family == ToolFamily::Clang {
+                // Mirror what clang does at link stage
+                self.print(&format_args!("cargo:rustc-link-arg=-defaultlib:oldnames"));
+            }
         }
 
         if self.link_lib_modifiers.is_empty() {
@@ -1710,13 +1715,6 @@ impl Build {
                     cmd.push_opt_unless_duplicate("-Os".into());
                 } else {
                     cmd.push_opt_unless_duplicate(format!("-O{}", opt_level).into());
-                }
-
-                if cmd.family == ToolFamily::Clang && target.contains("windows") {
-                    // Disambiguate mingw and msvc on Windows. Problem is that
-                    // depending on the origin clang can default to a mismatchig
-                    // run-time.
-                    cmd.push_cc_arg(format!("--target={}", target).into());
                 }
 
                 if cmd.family == ToolFamily::Clang && target.contains("android") {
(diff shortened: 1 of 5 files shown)


ChrisDenton:
Hm, unsure about the implementation of this. Since we're forcing this library to be included, we need to be careful that it doesn't override anything the user has implemented themselves. clang does this using `-defaultlib:oldnames.lib` to give the library lower priority than non-default libs.

dot-asm:
`cargo:rustc-link-arg=-defaultib:oldnames` then. Deeper dive into `-vvv` logs for a multi-file project revealed that it's also a wrong spot for the link arguments. [As well as --target duplication thing.]

src/lib.rs
```diff
@@ -1172,6 +1172,11 @@ impl Build {
                     atlmfc_lib.display()
                 ));
             }
+
+            if compiler.family == ToolFamily::Clang {
+                // Mirror what clang does at link stage
+                self.print(&format_args!("cargo:rustc-link-arg=-defaultib:oldnames"));
```
ChrisDenton, line 1178: Typo
```suggestion
                self.print(&format_args!("cargo:rustc-link-arg=-defaultlib:oldnames"));
```

ChrisDenton:
This could really use a test to make sure it works and isn't accidentally broken in the future.

ChrisDenton:
Also cc @thomcc as using `rustc-link-arg` would require bumping the MSRV to >=1.56 (October 2021).

thomcc:
That MSRV bump doesn't bother me, but I would really like someone to double-check this works. My recollection is that there's some issue with using rustc-link-arg like this.

dot-asm:
> Also cc @thomcc as using `rustc-link-arg` would require bumping the MSRV to >=1.56 (October 2021).

Hmm, that would probably be undesirable... Two points in the context. One should keep in mind that the suggested modification affects only "fringe" use cases, an MSVC target with `$CC` set to `clang`, as opposed to `clang-cl`, which by itself is error-prone in the most general case. With this in mind, secondly, since the directive is ignored by an older version, then it wouldn't pose a problem for users who use the older version.

dot-asm:
Putting together a test case was challenging because one has to mix object files generated by cl and clang in the same executable, with the former smuggling in `oldnames.lib` through the linker directive. `-Zl` prevents it from doing this and thus shifts the burden of linking in its entirety on rustc. You can convince yourself that the test is affirmative by commenting the line that adds `-defaultlib:oldnames` link flag and noting that the cc-test fails to link with "unresolved external symbol strdup" message. Well, this is provided that you have `clang` on the `%PATH%`. Just in case, this is the case on Github Actions, which can be confirmed by noting `test msvcrt_here ... ok` in the log files.

dot-asm:
Re-based just in case.

PR #867, opened 2023-09-20
Don't pass compiler flags to MSVC assembers.

dot-asm:
While ml[64] ignores the compiler flags with warnings, armasm[64] rejects them and fails.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2220,9 +2220,6 @@ impl Build {
         if target.contains("i686") || target.contains("i586") {
             cmd.arg("-safeseh");
         }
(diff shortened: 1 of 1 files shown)


thomcc:
Doesn't that happen later on? Down here? https://github.com/rust-lang/cc-rs/blob/main/src/lib.rs#L1545

dot-asm:
> Doesn't that happen later on? Down here? https://github.com/rust-lang/cc-rs/blob/main/src/lib.rs#L1545

Yes, I've misread. Thanks! Fixed now. The commit message is reworded.

thomcc (approved):
Thanks.

PR #868, opened 2023-09-21
Pass self.definitions to armasm[64].exe.

dot-asm:


thomcc:
Do you have a link to docs for this?

dot-asm:
https://developer.arm.com/documentation/dui0802/latest/

dot-asm:
Well, the ARM doc talks about `--predefine` but for some reason so-far-case-insensitive MS decided that they accept specifically `--PreDefine`. This is in accordance with `armasm[64] -h`.

dot-asm:
Just in case, this is less important than #867.

dot-asm:
It occurred to me that you might have meant to suggest to add a commentary with the reference to the docs. Is this it?

thomcc:
No, I wanted the docs:

https://developer.arm.com/documentation/dui0802/b/Directives-Reference/SETA--SETL--and-SETS?lang=en
and
https://developer.arm.com/documentation/dui0802/b/armasm-Command-line-Options/--predefine

were what I was looking for. Thanks. I don't think a link in the code is needed -- my experience with ARM's docs are that the links don't tend to be that stable, so it'd probably bitrot.

thomcc (approved):
LGTM

PR #869, opened 2023-09-23
Be more specific about not passing -c to armasm[64].exe.

dot-asm:
4ce411709170197228d353aa2bbf385efe3b958a passes non-.asm files to the C compiler driver, which unlike armasm[64].exe does require -c flag.


thomcc (approved):
Thanks

PR #871, opened 2023-09-23
separated section for gcc or clang-only flags

youknowone:
No intention to change behavior.

whitespace-less diff: https://github.com/rust-lang/cc-rs/pull/871/files?w=1


src/lib.rs
```diff
@@ -1811,16 +1817,6 @@ impl Build {
             family.add_force_frame_pointer(cmd);
         }
 
-        if !cmd.is_like_msvc() {
```
youknowone, line 1814: is this flag order-sensitive?

PR #872, opened 2023-09-23
minimal maxosx deployment target is 10.9 when cpp

youknowone:
It was the cause of gnu_flag_if_supported_cpp failure on macOS


thomcc:
Ah, thank you!

thomcc (approved):

PR #878, opened 2023-09-30
feat: Use channels to maintain job tokens & reuse the implicit token without dropping it first

### LINKED ISSUE #858: Issues with releasing and reacquiring implicit jobserver token
I was trying to optimize build times and found that -sys crate dependencies seemed to be taking a long time according to the `cargo build --timings` output. Further investigation showed that these dependencies didn't take as long to build in isolation (with concurrency controlled for with `-j1`), which seemed odd.

More investigation led to https://github.com/rust-lang/cargo/issues/7689 and https://github.com/rust-lang/cargo/pull/7344 being the culprits.

Long story short, I think that releasing and reacquiring the implicit jobserver token causes the `cargo build --timings` output to be misleading.

Take this example:

```
cargo new example
cd example
cargo add bzip2-sys nng-sys zstd-sys 
cargo build -j1 --timings
```

And the resulting graph:

Screenshot from 2023-08-19 15-58-31

If I build each of the -sys dependencies in isolation, bzip2-sys takes ~0.5s, nng-sys takes ~14s, and zstd-sys takes ~15s. Note the inconsistency with the graph.

What appears to be happening, roughly, in this case is:

1. bzip2-sys build script starts
1. cc for bzip2-sys releases its implicit token before its processing loop
1. That allows zstd-sys build script to start
1. cc for zstd-sys releases its implicit token before its processing loop
1. That allows nng-sys build script to start
1. nng-sys build script runs to completion in ~14s
1. nng-sys starts and runs to completion
1. zstd-sys build script resumes and runs for ~2.5s
1. cc for zstd-sys releases a token
1. That allows bzip2-sys build script to resume and run to completion in ~0.5s
1. bzip2-sys starts and runs to completion
1. zstd-sys build script resumes and runs to completion in ~12.5s
1. zstd-sys starts and runs to completion

If you assume this is more or less what happened, the bars in the graph make sense and the timings are consistent with how long the -sys crates take to build in isolation.

Also, I don't know how cargo schedules dependency building, but I'm wondering if this issue could cause increases in compile times by delaying builds of dependents in an unexpected way. If cargo specifically chooses to compile one dependency earlier to unlock its ability to start compiling dependents, presumably with the goal of reducing overall build time, then I think this choice might be undermined by the behavior seen here, since we see other crates sort of cut in line.

I'm also wondering if, once this is fixed, the change introduced by https://github.com/rust-lang/cargo/pull/7344 should be reverted or modified somehow, or if doing so would cause problems with older versions of cc on older versions of cargo.

osiewicz:
This is a follow-up to #779 and addressing #858. This approach uses an additional thread for polling the jobserver. 
Per my local tests, it fixes #858

### DIFF
--- a/src/job_token.rs
+++ b/src/job_token.rs
@@ -1,4 +1,4 @@
-use jobserver::{Acquired, HelperThread};
+use jobserver::{Acquired, Client, HelperThread};
 use std::{
     env,
     sync::{
@@ -40,15 +40,17 @@ impl JobToken {
 /// gives out tokens without exposing whether they're implicit tokens or tokens from jobserver.
 /// Furthermore, instead of giving up job tokens, it keeps them around
 /// for reuse if we know we're going to request another token after freeing the current one.
-pub(crate) struct GlobalJobTokenServer {
+pub(crate) struct JobTokenServer {
     helper: HelperThread,
     tx: Sender<Option<Acquired>>,
     rx: Receiver<Option<Acquired>>,
 }
 
 impl JobTokenServer {
-    pub(crate) fn new() -> Result<Self, crate::Error> {
-        let client = jobserver();
+    pub(crate) fn new() -> &'static Self {
+        jobserver()
+    }
+    fn new_inner(client: Client) -> Result<Self, crate::Error> {
         let (tx, rx) = mpsc::channel();
         // Push the implicit token. Since JobTokens only give back what they got,
         // there should be at most one global implicit token in the wild.
@@ -76,11 +78,16 @@ impl JobTokenServer {
     }
 }
 
-/// Returns a suitable `jobserver::Client` used to coordinate
-/// parallelism between build scripts.
-fn jobserver() -> jobserver::Client {
+/// Returns a suitable `JobTokenServer` used to coordinate
+/// parallelism between build scripts. A global `JobTokenServer` is used as this ensures
+/// that only one implicit job token is used in the wild.
+/// Having multiple separate job token servers would lead to each of them assuming that they have control
+/// over the implicit job token.
+/// As it stands, each caller of `jobserver` can receive an implicit job token and there will be at most
+/// one implicit job token in the wild.
+fn jobserver() -> &'static JobTokenServer {
     static INIT: Once = Once::new();
-    static mut JOBSERVER: Option<jobserver::Client> = None;
+    static mut JOBSERVER: Option<JobTokenServer> = None;
 
     fn _assert_sync<T: Sync>() {}
     _assert_sync::<jobserver::Client>();
@@ -88,9 +95,10 @@ fn jobserver() -> jobserver::Client {
     unsafe {
         INIT.call_once(|| {
             let server = default_jobserver();
-            JOBSERVER = Some(server);
+            JOBSERVER =
+                Some(JobTokenServer::new_inner(server).expect("Job server initialization failed"));
         });
-        JOBSERVER.clone().unwrap()
+        JOBSERVER.as_ref().unwrap()
     }
 }
 
--- a/src/job_token.rs
+++ b/src/job_token.rs
@@ -39,7 +39,8 @@ pub(crate) struct JobTokenServer {
 impl JobTokenServer {
     pub(crate) fn new(client: Client) -> Result<Self, crate::Error> {
         let (tx, rx) = mpsc::channel();
-        // Initialize the
+        // Push the implicit token. Since JobTokens only give back what they got,
+        // there should be at most one global implicit token in the wild.
         tx.send(None).unwrap();
         let pool = tx.clone();
         let helper = client.into_helper_thread(move |acq| {
--- a/src/job_token.rs
+++ b/src/job_token.rs
@@ -14,13 +14,14 @@ pub(crate) struct JobToken {
     token: Option<Acquired>,
     /// A pool to which `token` should be returned. `pool` is optional, as one might want to release a token straight away instead
     /// of storing it back in the pool - see [`Self::forget()`] function for that.
-    pool: Option<Sender<Option<Acquired>>>,
+    pool: Option<Sender<Option<Result<Acquired, crate::Error>>>>,
 }
 
 impl Drop for JobToken {
     fn drop(&mut self) {
         if let Some(pool) = &self.pool {
-            let _ = pool.send(self.token.take());
+            // Always send back an Ok() variant as we know that the acquisition for this token has succeeded.
+            let _ = pool.send(self.token.take().map(|token| Ok(token)));
         }
     }
 }
@@ -43,8 +44,8 @@ impl JobToken {
 /// for reuse if we know we're going to request another token after freeing the current one.
 pub(crate) struct JobTokenServer {
     helper: HelperThread,
-    tx: Sender<Option<Acquired>>,
-    rx: Receiver<Option<Acquired>>,
+    tx: Sender<Option<Result<Acquired, crate::Error>>>,
+    rx: Receiver<Option<Result<Acquired, crate::Error>>>,
 }
 
 impl JobTokenServer {
@@ -58,12 +59,12 @@ impl JobTokenServer {
         tx.send(None).unwrap();
         let pool = tx.clone();
         let helper = client.into_helper_thread(move |acq| {
-            let _ = pool.send(Some(acq.unwrap()));
+            let _ = pool.send(Some(acq.map_err(|e| e.into())));
         })?;
         Ok(Self { helper, tx, rx })
     }
 
-    pub(crate) fn acquire(&self) -> JobToken {
+    pub(crate) fn acquire(&self) -> Result<JobToken, crate::Error> {
         let token = if let Ok(token) = self.rx.try_recv() {
             // Opportunistically check if there's a token that can be reused.
             token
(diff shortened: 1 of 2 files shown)


osiewicz:
I am wondering if we could also attempt removal of the current `wait_thread` in this code, as per #829  it is there mostly to avoid a deadlock

src/job_token.rs
```diff
@@ -0,0 +1,70 @@
+use jobserver::{Acquired, Client, HelperThread};
+use std::sync::mpsc::{self, Receiver, Sender};
+
+pub(crate) struct JobToken {
+    /// The token can either be a fresh token obtained from the jobserver or - if `token` is None - an implicit token for this process.
+    /// Both are valid values to put into queue.
+    token: Option<Acquired>,
+    pool: Sender<Option<Acquired>>,
```
NobodyXu, line 8: You could make this an `Option` and remove `should_return_to_queue`:

```suggestion
    pool: Option<Sender<Option<Acquired>>>,
```

If `pool` is `Some(..)`, then send the `token` back to the `pool`.

src/lib.rs
```diff
@@ -1432,7 +1442,7 @@ impl Build {
 
         /// Returns a suitable `jobserver::Client` used to coordinate
         /// parallelism between build scripts.
-        fn jobserver() -> &'static jobserver::Client {
+        fn jobserver() -> jobserver::Client {
```
NobodyXu, line 1445: I'd recommend moving this into mod `job_token`

src/job_token.rs
```diff
@@ -0,0 +1,70 @@
...
+        Ok(Self { helper, tx, rx })
+    }
+
+    pub(crate) fn acquire(&mut self) -> JobToken {
+        let token = if let Ok(token) = self.rx.try_recv() {
+            // Opportunistically check if there's a token that can be reused.
+            token
+        } else {
+            // Cold path, request a token and block
+            self.helper.request_token();
```
NobodyXu, line 61: Do we really need another help thread?
Can't we just acquire the token directly here?

NobodyXu (changes requested):

NobodyXu:
> I am wondering if we could also attempt removal of the current `wait_thread` in this code, as per #829 it is there mostly to avoid a deadlock with `acquire` - which happens in yet another separate thread anyways as of this PR.

`JobTokenServer::acquire` from this PR still blocks, so it would still be required.
However I don't think it's necessary for `JobTokenServer` to spawn a help thread, that's definitely not necessary.

osiewicz:
Eh, I've flirted with the idea of `Arc<AtomicBool>`, though I like a channel approach a bit more, as it makes obtaining a token opaque - you don't care whether the token is owned by the current process or if it was obtained from the jobserver. It also allows for token reuse.
I do agree though that additional helper thread is probably not necessary, so I'll get rid of that.

osiewicz:
Eh, I've tried to get rid of helperthread, leading to a deadlock of a build on one of my test projects. I believe the issue here is similar to what you've ran into in https://github.com/rust-lang/cc-rs/pull/779#discussion_r1258244080 - we can't just acquire, as that might lead to the deadlock. By having a separate thread and a channel we can request a token like we do with `acquire` - separate thread sounds redundant, right? Well, if `acquire` is never going to return, then a separate thread is blocked, not the main one - and we can grab the implicit token once that becomes available.
Attaching a patch on top of 52afaf1, perhaps I've done something incorrectly.
[no_helperthread.patch](https://github.com/rust-lang/cc-rs/files/12777742/no_helperthread.patch)

src/job_token.rs
```diff
@@ -68,3 +74,48 @@ impl JobTokenServer {
         }
     }
 }
+
+/// Returns a suitable `jobserver::Client` used to coordinate
+/// parallelism between build scripts.
+pub(super) fn jobserver() -> jobserver::Client {
```
NobodyXu, line 80: Why not return a `JobTokenServer` here?

NobodyXu: You can also make this an associated function of `JobTokenServer`

NobodyXu (changes requested):

NobodyXu:
> Well, if `acquire` is never going to return, then a separate thread is blocked, not the main one - and we can grab the implicit token once that becomes available.

I see why it doesn't work: Setting new values on `Arc<AtomicBool>` cannot stop `jobserver::Client` from blocking.

However in the current PR it calls mpsc recv, which will succeed and stop blocking when the implicit token is released.

(It's a shame there isn't built-in async executor in stdlib, otherwise this would be tons easier.)

NobodyXu:
Regarding the `wait_thread`, it can be removed by providing a non-blocking `try_acquire` using `mpsc::Receiver::try_receive`.

Since the spawning thread can consume all tokens (implicit and explicit), `wait_thread` is necessary unless we add `try_acquire`, and switch to waiting on spawned children on failure.

However this sounds so complex that I believe it should be a separate PR.

NobodyXu:
Regading jobserver, I think that maybe it's a good idea to have a vendored implementation that never blocks.

`cc` only needs to:
 - open a jobserver
 - create a jobserver
 - try acquire a token
 - release a token

src/job_token.rs
osiewicz: To be fair this code was already there, so I didn't touch it much - for a second I've removed it altogether and created the jobserver from scratch, but reverted that since. Moving this into `JobTokenServer` is a small change though, let's do it. :)

osiewicz: Come to think of it you're right. We should have a global JobTokenServer, as otherwise if someone initializes two `JobTokenServer`s, they'll both think that they have a implicit job token. The channel should be shared between different JobTokenServer instances.

src/job_token.rs
```diff
@@ -0,0 +1,130 @@
...
+        jobserver()
+    }
+    fn new_inner(client: Client) -> Result<Self, crate::Error> {
+        let (tx, rx) = mpsc::channel();
+        // Push the implicit token. Since JobTokens only give back what they got,
+        // there should be at most one global implicit token in the wild.
+        tx.send(None).unwrap();
+        let pool = tx.clone();
+        let helper = client.into_helper_thread(move |acq| {
+            let _ = pool.send(Some(acq.unwrap()));
```
NobodyXu, line 60: I'm a bit worried about panicking in another thread.

It won't even be reported and might just cause cc to stuck forever.

src/job_token.rs
```diff
@@ -0,0 +1,130 @@
...
+/// Returns a suitable `JobTokenServer` used to coordinate
+/// parallelism between build scripts. A global `JobTokenServer` is used as this ensures
+/// that only one implicit job token is used in the wild.
+/// Having multiple separate job token servers would lead to each of them assuming that they have control
+/// over the implicit job token.
+/// As it stands, each caller of `jobserver` can receive an implicit job token and there will be at most
+/// one implicit job token in the wild.
+fn jobserver() -> &'static JobTokenServer {
+    static INIT: Once = Once::new();
+    static mut JOBSERVER: Option<JobTokenServer> = None;
```
NobodyXu, line 90: I have always wanted to change this t
```suggestion
    static mut JOBSERVER: std::mem::MaybeUninit<JobTokenServer> = MaybeUninit::uninit();
```

I think this can be done in this PR.

NobodyXu (changes requested):

src/job_token.rs
osiewicz: Panics in build scripts should abort the build process straight away, right? Either way I'm fine with changing that.

src/job_token.rs
osiewicz: Heads up, `assume_init_ref` would bump MSRV to 1.55 per [docs](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.assume_init_ref)

NobodyXu: You can do:

```rust
        unsafe {
            &*self.as_ptr()
        }
```

since `cc` already uses `unsafe` here anyway.

NobodyXu (approved):
LGTM, but you would need a team member's approval to merge

thomcc:
I'll take a look later. I suspect it's mostly fine with @NobodyXu's approval.

NobodyXu (approved):

NobodyXu:
Thank you @osiewicz !

P.S. I was just granted the permission to approve and merge PRs in this repository.

osiewicz:
Thanks!

NobodyXu:
I just discovered a problem with this PR: The mpsc in the jobserver will never be dropped so the token will never be put back to jobserver.

I'm already working on a new PR which dramatically improves the parallel compilation and it will be in ready in a few days.

osiewicz:
Uh oh, that's no good.
Off the top of my head though, does the implicit token need to be returned? We never explicitly request it so it's effectively managed by a caller of cc crate. For the acquired tokens, we should be returning them to the pool by the impl of JobToken  for which `.forget` was called.

That's just off the top of my head though and I'm writing that reply at 3AM, so you know, I am probably wrong. :x

NobodyXu:
> Off the top of my head though, does the implicit token need to be returned?

No

> For the acquired tokens, we should be returning them to the pool by the impl of JobToken  for which `.forget` was called.

But for tokens which are still in the mpsc, they will never be returned to the jobserver.

NobodyXu:
I've opened #889 which fixes this and completely removes usage of thread, please have a look/review if you have time.

PR #704, opened 2022-07-27
Add AppleTVOS support

youknowone:

PR #880, opened 2023-10-03
fix: Add apple tvos support

lcruz99:
This PR allows using cc-rs lib to target [tier 3 *-apple builds](https://doc.rust-lang.org/nightly/rustc/platform-support/apple-tvos.html):
- Apple tvOS on aarch64
- Apple tvOS Simulator on x86_64

As clang should be used with this target, I've removed tests added on https://github.com/rust-lang/cc-rs/pull/704 that require gcc.


lcruz99:
CLOSED: wrong base branch

PR #881, opened 2023-10-03
fix: Add apple tvos support

lcruz99:
This PR allows using cc-rs lib to target [tier 3 *-apple builds](https://doc.rust-lang.org/nightly/rustc/platform-support/apple-tvos.html):

    Apple tvOS on aarch64
    Apple tvOS Simulator on x86_64

~~As clang should be used with this target, I've removed tests added on https://github.com/rust-lang/cc-rs/pull/704 that require gcc.~~ Restored tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,6 +92,24 @@ jobs:
     - run: cargo test ${{ matrix.no_run }} --manifest-path cc-test/Cargo.toml --target ${{ matrix.target }} --features parallel
     - run: cargo test ${{ matrix.no_run }} --manifest-path cc-test/Cargo.toml --target ${{ matrix.target }} --release
 
+  check-tvos:
+    name: Test aarch64-apple-tvos
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install Rust (rustup)
+      run: |
+        set -euxo pipefail
+        rustup toolchain install nightly --no-self-update --profile minimal
+        rustup component add rust-src --toolchain nightly
+        rustup default nightly
+      shell: bash
+    - run: cargo test -Z build-std=std --no-run --target aarch64-apple-tvos
+    - run: cargo test -Z build-std=std --no-run --features parallel --target aarch64-apple-tvos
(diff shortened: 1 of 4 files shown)


tests/test.rs
```diff
@@ -507,33 +507,3 @@ fn gnu_apple_darwin() {
         cmd.must_not_have("-isysroot");
     }
 }
-
```
thomcc, line 510: These tests didn't require GCC, just check that the target-specific stuff works. They shouldn't be deleted.

There should also be a test that clang works and has everything set appropriately.

thomcc (changes requested):
Forgot to mark this as needing changes.

.github/workflows/main.yml
```diff
@@ -38,6 +38,11 @@ jobs:
             rust: stable
             target: aarch64-apple-ios
             no_run: --no-run
+          - build: aarch64-tvos
+            os: macos-latest
+            rust: nightly
+            target: aarch64-apple-tvos
```
NobodyXu, line 44: There's no pre-built libstd for this target, so rustup failed in CI.

IMO it's better to create a new job since it woukd have to use `-Zbuild-std` to manually compile the libstd.

NobodyXu (changes requested):

lcruz99:
@thomcc @NobodyXu FYI

I've pushed the requested changes, sorry for the late reply :)

.github/workflows/main.yml
```diff
@@ -92,6 +92,24 @@ jobs:
     - run: cargo test ${{ matrix.no_run }} --manifest-path cc-test/Cargo.toml --target ${{ matrix.target }} --features parallel
     - run: cargo test ${{ matrix.no_run }} --manifest-path cc-test/Cargo.toml --target ${{ matrix.target }} --release
 
+  test-tvos:
```
NobodyXu, line 95: I think "check" is a better name than "test" since no test is actually run:

```suggestion
  check-tvos:
```

NobodyXu (changes requested):

NobodyXu (approved):
LGTM

thomcc (approved):
Looks good.

PR #884, opened 2023-10-06
added a function to remove flags

raisfeld-ori:
i have a project that uses cc in it's deep backend, and i need to remove one of the flags added deep inside (which i can't remove). i added this function (it has some shit time complexity, but it's like 2am in my country, so i don't care). so yeah, i made this function to fix it and remove this function

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -431,12 +431,11 @@ impl Build {
     /// removes a compiler flag that was added by [`Build::flag`]
     ///
     /// # Example
-    /// ```no_run
+    /// ```
     /// cc::Build::new()
     ///     .file("src/foo.c")
     ///     .flag("unwanted_flag")
-    ///     .remove_flag("unwanted_flag")
-    ///     .compile("bar");
+    ///     .remove_flag("unwanted_flag");
     /// ```
 
     pub fn remove_flag(&mut self, flag: &str) -> &mut Build {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -440,7 +440,7 @@ impl Build {
     /// ```
 
     pub fn remove_flag(&mut self, flag: &str) -> &mut Build {
-        self.flags.retain(|other_flag| other_flag == &flag.into());
+        self.flags.retain(|other_flag| **other_flag == flag);
         self
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -428,7 +428,7 @@ impl Build {
         self
     }
 
-    /// removes a flag from self.flags
+    /// removes a flag from self.flags that was added by Build.flag()
     ///
     /// # example
     /// ```no_run
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -428,6 +428,24 @@ impl Build {
         self
     }
 
+    /// removes a flag from the existing flags.
+    /// example:
+    /// ```no_run
+    /// c::Build::new()
+    ///     .file("src/foo.c")
+    ///     .flag("-fake_flag")
+    ///     .remove_flag("-fake_flag")
+    ///     .compile("foo");
+    /// ```
+    pub fn remove_flag(&mut self, flag: &'static str) -> &mut Self{
+        let flag: Arc<str> = flag.into();
+        if self.flags.contains(&flag){
+            self.flags.retain(|other_flag| other_flag != &flag);
+        }
+        
+        return self;
+    }
+
     /// Add a flag to the invocation of the ar
     ///
     /// # Example
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -440,7 +440,7 @@ impl Build {
     /// ```
 
     pub fn remove_flag(&mut self, flag: &str) -> &mut Build {
-        self.flags.retain(|other_flag| **other_flag == flag);
+        self.flags.retain(|other_flag| *other_flag == flag);
         self
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -428,22 +428,20 @@ impl Build {
         self
     }
 
-    /// removes a flag from the existing flags.
-    /// example:
+    /// removes a flag from self.flags
+    ///
+    /// # example
     /// ```no_run
     /// cc::Build::new()
     ///     .file("src/foo.c")
-    ///     .flag("-fake_flag")
-    ///     .remove_flag("-fake_flag")
-    ///     .compile("foo");
+    ///     .flag("unwanted_flag")
+    ///     .remove_flag("unwanted_flag")
+    ///     .compile("bar");
     /// ```
-    pub fn remove_flag(&mut self, flag: &str) -> &mut Self{
-        let flag: Arc<str> = flag.into();
-        if self.flags.contains(&flag){
-            self.flags.retain(|other_flag| other_flag != &flag);
-        }
-        
-       self
+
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -428,6 +428,24 @@ impl Build {
...
+    /// removes a flag from the existing flags.
+    /// example:
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     .flag("-fake_flag")
+    ///     .remove_flag("-fake_flag")
+    ///     .compile("foo");
+    /// ```
+    pub fn remove_flag(&mut self, flag: &'static str) -> &mut Self{
```
NobodyXu, line 440: The `flag` doesn't have to be a `'static`

```suggestion
    pub fn remove_flag(&mut self, flag: &str) -> &mut Self{
```

src/lib.rs
```diff
@@ -428,6 +428,24 @@ impl Build {
...
+        let flag: Arc<str> = flag.into();
+        if self.flags.contains(&flag){
+            self.flags.retain(|other_flag| other_flag != &flag);
+        }
```
NobodyXu, line 444: You don't have to convert it into an `Arc` (which creates a heap allocation) and calling `contains` and `retain` causes two iteration to be made:

```suggestion
        self.flags.retain(|other_flag| &**other_flag != flag);
```

src/lib.rs
```diff
@@ -428,6 +428,24 @@ impl Build {
...
+        
+        return self;
```
NobodyXu, line 446:
```suggestion
       self
```

NobodyXu (changes requested):

NobodyXu:
There's also a format issue, please run `cargo fmt` locally and then commit the changes made by it.

I recommend you to setup your IDE/editor to run `cargo-fmt` automatically on save.

src/lib.rs
```diff
@@ -428,6 +428,22 @@ impl Build {
         self
     }
 
+    /// removes a flag from self.flags
```
NobodyXu, line 431:
```suggestion
    /// Removes a flag added by [`Build::flag`].
```

src/lib.rs
```diff
@@ -428,6 +428,22 @@ impl Build {
...
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     .flag("unwanted_flag")
+    ///     .remove_flag("unwanted_flag")
+    ///     .compile("bar");
+    /// ```
+
+    pub fn remove_flag(&mut self, flag: &str) -> &mut Build {
+        self.flags.retain(|other_flag| other_flag == &flag.into());
```
NobodyXu, line 443: Please don't use `flag.into()`, it will convert `&str` to `Arc<str>` and thus creates an unnecessary heap allocation:

```suggestion
        self.flags.retain(|other_flag| &**other_flag == flag);
```

NobodyXu (changes requested):

src/lib.rs
raisfeld-ori: it removes flags directly from the Vec<Arc<&str>> of flags, so no, it doesn't just remove from Build::flag. it can also remove ar_flag, asm_flag, and (i think, haven't tried it) even file

NobodyXu: From [`cc::Build`](https://docs.rs/cc/latest/src/cc/lib.rs.html#95-135):

```rust
pub struct Build {
    flags: Vec<Arc<str>>,
    flags_supported: Vec<Arc<str>>,
    known_flag_support_status: Arc<Mutex<HashMap<String, bool>>>,
    ar_flags: Vec<Arc<str>>,
    asm_flags: Vec<Arc<str>>,
    no_default_flags: bool,
    files: Vec<Arc<Path>>,
    // ...
}
```

Since the function here only modifies `self.flags`, I don't think it will affect ar/asm flag or file

NobodyXu: @raisfeld-ori I still want this to change because the current doc includes implementation details and the function only removes flags added by `Build::flag`

NobodyXu (changes requested):

src/lib.rs
raisfeld-ori: sorry, i didn't know that, i guess didn't read the code well enough, so i thought all flags would be part of a single structure. i'l change it now

src/lib.rs
```diff
@@ -428,6 +428,22 @@ impl Build {
         self
     }
 
+    /// removes a flag from self.flags that was added by Build.flag()
```
NobodyXu, line 431: Please don't mention `self.flags` since it is an implementation details, and also it's better to use [`Build::flag`] since it will create a url in the doc generated:

```suggestion
    /// removes a compiler flag that was added by [`Build::flag`]
```

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -428,6 +428,22 @@ impl Build {
...
+    ///
+    /// # example
```
NobodyXu, line 433: Since doc for other functions uses capitalized `Example`, I suggest doing the same here:

```suggestion
    /// # Example
```

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -428,6 +428,22 @@ impl Build {
...
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     .flag("unwanted_flag")
+    ///     .remove_flag("unwanted_flag")
+    ///     .compile("bar");
+    /// ```
+
+    pub fn remove_flag(&mut self, flag: &str) -> &mut Build {
+        self.flags.retain(|other_flag| &**other_flag == flag);
```
NobodyXu, line 443: I didn't noticed that you changed `!=` to `==`, `retain` only keeps element that returns `true` so it should be `!=`:

```suggestion
        self.flags.retain(|other_flag| &**other_flag != flag);
```

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -428,6 +428,22 @@ impl Build {
         self
     }
 
+    /// removes a compiler flag that was added by [`Build::flag`]
```
NobodyXu, line 431: Minor doc suggestion:

```suggestion
    /// Removes a compiler flag that was added by [`Build::flag`]
```

src/lib.rs
```diff
@@ -428,6 +428,22 @@ impl Build {
...
+    ///
+    /// # Example
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     .flag("unwanted_flag")
+    ///     .remove_flag("unwanted_flag")
+    ///     .compile("bar");
+    /// ```
```
NobodyXu, line 440: I think it's better to remove `.compile` so that the doc can be run.
While it cannot prevent the bug that happened, it can at least verify that calling it at runtime doesn't panic.

```suggestion
    /// ```
    /// cc::Build::new()
    ///     .file("src/foo.c")
    ///     .flag("unwanted_flag")
    ///     .remove_flag("unwanted_flag");
    /// ```
```

NobodyXu (changes requested):

NobodyXu (approved):
LGTM, however you would need approval from a team member to merge.

cc @thomcc

PR #885, opened 2023-10-07
added a function to remove flags

raisfeld-ori:
i accidentally deleted the original pull request, so i made here a new one

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -428,7 +428,10 @@ impl Build {
         self
     }
 
-    /// removes a compiler flag that was added by [`Build::flag`]
+    /// Removes a compiler flag that was added by [`Build::flag`].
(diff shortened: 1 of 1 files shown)


NobodyXu (approved):

src/lib.rs
```diff
@@ -428,6 +428,21 @@ impl Build {
         self
     }
 
+    /// removes a compiler flag that was added by [`Build::flag`]
```
thomcc, line 431:
```suggestion
    /// Removes a compiler flag that was added by [`Build::flag`].
    ///
    /// Will not remove flags added by other means (default flags,
    /// flags from env, and so on).
```

thomcc (approved):
Seems good, reworded comment.

PR #889, opened 2023-10-22
Optimization: Vendor jobserver impl and rm thread spawning in parallel compile_objects

NobodyXu:
This PR contains a simpler, vendored implementation of jobserver optimized for cc, it supports non-blocking `try_acquire`.

By using the non-blocking `try_acquire` and async-as-state-machine, I was able to completely eliminate usage of additional thread inside parallel version of `compile_objects`.


thomcc:
This looks great at first glance, but it might not be for a week or two before I can actually review it.

NobodyXu:
> but it might not be for a week or two before I can actually review it (it's a big PR).

That's totally fine, it's huge and change a lot of stuff.

I might also polish it a bit in the meantime and added more testing for the parallel `compile_objects` to detect regression we previously seen (e.g. passing more objects than the cores, msrv).

thomcc:
Yep, feel free.

src/job_token.rs
```diff
@@ -1,139 +1,160 @@
...
+            Self(AtomicU32::new(parallelism))
+        }
+
+        pub(super) fn try_acquire(&self) -> Option<JobToken> {
+            let res = self.0.fetch_update(Relaxed, Relaxed, |tokens| {
+                if tokens > 0 {
+                    Some(tokens - 1)
+                } else {
+                    None
+                }
```
osiewicz, line 146:
```suggestion
                tokens.checked_sub(1)
```

src/job_token.rs
```diff
@@ -1,139 +1,160 @@
...
+            });
+
+            if res.is_ok() {
+                Some(JobToken())
+            } else {
+                None
+            }
```
osiewicz, line 153:
```suggestion
            res.ok().map(|_| JobToken())
```

src/lib.rs
```diff
@@ -1448,6 +1489,45 @@ impl Build {
                 child.kill().ok();
             }
         }
+
+        #[derive(Default)]
```
osiewicz, line 1493: I know it's not that much of code, but this feels a bit like an implementation detail; I think we should have it in a separate module with an explanation for why we can't just pull an async runtime in and that we have to roll our own async executor.

src/lib.rs
```diff
@@ -1328,116 +1334,151 @@ impl Build {
...
+                        std::time::Duration::from_millis(100 * (backoff_cnt - 3).min(10));
+                    thread::sleep(duration);
                 } else {
-                    backoff_cnt + 1
-                };
+                    // Given that we spawned a lot of compilation tasks, it is unlikely
+                    // that OS cannot find other ready task to execute.
+                    //
+                    // If all of them are done, then we will yield them and spawn more,
+                    // or simply returns.
```
osiewicz, line 1468:
```suggestion
                    // or simply return.
```

osiewicz:
Overall I like your approach and I think it solves the issue that we've had with initial channel-based implementation nicely, though I think it would be nice to explain:
- why do we have a static token queue in the first place
- why we can't really store the tokens in the queue and have to use an external jobserver for that (that's due to Rust not supporting Drop for static).
I didn't look much into the jobserver reimplementation, but I'm assuming it's a verbatim copy of `jobserver` with `try_acquire` slapped on top; if so, I'd suggest to state which version it is based on so that one could verify in the future if there's a good reason (e.g. a bugfix) to sync them again.
I also think that it would be pretty cool to split this up a bit.
Overall though this looks pretty cool to me, good job. =)

src/job_token.rs
NobodyXu: `token` is an atomic and it doesn't have `checked_sub`, using `fetch_update` is the most simple solution I can find.

osiewicz: I'm not suggesting to change the `fetch_update`; though `tokens` in this function is a `u32`, so we can use `checked_sub`.

src/lib.rs
thomcc: I would definitely prefer we don't pull an async runtime in.

osiewicz: Oh yeah, I'm not suggesting that - it's just that at first glance one could ask why we're rolling a custom executor instead of pulling something in. Given how `cc` is at the very bottom of most dependencies graphs, that would be a disaster I guess. =)
All I'm saying is that this decision should be documented to avoid having that discussion in the future.

NobodyXu:
> - why do we have a static token queue in the first place

The static token queue is for cases where there's no jobserver configured and we need to create a new jobserver.

Instead of porting the code to create a new jobserver, it's much simpler to have a static token queue implemented using atomics.

> - why we can't really store the tokens in the queue and have to use an external jobserver for that (that's due to Rust not supporting Drop for static).

I think there's some comments in the inherited_jobserver mod.

> I didn't look much into the jobserver reimplementation, but I'm assuming it's a verbatim copy of `jobserver` with `try_acquire` slapped on top; if so, I'd suggest to state which version it is based on so that one could verify in the future if there's a good reason (e.g. a bugfix) to sync them again.

It's actually based on [jobslot](https://docs.rs/jobslot), a fork of jobserver.

But even so it's not identical apart from try_acquire, I have simplified the code a bit and I will port that back to jobslot later.

> I also think that it would be pretty cool to split this up a bit.

Can you please elaborate on which part to split up?

> Overall though this looks pretty cool to me, good job. =)

Thank you and thanks for reviewing my code!

src/job_token.rs
NobodyXu: Thanks!
Sorry for the misunderstanding

src/job_token.rs
```diff
@@ -1,139 +1,152 @@
...
+
+#[cfg(unix)]
+#[path = "job_token/unix.rs"]
+mod sys;
+
+#[cfg(windows)]
+#[path = "job_token/windows.rs"]
+mod sys;
+
+pub(super) struct JobToken();
```
osiewicz, line 13: I guess it could be an unit struct instead of an empty tuple struct, what'd you say?
```suggestion
pub(super) struct JobToken;
```

osiewicz:
Oh yeah, I've meant that we should probably have comments for the first 2 points; by static I've meant that we have this `JOBSERVER` global that we've introduced in the previous PR; what problems does it solve? Why can't we use the job queue there? The answer to both of these questions is obvious to me and you given that we've cooperated and we have the understanding of the problem, but it's not going to be obvious to someone who reads this code for the first time.

> Can you please elaborate on which part to split up?
I think the async executor should be in a separate module (something like `executor`, idk).  Backoff logic could be hidden away as well, as this code (IMO) mixes the job scheduling (high level) with polling for the futures.

src/job_token.rs
osiewicz: No worries =)

src/job_token.rs
NobodyXu: I deliberately wrote it like that, to prevent other modules from creating `JobToken` accidentally without calling `try_acquire`.

osiewicz: Ah, I see! Thanks for clearing that up, let's keep it then.

src/lib.rs
NobodyXu: I've added the doc and extract async-executor out as a module as suggested.

NobodyXu:
cc @nrc since this PR creates a small async executor and it might help your "Async fundamentals initiative: portable and interoperable" in adding an small single-thread async executor to libstd.

osiewicz (approved):
LGTM

NobodyXu:
It seems that this PR have some MSRV issues, will wait for #890 to be merged and rebase.

NobodyXu:
Running `cargo +1.46 c --all-features` gives me the following errors:

```
error[E0658]: or-patterns syntax is experimental
  --> src/job_token/unix.rs:88:17
   |
88 |                 Some(libc::O_RDONLY) | Some(libc::O_RDWR),
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #54883 <https://github.com/rust-lang/rust/issues/54883> for more information

error[E0658]: or-patterns syntax is experimental
  --> src/job_token/unix.rs:89:17
   |
89 |                 Some(libc::O_WRONLY) | Some(libc::O_RDWR),
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #54883 <https://github.com/rust-lang/rust/issues/54883> for more information

error[E0658]: use of unstable library feature 'slice_strip'
  --> src/job_token/unix.rs:28:21
   |
28 |                 arg.strip_prefix(b"--jobserver-fds=")
   |                     ^^^^^^^^^^^^
   |
   = note: see issue #73413 <https://github.com/rust-lang/rust/issues/73413> for more information

error[E0658]: use of unstable library feature 'slice_strip'
  --> src/job_token/unix.rs:29:37
   |
29 |                     .or_else(|| arg.strip_prefix(b"--jobserver-auth="))
   |                                     ^^^^^^^^^^^^
   |
   = note: see issue #73413 <https://github.com/rust-lang/rust/issues/73413> for more information

error[E0658]: use of unstable library feature 'slice_strip'
  --> src/job_token/unix.rs:33:31
   |
33 |         if let Some(fifo) = s.strip_prefix(b"fifo:") {
   |                               ^^^^^^^^^^^^
   |
   = note: see issue #73413 <https://github.com/rust-lang/rust/issues/73413> for more information

error[E0599]: no method named `split_once` found for reference `&str` in the current scope
  --> src/job_token/unix.rs:64:31
   |
64 |         let (read, write) = s.split_once(',')?;
   |                               ^^^^^^^^^^ method not found in `&str`

error: aborting due to 6 previous errors

Some errors have detailed explanations: E0599, E0658.
For more information about an error, try `rustc --explain E0599`.
error: could not compile `cc`.

To learn more, run the command again with --verbose.
```

The first one is https://github.com/rust-lang/rust/pull/79278 stablised in 1.53, the second one is https://github.com/rust-lang/rust/pull/77853 stablised in 1.51 , the last one is https://github.com/rust-lang/rust/pull/81940 stablised in 1.52

Given that rust 1.53 is released on Jun 18, 2021 and it has been more than two years, @thomcc would it makes sense to bump the MSRV (again)?

thomcc:
> would it makes sense to bump the MSRV (again)?

I'm fine with it.

(I'll be actually reviewing this this weekend, just was skimming through the comments)

src/job_token/unix.rs
```diff
@@ -0,0 +1,189 @@
...
+                }
+                Err(e) => Err(e),
+            }
+        } else {
+            Ok(None)
+        }
+    }
+
+    pub(super) fn release(&self) -> io::Result<()> {
+        // Note that the fd may be nonblocking but we're going to go ahead
```
thomcc, line 134: I find this a little confusing. Should it say "Note that the fd may not be nonblocking"? (e.g. the fd is probably non-blocking but also may not be). Or am I mistaken?

src/job_token.rs
```diff
@@ -1,139 +1,163 @@
...
+                }
+            }
+
+            Self(AtomicU32::new(parallelism))
+        }
+
+        pub(super) fn try_acquire(&self) -> Option<JobToken> {
+            let res = self
+                .0
+                .fetch_update(Relaxed, Relaxed, |tokens| tokens.checked_sub(1));
```
thomcc, line 154: Kind of a clever way to write this. I don't hate it though.

src/job_token.rs
```diff
@@ -1,139 +1,163 @@
...
+
+            res.ok().map(|_| JobToken())
+        }
+
+        pub(super) fn release_token_raw(&self) {
+            self.0.fetch_add(1, Relaxed);
```
thomcc, line 160: IIUC the relaxed ops in this patch are only correct because we're single threaded. The overhead of using stronger atomics is way smaller than almost anything else `cc` does, so I'm not sure the savings are worth it. More broadly, it's possible that we may find a reason in the future we need to use a thread in some case in the future, on some platform. I'd like it if changing another part of the code doesn't make this slightly wrong.

Assuming I'm not mistaken there, presumably using acquire/release as implied by the function name is right, but feel free to use stronger if that's incorrect.

src/job_token/unix.rs
```diff
@@ -0,0 +1,189 @@
...
+            _ => Err(io::Error::from(io::ErrorKind::UnexpectedEof)),
+        }
+    }
+}
+
+fn set_nonblocking(file: &File) -> Option<()> {
+    // F_SETFL can only set the O_APPEND, O_ASYNC, O_DIRECT, O_NOATIME, and
+    // O_NONBLOCK flags.
+    //
+    // For pipe, only O_NONBLOCK is meaningful, so it is ok to
```
thomcc, line 159: Is this guaranteed to be true on all unix (including future ones?)

src/job_token/windows.rs
```diff
@@ -0,0 +1,67 @@
...
+unsafe impl Sync for JobServerClient {}
+unsafe impl Send for JobServerClient {}
+
+impl JobServerClient {
+    pub(super) unsafe fn open(var: OsString) -> Option<Self> {
+        let s = var
+            .to_str()?
+            .split_ascii_whitespace()
+            .filter_map(|arg| {
+                arg.strip_prefix("--jobserver-fds=")
```
thomcc, line 28: On windows I think we should probably check `--jobserver-auth` first, and possibly not check `--jobserver-fds` at all. That's what https://www.gnu.org/software/make/manual/html_node/Windows-Jobserver.html implies, anyway. Maybe I'm missing some older docs that say that it can be provided in --jobserver-fds.

src/job_token/windows.rs
```diff
@@ -0,0 +1,67 @@
...
+                    .or_else(|| arg.strip_prefix("--jobserver-auth="))
+            })
+            .find(|s| !s.is_empty())?;
+
+        let name = CString::new(s).ok()?;
+
+        let sem = OpenSemaphoreA(
```
thomcc, line 35: I wonder if it's worth complaining if the semaphore name isn't ascii (since we aren't using OpenSemaphoreW). It's hard to imagine that'd ever be the case in reality though.

src/lib.rs
```diff
@@ -1330,116 +1334,93 @@ impl Build {
...
-                        Err(err) => {
-                            // Task fail, remove the entry.
-                            has_made_progress = true;
-                            if is_disconnected {
-                                token.forget();
+                let mut pendings_is_empty = false;
+
+                cell_update(&pendings, |mut pendings| {
+                    // Try waiting on them.
+                    retain_unordered_mut(&mut pendings, |(cmd, program, child, _token)| {
```
thomcc, line 1363: It'd be nice if we could make this not O(n) per wait. I can think of ways to do that but I don't really want to land them at the same time as this. They also are plausibly overkill.

thomcc (commented):
This is really good. Clean and nice to read, especially for code that could be much more complex. I left a lot of comments but I don't think any should be tough to resolve (anything that is a big change we should punt on until after this release, to avoid releasing too many possibly-breaking things at once).

src/job_token.rs
```diff
@@ -1,139 +1,163 @@
...
-    return client;
+    impl JobServer {
+        pub(super) fn new() -> Self {
+            // Use `NUM_JOBS` if set (it's configured by Cargo) and otherwise
+            // just fall back to a semi-reasonable number.
+            //
+            // Note that we could use `num_cpus` here but it's an extra
+            // dependency that will almost never be used, so
+            // it's generally not too worth it.
+            let mut parallelism = 4;
```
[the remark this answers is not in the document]
NobodyXu: Yes, @dot-asm is against bumping msrv past 1.56

src/job_token/unix.rs
NobodyXu: Ohh I colied this from jobslot and it's an old comment I forgot to remove

src/job_token/unix.rs
NobodyXu: There's no guarantee but I think it should be ok, since I read in manpage it isn't ok to change flags other than `O_NONBLOCK`

src/job_token/windows.rs
NobodyXu: hmm I can switch to that version for maximum compatibikity

src/lib.rs
NobodyXu: unfortunately looping through the array is necessary, even if we use poll/epoll to poll on signalfd, unless we obtain a pidfd somehow.

Currently the method for obtaining pidfd via libstd is unstable so we would have to mess with syscalls directly to do this and I'm reluctant to try it.

src/job_token.rs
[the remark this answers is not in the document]
thomcc: @dot-asm can you make a defense as to why? It would be pretty nice to use available_parallelism here to avoid oversaturating the CPU on machines without a lot of concurrency (which includes a lot of vms/containers and machines people use for CI)

src/lib.rs
thomcc: Yeah, it probably needs more thought.

thomcc: That said, looping over it only in cases where we know there are changes is a lot better than looping when there might be none.

src/job_token.rs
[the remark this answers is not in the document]
osiewicz: Isn't `NUM_JOBS` going to be set most of the time when `cc` is being invoked through Cargo? Thus making `parallelism`'s default value less likely to be used.

src/job_token/windows.rs
NobodyXu: jobserver implementation always checks both so I assumed that both can be used https://docs.rs/jobserver/latest/x86_64-pc-windows-msvc/src/jobserver/lib.rs.html#282

It actually makes sense that only `--jobserver-auth` is used since windows doesn't have to concept of fd anyway.

I will remove checking for `--jobswerver-fds` on windows

src/job_token/windows.rs
NobodyXu: Since `jobserver` and `jobslot` also uses `OpenSemaphoreA`, I agree that it's a good idea to just check for ascii.

Switching to `OpenSemaphoreW` would make the code a bit harder to maintain since now it differs from upstream in more ways and since `OpenSemaphoreA` already works, it probably doesn't matter.

NobodyXu: Done.

src/job_token/windows.rs
NobodyXu: Done.

src/job_token/unix.rs
NobodyXu: Done.

src/job_token.rs
NobodyXu: Done, I've changed them to use acquire/release.

src/job_token/unix.rs
NobodyXu: Besides, we really don't want anyflags other than `O_NONBLOCK` to be set.

Pipe is always fifo, so `O_APPEND` means nothing.
Pipe does not write to disk, so `O_ASYNC`, `O_NOATIME` and `O_DIRECT` is useless.

`O_ASYNC` is for fs async I/O which is usually not used, `O_DIRECT` is for direct disk I/O which is also usually not used and meanless for in-memory only pipe.

src/job_token/windows.rs
```diff
@@ -0,0 +1,75 @@
...
+    ffi::{CString, OsString},
+    io, ptr,
+};
+
+use crate::windows_sys::{
+    OpenSemaphoreA, ReleaseSemaphore, WaitForSingleObject, FALSE, HANDLE, SEMAPHORE_MODIFY_STATE,
+    THREAD_SYNCHRONIZE, WAIT_ABANDONED, WAIT_FAILED, WAIT_OBJECT_0, WAIT_TIMEOUT,
+};
+
+const WAIT_ABANDOEND_ERR_MSG: &str = r#" The specified object is a mutex object that was not released by the thread that owned the mutex object before the owning thread terminated. Ownership of the mutex object is granted to the calling thread and the mutex state is set to nonsignaled.
```
thomcc, line 11: This is way more error information than we should provide, and none of it is likely to be useful for us. It's also wrong, because we're waiting on a semaphore object, but it says it was a mutex object, and describes some of the concerns around abandoned mutex objects.

Also, Windows semaphore objects aren't owned by a process/thread (unlike windows mutex objects), so I am pretty sure that WAIT_ABANDONED can never be returned for one anyway.

So we should probably just change this to something like `"Wait on jobserver semaphore returned WAIT_ABANDONED"` (it probably doesn't need to be const this way too), and add a comment in try_acquire that we believe this to be impossible for a semaphore.

src/job_token/windows.rs
NobodyXu: Yes, since the name is almost always auto-generated by `make`, or `cargo` using `jobserver`.

src/job_token/windows.rs
NobodyXu: Thanks, I've updated the errmsg and added the comment.

src/job_token.rs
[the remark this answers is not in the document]
NobodyXu: Yes, I've added a "TODO" comment.

thomcc (approved):
Thanks for all the work, this is great!

PR #890, opened 2023-10-27
Fix msrv CI: Check for `--all-features`

NobodyXu:
To make sure that enabling `parallel` feature would not require bumping msrv.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -128,7 +128,7 @@ jobs:
     - name: Create Cargo.lock with minimal version
       run: cargo +nightly update -Zminimal-versions
     - name: Fetch dependencies crates using stable since fetching them using 1.46.0  is very slow
-      run: cargo +stable fetch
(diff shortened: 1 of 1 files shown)


NobodyXu:
@thomcc Pinging since I'd like this PR to merge first.
I would need this to be merged so that I can bump MSRV (and possibly re-enable macos msrv check) sas otherwise it would cause a merge conflict.

This PR is quite simple so it won't take you long to review it.

dot-asm:
> so that I can bump MSRV

Not that my opinion matters, I'm just a user, but I would like to speak up **against** bumping MSRV if it goes above 1.56 [which matches 2021 edition].

NobodyXu:
@dot-asm I planned to bump it in 1.53 https://github.com/rust-lang/cc-rs/pull/889#issuecomment-1788887737

thomcc (approved):
Thanks!

PR #894, opened 2023-11-04
Bump MSRV to 1.53.0 and re-enable macos MSRV check

NobodyXu:
To solve the issue mentioned https://github.com/rust-lang/cc-rs/pull/889#issuecomment-1788887737

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -116,7 +116,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest]
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


thomcc (approved):
Looks great, thanks.

NobodyXu:
Oops, it looks like 1.53 still has problem with linking on latest macos.
Will revert that change.

thomcc:
Hm, do you know which version fixes it?

NobodyXu:
> Hm, do you know which version fixes it?

Probably 1.56 or 1.58, had the same issue in https://github.com/cargo-bins/jobslot , however even 1.56 might be too new for some users (@dot-asm ).

Given that there's only one line of `cfg!(target_os = "macos")` which does not conditionally pull in new code but instead evalutes to a boolean, I think it's fine not checking on macos.

Issue #900, opened 2023-11-13
New minimum MacOS version logic causes build error


helixbass:
Hi,
I have a somewhat fuzzy grasp of the problem but think I've tracked this down to being an issue introduced by #848 

The problem I'm running into is that when trying to build a Rust project on my Mac (Intel, macos 12.6.9) that uses a build script + `cc` to build a C++ file it is failing on not being able to find `libstdc++`. Which I guess is because it "should" be looking for `libc++` instead?

For example [`tree-sitter-ruby`](https://github.com/tree-sitter/tree-sitter-ruby/tree/master), or [this test repo](https://github.com/helixbass/test-build-script-cpp)

These fail when running `cargo build` with something like:
```
  running: "c++" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "--target=x86_64-apple-darwin" "-mmacosx-version-min=10.7" "-I" "src" "-Wall" "-Wextra" "-o" "/Users/jrosse/prj/tree-sitter-lint-plugin-eslint-builtin/.tree-sitter-lint/tree-si
tter-lint-local/target/release/build/tree-sitter-ruby-3ed10c8d09b2beda/out/src/scanner.o" "-c" "src/scanner.cc"
  cargo:warning=clang: warning: include path for libstdc++ headers not found; pass '-stdlib=libc++' on the command line to use the libc++ standard library instead [-Wstdlibcxx-not-found]

  cargo:warning=src/scanner.cc:4:10: fatal error: 'cstring' file not found
```

So my rough analysis is that:

I just started running into this because `cc@1.0.84` just got released, which includes #848 

#848 appears to introduce the use of that `macosx-version-min` flag. Then #872 appears to try and change that logic to force a minimum `macosx-version-min` of `10.9` when you're compiling C++

But it appears that the fix from #872 isn't "kicking in" (and so per above logging, it is using `macosx-version-min=10.7`, which I guess causes it to use `libstdc++` instead of `libc++` or something?) because `rustc_provided_target()` (which runs `rustc --target whatever --print deployment-target`) is returning `10.7`

So I'm guessing the fix here is to refine that #872 logic to eg also force `10.9` when `rustc_provided_target()` returns something less than `10.9` (eg `10.7`)?

Riolku:
In case anyone wants a MWE:

```
[package]
name = "testing"
version = "0.1.0"
edition = "2021"

[dependencies]

[build-dependencies]
cc = "1.0"
```

[...]

PR #901, opened 2023-11-14
Fix Apple deployment version floor when linking C++

BlackHoleFox:
Like the title says :) This fixes an unfortunate mixture of https://github.com/rust-lang/cc-rs/pull/848 and https://github.com/rust-lang/cc-rs/pull/872 so that the baseline version for modern C++ on Apple targets is respected again, so they now link the right C++ stdlib again.

Fixes https://github.com/rust-lang/cc-rs/issues/900

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3407,6 +3407,8 @@ impl Build {
         target: &str,
         arch_str: Option<&str>,
     ) -> String {
+        const OLD_IOS_MINIMUM_VERSION: &str = "7.0";
+
         fn rustc_provided_target(rustc: Option<&str>, target: &str) -> Option<String> {
             let rustc = rustc?;
             let output = Command::new(rustc)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -3427,6 +3429,74 @@ impl Build {
...
+        // A long time ago they used libstdc++, but since macOS 10.9 and iOS 7 libc++ has been the library the SDKs provide to link against.
+        // If a `cc`` config wants to use C++, we round up to these versions as the baseline.
+        let maybe_cpp_version_baseline = |deployment_target_ver: String| -> String {
+            if !self.cpp {
+                return deployment_target_ver;
+            }
+
+            let mut deployment_target = deployment_target_ver
+                .split('.')
+                .map(|v| v.parse::<u8>().expect("integer version"));
```
NobodyXu, line 3453: Let's parse it as a u32 just to be safe:

```suggestion
                .map(|v| v.parse::<u32>().expect("integer version"));
```

NobodyXu (changes requested):

src/lib.rs
BlackHoleFox: Switched.

src/lib.rs
```diff
@@ -3427,6 +3429,74 @@ impl Build {
...
+                .split('.')
+                .map(|v| v.parse::<u32>().expect("integer version"));
+
+            match os {
+                AppleOs::MacOs => {
+                    // Targeting higher then 10.x (11.0+)?
+                    let major_supported = deployment_target
+                        .next()
+                        .map(|major| major > 10)
+                        .unwrap_or_default();
```
NobodyXu, line 3461:
```suggestion
                    let major = deployment_target.next().unwrap_or(0);
```

I think it's more readable to obtain the major and minor, then test it in `if`.

NobodyXu (changes requested):

thomcc (approved):

NobodyXu (approved):

Issue #797, opened 2023-03-10
Failed to cross compile rust-openssl(openssl-src-rs) for windows msvc targets via cross-rs


cavivie:
`cross-rs`: At present, [cross-rs](https://github.com/cross-rs/cross) uses the Docker Ubuntu environment when compiling, and will download the wine and msvc tool chain in Ubuntu container to compile by mapping the source code into the container.

`openssl`: When the vendored feature is enabled, [rust-openssl](https://github.com/sfackler/rust-openssl) will call the [openssl-src-rs](https://github.com/alexcrichton/openssl-src-rs) to build the openssl c library. `openssl-src-rs` uses `rust-lang/cc-rs` to find the msvc tool when compiling the openssl library.

```rust
cc::windows_registry::find(target, "nmake.exe");
```

```
thread 'main' panicked at 'failed to find nmake', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-src-111.25.1+1.1.1t/src/lib.rs:454
```

Although the correct VC environment has been set, the find result is still None. Then I checked ` find_tool' function:
```rust
#[cfg(not(windows))]
pub fn find_tool(_target: &str, _tool: &str) -> Option<Tool> {
    None
}
```

When compiling cross platforms, should we use below the code to check?
```rust
// use these
#[cfg(target_os=windows)]
#[cfg(not(target_os=windows))]

// not use these
#[cfg(windows)]
#[cfg(not(windows))]
```

Or should we manually pass `--cfg windows` to compile?

cavivie:
I'm so sorry. I think my description is wrong. We should use `CARGO_ CFG_ TARGET_ OS` option to check the target instead of `cfg` conditional compilation, because `#[cfg (windows)]` detects the runtime environment. So in order to support cross compilation, I think this conditional compilation should be modified.

```rust
/// Similar to the `find` function above, this function will attempt the same
/// operation (finding a MSVC tool in a local install) but instead returns a
/// `Tool` which may be introspected.
#[cfg(not(windows))]
pub fn find_tool(_target: &str, _tool: &str) -> Option<Tool> {
    // print some information for debugging
    println!("cargo:warning=target is not windows, but env `CARGO_CFG_TARGET_OS` is {:?}", "", std::env::var("CARGO_CFG_TARGET_OS"));
    None
}
```

When I call `cc` to build for the compilation target is `x86_ 64-pc-windows-msvc` on Ubuntu, this code will be executed and the following information will be output:
[...]

PR #907, opened 2023-11-28
Support finding windows tools on non-Windows host

cavivie:
This PR contains a simple implementation of building targets on non-Windows platform. it supports manually set environment variables to find Windows msvc tools.

By setting `VCINSTALLDIR` environment variable, we can search msvc tools simply, and resolve related issues:
- https://github.com/rust-lang/cc-rs/issues/797

Currently, this PR can be regarded as a draft implementation, and the implementation details may still be discussed.


cavivie:
@ChrisDenton It would be great if you could give me some advice.

src/lib.rs
```diff
@@ -87,7 +87,7 @@ mod vs_instances;
 #[cfg(windows)]
 mod windows_sys;
 
-pub mod windows_registry;
+pub mod windows_helper;
```
NobodyXu, line 90: This is a breaking change, I suggest you to rename it baxk to `windows_registry`

NobodyXu (changes requested):

src/lib.rs
cavivie: **resolved**

src/windows_registry.rs
```diff
@@ -957,3 +947,73 @@ mod impl_ {
...
+    // Finding devenv.exe tool under unix system is not currently supported.
+    // Maybe can check it using an environment variable looks like `DEVENV_BIN`.
+    pub(super) fn find_devenv(_target: TargetArch<'_>) -> Option<Tool> {
+        None
+    }
+
+    /// Attempt to find the tool using environment variables set by vcvars.
+    pub(super) fn find_msvc_environment(tool: &str, _target: TargetArch<'_>) -> Option<Tool> {
+        // Early return if the environment doesn't contain a VC install.
+        let _vc_install_dir = env::var_os("VCINSTALLDIR")?;
```
NobodyXu, line 974: Seem to be another duplicate.

NobodyXu (changes requested):

src/windows_registry.rs
```diff
@@ -8,22 +8,39 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! A helper module to probe the Windows Registry when looking for
-//! windows-specific tools.
+//! A helper module to looking for windows-specific tools:
+//! 1. On Windows host, probe the Windows Registry if needed;
+//! 2. On non-Windows host, check specified environment variables.
 
+#![allow(clippy::upper_case_acronyms)]
```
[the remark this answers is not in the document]
cavivie: yes, it is a duplicate, I will remove it.

src/windows_registry.rs
cavivie: We can not do this at a non-Windows platform, so we should remove it, or find out any way to do like this.

NobodyXu: I mean there are two lines of

```
        let _vc_install_dir = env::var_os("VCINSTALLDIR")?;
```

cavivie (commented):
done

cavivie (commented):
done

cavivie (commented):
done

cavivie (commented):
duplicate

src/windows_registry.rs
NobodyXu: Thanks, these two lines looks really similar that my brain mixed them up

NobodyXu: > We have to make sure `VCINSTALLDIR ` exists because the build tools are under `VCINSTALLDIR`.

If the build tools are under them, does it make sense to use them to find the tools in the same way it's done below?

cavivie: I think it makes a lot of sense, but I'm not sure if it works or if it's really consistent with real usage.

cavivie: In theory, the VS installation directory structure of the simulation environment (such as Wine) should be consistent with the host environment (Windows host), but I am not sure whether other custom installation behaviors are completely consistent with those installed on the host using vsinstaller.

cavivie: Also, whether we distinguish between non-Windows host environments should not be confused with the assumption that it must be wine here, if a non-Windows host environment is not necessarily by wine, but at least for now.

cavivie: Then again, I think once we have a simulated environment, the entire execution method should not be much different from the real windows environment, so the windows implementation should become available here as well, right? I don't know if the distinction between windows and non-windows is really needed here.

NobodyXu: I think setting up a new CI for testing cross compilation to Windows on Linux using cross-rs or wine will be helpful.

It would verify that the code you wrote indeed works and gives us confidence in merging in the code, and detect future regression.

cavivie: Unfortunately, since the msvc environment is not redistributable, we don't have a pre-built image to perform CI, and we can't simply do this? May only be made on a non-disclosure and non-redistribution basis.

NobodyXu: Ahhh yes,that's really annoying.

Since we are only using in the CI, maybe that term doesn't apply to this use case/microsoft won't sue us?

NobodyXu: It would be great if we can get support from people who are more familiar with them or microsoft employers

cavivie: Although only used in CI, image storage containing msvc tools appears to be against Microsoft policy.

NobodyXu: That's annoying...

NobodyXu:
cc @thomcc I will need your help on how to move forward with this PR

maaft:
subbing. very interested to be able to use cross for msvc compilation.

NobodyXu (commented):
Going through this again, I think it's ok to merge in this PR.
At the very least it shouldn't regress anything, then it will be perfectly fine for it to merge in.

If we find any issue with it lately, we could just fix it and cut a new release.

NobodyXu:
cc @cavivie Can you rebase this please?

I'm willing to merge it in.

cavivie:
> cc @cavivie Can you rebase this please?
> 
> I'm willing to merge it in.

Okay, I will do it as soon as possible.

src/windows/find_tools.rs
```diff
@@ -1041,3 +1026,73 @@ mod impl_ {
...
+    }
+
+    /// Attempt to find the tool using environment variables set by vcvars.
+    pub(super) fn find_msvc_environment(tool: &str, _target: TargetArch<'_>) -> Option<Tool> {
+        // Early return if the environment doesn't contain a VC install.
+        let _vc_install_dir = env::var_os("VCINSTALLDIR")?;
+        let _vs_install_dir = env::var_os("VSINSTALLDIR")?;
+
+        // Should we take the path of tool for the v[c|s] install dir?
+        // Both `VCINSTALLDIR` / `VSINSTALLDIR` are unused currently.
```
NobodyXu, line 1057: I feel like we should use them, if we check for existence of these env vars.

NobodyXu (changes requested):

NobodyXu (approved):
Thank you!

PR #910, opened 2023-12-06
Add support for ARM64EC when building with MSVC

dpaoliello:
* Building for [ARM64EC](https://learn.microsoft.com/en-us/windows/arm/arm64ec) requires passing the `-arm64EC` flag to cl.exe and `/machine:arm64ec` to lib.exe: <https://learn.microsoft.com/en-us/windows/arm/arm64ec-build#developer-command-prompt>
* Updated tool discovery to handle `arm64ec` as a target arch by using the `aarch64` paths.
* When targeting `arm64ec` the lib path must ALSO include the `arm64ec` VC libs BEFORE the `aarch64` libs.
* Added a new `TargetArch` wrapper type to avoid splitting the target string multiple times or calling `contains` when we could be using string comparison.

### DIFF
--- a/src/windows_registry.rs
+++ b/src/windows_registry.rs
@@ -55,6 +55,9 @@ pub fn find_tool(target: &str, tool: &str) -> Option<Tool> {
         return None;
     }
 
+    // Split the target to get the arch.
+    let target = impl_::TargetArch(target.split_once('-')?.0);
+
     // Looks like msbuild isn't located in the same location as other tools like
     // cl.exe and lib.exe. To handle this we probe for it manually with
     // dedicated registry keys.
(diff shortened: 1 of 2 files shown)


dpaoliello:
@ChrisDenton would you mind having a look at this PR? It's blocking adding ARM64EC as a new target, and also provides some cleanups to avoid splitting the target triple over-and-over again within `windows_registry`.

ChrisDenton:
For sure, I'm still catching up on review but I'll prioritize this.

src/windows_registry.rs
```diff
@@ -451,9 +477,9 @@ mod impl_ {
     }
 
     fn vs15plus_vc_paths(
-        target: &str,
+        target: TargetArch<'_>,
         instance_path: &Path,
-    ) -> Option<(PathBuf, PathBuf, PathBuf, PathBuf, PathBuf)> {
+    ) -> Option<(PathBuf, PathBuf, PathBuf, PathBuf, Option<PathBuf>, PathBuf)> {
```
ChrisDenton, line 482: I think this return type is getting increasingly out of hand and should probably be a proper struct or otherwise rearchitectured. However, `vs15plus_vc_paths` is only called from one place and that immediately deconstructs it into named variables so I'm not minded to block this PR on that.

ChrisDenton (approved):
Looks good to me. I have one observation but nothing that should block merging this.

I do think using `TargetArch` is a clear improvement even without the other changes in this PR.

PR #914, opened 2024-01-02
Add new compile_intermediates function.

### LINKED ISSUE #912: Expose function compile to an object file
Currently, rustc's bootstrap uses `cc` to generate object file by asking it to generate a `.a` archive, discard the archive, and "guessing" where the temporary object files are to use them. Things like [this](https://github.com/rust-lang/rust/blob/c5208518faa423b96e6d1fecaa6428a2fa4eb9d1/src/bootstrap/src/core/build_steps/llvm.rs#L1214). This, however, has been broken since either #684 or #786, as bootstrap does not understand the hashing mechanism used by cc-rs. This prevents upgrading cc in rustc's bootstrap, which is not ideal as `cc` has since included some bugfixes that are necessary to properly handle some situations (such as [this](https://github.com/rust-lang/cc-rs/pull/676)).

I wonder if cc would accept a PR adding a public `Build::compile_objects()` function that would return a `Vec<Path>` of all the object files?

roblabla:
This new function can be used to just compile the files to a bunch of .o files. It won't create an archive file, nor print any of the cargo link directives. This is meant to be used by rustc to:

- Compile the MUSL `crtbegin.c` [here](https://github.com/rust-lang/rust/blob/c5208518faa423b96e6d1fecaa6428a2fa4eb9d1/src/bootstrap/src/core/build_steps/llvm.rs#L1214)
- Use different set of arguments when building the C and C++ parts of libunwind [here](https://github.com/rust-lang/rust/blob/c5208518faa423b96e6d1fecaa6428a2fa4eb9d1/src/bootstrap/src/core/build_steps/llvm.rs#L1358)

Currently, rustc guesses the output path of the object files, even though it's an unexposed implementation detail of `cc`. The latest version of `cc` added some hash to the generated object file name, which broke rustc's assumptions.

By providing a proper way to generate an object file using `cc`, this should hopefully make the system more robust.

Fixes #912

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3846,13 +3846,6 @@ fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<Object>, Er
         } else {
             dst.join(file).with_extension("o")
         };
-        let obj = if !obj.starts_with(&dst) {
-            dst.join(obj.file_name().ok_or_else(|| {
-                Error::new(ErrorKind::IOError, "Getting object file details failed.")
-            })?)
-        } else {
-            obj
-        };
 
         match obj.parent() {
             Some(s) => fs::create_dir_all(s)?,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2364,6 +2364,7 @@ impl Build {
     }
 
     fn apple_flags(&self, cmd: &mut Tool) -> Result<(), Error> {
+        #[allow(dead_code)]
         enum ArchSpec {
             Device(&'static str),
             Simulator(&'static str),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3827,25 +3827,22 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(), 
 fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<Object>, Error> {
     let mut objects = Vec::with_capacity(files.len());
     for file in files {
-        let obj = if file.has_root() || file.components().any(|x| x == Component::ParentDir) {
-            // If `file` is an absolute path or might not be usable directly as a suffix due to
-            // using "..", use the `basename` prefixed with the `dirname`'s hash to ensure name
-            // uniqueness.
-            let basename = file
-                .file_name()
-                .ok_or_else(|| Error::new(ErrorKind::InvalidArgument, "file_name() failure"))?
-                .to_string_lossy();
-            let dirname = file
-                .parent()
-                .ok_or_else(|| Error::new(ErrorKind::InvalidArgument, "parent() failure"))?
-                .to_string_lossy();
-            let mut hasher = hash_map::DefaultHasher::new();
-            hasher.write(dirname.to_string().as_bytes());
-            dst.join(format!("{:016x}-{}", hasher.finish(), basename))
-                .with_extension("o")
-        } else {
-            dst.join(file).with_extension("o")
-        };
+        let basename = file
+            .file_name()
+            .ok_or_else(|| Error::new(ErrorKind::InvalidArgument, "file_name() failure"))?
+            .to_string_lossy();
+        let dirname = file
+            .parent()
+            .ok_or_else(|| Error::new(ErrorKind::InvalidArgument, "parent() failure"))?
+            .to_string_lossy();
+
+        // Hash the dirname. This should prevent conflicts if we have multiple
+        // object files with the same filename in different subfolders.
+        let mut hasher = hash_map::DefaultHasher::new();
+        hasher.write(dirname.to_string().as_bytes());
+        let obj = dst
+            .join(format!("{:016x}-{}", hasher.finish(), basename))
+            .with_extension("o");
 
         match obj.parent() {
             Some(s) => fs::create_dir_all(s)?,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3825,7 +3825,7 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(), 
 /// Find the destination object path for each file in the input source files,
 /// and store them in the output Object.
 fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<Object>, Error> {
-    let mut objects = Vec::new();
+    let mut objects = Vec::with_capacity(files.len());
     for file in files {
         let obj = if file.has_root() || file.components().any(|x| x == Component::ParentDir) {
             // If `file` is an absolute path or might not be usable directly as a suffix due to
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1131,48 +1131,7 @@ impl Build {
         };
         let dst = self.get_out_dir()?;
 
-        let mut objects = Vec::new();
-        for file in self.files.iter() {
-            let obj = if file.has_root() || file.components().any(|x| x == Component::ParentDir) {
-                // If `file` is an absolute path or might not be usable directly as a suffix due to
-                // using "..", use the `basename` prefixed with the `dirname`'s hash to ensure name
-                // uniqueness.
-                let basename = file
-                    .file_name()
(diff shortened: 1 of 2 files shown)


thomcc:
Thanks. I won't be able to take a look at this until the weekend.

src/lib.rs
```diff
@@ -1316,6 +1275,32 @@ impl Build {
...
+        }
+    }
+
+    /// Run the compiler, generating intermediate files, but without linking
+    /// them into an archive file.
+    ///
+    /// This will return a result instead of panicing; see `compile_intermediates()` for the complete description.
+    pub fn try_compile_intermediates(&self) -> Result<Vec<PathBuf>, Error> {
+        let dst = self.get_out_dir()?;
+        let objects = objects_from_files(self.files.iter().map(|v| &**v), &dst)?;
```
NobodyXu, line 1296: I think it worth extracting `self.files.iter().map(|v| &**v)` into another function, or add another associated function `objects` which calculate this.

The `objects_from_file` is a generic function and since each closure has a different type, the function is mono-ed twice instead of once.

NobodyXu (changes requested):

src/lib.rs
roblabla: I guess I could remove the genericity and just have object_from_files take an `&[Arc<Path>]`

src/lib.rs
```diff
@@ -3830,6 +3815,54 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(),
     }
 }
 
+/// Find the destination object path for each file in the input source files,
+/// and store them in the output Object.
+fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<Object>, Error> {
+    let mut objects = Vec::new();
+    for file in files.into_iter() {
```
NobodyXu, line 3822:
```suggestion
    for file in files {
```

src/lib.rs
```diff
@@ -3830,6 +3815,54 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(),
     }
 }
 
+/// Find the destination object path for each file in the input source files,
+/// and store them in the output Object.
+fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<Object>, Error> {
+    let mut objects = Vec::new();
```
NobodyXu, line 3821:
```suggestion
    let mut objects = Vec::with_capacity(files.len());
```

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -3830,6 +3815,54 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(),
...
+                .ok_or_else(|| Error::new(ErrorKind::InvalidArgument, "parent() failure"))?
+                .to_string_lossy();
+            let mut hasher = hash_map::DefaultHasher::new();
+            hasher.write(dirname.to_string().as_bytes());
+            dst.join(format!("{:016x}-{}", hasher.finish(), basename))
+                .with_extension("o")
+        } else {
+            dst.join(file).with_extension("o")
+        };
+        let obj = if !obj.starts_with(&dst) {
```
NobodyXu, line 3842: Is this really necessary?

`obj` is created by joining `dst` with a name, so won't this always evaluates to `false`?

NobodyXu (changes requested):

src/lib.rs
roblabla: I assume that this check is a remnant from before https://github.com/rust-lang/cc-rs/pull/684 was merged. Since there was no `file.has_root()` check before, the `dst.join(file)` could result in a path outside dst.

But yeah, it can probably be removed now.

src/lib.rs
```diff
@@ -3830,6 +3815,47 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(),
     }
 }
 
+/// Find the destination object path for each file in the input source files,
+/// and store them in the output Object.
+fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<Object>, Error> {
+    let mut objects = Vec::with_capacity(files.len());
+    for file in files {
+        let obj = if file.has_root() || file.components().any(|x| x == Component::ParentDir) {
```
NobodyXu, line 3823: I'm thinking that we need to do some normalisation before running this line, e.g. `././a` => `a`, `a/../b` => `b`

cc @thomcc

NobodyXu (changes requested):

src/lib.rs
roblabla: Probably. Though it's worth noting that `a/../b` does not necessarily == `b`, in particular if `a` is a symlink.

TBH I think we should just remove the check entirely and always take the codepath that does the hashing. It's simpler, fixes some problems where object filename conflcits can cause issue[^1], and sidesteps potential issues we could have with normalisation.

[^1]: for instance, today, if you try to build a lib with something like `cc::Build::new().file("a/object.c").file("b/object.c").compile("lib")`, it will only contain a single file named object.o - one of the two objects will be discarded.

NobodyXu: I agree, it's probably a good idea to always hash the prefix.

NobodyXu: Though I still think we can do some normalisation to remove `./` as it is no-op and not affected by symlink or whatever.

NobodyXu:
pinging @roblabla just in case you forgot this PR/comment

src/lib.rs
roblabla: I removed the condition, and make it so it _always_ does the hashing.

I'm not sure what the normalization is supposed to achieve, and worry that it'll create more problems than it solves. @NobodyXu can you explain why you want the path to be normalised?

src/lib.rs
```diff
@@ -3837,6 +3823,43 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(),
...
 }
 
+/// Find the destination object path for each file in the input source files,
+/// and store them in the output Object.
+fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<Object>, Error> {
+    let mut objects = Vec::with_capacity(files.len());
+    for file in files {
+        let basename = file
+            .file_name()
+            .ok_or_else(|| Error::new(ErrorKind::InvalidArgument, "file_name() failure"))?
```
NobodyXu, line 3833:
```suggestion
            .ok_or_else(|| Error::new(ErrorKind::InvalidArgument, "No file_name for object file path!"))?
```

src/lib.rs
```diff
@@ -3837,6 +3823,43 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(),
...
+            .to_string_lossy();
+        let dirname = file
+            .parent()
+            .ok_or_else(|| Error::new(ErrorKind::InvalidArgument, "parent() failure"))?
```
NobodyXu, line 3837:
```suggestion
            .ok_or_else(|| Error::new(ErrorKind::InvalidArgument, "No parent for object file path!"))?
```

src/lib.rs
```diff
@@ -3837,6 +3823,43 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(),
...
+        let mut hasher = hash_map::DefaultHasher::new();
+        hasher.write(dirname.to_string().as_bytes());
+        let obj = dst.join(format!("{:016x}-{}", hasher.finish(), basename))
+            .with_extension("o");
+
+        match obj.parent() {
+            Some(s) => fs::create_dir_all(s)?,
+            None => {
+                return Err(Error::new(
+                    ErrorKind::IOError,
```
NobodyXu, line 3851: There's no I/O error:

```suggestion
                    ErrorKind::InvalidArgument,
```

NobodyXu (changes requested):
Thank you!

Could you please run cargo-fmt before committing?

src/lib.rs
NobodyXu: I don't think it is necessary now that I think about it again, all we care is unique object file name here.

NobodyXu (approved):
Thank you!

PR #915, opened 2024-01-04
Link against libc++ on *-linux-ohos

DianQK:
I can confirm that openharmony is utilizing the runtime in llvm-project, e.g. libc++.

Refer:
- https://doc.rust-lang.org/beta/rustc/platform-support/openharmony.html
- https://github.com/llvm/llvm-project/blob/0cdaadf15aaa6609f93e3508417b47baa3891996/clang/lib/Driver/ToolChains/OHOS.h#L33


JohnTitor (approved):
LGTM, thanks1

Issue #916, opened 2024-01-13
Allow suppression of cargo warning instructions in output


scootermon:
Right now cc allows for the suppression of specific cargo instructions using `.cargo_metdata(false)`, but this doesn't affect the output of `cargo:warning=` lines.
I'm working on a bindings library which uses compilation tests to determine whether certain features are supported or not. CC already almost fully supports this use-case[^1], the only issue being that the warnings and errors show up as though they're part of the "main compilation".

Since I belive this feature is unrelated to the behaviour of `cargo_metadata`, how would you feel about adding a `.cargo_warnings(false)` setting that allows us to disable the warnings?
I'm of course happy to contribute the implementation.

[^1]: Shoutout to #912, which will also help here because it helps remove an otherwise unnecessary step from the compilation tests

NobodyXu:
Yes I'm willing to review a PR disabling the warnings.

PR #917, opened 2024-01-14
Add `cargo_warnings` config to control the use of the cargo warning instruction

scootermon:
Closes: #916 

I couldn't come up with a clean abstraction that doesn't needlessly bloat the code so I decided to keep it simple. Because of the lack of abstraction I'm somewhat concerned about maintainability though, as it might become harder to keep track of the warning instructions.

I also found some instances of `println!`s, which looked to me like they're supposed to be using the cargo warning instruction. I adjusted those, but let me know if that's incorrect.


src/lib.rs
```diff
@@ -2382,7 +2405,7 @@ impl Build {
         enum ArchSpec {
             Device(&'static str),
             Simulator(&'static str),
-            Catalyst(&'static str),
+            Catalyst(#[allow(dead_code)] &'static str),
```
scootermon, line 2408: This is only here because the new nightly versions correctly detect that this field is never used, which causes the CI to fail.

src/lib.rs
```diff
@@ -1393,7 +1405,10 @@ impl Build {
                                 // sure users always see all the compilation failures.
                                 has_made_progress.set(true);
 
-                                let _ = writeln!(stdout, "cargo:warning={}", err);
+                                if self.cargo_warnings {
+                                    let _ = write!(stdout, "cargo:warning=");
+                                }
+                                let _ = writeln!(stdout, "{}", err);
```
NobodyXu, line 1411: You should move this inside the `if`, IMO you could actually write a function taking `fmt::Arguments<'_>` and then have a macro to wrap it

src/lib.rs
```diff
@@ -2167,11 +2182,19 @@ impl Build {
...
                 _ => {
-                    println!(
-                        "cargo:warning=cpp_set_stdlib is specified, but the {:?} compiler \
+                    let stdout = io::stdout();
+                    {
+                        let mut stdout = stdout.lock();
+                        if self.cargo_warnings {
+                            let _ = write!(stdout, "cargo:warning=");
+                        }
+                        let _ = writeln!(
```
NobodyXu, line 2191: Same here

src/lib.rs
```diff
@@ -2726,7 +2749,14 @@ impl Build {
...
 
         if target.contains("msvc") && tool.family == ToolFamily::Gnu {
-            println!("cargo:warning=GNU compiler is not supported for this target");
+            let stdout = io::stdout();
+            {
+                let mut stdout = stdout.lock();
+                if self.cargo_warnings {
+                    let _ = write!(stdout, "cargo:warning=");
+                }
+                let _ = writeln!(stdout, "GNU compiler is not supported for this target");
```
NobodyXu, line 2758: Also here

src/lib.rs
```diff
@@ -3597,7 +3633,7 @@ impl Tool {
             } else {
                 // --version doesn't include clang for GCC
                 println!(
-                    "cargo-warning:Compiler version doesn't include clang or GCC: {:?}",
+                    "cargo:warning=Compiler version doesn't include clang or GCC: {:?}",
```
NobodyXu, line 3636: Thanks for fixing these warnings!

src/lib.rs
```diff
@@ -4169,11 +4211,11 @@ struct PrintThread {
 }
 
 impl PrintThread {
-    fn new() -> Result<Self, Error> {
+    fn new(cargo_warnings: bool) -> Result<Self, Error> {
         let (pipe_reader, pipe_writer) = os_pipe::pipe()?;
```
NobodyXu, line 4215: I recommend to skip spawaning the thread to reduce cost of compilation, if warning is disabled.

You could also skip creating pipes, but then you need to change code using `pipe_writer*`

NobodyXu (changes requested):
Thank you for the PR!

Can you please add a test for which warnings are disabled?

src/lib.rs
scootermon: Just to be clear, you'd prefer that `cargo_warnings(false)` fully disables the output of warnings, not just the 'cargo:warning' prefix?
I don't mind either way.

src/lib.rs
scootermon: I saw this too but cautiously decided against it because `PrintThread` also does line-by-line locking of stdout. I'm worried that removing this will cause garbled output when using the `parallel` feature.

src/lib.rs
NobodyXu: > Just to be clear, you'd prefer that `cargo_warnings(false)` fully disables the output of warnings, not just the 'cargo:warning' prefix? I don't mind either way.

Yes, the output is there as warning, so if the user disables the warning we should also discard the output.

src/lib.rs
NobodyXu: PrintThread only lock stdout when it receives a line to print to stdout, tomprevent interleaving.

scootermon: Oh but if we take the approach of fully disabling output instead of just removing the instruction prefix, we can just pass around an `Option<PrintThread>` and close stderr if it's `None`.

src/lib.rs
```diff
@@ -295,6 +295,25 @@ impl Object {
...
 }
 
+#[cfg(feature = "parallel")]
+macro_rules! writeln_warning {
+    ($dst:expr $(,)?) => {
+        $crate::writeln_warning!($dst, "")
+    };
+    ($dst:expr, $($arg:tt)*) => {
+        ::std::writeln!($dst, "cargo:warning={}", ::std::format_args!($($arg)*))
+    };
```
NobodyXu, line 305: `format_args` expects at least one argument, with the first one being a literal string:

```suggestion
    ($dst:expr, $s:literal) => {
        ::std::writeln!($dst, "cargo:warning={}", s)
    };
    ($dst:expr, $fmt:literal, $($arg:tt)*) => {
        ::std::writeln!($dst, "cargo:warning={}", ::std::format_args!($fmt, $($arg)*))
    };
```

src/lib.rs
```diff
@@ -295,6 +295,25 @@ impl Object {
...
+}
+
+macro_rules! println_warning {
+    () => {
+        $crate::println_warning!("")
+    };
+    ($($arg:tt)*) => {
```
NobodyXu, line 312: Same here.

src/lib.rs
```diff
@@ -2182,18 +2204,8 @@ impl Build {
...
-                        }
-                        let _ = writeln!(
-                            stdout,
-                            "cpp_set_stdlib is specified, but the {:?} compiler \
-                         does not support this option, ignored",
-                            cmd.family,
-                        );
+                    if self.cargo_warnings {
+                        println_warning!("cpp_set_stdlib is specified, but the {:?} compiler does not support this option, ignored", cmd.family);
                     }
```
NobodyXu, line 2209: Sorry for the confusion, but I was actually thinking of extracting this as a function and macro:

```rust
impl Build {
    fn println_warning(&self, arg: &fmt::Argument<'_>) {
        if self.cargo_warnings {
            println!("cargo:warning={}", arg);
        }
    }
}

macro_rules! println_warning {
    ($this:expr, $s:literal) => {
        this.println_warning(format_args("{}", s))
    };
    ($dst:expr, $fmt:literal, $($arg:tt)*) => {
        this.println_warning(format_args!($fmt, $($arg)*)))
    };
}
```

src/lib.rs
```diff
@@ -4211,11 +4223,11 @@ struct PrintThread {
 }
 
 impl PrintThread {
-    fn new(cargo_warnings: bool) -> Result<Self, Error> {
+    fn new() -> Result<Self, Error> {
         let (pipe_reader, pipe_writer) = os_pipe::pipe()?;
```
NobodyXu, line 4227: I'm actually thinking of something like this:

```rust
impl PrintThread {
    fn new(cargo_warnings: bool) -> Result<Self, Error> {
        if !cargo_warning {
            return Ok(Self { handle: None, pipe_writer: None });
        }
        
        // ...
    }

    fn take_pipe_writer(&mut self) -> Option<File> {
        self.pipe_writer.take()
    }

    fn clone_pipe_writer(&self) -> Result<Option<File>, Error> {
        self.try_clone_pipe_writer()
    }
```

Using this will reduce the diff which simply changes `PrintThread` to `Option<PrintThread>`, while also removiung the use of `print_thread.map(...).transpose()?`

NobodyXu (changes requested):

src/lib.rs
scootermon: The only problem I see with this approach is that it makes it impossible to distinguish between an empty `PrintThread` and one that's had its pipe writer taken. If misused, this could lead to an annoyingly hard-to-track bug down the line.

NobodyXu: That's true, I will resolve this comment.

src/lib.rs
scootermon: I went with this signature to mirror the macros from `std`. While yours is technically more correct, it produces slightly worse error messages in case the first argument isn't a literal string (because format_args! has special handling for that case).

```rust
1  | macro_rules! a {
   | -------------- when calling this macro
...
15 |     a!(VALUE, "world");
   |        ^^^^^ no rules expected this token in macro call
   |
note: while trying to match meta-variable `$fmt:literal`
  --> src/main.rs:2:6
   |
2  |     ($fmt:literal, $($arg:tt)*) => {
```

vs:

```rust
16 |     b!(VALUE, "world");
   |        ^^^^^
   |
help: you might be missing a string literal to format with
   |
16 |     b!("{} {}", VALUE, "world");
   |        ++++++++
```

NobodyXu: Thanks, in that case I will keep your signature

src/lib.rs
NobodyXu: Yes, sorry for the confusion.

NobodyXu: > I see the appeal of having this println_warning method mirroring the design of the print method, but that means that the macro can only be used when you have access to Build.

I think having a `Build::println_warning` that accepts `& dyn Display`, user can pass in `&str` or `&fmt::Arguments<'_>`:

```
impl Build {
    fn println_warning(&self, arg: &dyn Display) {
        if self.cargo_warnings {
            println!("cargo:warning={}", arg);
        }
    }
}


```

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -1173,10 +1204,13 @@ impl Build {
             objects.push(Object::new(file.to_path_buf(), obj));
         }
 
-        let print = PrintThread::new()?;
+        let print = self
+            .cargo_warnings
+            .then(|| PrintThread::new())
```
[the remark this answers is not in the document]
scootermon: Thanks, I rely too much on Clippy to automatically fix these things 😅

src/lib.rs
```diff
@@ -1393,7 +1405,7 @@ impl Build {
                                 // sure users always see all the compilation failures.
                                 has_made_progress.set(true);
 
-                                let _ = writeln!(stdout, "cargo:warning={}", err);
+                                self.print_warning(&err);
```
scootermon, line 1408: Just want to point out that this now behaves slightly differently because it no longer goes through the BufWriter and instead writes to stdout directly.
I don't really like this. Are you okay with keeping this as:

```rust
if self.cargo_warnings {
    let _ = writeln!(stdout, "cargo:warning={}", err);
}
```

cc-test/build.rs
```diff
@@ -103,4 +103,12 @@ fn main() {
...
     assert!(out.contains("hello world"));
+
+    // Compilations _should_ fail, but we don't want it to show up as a warning.
+    cc::Build::new()
+        .cargo_metadata(false)
+        .cargo_warnings(false)
+        .file("src/suppress_warnings.c")
+        .try_compile("suppress_warnings")
+        .unwrap_err();
 }
```
scootermon, line 114: At the moment this "test" can't really be enforced. If it fails, you will see a warning in the output, but AFAIK there's no way to force that into an error.
Do you have an idea how else we could design a test for this?

src/lib.rs
scootermon: Makes sense. Sorry for the confusion on my part :)

src/lib.rs
NobodyXu: Yes, I'm ok with keeping the if and writeln

cc-test/build.rs
NobodyXu: I think you can "fork" the process by writing a test that execute `std::env::current_exe()` then capture stderr/stdout

NobodyXu:
@scootermon if it's too hard to add testing, I'm ok with removing all the testing.

scootermon:
> @scootermon if it's too hard to add testing, I'm ok with only test them on unix (since tests failed on windows)

I added testing in the latest commits and there shouldn't be any problems on Windows because I'm not actually forking the process. The CI is failing because the tests are working :) We're still writing the warning "Compiler version doesn't include clang or GCC" to the console (because that's handled in `Tool` not in `Build`).
I took some time off to think about how to deal with it but I would like to hear your input on it:
1. allow specific warnings to be written in the unit test
2. pass down the cargo_warnings flag into `Tool` to suppress it. Not as clean because we can't use the `Build::print_warning` method.

NobodyXu:
I suppose we should pass down the `cargo_warnings` flag to `Tool.

I suppose we can add a new type:

```
struct CargoWarning(bool);

impl CargoWarning {
    fn print_warning(&self, msg: &dyn Display) { .. }
```

So that the `print_warning` can be shared between `Build` and `Tool`.

scootermon:
@NobodyXu, apologies for the delay, got caught up with other stuff.

I did end up having to skip the warnings_on test case on MSVC though because cl.exe apparently doesn't write compilation warnings to stderr...

cc-test/tests/output.rs
```diff
@@ -0,0 +1,68 @@
+use std::fs;
+use std::path::PathBuf;
+
+#[test]
+fn cargo_warnings_on() {
+    if env!("TEST_WARNINGS_ON") == "0" {
```
NobodyXu, line 6: We already skip `run_forked_capture_output(&out, "warnings-on");` if `TEST_WARNINGS_ON` is ser to 0, do we still need this?

src/lib.rs
```diff
@@ -4202,11 +4276,21 @@ impl PrintThread {
...
+    /// Will panic if the pipe writer has already been taken.
+    fn take_pipe_writer(&mut self) -> File {
+        self.pipe_writer.take().unwrap()
+    }
+
+    /// # Panics
+    ///
+    /// Will panic if the pipe writer has already been taken.
+    fn clone_pipe_writer(&self) -> Result<File, Error> {
+        self.try_clone_pipe_writer().map(|writer| writer.unwrap())
```
NobodyXu, line 4290:
```suggestion
        self.try_clone_pipe_writer().map(Option::unwrap)
```

src/lib.rs
```diff
@@ -3871,22 +3896,31 @@ fn try_wait_on_child(
...
+) -> Result<Vec<u8>, Error> {
     cmd.stdout(Stdio::piped());
 
-    let mut print = PrintThread::new()?;
-    let mut child = spawn(cmd, program, print.pipe_writer().take().unwrap())?;
+    let mut print = cargo_output.print_thread()?;
+    let mut child = spawn(
+        cmd,
+        program,
+        print.as_mut().map(|print| print.take_pipe_writer()),
```
NobodyXu, line 3922:
```suggestion
        print.as_mut().map(PrintThread::take_pipe_writer),
```

src/lib.rs
```diff
@@ -3871,22 +3896,31 @@ fn try_wait_on_child(
...
-fn run_inner(cmd: &mut Command, program: &str, pipe_writer: File) -> Result<(), Error> {
+fn run_inner(cmd: &mut Command, program: &str, pipe_writer: Option<File>) -> Result<(), Error> {
     let mut child = spawn(cmd, program, pipe_writer)?;
     wait_on_child(cmd, program, &mut child)
 }
 
-fn run(cmd: &mut Command, program: &str, print: &PrintThread) -> Result<(), Error> {
-    run_inner(cmd, program, print.pipe_writer_cloned()?.unwrap())?;
+fn run(cmd: &mut Command, program: &str, print: Option<&PrintThread>) -> Result<(), Error> {
+    let pipe_writer = print.map(|print| print.clone_pipe_writer()).transpose()?;
```
NobodyXu, line 3905:
```suggestion
    let pipe_writer = print.map(PrintThread::clone_pipe_writer).transpose()?;
```

NobodyXu (changes requested):

cc-test/tests/output.rs
scootermon: I guess we could use the presence of the build artifacts (i.e. the stdout and stderr file) to detect whether to run the test, but then that could also be caused by someone modifying the build script and forgetting to update it here. Since the test would be silently skipped in that case, I would argue that we should keep it as explicit as possible.

src/lib.rs
```diff
@@ -1425,7 +1440,8 @@ impl Build {
                         YieldOnce::default().await
                     }
                 };
-                let child = spawn(&mut cmd, &program, print.pipe_writer_cloned()?.unwrap())?;
+                let pipe_writer = print.map(|print| print.clone_pipe_writer()).transpose()?;
```
NobodyXu, line 1443: One last nit:

```suggestion
                let pipe_writer = print.map(PrintThread::clone_pipe_writer).transpose()?;
```

NobodyXu (changes requested):

NobodyXu (approved):

NobodyXu:
Thank you!

Issue #847, opened 2023-08-03
Encoded `CFLAGS` for supporting spaces


MarijnS95:
According to https://github.com/rust-lang/cc-rs/#external-configuration-via-environment-variables spaces nor escapes for spaces are not supported in e.g. `C(XX)FLAGS`. In [`xbuild`](https://github.com/rust-mobile/xbuild) we [use `CFLAGS` to set a `--sysroot` path](https://github.com/rust-mobile/xbuild/blob/875f933f5908b49436e7176f2c0cfd7f5233ee24/xbuild/src/cargo/mod.rs#L444-L445), and this can occasionally contain spaces on Windows machines (because the toolchain is unpacked to a user folder, and users like to have `Firstname Surname` as username and profile directory).

A likable solution is new `CARGO_ENCODED_RUSTFLAGS`-like variables, where spaces within individual args are supported by requiring the user to replace space delimiters in between separate arguments with the `\x1f` ASCII Unit Separator: https://doc.rust-lang.org/cargo/reference/environment-variables.html, to allow more mechanical piecing-together of these variables without messing around with nested quotes or backslash-escapes.

Referencing https://github.com/rust-mobile/xbuild/issues/124, where this is one of the deciding factors to build our Android compiler support _differently_.

joshtriplett:
While that might be possible to address in cc-rs, it still seems likely to break in other build systems, such as invocations of make, which don't have a solution to this problem.

MarijnS95:
@joshtriplett does `cc` forward to other build systems or only expand these flags into `argv` before invoking the compiler (which _should_ be possible and support spaces by splitting on the separator char)? Or is this to maintain feature parity with other build systems outside of `cc-rs` that also process `CFLAGS`?

joshtriplett:
I'm talking about the latter: cc-rs doesn't seem like the right place to unilaterally propagate a new standard for passing CFLAGS that no other build system will understand.

thomcc:
There admittedly are plenty of people just using `cc` so it's not entirely out of the question. I have mixed feelings about it though, for sure.

xbjfk:
> such as invocations of make, which don't have a solution to this problem.

make does have a "solution" for this problem, if you surround the flag containing spaces in quotes or use backslashes (like shell escaping), it works (could be done using e.g. shlex): https://asciinema.org/a/oR0RRBOmtEX1W9p4GzVQk7mIt
[...]

PR #918, opened 2024-01-24
Use RUSTC_WRAPPER if no other wrapper is provided

LeonMatthesKDAB:
Previously, the `rustc_wrapper_fallback` was only called if the tool
 was **not** a full path.
With this patch, the `RUSTC_WRAPPER` is always used, even if the tool provided is an exact path.
Providing a wrapper manually is still possible and overrides the `RUSTC_WRAPPER`.

If the path to the tool includes spaces it is otherwise impossible to provide a compiler cache like sccache through the CXX or CC environment variables, as the arguments will be split up incorrectly.

See also: https://github.com/corrosion-rs/corrosion/pull/474

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2761,7 +2761,11 @@ impl Build {
         // interpretation at all, just pass it on through. This'll hopefully get
         // us to support spaces-in-paths.
         if Path::new(&*tool).exists() {
-            return Some((PathBuf::from(&*tool), None, Vec::new()));
+            return Some((
+                PathBuf::from(&*tool),
+                Self::rustc_wrapper_fallback(),
+                Vec::new(),
+            ));
         }
 
         // Ok now we want to handle a couple of scenarios. We'll assume from
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2382,6 +2382,7 @@ impl Build {
         enum ArchSpec {
             Device(&'static str),
             Simulator(&'static str),
+            #[allow(dead_code)]
             Catalyst(&'static str),
         }
 


LeonMatthesKDAB:
> I'd recommend you to add `#[allow(dead_code)]` for line 2385

Hi, thank you. That has fixed the CI issues.
Unsure where that issue was coming from in the first place, as I haven't touched that code at all...

Probably just an update in nightly.

NobodyXu:
> If the path to the tool includes spaces it is otherwise impossible to provide a compiler cache like sccache through the CXX or CC environment variables, as the arguments will be split up incorrectly.

P.S. Looking at the source code, seems like cc actually supports splitting env `CC='sccache cc'`.

This is just a rant from me, doesn't affect merging of this PR.

NobodyXu (commented):
It would be great if you can setup a new CI that tests using `RUSTC_WRAPPER` for tools with relative and absolute pathas you describe, honestly a lot of stuff in cc are untested, so I'd prefer to use this opportunity to add some test for this.

LeonMatthesKDAB:
> > If the path to the tool includes spaces it is otherwise impossible to provide a compiler cache like sccache through the CXX or CC environment variables, as the arguments will be split up incorrectly.
> 
> P.S. Looking at the source code, seems like cc actually supports splitting env `CC='sccache cc'`.
> 
> This is just a rant from me, doesn't affect merging of this PR.

Yes, it does, but only if the path to `cc` doesn't include spaces.

On windows, our CI finds the path to `cc` is something like:
`C:\Program Files\...\cc`
If this is encoded in `CC`, what cc-rs ends up generating is:
`sccache "C:\Program" "Files\...\cc"`.
Which then means sccache thinks `C:\Program` is the compiler, which of course doesn't make sense. 

For this to work correctly, we'd need #847 .
But the sentiment seems to be that it's not an ideal solution.
The patch here can work with the current behavior, as spaces are supported if `CC` is a direct file path.
Just the `RUSTC_WRAPPER` wasn't supported in that case, which would still be useful to support.
It can speed up our CI builds by more than half!

NobodyXu:
Thanks I see, this is something I didn't realize could happen in production where path includes spaces.

NobodyXu:
It would be great if you could verify it works, then I would have the confidence of approving and merging this PR.

LeonMatthesKDAB:
Hi, of course.
We have tried this here: https://github.com/KDAB/cxx-qt/pull/820 (with a backported version of this patch in cc-rs 1.0.79)
And it seems to work.

For some reason our Windows CI doesn't gain much speed when running under ctest, which we haven't fully investigated.
But e.g. our macOS CI has speed up from ~12 min. to ~6 min :partying_face:

Be-ing:
> It would be great if you could verify it works, then I would have the confidence of approving and merging this PR.

This behavior should also be documented, though considering that none of the environment variable usage is documented, I don't think that should hold up this PR. I opened a new issue for that: #920.

NobodyXu (approved):
Thank you!

Issue #920, opened 2024-01-26
documentation on environment variables is lacking


Be-ing:
Currently, the documentation on what environment variables this crate reads and what effects they have is lacking. For example, the [try_flags_from_environment documentation](https://docs.rs/cc/latest/cc/struct.Build.html#method.try_flags_from_environment) merely says "Normally the cc crate will consult with the standard set of environment variables (such as CFLAGS and CXXFLAGS) to construct the compiler invocation." cc-rs uses more than `CFLAGS` and `CXXFLAGS`. How to use `CC`/`CXX` is not documented, nor how to use target-specific environment variables, nor how `CC`/`CXX` can specify compiler caches and compiler flags, nor how `RUSTC_WRAPPER` can be used for C & C++ compiler caching.

NobodyXu:
There are a few more from https://github.com/rust-lang/cc-rs/blob/8cf5455c22b297f95dcbbfc06e1dfd68c34e19e9/src/lib.rs#L3238 :

 - `TARGET`: set the default target if user does not specify one
 - `HOST`
 - `OPT_LEVEL`
 - `DEBUG`

NobodyXu:
Also

 - `CRATE_CC_NO_DEFAULTS` https://github.com/rust-lang/cc-rs/blob/8cf5455c22b297f95dcbbfc06e1dfd68c34e19e9/src/lib.rs#L1628
 - `CARGO_CFG_TARGET_FEATURE` https://github.com/rust-lang/cc-rs/blob/8cf5455c22b297f95dcbbfc06e1dfd68c34e19e9/src/lib.rs#L1713
 - `RUSTC_LINKER` https://github.com/rust-lang/cc-rs/blob/8cf5455c22b297f95dcbbfc06e1dfd68c34e19e9/src/lib.rs#L3080
 - `CROSS_COMPILE` https://github.com/rust-lang/cc-rs/blob/8cf5455c22b297f95dcbbfc06e1dfd68c34e19e9/src/lib.rs#L3083

NobodyXu:
There are also prefixed version of `CFLAGS`/`CXXFLAGS`

https://github.com/rust-lang/cc-rs/blob/8cf5455c22b297f95dcbbfc06e1dfd68c34e19e9/src/lib.rs#L2180

NobodyXu:
Also prefixed  in `CC` and `CXX`

[...]

PR #921, opened 2024-01-28
move documentation from README.md to lib.rs

Be-ing:
so it appears on https://docs.rs and inconsistencies aren't created between README.md and lib.rs

per discussion in https://github.com/rust-lang/cc-rs/issues/920

The old introductory text in lib.rs has been replaced with the text from README.md and the introductory paragraph revised. Otherwise no major changes to the text have been made.

I will follow this PR with more substantial edits to the text.


NobodyXu (approved):
Thanks, it LGTM but I would like feedback from others as well.

NobodyXu:
cc @thomcc I would like your review on this PR

NobodyXu:
I will merge this before next release to minimize the disruption.

BlackHoleFox:
potentially both could be kept by using the [doc = include_str!](https://github.com/slint-ui/slint/blob/dd8fefd6a21cb319e28f956b415e36322fa1a810/tools/viewer/main.rs#L4) trick. I think `cc`'s MSRV is high enough for that now?

Be-ing:
The trouble with `doc = include_str!("README.md")` is that https://github.com/rust-lang/cc-rs will show documentation for unreleased changes. I have been confused before by [another library](https://github.com/pdeljanov/Symphonia) doing that. Linking to https://docs.rs/cc doesn't have that issue; people will only see the documentation for published releases.

PR #922, opened 2024-01-29
fix broken intradoc links

Be-ing:


src/lib.rs
```diff
@@ -2883,7 +2883,7 @@ impl Build {
     /// Get the archiver that's in use for this configuration.
     ///
     /// This will return a result instead of panicing;
-    /// see [`get_archiver()`] for the complete description.
+    /// see [Self::get_archiver] for the complete description.
```
NobodyXu, line 2886: I thought the infra link needs to be quoted by `

Be-ing: That is not necessary, it's just for formatting.

NobodyXu: thx, though I prefer it to be quoted

NobodyXu (approved):
Thank you!

PR #925, opened 2024-01-29
move src/os_pipe.rs to src/os_pipe/mod.rs

Be-ing:


NobodyXu (approved):
Thanks, that does simplify the path in the `mod` declaration.

Issue #902, opened 2023-11-14
Compilation in a macOS GitHub runner fails with cc v1.0.84


twistedfall:
Hello!

For my crate `opencv` I see that starting with `cc` version 1.0.84 the macOS runners are failing: https://github.com/twistedfall/opencv-rust/actions/runs/6853515550

I must also note that I haven't used `cc` versions `1.0.80..=1.0.83` because of the another issue that was causing hangs due to the interaction of `jobserver` in both `cc` and `opencv` (https://github.com/rust-lang/cc-rs/issues/844)

The specific error is:
```
fatal error: 'limits' file not found
```
it is caused by the missing stdlib which clang warns about too:
```
clang: warning: include path for libstdc++ headers not found; pass '-stdlib=libc++' on the command line to use the libc++ standard library instead [-Wstdlibcxx-not-found]
```

As far as I can see between version 1.0.79 and 1.0.84 the C++ compiler is called with different arguments:
* `1.0.79`
```
"c++" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-2" "-fno-omit-frame-pointer" "-m64" "-arch" "x86_64" "-I" "/Users/runner/work/opencv-rust/opencv-rust/src_cpp" "-I" "/Users/runner/work/opencv-rust/opencv-rust/target/debug/build/opencv-fb8aea6a002a5c33/out" "-I" "." "-I" "/usr/local/opt/opencv/include/opencv4" "-Wall" "-Wextra" "-std=c++14" "-Wno-deprecated-declarations" "-Wno-deprecated-copy" "-Wno-unused-parameter" "-Wno-sign-compare" "-Wno-comment" "-Wno-unused-variable" "-Wno-ignored-qualifiers" "-Wno-return-type-c-linkage" "-Wno-overloaded-virtual" "-F/usr/local/opt/opencv/include/opencv4" "-o" "/Users/runner/work/opencv-rust/opencv-rust/target/debug/build/opencv-fb8aea6a002a5c33/out/d25e1c8da0c38997-core.o" "-c" "/Users/runner/work/opencv-rust/opencv-rust/target/debug/build/opencv-fb8aea6a002a5c33/out/core.cpp"
```
* `1.0.84`
```
[...]

PR #938, opened 2024-01-30
add Build::apple_deployment_target function

Be-ing:
flag_if_supported doesn't work for this because try_get_compiler adds the Apple deployment target argument before flags_supported is processed, so users can only append a second -mmacosx-version-min flag via flag_if_supported. To work around this, add a function specifically for this purpose.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,6 +88,7 @@ use tool::ToolFamily;
 /// documentation on each method itself.
 #[derive(Clone, Debug)]
 pub struct Build {
+    apple_deployment_target: Option<Arc<str>>,
     include_directories: Vec<Arc<Path>>,
     definitions: Vec<(Arc<str>, Option<Arc<str>>)>,
     objects: Vec<Arc<Path>>,
@@ -202,6 +203,7 @@ impl Build {
     /// [`compile`]: struct.Build.html#method.compile
     pub fn new() -> Build {
         Build {
+            apple_deployment_target: None,
             include_directories: Vec::new(),
             definitions: Vec::new(),
             objects: Vec::new(),
@@ -1727,7 +1729,7 @@ impl Build {
                             map_darwin_target_from_rust_to_compiler_architecture(target)
                         {
                             let deployment_target =
-                                self.apple_deployment_version(AppleOs::Ios, target, None);
+                                self.get_apple_deployment_target(AppleOs::Ios, target, None);
                             cmd.args.push(
                                 format!(
                                     "--target={}-apple-ios{}-simulator",
@@ -1741,7 +1743,7 @@ impl Build {
                             map_darwin_target_from_rust_to_compiler_architecture(target)
                         {
                             let deployment_target =
-                                self.apple_deployment_version(AppleOs::WatchOs, target, None);
+                                self.get_apple_deployment_target(AppleOs::WatchOs, target, None);
                             cmd.args.push(
                                 format!(
                                     "--target={}-apple-watchos{}-simulator",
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3321,7 +3333,7 @@ impl Build {
         Ok(ret)
     }
 
-    fn apple_deployment_version(
+    fn get_apple_deployment_target(
```
NobodyXu, line 3336: Can we change the return type to `Cow<'_, str>` to avoid cloning `str`?

NobodyXu (changes requested):

NobodyXu:
This seems like related to #902

NobodyXu:
cc @Be-ing if you don't mind it, I can update the PR for you.

Be-ing:
I did some more testing, and the existing `flag_if_supported` API does work. I am not sure exactly how the CLI arguments are handled by clang. When `-mmacosx-version-min` is passed twice (first time automatically by cc, second time by user-specified `flag_if_supported`), it seems to either prefer the second argument or the higher specified version. It might still be a good idea to add this API though because that behavior of clang is not documented and I don't know if it can be relied on long term. Thoughts?

NobodyXu:
IMO it's better to add this API, not just due to the order of cmdline arg, but cc-rs seems to have different behavior based on apple deployment target version and build target.

Be-ing:
> cc-rs seems to have different behavior based on apple deployment target version and build target

Yes, but that can be addressed using the existing `flag_if_supported` API, so should we add a new API?

NobodyXu:
If that can be reliably solved by `flag_if_supported()` on all compilers, then it is indeed unnecessary.

But I'm not sure if it would work on gcc or msvc.

src/lib.rs
```diff
@@ -3426,15 +3439,18 @@ impl Build {
                 }),
 
             AppleOs::Ios => deployment_from_env("IPHONEOS_DEPLOYMENT_TARGET")
+                .or_else(|| self.apple_deployment_target.as_ref().map(|a| a.to_string()))
                 .or_else(|| rustc_provided_target(rustc, target))
                 .map(maybe_cpp_version_baseline)
                 .unwrap_or_else(|| OLD_IOS_MINIMUM_VERSION.into()),
 
-            AppleOs::WatchOs => deployment_from_env("WATCHOS_DEPLOYMENT_TARGET")
+            AppleOs::WatchOs => deployment_from_env("_DEPLOYMENT_TARGET")
```
BlackHoleFox, line 3447: This looks like an error maybe?

Be-ing: whoops, yeah that was an accident

BlackHoleFox:
@Be-ing On the contrary I think this should still be added. `MACOSX_DEPLOYMENT_TARGET` only works if you either set it outside of the process in a Makefile, cargo env config, etc or use `std::env::set_var` at runtime inside `build.rs`. 

That last option is the only programmatic way to do it and manipulating the process' environment variables is an uh, [dicey area](github.com/rust-lang/rust/issues/90308). Having a proper library API to control this for each target would be nice imo.

Be-ing:
Now that the old behavior of defaulting to the SDK's default deployment target has been restored, what is the need for this API? The default is a recent deployment target, so old code builds fine. A crate specifying a lower minimum than the default wouldn't really change anything.

BlackHoleFox:
> A crate specifying a lower minimum than the default wouldn't really change anything.

What do you mean?

Be-ing:
If a crate specifies a lower deployment target than the default, it's effectively superfluous because it would have built anyway. Only if *every* crate in the dependency graph that uses cc specifies the same lower deployment target will the resulting binary actually be usable on the lower deployment target. That's best achieved via the environment variable because that effects every crate that uses cc.

BlackHoleFox:
Seems like a reasonable take, the well-stated counter-argument is appreciated :)

PR #939, opened 2024-01-30
Add support for riscv32imac-esp-espidf

madmo:
esp32c6 and esp32h2 use the riscv32imac-esp-espidf target

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3048,6 +3048,7 @@ impl Build {
                 "riscv64-unknown-elf",
                 "riscv-none-embed",
             ]),
+            "riscv32imac-esp-espidf" => Some("riscv32-esp-elf"),
             "riscv32imac-unknown-none-elf" => self.find_working_gnu_prefix(&[
                 "riscv32-unknown-elf",
                 "riscv64-unknown-elf",


NobodyXu (commented):
Thanks lgtm, but I'm not very familiar with riscv targets, so I would also like review from *-esp-espidf target maintainers @ivmarkov @MabezDev

ivmarkov:
Hmmm, let me take a deeper look tomorrow. For one, I don't see there a definition of the older `riscv32imc-esp-espidf` target (which is used for esp32c3). Or am I blind? :)

@madmo Can you elaborate a bit what is your use case?

madmo:
> Hmmm, let me take a deeper look tomorrow. For one, I don't see there a definition of the older `riscv32imc-esp-espidf` target (which is used for esp32c3). Or am I blind? :)
> 
> @madmo Can you elaborate a bit what is your use case?

@ivmarkov riscv32imc-esp-espidf is some 10 lines down in the same file, but was already there. Probably because the c3 was released way earlier.

Sure, my use case is that I have a cargo based esp32c6 project which is mostly rust code :partying_face:. For some tiny functionality I want to re use a existing single file C library. Switching the whole project to a CMake based build feels overkill.

ivmarkov:
> > Hmmm, let me take a deeper look tomorrow. For one, I don't see there a definition of the older `riscv32imc-esp-espidf` target (which is used for esp32c3). Or am I blind? :)
> > @madmo Can you elaborate a bit what is your use case?
> 
> @ivmarkov riscv32imc-esp-espidf is some 10 lines down in the same file, but was already there. Probably because the c3 was released way earlier.

@madmo Ah yes, I'm blind indeed. :)

If I could ask you a favor: while we are at it, would you mind extending your PR to also include mapping for the `riscv32imafc-esp-espidf` target as well? This target was [upstreamed recently](https://github.com/rust-lang/rust/commit/43ce53375c976ad31629282e17ca7861c0d23a2c), but we forgot about `cc-rs` there as well. Thanks in advance!

> 
> Sure, my use case is that I have a cargo based esp32c6 project which is mostly rust code 🥳. For some tiny functionality I want to re use a existing single file C library. Switching the whole project to a CMake based build feels overkill.

@NobodyXu I'm fine with the change - and as per above - I would also suggest we upstream support for the `riscv32imafc-esp-espidf` too.

NobodyXu (approved):

NobodyXu:
Thank you @madmo @ivmarkov !

> and as per above - I would also suggest we upstream support for the riscv32imafc-esp-espidf too.

I would accept a separate PR adding support to it!

PR #940, opened 2024-01-31
Forward lines from stderr of child process to stdout on the same thread, instead of spawning a thread

### LINKED ISSUE #913: Upgrading to cc@1.0.84 breaks libstd bootstrap
Right now, rust-lang/rust [uses cc@1.79](https://github.com/rust-lang/rust/blob/3d0e6bed600c0175628e96f1118293cf44fb97bd/Cargo.lock#L439-L443). If I attempt to upgrade to cc@1.84, I get a bunch of errors from bootstrap.

<details>

```
Building stage0 library artifacts (x86_64-unknown-linux-gnu)
  Downloaded cc v1.0.84
  Downloaded 1 crate (74.9 KB) in 0.53s
   Compiling libc v0.2.150
   Compiling core v0.0.0 (/home/wesley/rust/library/core)
   Compiling memchr v2.5.0
   Compiling std v0.0.0 (/home/wesley/rust/library/std)
error: unexpected `cfg` condition name
  --> library/core/src/num/mod.rs:29:11
   |
29 | #[cfg(not(no_fp_fmt_parse))]
   |           ^^^^^^^^^^^^^^^
   |
   = note: `-D unexpected-cfgs` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unexpected_cfgs)]`

error: unexpected `cfg` condition name
  --> library/core/src/num/mod.rs:31:11
   |
31 | #[cfg(not(no_fp_fmt_parse))]
   |           ^^^^^^^^^^^^^^^

error: unexpected `cfg` condition name
  --> library/core/src/num/mod.rs:33:11
   |
33 | #[cfg(not(no_fp_fmt_parse))]
   |           ^^^^^^^^^^^^^^^

error: unexpected `cfg` condition name
  --> library/core/src/num/mod.rs:35:11
   |
35 | #[cfg(not(no_fp_fmt_parse))]
   |           ^^^^^^^^^^^^^^^

error: unexpected `cfg` condition name
  --> library/core/src/num/mod.rs:56:11
   |
56 | #[cfg(not(no_fp_fmt_parse))]
   |           ^^^^^^^^^^^^^^^

error: unexpected `cfg` condition value
   --> library/core/src/mem/mod.rs:733:19
    |
733 |     #[cfg(not(any(target_arch = "spirv")))]
    |                   ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `target_arch` are: `aarch64`, `arm`, `avr`, `bpf`, `csky`, `hexagon`, `loongarch64`, `m68k`, `mips`, `mips32r6`, `mips64`, `mips64r6`, `msp430`, `nvptx64`, `powerpc`, `powerpc64`, `riscv32`, `riscv64`, `s390x`, `sparc`, `sparc64`, `wasm32`, `wasm64`, `x86`, `x86_64`

error: unexpected `cfg` condition value
   --> library/core/src/ffi/mod.rs:244:5
    |
244 |     target_arch = "asmjs",
    |     ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `target_arch` are: `aarch64`, `arm`, `avr`, `bpf`, `csky`, `hexagon`, `loongarch64`, `m68k`, `mips`, `mips32r6`, `mips64`, `mips64r6`, `msp430`, `nvptx64`, `powerpc`, `powerpc64`, `riscv32`, `riscv64`, `s390x`, `sparc`, `sparc64`, `wasm32`, `wasm64`, `x86`, `x86_64`

error: unexpected `cfg` condition value
   --> library/core/src/ffi/mod.rs:273:5
    |
273 |     target_arch = "asmjs",
    |     ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `target_arch` are: `aarch64`, `arm`, `avr`, `bpf`, `csky`, `hexagon`, `loongarch64`, `m68k`, `mips`, `mips32r6`, `mips64`, `mips64r6`, `msp430`, `nvptx64`, `powerpc`, `powerpc64`, `riscv32`, `riscv64`, `s390x`, `sparc`, `sparc64`, `wasm32`, `wasm64`, `x86`, `x86_64`

error: unexpected `cfg` condition value
   --> library/core/src/ffi/mod.rs:434:5
    |
434 |     target_arch = "asmjs",
    |     ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `target_arch` are: `aarch64`, `arm`, `avr`, `bpf`, `csky`, `hexagon`, `loongarch64`, `m68k`, `mips`, `mips32r6`, `mips64`, `mips64r6`, `msp430`, `nvptx64`, `powerpc`, `powerpc64`, `riscv32`, `riscv64`, `s390x`, `sparc`, `sparc64`, `wasm32`, `wasm64`, `x86`, `x86_64`

error: unexpected `cfg` condition value
   --> library/core/src/ffi/mod.rs:464:9
    |
464 |     not(target_arch = "asmjs"),
    |         ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `target_arch` are: `aarch64`, `arm`, `avr`, `bpf`, `csky`, `hexagon`, `loongarch64`, `m68k`, `mips`, `mips32r6`, `mips64`, `mips64r6`, `msp430`, `nvptx64`, `powerpc`, `powerpc64`, `riscv32`, `riscv64`, `s390x`, `sparc`, `sparc64`, `wasm32`, `wasm64`, `x86`, `x86_64`

error: unexpected `cfg` condition value
   --> library/core/src/ffi/mod.rs:398:9
    |
398 |         target_arch = "asmjs",
    |         ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `target_arch` are: `aarch64`, `arm`, `avr`, `bpf`, `csky`, `hexagon`, `loongarch64`, `m68k`, `mips`, `mips32r6`, `mips64`, `mips64r6`, `msp430`, `nvptx64`, `powerpc`, `powerpc64`, `riscv32`, `riscv64`, `s390x`, `sparc`, `sparc64`, `wasm32`, `wasm64`, `x86`, `x86_64`

error: unexpected `cfg` condition value
   --> library/core/src/ffi/mod.rs:416:13
    |
416 |         not(target_arch = "asmjs"),
    |             ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `target_arch` are: `aarch64`, `arm`, `avr`, `bpf`, `csky`, `hexagon`, `loongarch64`, `m68k`, `mips`, `mips32r6`, `mips64`, `mips64r6`, `msp430`, `nvptx64`, `powerpc`, `powerpc64`, `riscv32`, `riscv64`, `s390x`, `sparc`, `sparc64`, `wasm32`, `wasm64`, `x86`, `x86_64`

error: unexpected `cfg` condition name
  --> library/core/src/fmt/mod.rs:16:11
   |
16 | #[cfg(not(no_fp_fmt_parse))]
   |           ^^^^^^^^^^^^^^^

error: unexpected `cfg` condition name
  --> library/core/src/fmt/mod.rs:18:7
   |
18 | #[cfg(no_fp_fmt_parse)]
   |       ^^^^^^^^^^^^^^^

error: unexpected `cfg` condition name
   --> library/core/src/../../stdarch/crates/core_arch/src/x86/mod.rs:782:11
    |
782 | #[cfg(not(stdarch_intel_sde))]
    |           ^^^^^^^^^^^^^^^^^

error: unexpected `cfg` condition name
   --> library/core/src/../../stdarch/crates/core_arch/src/x86/mod.rs:784:11
    |
784 | #[cfg(not(stdarch_intel_sde))]
    |           ^^^^^^^^^^^^^^^^^

error: unexpected `cfg` condition name
   --> library/core/src/../../stdarch/crates/core_arch/src/x86/mod.rs:787:11
    |
787 | #[cfg(not(stdarch_intel_sde))]
    |           ^^^^^^^^^^^^^^^^^

error: unexpected `cfg` condition name
   --> library/core/src/../../stdarch/crates/core_arch/src/x86/mod.rs:789:11
    |
789 | #[cfg(not(stdarch_intel_sde))]
    |           ^^^^^^^^^^^^^^^^^

error: could not compile `core` (lib) due to 18 previous errors
```

</details>

I believe the issue is the dependency on `libc`

https://github.com/rust-lang/cc-rs/blob/802399f3fab0820707dc8e562f206594e9a9e1db/Cargo.toml#L24

which does not activate the `rustc-dep-of-std` feature in this case.

dpaoliello:
`cc-rs` was redirecting output from `stderr` of its child processes to a pipe, which was then being read on dedicated thread and written to `stdout` with a wrapper message.

This pattern caused a few issues:
* It created a dependency from `cc-rs` to `libc` when building the Rust standard library with optimized compiler intrinsics, resulting in the Rust build breaking (see #913).
* With the `parallel` feature enabled, there was no synchronization between different child processes writing to the pipe, potentially resulting in corrupted output.
* The thread that was forwarding the lines was wasted, since the main thread was blocking waiting for the child process(es) to exit anyway.

The fix for this is to forward output from `stderr` to `stdout` on the current thread. For the `parallel` feature this required implementing a non-blocking version of `BufReader::read_until` that consumes whatever data is available and then continues.

Fixes #913


NobodyXu:
There're a lot of changes here, it might take me some time to review it.

dpaoliello:
If it helps, I have a build of the Rust toolchain that uses this: https://github.com/rust-lang/rust/actions/runs/7719025147/job/21041489650?pr=119199

src/command_helpers.rs
```diff
@@ -41,83 +40,55 @@ impl CargoOutput {
...
-pub(crate) struct PrintThread {
-    handle: Option<JoinHandle<()>>,
-    pipe_writer: Option<File>,
-}
 
-impl PrintThread {
-    pub(crate) fn new() -> Result<Self, Error> {
-        let (pipe_reader, pipe_writer) = crate::os_pipe::pipe()?;
+    #[cfg(feature = "parallel")]
+    pub(crate) fn reader_for_stderr_as_warnings(
```
NobodyXu, line 52: I think it doesn't have to be an associated fn of CargoOutput, child.stdout() already has the information required.

NobodyXu (changes requested):

src/command_helpers.rs
```diff
@@ -41,83 +40,55 @@ impl CargoOutput {
...
+            .then(|| BufReader::with_capacity(4096, child.stderr.take().unwrap()))
+    }
 
-        // Capture the standard error coming from compilation, and write it out
-        // with cargo:warning= prefixes. Note that this is a bit wonky to avoid
-        // requiring the output to be UTF-8, we instead just ship bytes from one
-        // location to another.
-        let print = thread::spawn(move || {
-            let mut stderr = BufReader::with_capacity(4096, pipe_reader);
+    fn forward_stderr_as_warnings(&self, child: &mut Child) {
```
NobodyXu, line 60: Same here, it doesn't have to be an associated fn.

src/command_helpers.rs
```diff
@@ -307,8 +263,27 @@ pub(crate) fn try_wait_on_child(
...
-    match child.try_wait() {
+    // Check the child status THEN forward stderr messages, so that we don't race
+    // between the child printing messages and then exiting.
+    let wait_result = child.try_wait();
+
+    if let Some(child_stderr_reader) = child_stderr_reader.as_mut() {
+        if let Ok(mut available) = child_stderr_reader.fill_buf() {
+            let stdout = io::stdout();
+            let original_size = available.len();
+            while let Some(i) = memchr::memchr(b'\n', available) {
```
NobodyXu, line 276: I'd prefer not to add dep `memchr`.

src/command_helpers.rs
```diff
@@ -307,8 +263,27 @@ pub(crate) fn try_wait_on_child(
...
     child: &mut Child,
     stdout: &mut dyn io::Write,
+    child_stderr_reader: &mut Option<BufReader<std::process::ChildStderr>>,
 ) -> Result<Option<()>, Error> {
-    match child.try_wait() {
+    // Check the child status THEN forward stderr messages, so that we don't race
+    // between the child printing messages and then exiting.
+    let wait_result = child.try_wait();
+
+    if let Some(child_stderr_reader) = child_stderr_reader.as_mut() {
```
NobodyXu, line 272: I'm thinking if it makes more sense to do this in the async runtime, i.e. if we add poll/select/epoll/waitfor_multiple_objects in the future to speed it up, then it makes sense to do it there.

However since it'd be done in the future, you don't have to change this code.

Some refactoring to extract a new fn is enough.

NobodyXu (changes requested):

src/command_helpers.rs
```diff
@@ -41,83 +40,51 @@ impl CargoOutput {
...
-    handle: Option<JoinHandle<()>>,
-    pipe_writer: Option<File>,
+#[cfg(feature = "parallel")]
+pub(crate) fn reader_for_stderr_as_warnings(
+    child: &mut Child,
+) -> Option<BufReader<std::process::ChildStderr>> {
+    child
+        .stderr
+        .take()
+        .map(|stderr| BufReader::with_capacity(4096, stderr))
```
NobodyXu, line 59: I feel like we don't need 4096 B for the buffer, IMO I rarely see any warning for my project using cc-rs indirectly, and even if it does it's unlikely to print 4096B of characters.

If that happens, it will likely be a compilation error in C++ which dumps a lot of messages due to overloading.

It's so rare and usually signals a failure, so I don't think we need to optimize for that particular use case.

We should instead tune it down to save memory, allocating a whole page is going to be quite expensive since it's large.

I'd recommend something like 100B, should be good enough.

src/command_helpers.rs
```diff
@@ -41,83 +40,51 @@ impl CargoOutput {
...
-            while stderr.read_until(b'\n', &mut line).unwrap() != 0 {
-                {
-                    let mut stdout = stdout.lock();
-
-                    stdout.write_all(b"cargo:warning=").unwrap();
-                    stdout.write_all(&line).unwrap();
-                    stdout.write_all(b"\n").unwrap();
-                }
+fn forward_all_stderr_as_warnings(child: &mut Child) {
+    if let Some(stderr) = child.stderr.as_mut() {
```
NobodyXu, line 63: Let's use `reader_for_stderr_as_warning` here, to reuse buffer configuration.

src/command_helpers.rs
```diff
@@ -41,83 +40,51 @@ impl CargoOutput {
...
-impl Drop for PrintThread {
-    fn drop(&mut self) {
-        // Drop pipe_writer first to avoid deadlock
-        self.pipe_writer.take();
+fn write_warning(stdout: &Stdout, line: &[u8]) {
+    let mut stdout = stdout.lock();
 
-        self.handle.take().unwrap().join().unwrap();
-    }
+    stdout.write_all(b"cargo:warning=").unwrap();
```
NobodyXu, line 81: Shame that we don't have write_all_vectored

src/command_helpers.rs
```diff
@@ -301,14 +253,38 @@ pub(crate) fn command_add_output_file(
...
 
+#[cfg(feature = "parallel")]
+fn forward_available_stderr_as_warnings(
+    child_stderr_reader: &mut Option<BufReader<std::process::ChildStderr>>,
+) {
+    if let Some(child_stderr_reader) = child_stderr_reader.as_mut() {
+        if let Ok(available) = child_stderr_reader.fill_buf() {
+            let stdout = io::stdout();
+            let mut consumed = 0;
+            for line in available.split(|&b| b == b'\n') {
```
NobodyXu, line 264: I think we need to handle cases where `available` does not include a newline and there's more data to come.

In this case, I think you should keep the data until newline or eof, because if you prints part of the line as a warning, then you would be adding newline to the originaly message.

NobodyXu (changes requested):

src/command_helpers.rs
dpaoliello: I'll switch this to `BufReader::new` instead of explicitly setting a capacity.

src/command_helpers.rs
dpaoliello: Ah, yep, I forgot that `split` will provide the rest of the items in last iteration.

src/command_helpers.rs
```diff
@@ -301,14 +248,41 @@ pub(crate) fn command_add_output_file(
...
+fn forward_available_stderr_as_warnings(
+    child_stderr_reader: &mut Option<BufReader<std::process::ChildStderr>>,
+) {
+    if let Some(child_stderr_reader) = child_stderr_reader.as_mut() {
+        if let Ok(available) = child_stderr_reader.fill_buf() {
+            let stdout = io::stdout();
+            let mut consumed = 0;
+            for line in available.split_inclusive(|&b| b == b'\n') {
+                // Only forward complete lines, leave the rest in the buffer.
+                if let Some((b'\n', line)) = line.split_last() {
```
NobodyXu, line 261: Isn't it more efficient to check the last byte of the buffer directly?

src/command_helpers.rs
```diff
@@ -301,14 +248,41 @@ pub(crate) fn command_add_output_file(
...
+                    consumed += line.len() + 1;
+                    write_warning(&stdout, line);
```
NobodyXu, line 263: I think you also need to handle incomplete line on eof.

Tip: You could convert this function to an async one to make it easier to write, compile_objects does have some async support if you follow the example

NobodyXu (changes requested):

src/command_helpers.rs
```diff
@@ -41,83 +40,46 @@ impl CargoOutput {
...
     }
 }
 
-pub(crate) struct PrintThread {
-    handle: Option<JoinHandle<()>>,
-    pipe_writer: Option<File>,
+pub(crate) fn reader_for_stderr_as_warnings(
+    child: &mut Child,
+) -> Option<BufReader<std::process::ChildStderr>> {
+    child.stderr.take().map(BufReader::new)
```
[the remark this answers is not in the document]
dpaoliello: Ok, back to 100b buffer.

The `BufReader` makes things more convenient, otherwise if we were using the `ChildStderr` directly we'd need to shuffle any remaining bytes after the final `\n` back to the beginning of the `Vec` between each `read` OR we would have to `read_to_end` which could result in a very large `Vec`.

src/command_helpers.rs
dpaoliello: `write_warning` doesn't expect the line to end with `\n`, so I'd have to trim it off anyway. Looking at the implementation for `split_last`, it doesn't seem that expensive.

src/command_helpers.rs
dpaoliello: Ugh, I think the only way that I can handle that is to use `forward_all_stderr_as_warnings` once the process has exited, otherwise there's no way with `BufReader` to know if the underlying data source still has data left or has hit EOF without consuming all the bytes currently in the buffer.

dpaoliello: I also don't know how async would help here...

src/command_helpers.rs
NobodyXu: Thanks, I misunderstood the API of split_last

src/command_helpers.rs
NobodyXu: async would make handling internal state easier, so instead of using BufReader you can just set to childstderr to non-blocking, keep reading from the child into a Vec and find newline until it returns `EWOULDBLOCK`, then yield control until it resumes.

src/command_helpers.rs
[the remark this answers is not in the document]
NobodyXu: > otherwise if we were using the ChildStderr directly we'd need to shuffle any remaining bytes after the final \n back 

Hmm I think this isn't very hard with [`Vec::drain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain), just two lines for splitting based on newline and remove it using drain after printing.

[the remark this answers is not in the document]
dpaoliello: This feels like a premature optimization: we'd have to choose a size for the `Vec` to read into, which is no different to the buffer in the `BufReader`. It would avoid the second allocation, but this is in the context of spawning and waiting for a process so the execution time is dominated by that rather than the work here. I'd prefer to keep the code simple than make the optimization to avoid a `BufReader`. 

We've already made improvements: 100b for the `BufReader` instead of 4kb and no allocations for the `Vec` if there's no warnings.

src/command_helpers.rs
dpaoliello: I don't see how to do this with the API exposed by `ChildStdErr`, unless you mean to convert it into its underlying fd or handle, which doesn't seem portable.

src/command_helpers.rs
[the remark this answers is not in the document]
NobodyXu: > It would avoid the second allocation, but this is in the context of spawning and waiting for a process so the execution time is dominated by that rather than the work here. I'd prefer to keep the code simple than make the optimization to avoid a `BufReader`. 

Understood, that is indeed the case.

> We've already made improvements: 100b for the `BufReader` instead of 4kb and no allocations for the `Vec` if there's no warnings. 

Yeah that's probably good enough.

Thanks.

src/command_helpers.rs
NobodyXu: That's true...it's only possible if you create the pipe manually.

Perhaps just add a detection of eof and then flush all pending data in the buffer?

dpaoliello:
@NobodyXu I'm going to have to rework this, since there are two issues:

The first, fixable, issue is that `BufReader.fill_buf` will only `read` more if all data in the buffer has been consumed - so if we don't consume everything then `fill_buf` won't read any more data and so each subsequent call to `fill_buf` will keep returning the same data. 

The second, fundamental, issue is that `read` will block until data is available or the child exits - so in parallel mode we'll block waiting for the first child's `stderr` to return something or until that process exits.

There's a few different ways that we can address these:
1. Accept the limitations of the second issue, and workaround the first issue by replacing the `BufReader` with our own calls to `read` into a `Vec` (as you had suggested for the non-parallel code).
2. Give up on trying to forward the `stderr` output while the children are running, and instead forward them all once the child has exited.
3. Do something like the original code but create one thread per child instead of using a single thread.

Thoughts?

NobodyXu:
> The second, fundamental, issue is that read will block until data is available or the child exits - so in parallel mode we'll block waiting for the first child's stderr to return something or until that process exits.

If you bring back os_pipe, then you could read without blocking.

dpaoliello:
@NobodyXu ok, I think I have this figured out: for Unix, I'm setting the pipe as non-blocking and checking for `WouldBlock`. For Windows, I'm using `PeekNamedPipe` to check for data and returning early if there is none.

I also removed the `BufReader` and switched to directly reading into the `Vec`.

src/windows/windows_sys.rs
```diff
@@ -55,16 +55,18 @@ extern "system" {
 }
 #[link(name = "kernel32")]
 extern "system" {
-    pub fn CreatePipe(
-        hreadpipe: *mut HANDLE,
-        hwritepipe: *mut HANDLE,
-        lppipeattributes: *const SECURITY_ATTRIBUTES,
-        nsize: u32,
-    ) -> BOOL;
+    pub fn OpenSemaphoreA(dwdesiredaccess: u32, binherithandle: BOOL, lpname: PCSTR) -> HANDLE;
```
NobodyXu, line 58: Please update https://github.com/rust-lang/cc-rs/blob/main/dev-tools/gen-windows-sys-binding/windows_sys.list and then run the crate to generate this file, this file is not meant for modification directly.

src/command_helpers.rs
```diff
@@ -41,83 +40,175 @@ impl CargoOutput {
...
+                stdout.write_all(b"\n").unwrap();
+            };
 
-                    stdout.write_all(b"cargo:warning=").unwrap();
-                    stdout.write_all(&line).unwrap();
-                    stdout.write_all(b"\n").unwrap();
+            #[cfg(all(windows, feature = "parallel"))]
+            let is_non_blocking = self.is_non_blocking;
+            let mut read_stderr = move |buf: &mut [u8]| -> Result<usize, io::Error> {
+                // On Unix, the pipe is non-blocking, so we can just read.
```
NobodyXu, line 88: You need to set the pipe to non-blocking on unix, it's not non-blocking by default.

src/command_helpers.rs
```diff
@@ -41,83 +40,175 @@ impl CargoOutput {
...
-
-        // Capture the standard error coming from compilation, and write it out
-        // with cargo:warning= prefixes. Note that this is a bit wonky to avoid
-        // requiring the output to be UTF-8, we instead just ship bytes from one
-        // location to another.
-        let print = thread::spawn(move || {
-            let mut stderr = BufReader::with_capacity(4096, pipe_reader);
-            let mut line = Vec::with_capacity(20);
-            let stdout = io::stdout();
+#[cfg(all(feature = "parallel", not(unix), not(windows)))]
```
NobodyXu, line 58: At this point, why not bring back os_pipe mod and move the unsafe function there?

NobodyXu (changes requested):

src/windows/windows_sys.rs
dpaoliello: Ah, I was wondering where that file was - thanks!

src/command_helpers.rs
dpaoliello: It's set to non-blocking by `set_non_blocking` - I've updated the comment to match and added a debug assert to check the flags.

src/command_helpers.rs
dpaoliello: I'll move these unsafe functions into their own module under `parallel`, but there's no need to resurrect the rest of `os_pipe`.

src/parallel/stderr.rs
```diff
@@ -0,0 +1,84 @@
...
+pub fn set_non_blocking(stderr: &mut ChildStderr) -> Result<(), Error> {
+    // On Unix, switch the pipe to non-blocking mode.
+    // On Windows, we have a different way to be non-blocking.
+    #[cfg(unix)]
+    {
+        use std::os::unix::io::AsRawFd;
+        let fd = stderr.as_raw_fd();
+        let flags = get_flags(&fd)?;
+
+        if unsafe { libc::fcntl(fd, libc::F_SETFL, flags | libc::O_NONBLOCK) } != 0 {
```
NobodyXu, line 34: Honestly I think you don't need get_flags().

From the [man page of fcntl](https://man7.org/linux/man-pages/man2/fcntl.2.html)

>  On Linux, this command can change
>               only the O_APPEND, O_ASYNC, O_DIRECT, O_NOATIME, and
>               O_NONBLOCK flags.  It is not possible to change the
>               O_DSYNC and O_SYNC flags; see BUGS, below.

It's a pipe on unix, so only O_NONBLOCK makes sense IMHO.

src/parallel/stderr.rs
```diff
@@ -0,0 +1,84 @@
...
+                null_mut(),
+                &mut bytes_available,
+                null_mut(),
+            ) == 0
+            {
+                return Err(std::io::Error::last_os_error());
+            }
+        }
+        if bytes_available == 0 {
+            return Err(std::io::Error::new(
```
NobodyXu, line 70: Recommend  to use <io::Error as From> impl for io::ErrorKind, io::Error::new allocates internally while the from errorkind does not.

NobodyXu (changes requested):

src/parallel/stderr.rs
dpaoliello: Fair enough - I've added debug asserts just to be safe.

src/command_helpers.rs
```diff
@@ -41,83 +40,122 @@ impl CargoOutput {
...
+                buffer.resize(buffer.capacity(), 0);
+                match read_stderr(stderr, &mut buffer[old_data_end..]) {
+                    Err(err) if err.kind() == std::io::ErrorKind::WouldBlock => {
+                        // No data currently, yield back.
+                        buffer.truncate(old_data_end);
+                        return false;
+                    }
+                    Err(err) if err.kind() == std::io::ErrorKind::Interrupted => {
+                        // Interrupted, try again.
+                        buffer.truncate(old_data_end);
```
NobodyXu, line 104: resize is honestly a bit expensive as it have to set the rest of the buffer to 0.

You can either:
 - move the resize outside the loop, and make sure every match returns.
 - use `MaybeUninit`, but given that the `std::io::Read` trait only takes `&[u8]`, it in theory can be a UB if the implementator of `Read` actually reads from it, instead of just writing to it.

src/command_helpers.rs
```diff
@@ -41,83 +40,122 @@ impl CargoOutput {
...
+                    }
+                    Ok(0) | Err(_) => {
+                        // End of stream: flush remaining data and bail.
+                        if old_data_end > 0 {
+                            write_warning(&buffer[..old_data_end]);
+                        }
+                        return true;
```
NobodyXu, line 111: Let's set `self.inner = 0;` here, so that once `true` is returned, it will always returns `true` in the future (because in parallel `compile_objects`, it can be called after `true` is returned).

NobodyXu (changes requested):

src/command_helpers.rs
```diff
@@ -41,83 +40,122 @@ impl CargoOutput {
...
+            #[cfg(feature = "parallel")]
+            is_non_blocking: false,
+        }
+    }
+
+    fn forward_available(&mut self) -> bool {
+        if let Some((stderr, buffer)) = self.inner.as_mut() {
+            fn write_warning(line: &[u8]) {
+                let stdout = io::stdout();
+                let mut stdout = stdout.lock();
```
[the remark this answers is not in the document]
dpaoliello: Unfortunately this causes errors with older versions of Rust: https://github.com/rust-lang/cc-rs/actions/runs/7764282806/job/21177496391

src/command_helpers.rs
```diff
@@ -41,83 +40,122 @@ impl CargoOutput {
...
+            let read_stderr = if self.is_non_blocking {
+                crate::parallel::stderr::read_non_blocked
+            } else {
+                ChildStderr::read
+            };
+            #[cfg(not(feature = "parallel"))]
+            let read_stderr = ChildStderr::read;
+
+            loop {
+                buffer.reserve(MIN_BUFFER_CAPACITY);
```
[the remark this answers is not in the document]
dpaoliello: If the buffer is full but we haven't seen a `\n` yet, then we need to expand it to read more to find that `\n` (since we can't write a partial line).

src/command_helpers.rs
dpaoliello: I switched this to `set_len` since I'm always truncating to the initialized data after reading - I can't move the `resize` outside of the loop (in case we've filled the buffer but haven't seen a `\n`), and I don't think that `MaybeUninit` supports dynamic sized buffers.

src/command_helpers.rs
dpaoliello: Good call.

src/command_helpers.rs
NobodyXu: I was thinking of `Vec<MaybeUninit<u8>>`

dpaoliello: I don't think that works, `MaybeUninit` isn't guaranteed to be the same layout as the underlying type: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#layout-1 so you can't get a `&[u8]` from a `Vec<MaybeUninit<u8>>`.

NobodyXu: Thanjs, I didn't know that.

src/parallel/stderr.rs
```diff
@@ -0,0 +1,74 @@
...
+                0,
+                null_mut(),
+                &mut bytes_available,
+                null_mut(),
+            ) == 0
+            {
+                return Err(std::io::Error::last_os_error());
+            }
+        }
+        if bytes_available == 0 {
```
NobodyXu, line 58: I wonder if it makes sense to pass the `Vec` here, then `.reserve(bytes_available)` to make sure all remaining data is read.

src/parallel/stderr.rs
```diff
@@ -0,0 +1,74 @@
...
+            return Err(std::io::ErrorKind::WouldBlock.into());
+        }
+    }
+    // On Unix, the pipe set to non-blocking by set_non_blocking, so we can just read.
+    #[cfg(unix)]
+    {
+        use std::os::unix::io::AsRawFd;
```
NobodyXu, line 65: Nit: ioctl supports FIONREAD on non-blocking pipe, to retrieve data left for reading.

NobodyXu (changes requested):
Thanks, it overall LGTM, just two nitpicks I want to discuss with you before I approve and merge it.

src/parallel/stderr.rs
dpaoliello: Oh, nice! this really simplifies the code since Windows and Unix can then use the same pattern. Thanks!

src/parallel/stderr.rs
dpaoliello: With the change to use `FIONREAD`, I've modified the code to split checking available data from doing the actual read, so it's pretty easy to reserve at least the bytes currently available.

src/command_helpers.rs
```diff
@@ -41,83 +40,110 @@ impl CargoOutput {
...
-        self.pipe_writer.take().unwrap()
-    }
-
-    /// # Panics
-    ///
-    /// Will panic if the pipe writer has already been taken.
-    pub(crate) fn clone_pipe_writer(&self) -> Result<File, Error> {
-        self.try_clone_pipe_writer().map(Option::unwrap)
+impl StderrForwarder {
+    pub(crate) fn new(child: &mut Child) -> Self {
```
NobodyXu, line 59: You still need to set the childstderr as non-blocking.

I read tht FIONREAD only works on pipe if it's non-blocking.

Also, I want to play safe and always use non-blocking to prevent deadlock.

src/parallel/stderr.rs
```diff
@@ -0,0 +1,47 @@
...
+                    "PeekNamedPipe failed with {}",
+                    std::io::Error::last_os_error()
+                ),
+            ));
+        }
+    }
+    #[cfg(unix)]
+    {
+        use std::os::unix::io::AsRawFd;
+        if unsafe { libc::ioctl(stderr.as_raw_fd(), libc::FIONREAD, &mut bytes_available) } != 0 {
```
NobodyXu, line 39: Stackoverflow says some implementation is broken and fail none-the-less.

I'd like it to fallback to reserving 100B on failure, and have an atomic for checking is the impl is buggy/supported.

NobodyXu (changes requested):

src/command_helpers.rs
dpaoliello: Ok, I resurrected this non-blocking code.

src/parallel/stderr.rs
dpaoliello: Ugh, ok.

I'm also noticing that I'm getting spurious 0-bytes available on Linux (Ubuntu 20.04 in WSL) as well...

dpaoliello: I'm also not a fan of disabling this across the entire process after a single pipe fails, so I've added a per-pipe flag if it fails or returns 0.

src/parallel/stderr.rs
```diff
@@ -0,0 +1,75 @@
...
+            return Err(Error::new(
+                ErrorKind::IOError,
+                format!("ioctl failed with {}", std::io::Error::last_os_error()),
+            ));
+        }
+    }
+    Ok(bytes_available as usize)
```
NobodyXu, line 74: I don't like `as` casting, please use `.try_into().unwrap()` so that it would fail if casting negative number to `usize`, or when casting number too large to fit in `usize`

NobodyXu (changes requested):

NobodyXu (approved):

NobodyXu:
Thanj you!

dpaoliello:
Can we please get a release with this change?

NobodyXu:
Yes, I definitely want to have a release.

There's a regression #902 that needs to be fixed before a release can be done.

PR #943, opened 2024-02-12
Fix default deployment target behavior for Apple targets

BlackHoleFox:
In hindsight defaulting to rustc's default deployment targets when no explicit ones were provided wasn't a great call. Those versions are older then what most people target and interact badly with third-party C libraries and Apple's SDKs that expect *usually* higher versions instead. So this goes ahead and reverts to `cc`'s older behavior of using the target's current platform SDK `DefaultDeploymentTarget` when no explicit version is given to `cc`. 

This afaict is functionally identical to passing nothing and letting `clang` figure out the default except this keeps the code cleaner because optionals don't need handled everywhere :)

Fixes https://github.com/rust-lang/cc-rs/issues/902


NobodyXu (approved):
Thank  you!

PR #947, opened 2024-02-13
CI: Run test on aarch64-apple-darwin

NobodyXu:
Also uses Swatinem/rust-cache@v2 to speedup compilation of dependencies.


BlackHoleFox (approved):
` Swatinem/rust-cache` seems widely used and well-maintained, but an actual `cc` maintainer should confirm what the dependency policy is. On that point though splitting the cache addition into its own commit would make it easier to see.

Amanieu (approved):

NobodyXu:
Thanks!

PR #951, opened 2024-02-16
Added tvos-sim support

ErikEverson:
This PR allows using cc-rs lib to target [tier 3 *-apple builds](https://doc.rust-lang.org/nightly/rustc/platform-support/apple-tvos.html) simulator builds

- The aarch64-apple-tvos-sim target was added so both tvos-sim and x86 versions are simulator versions.
- I removed the target.contains("aarch64-apple-tvos") as it seems to be redundant and iOS and watchOS implementations do not include it. The build was not effected with this removal.
- Added a new check-tvos-sim step to actions

how was this tested: 

- cargo test -Z build-std=std --no-run --workspace --target aarch64-apple-tvos{and -sim}
- cargo test -Z build-std=std --no-run --workspace --target aarch64-apple-tvos{and -sim} --release
- cargo test -Z build-std=std --no-run --workspace --target aarch64-apple-tvos{and -sim} --features parallel

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -95,41 +95,42 @@ jobs:
     - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --release
     - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --features parallel
 
-  # This is separate from the matrix above because there is no prebuilt rust-std component for the target.
+  # This is separate from the matrix above because there is no prebuilt rust-std component for these targets.
   check-tvos:
-    name: Test aarch64-apple-tvos
-    runs-on: macos-latest
+    name: Test build-std
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        build: [nightly, macos, aarch64-macos]
+        include:
+          - build: aarch64-tvos
+            os: macos-latest
+            rust: nightly
+            target: aarch64-apple-tvos
+            no_run: --no-run
+          - build: aarch64-ios
+            os: macos-latest
+            rust: nightly
+            target: aarch64-apple-tvos-sim
+            no_run: --no-run
+          - build: x86_64-tvos
+            os: macos-latest
+            rust: nightly
+            target: x86_64-apple-tvos
+            no_run: --no-run
     steps:
     - uses: actions/checkout@v4
     - name: Install Rust (rustup)
       run: |
         set -euxo pipefail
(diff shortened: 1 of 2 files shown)


ErikEverson:
CC: @phatblat for eventual visionOS support additions

src/lib.rs
```diff
@@ -1948,21 +1948,6 @@ impl Build {
...
-                            let sdk_details =
-                                apple_os_sdk_parts(AppleOs::TvOs, &AppleArchSpec::Device(""));
-                            let deployment_target = self.apple_deployment_version(
-                                AppleOs::TvOs,
-                                None,
-                                &sdk_details.sdk,
-                            );
-                            cmd.args.push(
-                                format!("--target={}-apple-tvos{}", arch, deployment_target).into(),
-                            );
```
BlackHoleFox, line 1964: This change seems to misalign with XCode at least. When building Objective-C with `clang` via XCode it passes the `-target` flag to convey the deployment target: `-target arm64-apple-tvos16.0` instead of the `os-version-min` flag we pass (which doesn't seem documented for tvOS/watchOS? ignoring that though...)

Did you try building any libs with this just to make sure?

.github/workflows/main.yml
```diff
@@ -113,6 +113,24 @@ jobs:
     - run: cargo test -Z build-std=std --no-run --workspace --target aarch64-apple-tvos --release
     - run: cargo test -Z build-std=std --no-run --workspace --target aarch64-apple-tvos --features parallel
 
+  # This is separate from the matrix above because there is no prebuilt rust-std component for the target.
```
Be-ing, line 116: Could you create a new matrix to combine this with the aarch64-apple-tvos test above?

src/lib.rs
ErikEverson: That is a good point. I will add back this code as everything worked fine either way and that will align it with XCode's behavior.
As for libraries that I built this on there are two. [Blake3](https://github.com/BLAKE3-team/BLAKE3) and [Ring](https://github.com/briansmith/ring). Both are dependencies of the project I am working on [Ditto](https://github.com/getditto/DittoSwiftPackage) which is included in my iPad/tvOS application/s.

.github/workflows/main.yml
ErikEverson: @Be-ing I updated to have a matrix for all -Z build-std targets. The only ones right now are tvOS targets but any tier 3 targets would fall in this same bucket as well.

BlackHoleFox (approved):

NobodyXu (approved):
Thank you!

PR #955, opened 2024-02-20
Optimize `set_non_blocking`: Avoid dup mono

NobodyXu:
`set_non_blocking` takes a generic parameter and it is used in two different places with different types.

Since it only uses the fd returned by `pipe.as_raw_fd()`, we could create an `_inner` function that takes the fd and do the job.

### DIFF
--- a/src/parallel/job_token/unix.rs
+++ b/src/parallel/job_token/unix.rs
@@ -3,10 +3,7 @@ use std::{
     fs::{self, File},
     io::{self, Read, Write},
     mem::ManuallyDrop,
-    os::{
-        raw::c_int,
-        unix::{ffi::OsStrExt, prelude::*},
-    },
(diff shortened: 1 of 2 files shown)


Amanieu:
Does this even make a noticeable impact? This is a single function that is only monomorphized twice, it's not like `HashMap` which has many methods and can be monomorphized for thousands of different types.It makes the code harder to read on the other hand.

NobodyXu:
It probably won't have much effect and I'm just being pedantic.

Though it really doesn't hurt, the code is quite simple and the change would make sure future use of it won't cause more monomorphisation.

Amanieu:
Even if there are a few more monomorphizations, the impact of compiling one extra monomorphizationwill be unnoticeable. I'm still inclined not to merge this.

NobodyXu:
Got it, I will close this as not planned.

PR #957, opened 2024-02-21
Support native ARM64 MSVC toolchain, and fallback to x64 if emulation is available

dpaoliello:
VS 17.4 introduced a native ARM64 toolchain (<https://devblogs.microsoft.com/visualstudio/arm64-visual-studio-is-officially-here/>), and Windows 11 supports running x64 binaries on ARM64 devices.

As such, we should try to use the ARM64 toolchain if it is available, fallback to x64 if support is available on the device (as performance is better than x86) and use x86 as a last resort.

This supersedes <https://github.com/rust-lang/cc-rs/pull/732> (but thanks to @Alovchin91 for working on that PR, as it is the basis for this PR).

Tested on a Surface Pro X running Windows 11.


src/windows/find_tools.rs
```diff
@@ -482,28 +493,42 @@ mod impl_ {
...
 
-        let host = match host_arch() {
-            X86 => "X86",
-            X86_64 => "X64",
-            // There is no natively hosted compiler on ARM64.
-            // Instead, use the x86 toolchain under emulation (there is no x64 emulation).
-            AARCH64 => "X86",
+        let hosts = match host_arch() {
+            X86 => &["X86"],
+            X86_64 => &["X64"],
```
dpaoliello, line 498: Question for reviewers: now that we have a "fallback" mechanism in place, should x64 fallback to x86?

NobodyXu: IMO think it's ok to have it, though given that x64 is widely available it's unlikely to be used.

src/windows/find_tools.rs
```diff
@@ -482,28 +493,42 @@ mod impl_ {
...
+        let bin_path = host_path.join(&target);
         // But! we also need PATH to contain the target directory for the host
         // architecture, because it contains dlls like mspdb140.dll compiled for
         // the host architecture.
-        let host_dylib_path = path
-            .join("bin")
-            .join(format!("Host{}", host))
-            .join(host.to_lowercase());
-        let lib_path = path.join("lib").join(target);
+        let host_dylib_path = host_path.join(&host.to_lowercase());
```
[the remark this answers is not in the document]
dpaoliello: No, because `host_path` is to the host directory which contains the subdirectories for each target, whereas `host_dyn_path` points to the target directory in there for the current host.

For example:
`host_path` = `...\bin\Hostx64`
`host_dyn_path` = `...\bin\Hostx64\x64`

src/windows/find_tools.rs
```diff
@@ -199,6 +203,31 @@ mod impl_ {
         include: Vec<PathBuf>,
     }
 
+    fn is_amd64_emulation_supported() -> bool {
+        // In theory we should free the library handle, but since kernel32.dll is loaded in every
+        // process and we're a short-lived process, leaking the handles doesn't matter.
```
NobodyXu, line 208: Some users use cc-rs at runtime instead of in `build.rs`, so that assumption isn't true.

src/windows/find_tools.rs
```diff
@@ -199,6 +203,31 @@ mod impl_ {
...
+        let kernel32 = unsafe { LoadLibraryA(b"kernel32.dll\0".as_ptr() as _) };
+        // GetMachineTypeAttributes is only available on Win11 22000+.
+        if let Some(get_machine_type_attributes) =
+            unsafe { GetProcAddress(kernel32, b"GetMachineTypeAttributes\0".as_ptr() as _) }
+        {
+            let mut attributes = Default::default();
+            let get_machine_type_attributes: unsafe extern "system" fn(
```
NobodyXu, line 215: Please split the type out as a type alias, it's a bit hard to read.

src/windows/find_tools.rs
```diff
@@ -199,6 +203,31 @@ mod impl_ {
...
+                u16,
+                *mut MACHINE_ATTRIBUTES,
+            ) -> HRESULT = unsafe { std::mem::transmute_copy(&get_machine_type_attributes) };
```
NobodyXu, line 218: Is it necessary to use `transmute` here?

NobodyXu: BTW, I'd like some comment explaining why these unsafe block is safe.

NobodyXu (changes requested):

src/windows/find_tools.rs
dpaoliello: Ok, made a wrapping type with `Drop`.

src/windows/find_tools.rs
dpaoliello: Fair enough.

src/windows/find_tools.rs
dpaoliello: Unfortunately, yes, `GetProcAddress` returns a `unsafe fn() -> isize`, so we need to transmute it into the correct function type.

Ideally, we'd have windows-sys generate the type that we transmute to, so that we can guarantee that it is correct, but I'm not sure how to do that without it being linked in as a dependency.

NobodyXu: Maybe we need to update the windows generator.

There's a new version but it removes all useful APIs and becomes a binary crate.

Not sure what's the way forward.

src/windows/find_tools.rs
```diff
@@ -199,6 +203,47 @@ mod impl_ {
...
+    }
+
+    impl Drop for LibraryHandle {
+        fn drop(&mut self) {
+            unsafe { FreeLibrary(self.0) };
+        }
+    }
+
+    fn is_amd64_emulation_supported() -> bool {
+        let kernel32 = LibraryHandle::new(b"kernel32.dll\0").unwrap();
```
NobodyXu, line 222: Let's return `false` on failure to load `kernel32`?

Or use some fallback, or return error, panicking not a good way to report error.

NobodyXu (changes requested):

src/windows/find_tools.rs
dpaoliello: Ok, I cleaned this up
* Updated to the latest `windows_bindgen`.
* Created a new list where we remove the `#[link]` attribute, which allows me to generate the definition for `GetMachineTypeAttributes` without statically linking it.

dpaoliello: That... didn't work.

It looks like any use of `GetMachineTypeAttributes` outside of a `const` scope will cause it to be linked into the binary. So, I've gone back to the hand-written definition BUT I've added a `const` block to make sure that the hand-written signature and the generated one match.

NobodyXu (changes requested):

src/windows/find_tools.rs
```diff
@@ -199,6 +205,71 @@ mod impl_ {
...
+        fn new(name: &[u8]) -> Option<Self> {
+            let handle = unsafe { LoadLibraryA(name.as_ptr() as _) };
+            (!handle.is_null()).then(|| Self(handle))
+        }
+
+        /// Get a function pointer to a function in the library.
+        /// SAFETY: The caller must ensure that the function signature matches the actual function.
+        /// The easiest way to do this is to add an entry to windows_sys_no_link.list and use the
+        /// generated function for `func_signature`.
+        unsafe fn get_proc_address<F>(&self, name: &[u8]) -> Option<F> {
```
riking, line 220: This should be tying the lifetime of F to the lifetime of self, if any.

NobodyXu: Thanks, opened #1010 to update safety comment for the function.

I can't add lifetime to the return type directly, I can only add bound to the generic F.

However, I don't think we can add lifetime to function pointer, a newtype that dereferences to F would have the same problem since function pointer is copyable, so the only option is to add a safety comment for this unsafe function.

PR #960, opened 2024-02-21
fix: remove unnecessary `println`s

amaanq:
It seems like it was probably for debugging, but should've been removed before #931 was merged.

Thanks!

### DIFF
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -196,7 +196,12 @@ fn write_warning(line: &[u8]) {
     stdout.write_all(b"\n").unwrap();
 }
 
-fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(), Error> {
+fn wait_on_child(
+    cmd: &Command,
+    program: &str,
+    child: &mut Child,
+    cargo_output: &CargoOutput,
+) -> Result<(), Error> {
     StderrForwarder::new(child).forward_all();
 
     let status = match child.wait() {
@@ -211,7 +216,8 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(), 
             ));
         }
     };
-    println!("{}", status);
+
+    cargo_output.print_warning(&status);
 
     if status.success() {
         Ok(())
@@ -280,7 +286,7 @@ pub(crate) fn run(
     cargo_output: &CargoOutput,
 ) -> Result<(), Error> {
     let mut child = spawn(cmd, program, cargo_output)?;
-    wait_on_child(cmd, program, &mut child)
+    wait_on_child(cmd, program, &mut child, cargo_output)
 }
 
 pub(crate) fn run_output(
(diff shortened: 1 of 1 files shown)


NobodyXu (changes requested):
How about we turn them into warnings?

They are quite useful when debugging

NobodyXu:
> If it were a warning, would that be configurable in any way?

yes, [`Build::cargo_warning`](https://docs.rs/cc/latest/cc/struct.Build.html#method.cargo_warnings) can disable it.

amaanq:
perfect! should I just have an `if cargo_output.warnings` check for each case and print  if so?

NobodyXu:
You just need to use `cargo_output: &CargoOutput,`, it will handle it for you (you'll need to add it to params of `wait_on_child`)

amaanq:
fixed

src/command_helpers.rs
```diff
@@ -211,7 +216,10 @@ fn wait_on_child(cmd: &Command, program: &str, child: &mut Child) -> Result<(),
             ));
         }
     };
-    println!("{}", status);
+
+    if cargo_output.warnings {
```
NobodyXu, line 220: `CargoOutput.print_warning` is designed for this use case, it's better to use it instead of if here.

NobodyXu (changes requested):

src/command_helpers.rs
amaanq: oops

src/command_helpers.rs
```diff
@@ -320,7 +326,7 @@ pub(crate) fn spawn(
         }
     }
 
-    println!("running: {:?}", cmd);
+    cargo_output.print_warning(&format!("running: {:?}", cmd));
```
NobodyXu, line 329: please use `format_args!` to avoid heap alloc

NobodyXu (changes requested):

NobodyXu (approved):
Thank you!

amaanq:
Thanks a ton for the quick reviewing!!

amaanq:
Just a small request, if we could have a minor release somewhat soon that'd be really nice for tree-sitter to also have a minor patch bump as well to get rid of this runtime logging :grin:

NobodyXu:
I can cut a new release in one week, given that we already got one release this week.

amaanq:
that works great, thank you

NobodyXu:
Turns out that we need another logging level for such information.

amaanq:
I think that might be a good idea, to be honest I still sorta doubt the usefulness of logging of what's running and the output status code, but I can see why it might be nice to have. Do you think a `debug` level better suits this?

NobodyXu:
I suppose we just add another function `debugging_print(bool)` to disable it, specifically for your ise case.

It will be printed without `cargo:warnings=` just like how it was before this PR, but allows you to disable it to get back the old behavior.

amaanq:
Alright I've put up #972, I think that debug logging should be disabled by default - let's continue the conversation there

PR #961, opened 2024-02-22
Add `/System/iOSSupport` to header/library search paths on Mac Catalyst

madsmtm:
This allows using `#import <UIKit/UIKit.h>` in C code processed by `cc`.

See https://github.com/rust-lang/rust/pull/121430 for more context. In `cc`, we need a few more flags than `rustc`, as we also want to add header search paths.

CC @BlackHoleFox.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2582,14 +2582,42 @@ impl Build {
                 "Detecting {:?} SDK path for {}",
                 os, sdk_details.sdk
             ));
-            let sdk_path = if let Some(sdkroot) = env::var_os("SDKROOT") {
-                sdkroot
-            } else {
-                self.apple_sdk_root(&sdk_details.sdk)?
-            };
+            let sdk_path = self.apple_sdk_root(&sdk_details.sdk)?;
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2582,16 +2582,44 @@ impl Build {
...
+                // Framework header search path
+                OsString::from("-iframework"),
+                ios_support.join("/System/Library/Frameworks").into(),
+                // Library search path
+                {
+                    let mut s = OsString::from("-L");
+                    s.push(&ios_support.join("/usr/lib"));
+                    s
+                },
+                // Framework linker search path
```
BlackHoleFox, line 2611: Should this get moved up to right under `-iframework`?

BlackHoleFox (approved):

src/lib.rs
NobodyXu: Sorry didn't see this😂

madsmtm: It doesn't really matter, I went with keeping the header and the library search paths separated.

PR #973, opened 2024-02-24
Disable optimization for jobserver pipe

NobodyXu:
It could be the reason why the failure still happens in https://github.com/sfackler/rust-openssl/issues/2184#issuecomment-1962453714

### DIFF
--- a/src/parallel/job_token/unix.rs
+++ b/src/parallel/job_token/unix.rs
@@ -74,13 +74,6 @@ impl JobServerClient {
                 Some(libc::O_RDONLY) | Some(libc::O_RDWR),
                 Some(libc::O_WRONLY) | Some(libc::O_RDWR),
             ) => {
-                // Optimization: Try converting it to a fifo by using /dev/fd
-                if let Some(jobserver) =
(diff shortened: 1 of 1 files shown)


chen-hongzhi:
It still fails in macOS and I have to manually set the `CMAKE_BUILD_PARALLEL_LEVEL` to the number of CPU cores to bypass this issue.

`make[2]: *** read jobs pipe: Resource temporarily unavailable.  Stop.`

NobodyXu:
Thanks, that's strange.

We only set `O_NONBLOCK` when cc is compiling and cleared that afterwards, so perhaps `set_blocking` fails to clear `O_NONBLOCK`

NobodyXu:
@n-eq Can you test this PR?

I kind of suspect macOS's `/dev/fd/$fd` is different from Linux.

On Linux, it opens the pipe again, with a different state, so that we can set `O_NONBLOCK` independently.

However I suspect that things are different on MacOS,.

NobodyXu:
Never mind, tried locally and it doesn't work

PR #974, opened 2024-02-25
Add logging to done_acquiring

NobodyXu:


NobodyXu:
@chen-hongzhi can you try this patch and provide me with the output of build.rs please?

Thank you!

chen-hongzhi:
I believe the issue is here: [https://github.com/rust-lang/cc-rs/blob/022f729fa19b55f316bfbc20753071cad0614b09/src/parallel/stderr.rs#L28](url)

It leads to no unset O_NONBLOCK.

NobodyXu:
Thank you so much, looks like I forgot to update it when copy pasing the code!

We really need more maintainers for cc, to spot the mistakes I'm making.

PR #979, opened 2024-02-25
Ci: Add integration test with other widely used crates

NobodyXu:


NobodyXu:
Turns out it's a bit difficult to get openssl to use parallel cc.

Since openssl use openssl-src for vendored build anyway, we could directly run it in our build.rs of cc-test

NobodyXu:
Closed as I think it's better to add testing for each specific targets.

The bug this PR was trying to prevent was actually a jobserver stuff, which we now use crate jobserver for instead of using vendored implementation.

PR #980, opened 2024-02-26
Add regression test for openssl-src

NobodyXu:
Build openssl as regression testing


NobodyXu:
A simplified version of #979

NobodyXu:
We switch back to using jobserver, so this should not be of a concern to us.

PR #985, opened 2024-03-02
Fix make resource failure when cc parallel is enabled

### LINKED ISSUE #962: make resource failure when cc parallel is enabled
Since cc >=1.0.80, we set the jobserver inherited to `O_NONBLOCK`, make probably isn't designed for this, which is why it failed.

I will open a new ticket in cc and fix it.

_Originally posted by @NobodyXu in https://github.com/rust-lang/cargo/issues/13476#issuecomment-1959123242_

NobodyXu:
Fixed #962


NobodyXu:
Verified locally by building `cargo-near`, that this PR indeed fixed #962

weihanglo:
Thank you for your hard work chasing this down! 

Just fly-by. Could we have one or two simple paragraphs describing what was wrong and how this PR gets it correct?

NobodyXu:
Yes, so the problem is that make < 4.3 pass jobserver using annoymous pipe fd, and thus they share the file description between parent and child processes, e.g. flags `O_NONBLOCK` are shared among `make`, `cargo`, `build.rs`, etc.

When `cc::Build::compile` in one `build.rs` sets `O_NONBLOCK` on the pipe, all other processes, including `make`, will also affected by this change.

Since make < 4.3 is not able to handle jobserver pipe with `O_NONBLOCK` set, it will fail.

make >= 4.3 uses `O_NONBLOCK` flag itself and it supports the named fifo mode, which is more robust than annoymous 
pipe, because they don't share the file description.

NobodyXu:
Note that jobserver currently has a problem: It doesn't clone the anonymous pipe passed via env.

So if someone calls `Build::compile`, creates a `jobserver::Client` and drop it, then calling `Build::compile` again will use the now closed pipe fd.


Since fd is reused, this end up creates some really strange bug (this happened in opencv before twistedfall/opencv-rust#480 , they fixed it by using jobslot, which does clone the annoymous pipe fd)

I've created rust-lang/jobserver-rs#57 but it has not been merged yet

Plus jobserver now has msrv 1.63, and cc is still on 1.56

Amanieu (approved):

PR #986, opened 2024-03-02
fix targets with zig toolchain

QuentinPerez:
When you use bazel + zig to cross compile a rust codebase, you got an 'invalid target error'
To fix the issue, a check has been added to know if it's a zig toolchain, not sure it's the right way to do it. 
Feel free to take over the PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1902,7 +1902,14 @@ impl Build {
 
         // Target flags
         match cmd.family {
-            ToolFamily::Clang { .. } => {
+            ToolFamily::Clang { zig_cc } => {
+                let target = if zig_cc {
+                    Cow::Owned(target.replace("-unknown-", "-"))
+                } else {
+                    Cow::Borrowed(target)
+                };
+                let target = target.as_ref();
+
                 if !cmd.has_internal_target_arg
                     && !(target.contains("android")
                         && android_clang_compiler_uses_target_arg_internally(&cmd.path))
@@ -1998,7 +2005,7 @@ impl Build {
                             );
                         }
                     } else if let Ok(index) = target_info::RISCV_ARCH_MAPPING
-                        .binary_search_by_key(&arch, |(arch, _)| &arch)
+                        .binary_search_by_key(&arch, |(arch, _)| arch)
                     {
                         cmd.args.push(
                             format!(


NobodyXu (changes requested):
It would be great if you could setup CI for testing zig-cc, to verify it indeed works as intended.

QuentinPerez:
> It would be great if you could setup CI for testing zig-cc, to verify it indeed works as intended.

I tried it, but I am not sure how to achieve it correctly. Do you have any ideas or guidance on how I could do it?

NobodyXu:
Simply install `pip3 install zig` in job test of `.github/workflows/main.yml`.

You would also need to add a `tests/zig-test.rs` and write all the testing for `zig-cc` in it.

src/tool.rs
```diff
@@ -96,7 +96,9 @@ impl Tool {
                 }
             };
             if stdout.contains("clang") {
-                ToolFamily::Clang
+                ToolFamily::Clang {
+                    is_zig: stdout.contains("ziglang"),
```
NobodyXu, line 100: Just note that I've updated how compiler family detection work in #1000

Recommend you to wait for it to merge and then rebase against it.

NobodyXu (changes requested):

NobodyXu:
@QuentinPerez #1000 has been merged, can you rebase against main please?

src/lib.rs
```diff
@@ -1902,7 +1902,13 @@ impl Build {
...
         // Target flags
         match cmd.family {
-            ToolFamily::Clang { .. } => {
+            ToolFamily::Clang { zig_cc } => {
+                let target = if zig_cc {
+                    Cow::Owned(target.replace("-unknown-", "-"))
+                } else {
+                    Cow::Borrowed(target)
+                };
+
```
NobodyXu, line 1911: Could do a

```suggestion
            let target = target.as_ref();

```

To avoid changing every if that uses target

NobodyXu (changes requested):

NobodyXu (commented):
Thanks, the code itself LGTM, can you add a test for zig-cc in CI?

You can just run cargo-test, while setting environment variable CC to zig-cc

NobodyXu:
Discussed this with @thomcc on zulip, he mentioned that the zig-cc target triple might be changed in the later version of zig-cc.

We are afraid that after merging this PR, zig-cc might change its target triple and break this, causing more code to require to support both newer version and older version.

We decided to hold off adding more workarounds for zig-cc until it reaches 1.0, when its target triple and cmdline args are stable.

PR #993, opened 2024-03-06
Fix the clang targets for RISC-V Rust targets

djkoloski:
Rust supports a variety of RISC-V targets which may include optional features in the first segment of the target triple. When invoking clang for these targets, we need to remove these features since clang doesn't use them.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1985,10 +1985,35 @@ impl Build {
                         cmd.args.push(
                             format!("--target={}", target.replace("riscv64gc", "riscv64")).into(),
                         );
+                    } else if target.starts_with("riscv64imac-") {
+                        cmd.args.push(
+                            format!("--target={}", target.replace("riscv64imac", "riscv64")).into(),
+                        );
                     } else if target.starts_with("riscv32gc-") {
                         cmd.args.push(
                             format!("--target={}", target.replace("riscv32gc", "riscv32")).into(),
                         );
+                    } else if target.starts_with("riscv32i-") {
+                        cmd.args.push(
+                            format!("--target={}", target.replace("riscv32i", "riscv32")).into(),
+                        )
+                    } else if target.starts_with("riscv32im-") {
+                        cmd.args.push(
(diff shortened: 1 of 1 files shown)


Amanieu (approved):
LGTM, but don't you also need to pass some flags to Clang to ensure it enables the correct features for the selected target?

djkoloski:
I'll take a second look tomorrow. I got it compiling but didn't test any feature functionality.

djkoloski:
It looks like all of the extensions are getting passed through `-march` by slicing them off of the target name, with the exceptions of 32-bit `linux` targets and 64-bit `linux`, `freebsd`, and `netbsd` targets. Those override `-march=rv32/64gc` exactly, as per [this PR](https://github.com/rust-lang/cc-rs/pull/583). In the default case though, these extensions are getting passed through generically. Here's the snippet in question:

```rust
if target.starts_with("riscv32") || target.starts_with("riscv64") {
    // get the 32i/32imac/32imc/64gc/64imac/... part
    let mut parts = target.split('-');
    if let Some(arch) = parts.next() {
        let arch = &arch[5..];
        if arch.starts_with("64") {
            if target.contains("linux")
                | target.contains("freebsd")
                | target.contains("netbsd")
                | target.contains("linux")
            {
                cmd.args.push(("-march=rv64gc").into());
                cmd.args.push("-mabi=lp64d".into());
            } else {
                cmd.args.push(("-march=rv".to_owned() + arch).into());
                cmd.args.push("-mabi=lp64".into());
            }
        } else if arch.starts_with("32") {
            if target.contains("linux") {
                cmd.args.push(("-march=rv32gc").into());
                cmd.args.push("-mabi=ilp32d".into());
            } else {
                cmd.args.push(("-march=rv".to_owned() + arch).into());
                cmd.args.push("-mabi=ilp32".into());
            }
        } else {
            cmd.args.push("-mcmodel=medany".into());
        }
    }
}
```

So this PR should be ready to merge. I was testing with `riscv32imc-unknown-none-elf` so I should be hitting those branches.

PR #997, opened 2024-03-07
allow setting `nvcc` program via `CUDACXX`

### LINKED ISSUE #995: feature request: Allow setting nvcc executable via env var.
The env vars `CC` and `CXX` can be set to point to C and c++ compilers respectively, I'd like to be able to set the `nvcc` executable in a similar way.

CMake uses `CUDACXX`, which seems reasonable enough and would work for my usecase. I currently just modify `PATH`.

Somthing analogous to `CFLAGS` and `CXXFLAGS` for Cuda would also be nice for completions sake, but I have not found a need for it.

I could probably implement and document this myself if this feature is in scope for the project.

MarcusDunn:
Don't know how the tests works (pointers welcome if you'd like a test).

How the other's (`CC` and `CXX`) get initialized seems very complex to support stuff like `ccache cc`. This is about as simple as could be.

Resolves #995

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,6 +88,8 @@
 //! be logged to stdout. This is useful for debugging build script issues, but can be
 //! overly verbose for normal use.
 //! * `CXX...` - see [C++ Support](#c-support).
+//! * `CUDACXX` - the nvcc program to invoke. Similar to `CC`, this must be an exact
+//! executable name and cannot contain extra flags.
 //!
 //! Furthermore, projects using this crate may specify custom environment variables
 //! to be inspected, for example via the `Build::try_flags_from_environment`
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2402,7 +2404,8 @@ impl Build {
             let dlink = out_dir.join(lib_name.to_owned() + "_dlink.o");
             let mut nvcc = self.get_compiler().to_command();
             nvcc.arg("--device-link").arg("-o").arg(&dlink).arg(dst);
-            run(&mut nvcc, "nvcc", &self.cargo_output)?;
+            let program = env::var("CUDACXX").unwrap_or_else(|_| String::from("nvcc"));
```
NobodyXu, line 2407: I think you might need to modify `cc::Build::get_base_compiler`.

This line you modified is for linking with cuda-specific part only.

P.S. I'm not so familiar with the cuda compilation part, so feel free to point out any mistake.

NobodyXu (changes requested):

src/lib.rs
MarcusDunn: I suspected this was too easy. I'll take a look.

dot-asm:
One can control which nvcc is invoked by setting NVCC environment variable.

PR #998, opened 2024-03-07
Fix: Accept non utf-8 characters in compiler and archiver

NobodyXu:
While the compiler automatically detected by cc always is a valid `str`, it is possible for user to override them using environment variable `CC`, `CXX` and `AR`, which could contain non utf-8 characters.

Before this PR is applied, cc uses `to_string_lossy()` to convert it to a valid `str` which might cause the creation of process to fail.

### DIFF
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -2,7 +2,6 @@
 
 use std::{
     collections::hash_map,
-    convert::AsRef,
     ffi::OsString,
     fmt::Display,
     fs,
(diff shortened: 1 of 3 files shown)


Amanieu (approved):

thomcc (commented):
Looks fine.

(I feel good enough to start easing myself back into reviews of small stuff like this)

thomcc:
Oh, it was already reviewed.

PR #1000, opened 2024-03-07
Fix compile family detection: Use C macros instead of `$compiler -v`

### LINKED ISSUE #958: Compiler version doesn't include clang or GCC: "cc" "--version"
On Debian Sid:
Output of `cc --version`, which is different from `gcc --version`:
```
cc (Debian 13.2.0-13) 13.2.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
Output of `cc -v`, which is the same as `gcc -v`:
```
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-linux-gnu/13/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Debian 13.2.0-13' --with-bugurl=file:///usr/share/doc/gcc-13/README.Bugs --enable-languages=c,ada,c++,go,d,fortran,objc,obj-c++,m2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-13 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/libexec --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-libstdcxx-backtrace --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --enable-libphobos-checking=release --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --enable-cet --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/reproducible-path/gcc-13-13.2.0/debian/tmp-nvptx/usr,amdgcn-amdhsa=/build/reproducible-path/gcc-13-13.2.0/debian/tmp-gcn/usr --enable-offload-defaulted --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 13.2.0 (Debian 13.2.0-13) 
```

So it's may be better to use `-v` to check the compiler type?

NobodyXu:
Fixed #958


src/tool.rs
```diff
@@ -74,70 +91,102 @@ impl Tool {
...
+            out_dir: Option<&Path>,
+        ) -> Result<ToolFamily, Error> {
+            let tmp = NamedTempfile::new(
+                &out_dir
+                    .map(Cow::Borrowed)
+                    .unwrap_or_else(|| Cow::Owned(env::temp_dir())),
+                "detect_compiler_family.c",
+            )?;
+            tmp.file()
+                .write_all(include_bytes!("detect_compiler_family.c"))?;
```
Be-ing, line 119: Is copying this to a temporary file necessary? Could you use `env!("CARGO_MANIFEST_DIR")/src/detect_compiler_family.c")` instead?

NobodyXu: That will only work if cc is used in `build.rs`

cc is also used in binary crates which has no access to the source code, so I'm afraid it won't work.

Be-ing: `env!` is evaluated when the file containing the macro is compiled, not when the code is executed. So I think it could work.

NobodyXu: Yes, but what if the source of cc is not available at all?

What if cc-rs is used in binary crate, and re-distribute without including cc-rs source files?

Be-ing: Why would being used in a library versus binary crate make a difference?

NobodyXu: Imagine this:

A binary crate pulls in cc-rs, compiles it into a single binary, and then distribute it to the user, who runs in on their computer.

`env!("CARGO_MANIFEST_DIR")` gives you the manifest dir for the source of cc-rs at compile-time, but it is not available in runtime, on the user's computer.

So we must use `include_bytes!` to copy the file into the final binary, and then write it to a named tempfile.

workingjubilee: Yeah, the approach here is regrettably necessary.

dev-tools/cc-test/build.rs
```diff
@@ -156,7 +156,13 @@ fn run_action_if_forked() -> bool {
     true
 }
 
+fn disable_debug_output() {
+    std::env::remove_var("CC_ENABLE_DEBUG_OUTPUT");
```
workingjubilee, line 160: Is there a reason we're mutating the environment here...?

NobodyXu: That env would break tests for warning/debug output

NobodyXu: And it is set in the CI, to make debugging CI failure easier.

workingjubilee: Ah, I see, this is purely testing code. Can you leave a comment about that?

src/tempfile.rs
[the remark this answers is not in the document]
NobodyXu: tempfile pulls in windows-sys on windows, plus fast-rand, it's definitely too much compared to this simple in-house implementation.

dev-tools/cc-test/build.rs
NobodyXu: Sure, will do

src/tempfile.rs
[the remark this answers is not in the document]
workingjubilee: Fair enough then.

workingjubilee (approved):
Seems to work locally. Can't attest to Windows correctness, but we'll find out very quickly if it isn't.

PR #1004, opened 2024-03-09
feat:  Query rustc for clang target triples instead of hardcoding them

### LINKED ISSUE #994: Query rustc for clang target triples
Right now, we pick the target triples to pass clang by defaulting to the first segment of the target triple and then manually fixing them up with a large set of branches. Is there a way we could query the target spec from rustc to get the LLVM target names? For example, by running:

```
$ rustc +nightly -Zunstable-options --print all-target-specs-json
```

We can obtain a JSON dump of all of the target configurations. Each target configuration contains an `"llvm-target"` field which we should be passing to clang. That's not stable right now, but maybe there's some other way to work with that?

NobodyXu:
Fixed #994

Add new workspace crate `gen-target-info` for generating `src/target_info.rs`, and use it to simplify riscv target arch mapping logic in `Build::add_default_flags`

### DIFF
--- a/dev-tools/gen-target-info/src/main.rs
+++ b/dev-tools/gen-target-info/src/main.rs
@@ -1,4 +1,4 @@
-use gen_target_info::{get_target_specs_from_json, write_string_mapping_to_file};
+use gen_target_info::{get_target_specs_from_json, write_target_tuple_mapping};
 use std::{fs::File, io::Write as _};
 
 fn main() {
@@ -22,7 +22,7 @@ fn main() {
         .collect::<Vec<_>>();
     riscv_target_mapping.sort_unstable_by_key(|(arch, _)| &**arch);
     riscv_target_mapping.dedup();
-    write_string_mapping_to_file(&mut f, "RISCV_ARCH_MAPPING", &riscv_target_mapping);
+    write_target_tuple_mapping(&mut f, "RISCV_ARCH_MAPPING", &riscv_target_mapping);
 
     // Flush the data onto disk
     f.flush().unwrap();
--- a/dev-tools/gen-target-info/src/main.rs
+++ b/dev-tools/gen-target-info/src/main.rs
@@ -1,0 +1,29 @@
+use gen_target_info::{get_target_specs_from_json, write_string_mapping_to_file};
+use std::{fs::File, io::Write as _};
+
+fn main() {
+    let target_specs = get_target_specs_from_json();
+
+    // Open file to write to
+    let manifest_dir = env!("CARGO_MANIFEST_DIR");
+
+    let path = format!("{manifest_dir}/../../src/target_info.rs");
+    let mut f = File::create(path).expect("failed to create src/target_info.rs");
+
+    // Wrute riscv target mapping
+    let mut riscv_target_mapping = target_specs
+        .0
+        .iter()
+        .filter_map(|(target, target_spec)| {
+            let arch = target.split_once('-').unwrap().0;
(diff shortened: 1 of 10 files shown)


src/target_info.rs
```diff
@@ -0,0 +1,10 @@
+pub const RISCV_ARCH_MAPPING: &[(&str, &str)] = &[
+    ("riscv32gc", "riscv32"),
+    ("riscv32i", "riscv32"),
+    ("riscv32im", "riscv32"),
+    ("riscv32imac", "riscv32"),
+    ("riscv32imafc", "riscv32"),
+    ("riscv32imc", "riscv32"),
+    ("riscv64gc", "riscv64"),
+    ("riscv64imac", "riscv64"),
+];
```
workingjubilee, line 10: People often try to edit these kinds of files directly, so I think this file should come with a comment like
```rust
// This file is generated code. Please edit the generator
// in dev-tools/gen-target-info if you need to make changes.
```

dev-tools/gen-target-info/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+
+fn main() {
+    let target_specs = get_target_specs_from_json();
+
+    // Open file to write to
+    let manifest_dir = env!("CARGO_MANIFEST_DIR");
+
+    let path = format!("{manifest_dir}/../../src/target_info.rs");
+    let mut f = File::create(path).expect("failed to create src/target_info.rs");
+
```
workingjubilee, line 12: Pursuant to my comment elsewhere, we can write the "header comment" here before we do any more writes.

workingjubilee (changes requested):
I am not entirely sure if this is an improvement, but I am certainly not against it, since it transforms a bunch of logic into data, which is much better to see.

dev-tools/gen-target-info/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+        .iter()
+        .filter_map(|(target, target_spec)| {
+            let arch = target.split_once('-').unwrap().0;
+            (arch.contains("riscv") && arch != &target_spec.arch)
+                .then_some((arch, &*target_spec.arch))
+        })
+        .collect::<Vec<_>>();
+    riscv_target_mapping.sort_unstable_by_key(|(arch, _)| &**arch);
+    riscv_target_mapping.dedup();
+    write_string_mapping_to_file(&mut f, "RISCV_ARCH_MAPPING", &riscv_target_mapping);
```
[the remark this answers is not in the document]
NobodyXu: Extracted as a function in `main.rs`

dev-tools/gen-target-info/src/write.rs
```diff
@@ -0,0 +1,14 @@
+use std::{fmt::Write as _, fs, io::Write as _};
+
+pub fn write_string_mapping_to_file(f: &mut fs::File, variable_name: &str, data: &[(&str, &str)]) {
```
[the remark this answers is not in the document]
NobodyXu: Renamed.

src/target_info.rs
NobodyXu: Added the prelude as a module document.

dev-tools/gen-target-info/src/main.rs
NobodyXu: Thanks, wrote the header as suggested.

workingjubilee (approved):

NobodyXu:
Thanks!

PR #1007, opened 2024-03-12
Fix xcrun sdk version arg

### LINKED ISSUE #963: 1.0.86 producing `clang: error: invalid version number in '-mmacosx-version-min=1.1'` on macos 10.14
We build for a variety of platforms and noticed build failures on macOS 10.14, `clang: error: invalid version number in '-mmacosx-version-min=1.1'`. After investigating, we found that the issue is related to cc 1.0.86. Manually editing our lock file and pinning to 1.0.83 resolves the issue. We do not directly depend on cc.

### LINKED ISSUE #1001: Console output is full of `xcrun: error: unable to lookup item 'PlatformVersion' from command line tools installation`
After rust-lang/rust updated cc from 1.0.79 to 1.0.90 (https://github.com/rust-lang/rust/pull/119199), I started seeing several copies of this message appearing on my console during a build:

```text
warning: profiler_builtins@0.0.0: xcrun: error: unable to lookup item 'PlatformVersion' from command line tools installation
warning: profiler_builtins@0.0.0: xcrun: error: unable to lookup item 'PlatformVersion' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'

warning: rustc_llvm@0.0.0: xcrun: error: unable to lookup item 'PlatformVersion' from command line tools installation
warning: rustc_llvm@0.0.0: xcrun: error: unable to lookup item 'PlatformVersion' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'
```

My build still seems to *work* just fine, so I'm trying to figure out the appropriate way to resolve these annoying console messages.

It seems that the changes made in #943 don't interact nicely with macOS systems that have the command-line developer tools installed, but don't have a full Xcode installation.

0123456789-jpg:
We should use `xcrun --show-sdk-version ...` to obtain targeted apple os version instead of `--show-sdk-platform-version`, as shown in my screenshot.
屏幕快照 2024-03-12 下午7 07 06

Closes #1001, closes #963

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3616,7 +3616,7 @@ impl Build {
 
             let version = run_output(
                 self.cmd("xcrun")
-                    .arg("--show-sdk-platform-version")
+                    .arg("--show-sdk-version")
                     .arg("--sdk")
                     .arg(sdk),
                 "xcrun",


BlackHoleFox:
Have a look at https://github.com/rust-lang/cc-rs/pull/1009 if you wouldn't mind as well. It ports some of clang's SDK reading logic into `cc` for this purpose. I'm worried about relying on `--show-sdk-version` always lining up with the deployment target and the other way also avoids needing `xcrun` to be present.

0123456789-jpg:
@NobodyXu

NobodyXu:
#1009 sounds like a better solution, if the json parser can be simplified in it.

NobodyXu:
Merged this as a workaround, while waiting for #1009

PR #1013, opened 2024-03-18
refactor: Optimize code based on cargo clippy suggestions 

InventiveCoder:


src/lib.rs
```diff
@@ -3415,11 +3415,11 @@ impl Build {
             })
             .map(|prefix| *prefix)
             .or_else(||
-            // If no toolchain was found, provide the first toolchain that was passed in.
-            // This toolchain has been shown not to exist, however it will appear in the
-            // error that is shown to the user which should make it easier to search for
-            // where it should be obtained.
-            prefixes.first().map(|prefix| *prefix))
+                // If no toolchain was found, provide the first toolchain that was passed in.
```
NobodyXu, line 3418: I'd recommend moving the comment outside the closure

src/lib.rs
```diff
@@ -236,14 +236,17 @@ use std::sync::{Arc, Mutex};
 #[cfg(feature = "parallel")]
 mod parallel;
 mod windows;
+
 // Regardless of whether this should be in this crate's public API,
 // it has been since 2015, so don't break it.
 pub use windows::find_tools as windows_registry;
 
 mod command_helpers;
+
```
NobodyXu, line 245: I personally think it's ok for `use` to on the next line of `mod`

NobodyXu (changes requested):

src/lib.rs
InventiveCoder: Thanks!

Modified and execute cargo fmt.

Please review again.

NobodyXu (approved):
Thanks!

PR #1015, opened 2024-03-19
fix llvm-ar as archiver for msvc targets; fix clang-cl detection; fix assembler output flag detection; add clang/clang-cl windows CI

russelltg:
This PR contains several clang on windows improvements:

1. properly detect clang vs clang-cl by invoking with `-?`, instead of checking for `_MSC_VER`. This is what cmake does, and `_MSC_VER` is always defined for msvc ABI targets, regardless of the frontend variant used
2. Fix detection of which output assembler flag to use. It used to use complex logic, including if the C compiler is clang, which was incorrect. Use the flag based on if it's using the msvc assembler
3. Add clang-cl/clang CI for windows
4. Detect usage of llvm-ar and pass GNU style flags to it

Original description:

this is somewhat a nasty hack here, but it does work. when (cross) compiling for msvc and you use llvm-ar as the archiver, then you need to use gnu-style flags.

currently it attemps to run llvm-ar like:

```
running: "/usr/bin/llvm-ar" "-out:/workspaces/ars/build_windows_clang_x86_64_relnoopt/./cargo/build/x86_64-pc-windows-msvc/release/build/link-cplusplus-b04e59fc086748c2/out/liblink-cplusplus.a" "-nologo" "/workspaces/ars/build_windows_clang_x86_64_relnoopt/./cargo/build/x86_64-pc-windows-msvc/release/build/link-cplusplus-b04e59fc086748c2/out/0466faf07398e270-dummy.o"
```

which doesn't work, as llvm-ar takes in GNU-style ar flags.

i'm definitely open to a more robust way to get the archiver "Family" (maybe something similar to how we detect compiler family? idk) if that would be better

I also moved some APIs to to use Path as it avoided a string->path conversion. Let me know if that's not of interest.

With this and the other PR, I can now cross-compile to windows using clang from linux :)

EDIT: also fixes the path to the archiver not being set on when env_tool finds it

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,7 +76,7 @@ jobs:
             os: windows-2019
             rust: stable
             target: x86_64-pc-windows-msvc
-            env: $Env:AR="llvm-ar" ; $Env:CC="clang" ; $Env:CXX="clang++" ; $Env:RUSTFLAGS="-Clinker=lld-link" ;
+            env: AR=llvm-ar CC=clang CXX=clang++ RUSTFLAGS=-Clinker=lld-link
     steps:
     - uses: actions/checkout@v4
     - name: Install Rust (rustup)
@@ -99,6 +99,8 @@ jobs:
     - name: add clang to path
       run: $Env:PATH += ";C:\msys64\mingw64\bin"
       if: matrix.build == 'windows-clang'
+    - run : echo ${{ matrix.env }} >> $GITHUB_ENV
+      shell: bash
     - uses: Swatinem/rust-cache@v2
     - run: ${{ matrix.env }} cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }}
     - run: ${{ matrix.env }} cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --release
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,12 +76,12 @@ jobs:
             os: windows-2019
             rust: stable
             target: x86_64-pc-windows-msvc
-            env: AR=llvm-ar CC=clang CXX=clang++ RUSTFLAGS=-Clinker=lld-link
+            env: $Env:AR="llvm-ar" ; $Env:CC="clang" ; $Env:CXX="clang++" ; $Env:RUSTFLAGS="-Clinker=lld-link" ;
           - build: windows-clang-cl
             os: windows-2019
             rust: stable
             target: x86_64-pc-windows-msvc
-            env: AR=llvm-ar CC=clang-cl CXX=clang-cl RUSTFLAGS=-Clinker=lld-link
+            env: $Env:AR="llvm-ar" ; $Env:CC="clang-cl" ; $Env:CXX="clang-cl" ; $Env:RUSTFLAGS="-Clinker=lld-link" ;
     steps:
     - uses: actions/checkout@v4
     - name: Install Rust (rustup)
@@ -104,14 +104,10 @@ jobs:
     - name: add clang to path
       run: $Env:PATH += ";C:\msys64\mingw64\bin"
       if: startsWith(matrix.build, 'windows-clang')
-    - run : echo ${{ matrix.env }} | xargs -I{} -n1 -- bash -c "echo {} >> $GITHUB_ENV"
-      shell: bash
-    - run: env
-      shell: bash
     - uses: Swatinem/rust-cache@v2
-    - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }}
-    - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --release
-    - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --features parallel
+    - run: ${{ matrix.env }} cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }}
+    - run: ${{ matrix.env }} cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --release
+    - run: ${{ matrix.env }} cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --features parallel
 
   # This is separate from the matrix above because there is no prebuilt rust-std component for these targets.
   check-tvos:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -106,6 +106,8 @@ jobs:
       if: startsWith(matrix.build, 'windows-clang')
     - run : echo ${{ matrix.env }} | xargs -I{} -n1 -- bash -c "echo {} >> $GITHUB_ENV"
       shell: bash
+    - run: env
+      shell: bash
     - uses: Swatinem/rust-cache@v2
     - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }}
     - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --release
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,7 +76,7 @@ jobs:
             os: windows-2019
             rust: stable
             target: x86_64-pc-windows-msvc
-            env: AR=llvm-ar CC=clang CXX=clang++ RUSTFLAGS=-Clinker=lld-link
+            env: $Env:AR="llvm-ar" ; $Env:CC="clang" ; $Env:CXX="clang++" ; $Env:RUSTFLAGS="-Clinker=lld-link" ;
     steps:
     - uses: actions/checkout@v4
     - name: Install Rust (rustup)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,21 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        build: [stable, beta, nightly, linux32, macos, aarch64-macos, aarch64-ios, win32, win64, mingw32, mingw64, windows-2019]
+        build:
+          [
+            stable,
+            beta,
+            nightly,
+            linux32,
+            macos,
+            aarch64-macos,
+            aarch64-ios,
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -2398,7 +2398,8 @@ impl Build {
         }
 
         let target = self.get_target()?;
-        if target.contains("msvc") {
+        let (mut ar, cmd, _any_flags) = self.get_ar()?;
+        if target.contains("msvc") && !cmd.to_str().unwrap().contains("llvm-") {
```
NobodyXu, line 2402:
```suggestion
        if target.contains("msvc") && !cmd.to_str_lossy().contains("llvm-") {
```

src/lib.rs
```diff
@@ -2435,8 +2435,8 @@ impl Build {
     fn assemble_progressive(&self, dst: &Path, objs: &[&Path]) -> Result<(), Error> {
         let target = self.get_target()?;
 
-        if target.contains("msvc") {
-            let (mut cmd, program, any_flags) = self.get_ar()?;
+        let (mut cmd, program, any_flags) = self.get_ar()?;
+        if target.contains("msvc") && !program.to_str().unwrap().contains("llvm-") {
```
NobodyXu, line 2439:
```suggestion
        if target.contains("msvc") && !program.to_str_lossy().contains("llvm-") {
```

NobodyXu (changes requested):
> i'm definitely open to a more robust way to get the archiver "Family" (maybe something similar to how we detect compiler family? idk) if that would be better

I don't know any better way than checking name, or its output of `--version`.

Perhaps we can try link something and check its output for signature of linker?

> With this and the other PR, I can now cross-compile to windows using clang from window

Is it possible to add testing for this in CI?

.github/workflows/main.yml
```diff
@@ -91,10 +96,13 @@ jobs:
...
         sudo apt-get install g++-multilib
       if: matrix.build == 'linux32'
+    - name: add clang to path
+      run: $Env:PATH += ";C:\msys64\mingw64\bin"
+      if: matrix.build == 'windows-clang'
     - uses: Swatinem/rust-cache@v2
-    - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }}
-    - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --release
-    - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }} --features parallel
+    - run: ${{ matrix.env }} cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }}
```
NobodyXu, line 103: Could use

```suggestion
    - run: echo "${{ matrix.env }}" >> "$GITHUB_ENV"
      shell: bash
    - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }}
```

NobodyXu (changes requested):

russelltg:
While working on that CI pass, I found that the clang-cl detection doesn't work at all....

Clang defines _MSC_VER when target is *msvc, and actually has no relation to the frontend variant. That combined with us not passing C[XX]FLAGS to the -E invocation just means that all it detects is if the host system is windows, lol....

russelltg:
While I'm working on clang cross compilation for linux->windows, I was going to add the CI pass to windows so I don't have to worry about getting windows SDK/msvc installed in Linux. But since the host is windows it ends up thinking clang is clang-cl!

russelltg:
I know you're waiting on this to release--I think it's still a good change of course, but that CI is *not* going to work yet. Want me to rip out CI and open a new PR with hopefully some way to fix clang-cl detection, that can be merged after the release?

NobodyXu:
> Want me to rip out CI and open a new PR with hopefully some way to fix clang-cl detection, that can be merged after the release?

I would prefer the clang-cl to be fixed as well before cutting a new release to avoid regression.

NobodyXu:
BTW, can you please update the PR title and description?

Thanks!!

russelltg:
Hmm it's hard to be sure that the env vars are actually being set....

russelltg:
Ah, I don't think it's possible to emit multiple env vars from GITHUB_ENV

which is kinda wild

https://github.com/actions/runner/blob/bc8b6e01528a1232341773be73078f7639f17187/src/Misc/layoutbin/hashFiles/index.js#L294

russelltg:
Ok, not quite ready I guess. Will look tomorrow

NobodyXu:
> Ah, I don't think it's possible to emit multiple env vars from GITHUB_ENV? At least I can't figure it out.

The doc says just put them in multiple lines, i.e. `echo -e "A=...\nB=..." >> "$GITHUB_ENV"`

.github/workflows/main.yml
```diff
@@ -72,6 +72,16 @@ jobs:
...
+          - build: windows-clang
+            os: windows-2019
+            rust: stable
+            target: x86_64-pc-windows-msvc
+            env: $Env:AR="llvm-ar" ; $Env:CC="clang" ; $Env:CXX="clang++" ; $Env:RUSTFLAGS="-Clinker=lld-link" ;
+          - build: windows-clang-cl
+            os: windows-2019
+            rust: stable
+            target: x86_64-pc-windows-msvc
+            env: $Env:AR="llvm-ar" ; $Env:CC="clang-cl" ; $Env:CXX="clang-cl" ; $Env:RUSTFLAGS="-Clinker=lld-link" ;
```
NobodyXu, line 84: If you write it like this:

```suggestion
            env: AR=llvm-ar\nCC=clang-cl\nCXX=clang-cl\nRUSTFLAGS=-Clinker=lld-link
```

Then use:

```
 - run: echo -e "${{ matrix.env }}" >> "$GITHUB_ENV"
   shell: bash
```

Then I think it will work

NobodyXu (changes requested):

NobodyXu:
Updated CI and code, still doesn't work

russelltg:
Ok, phew, there was some more cases to handle here, but it seems to be working.

russelltg:
I'm not sure if you want to keep the changes you made, notably the CI formatting changes. I'm happy to simplify that diff if you want.

NobodyXu:
> I'm not sure if you want to keep the changes you made, notably the CI formatting changes. I'm happy to simplify that diff if you want.

That'd be ok for me, thanks for yhe fix!

NobodyXu (approved):
Thank you, in LGTM, however since I'm not very familiar with cc test-suite, I'd like another approval from another maintainer @thomcc

russelltg:
Opened https://github.com/llvm/llvm-project/pull/87209 to address clang-cl not taking `/Fo:` but let's definitely stick with `/Fo` for maximum compatibility.

.github/workflows/main.yml
```diff
@@ -72,29 +86,53 @@ jobs:
...
+          rustup default ${{ matrix.rust }}
+        shell: bash
+      - name: Install g++-multilib
+        run: |
+          set -e
+          # Remove the ubuntu-toolchain-r/test PPA, which is added by default.
+          # Some packages were removed, and this is causing the g++multilib
+          # install to fail. Similar issue:
+          # https://github.com/scikit-learn/scikit-learn/issues/13928.
+          sudo add-apt-repository --remove ppa:ubuntu-toolchain-r/test
```
thomcc, line 116: Well, if it works...

src/command_helpers.rs
```diff
@@ -385,17 +385,20 @@ for help)"
...
-    dst: &Path,
-    cuda: bool,
-    msvc: bool,
-    clang: bool,
-    gnu: bool,
-    is_asm: bool,
-    is_arm: bool,
-) {
-    if msvc && !clang && !gnu && !cuda && !(is_asm && is_arm) {
+pub(crate) struct CmdAddOutputFileArgs {
```
thomcc, line 388: Thanks for this.

src/lib.rs
```diff
@@ -2435,8 +2448,8 @@ impl Build {
     fn assemble_progressive(&self, dst: &Path, objs: &[&Path]) -> Result<(), Error> {
         let target = self.get_target()?;
 
-        if target.contains("msvc") {
-            let (mut cmd, program, any_flags) = self.get_ar()?;
+        let (mut cmd, program, any_flags) = self.get_ar()?;
+        if target.contains("msvc") && !program.to_string_lossy().contains("llvm-") {
```
thomcc, line 2452: This will probably need adjusting but for now is fine.

src/lib.rs
```diff
@@ -3939,7 +3959,7 @@ fn which(tool: &Path, path_entries: Option<OsString>) -> Option<PathBuf> {
 }
 
 // search for |prog| on 'programs' path in '|cc| -print-search-dirs' output
-fn search_programs(cc: &mut Command, prog: &str, cargo_output: &CargoOutput) -> Option<PathBuf> {
+fn search_programs(cc: &mut Command, prog: &Path, cargo_output: &CargoOutput) -> Option<PathBuf> {
```
thomcc, line 3962: Ideally the change to use &Path would be in a different pr. But I'm not about to hold this up for that, since it is a good change.

src/tool.rs
```diff
@@ -135,11 +136,20 @@ impl Tool {
 
             cargo_output.print_debug(&stdout);
 
+            // https://gitlab.kitware.com/cmake/cmake/-/blob/69a2eeb9dff5b60f2f1e5b425002a0fd45b7cadb/Modules/CMakeDetermineCompilerId.cmake#L267-271
+            let accepts_cl_style_flags =
+                run(Command::new(path).arg("-?").stdout(Stdio::null()), path, &{
+                    // the errors are not errors!
```
thomcc, line 142: 🙃

thomcc (approved):
Thanks, the changes look good. That said, it's a big change, and I don't think we have great test coverage here, so it does make me worry a little. We should land it anyway though.

That said, maybe we should squash-land this so that we can revert more easily if it causes problems. WDYT @NobodyXu?

NobodyXu:
> That said, maybe we should squash-land this so that we can revert more easily if it causes problems.

Yeah, cc-rs defaults to squash merge so it should be pretty easy to revert.

PR #1016, opened 2024-03-19
Also shim llvm-ar in the clang_android test on non-Windows

### LINKED ISSUE #1012: Local tests fail on x86-64 Linux since rust-lang/cc-rs#991
Bisected a regression in running tests on main to 328f8d83458cbdf7c62469a206bdcf2aebd9eb68

```
running 40 tests
LLVM (http://llvm.org/):
  LLVM version 17.0.6
  Optimized build.
test gnu_extra_warnings0 ... ok
test compile_intermediates ... ok
test gnu_compile_assembly ... ok
test gnu_define ... ok
test gnu_debug_fp_auto ... ok
test gnu_debug_fp ... ok
test gnu_extra_warnings1 ... ok
cargo:warning=llvm-ar: error: /home/jubilee/rust/cc-rs/target/debug/gcc-test7gypcv/d1fba762150c532c-foo.o: No such file or directory
error: test failed, to rerun pass `--test test`

Caused by:
  process didn't exit successfully: `/home/jubilee/rust/cc-rs/target/debug/deps/test-9add85ae16a35a83` (exit status: 1)
```

I run x86-64 Linux and `cargo test` fails here. Not clear why. If the tests now require something be installed that wasn't installed before, that's a documentation bug, I would think.

cc @dpaoliello @NobodyXu

It is possible this may be related to the failure to upgrade cc-rs for rust-lang/rust's bootstrap to 1.0.90 in https://github.com/rust-lang/rust/pull/122504 but I don't know if it is.

dpaoliello:
In CI we detect and use `arm-linux-androideabi-ar` but some folks locally are seeing the test use `llvm-ar` - this is likely a difference in dependencies on the system (e.g., Android NDK installed or not) so we'll shim both.

Fixes #1012

### DIFF
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -642,7 +642,8 @@ fn clang_android() {
     {
         let test = Test::new();
         test.shim("arm-linux-androideabi-clang")
-            .shim("arm-linux-androideabi-ar");
+            .shim("arm-linux-androideabi-ar")
+            .shim("llvm-ar");
         test.gcc().target(target).file("foo.c").compile("foo");
         test.cmd(0).must_not_have("--target=arm-linux-androideabi");
     }


NobodyXu:
@workingjubilee Can you please try this PR?

workingjubilee:
Yep, passes on my machine.

workingjubilee (approved):

NobodyXu:
Thank you!

PR #1019, opened 2024-03-28
Add independent flags to C or C++ files

varphone:
This PR introduces a new feature for C/C++ flags, primarily to support the compilation of mixed C/C++ code.

Sometimes we use both C and C++ languages in a single project. However, the compilers for C and C++ are different, so we need to specify different compilers during compilation. This PR adds independent compiler flags for C and C++ files, allowing us to specify different compilers during compilation.

The main changes in this PR are as follows:

1. Add `Build::c_flag()` and `Build::cpp_flag()` methods to add independent flags to C or C++ files.
2. Add `src` argument to `Tool::to_command()` method to auto-detect c/cpp compiler by source file extension for Clang/Gnu toolset.


src/tool.rs
```diff
@@ -258,19 +264,72 @@ impl Tool {
...
+                    if is_c {
+                        path = PathBuf::from(s.replace("g++", "gcc"));
+                        extra_args = &self.c_args;
+                    }
+                    if is_cpp {
+                        path = PathBuf::from(s.replace("gcc", "g++"));
+                        extra_args = &self.cpp_args;
+                    }
+                }
+                _ => {}
```
NobodyXu, line 310: It seems that the `c_flags` and `cpp_flags` are ignored for msvc?

src/lib.rs
```diff
@@ -564,6 +568,36 @@ impl Build {
...
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.cpp")
+    ///     .cpp_flag("-std=c++17")
+    ///     .compile("foo");
+    /// ```
+    pub fn cpp_flag(&mut self, flag: &str) -> &mut Build {
```
NobodyXu, line 596: I think `cxx_flags` makes more sense.

`cpp` could be confused with pre-processors of C/C++ compiler.

src/lib.rs
```diff
@@ -1633,7 +1667,7 @@ impl Build {
     /// This will return a result instead of panicking; see expand() for the complete description.
     pub fn try_expand(&self) -> Result<Vec<u8>, Error> {
         let compiler = self.try_get_compiler()?;
-        let mut cmd = compiler.to_command();
+        let mut cmd = compiler.to_command(None);
```
NobodyXu, line 1670: Hmmm I wonder if there's any C/C++ that affects the expand of C/C++ program?

I can't think of any specific to C or specific to C++, AFAIK the only way to affect that is via `-D`.

Maybe changing std version would affect that, I recalled that newer C standard introduces more macros.

src/lib.rs
```diff
@@ -1790,6 +1824,14 @@ impl Build {
...
         }
 
+        for flag in self.c_flags.iter() {
+            cmd.c_args.push((**flag).into());
+        }
+
+        for flag in self.cpp_flags.iter() {
+            cmd.cpp_args.push((**flag).into());
+        }
+
```
NobodyXu, line 1834: I think you need to do this in `get_base_compiler`, there are some functions which call it directly.

NobodyXu (changes requested):
Thank you!

There are a few things I'd like you to change.

Also, can you add some tests for this functionality in `tests/test.rs` please?

src/tool.rs
varphone: MSVC has only one `cl.exe` and has no special parameters for `C` or `C++`, so it can be ignored.

NobodyXu: It doesn't have `-std=c11` which is used in example?

src/tool.rs
```diff
@@ -284,19 +290,72 @@ impl Tool {
...
     /// Converts this compiler into a `Command` that's ready to be run.
     ///
     /// This is useful for when the compiler needs to be executed and the
     /// command returned will already have the initial arguments and environment
     /// variables configured.
-    pub fn to_command(&self) -> Command {
+    ///
+    /// The `src` argument is used to determine the preferred compiler for the
+    /// source file. If `None`, the default compiler is used.
+    pub fn to_command(&self, src: Option<&PathBuf>) -> Command {
```
NobodyXu, line 350: This is breaking change, please move this to a new function.

NobodyXu (changes requested):

PR #1021, opened 2024-04-01
fix(env_tool): return None if env is empty

cathaysia:
Sometime env maybe a empty value, we should return None in this case:

image

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2948,16 +2948,17 @@ impl Build {
     /// Returns compiler path, optional modifier name from whitelist, and arguments vec
     fn env_tool(&self, name: &str) -> Option<(PathBuf, Option<String>, Vec<String>)> {
         let tool = match self.getenv_with_target_prefixes(name) {
-            Ok(tool) => tool,
-            Err(_) => return None,
+            Ok(tool) if !tool.trim().is_empty() => tool,
+            _ => return None,
         };
+        let tool = tool.trim();
 
         // If this is an exact path on the filesystem we don't want to do any
         // interpretation at all, just pass it on through. This'll hopefully get
         // us to support spaces-in-paths.
-        if Path::new(&*tool).exists() {
+        if Path::new(tool).exists() {
             return Some((
-                PathBuf::from(&*tool),
+                PathBuf::from(tool),
                 Self::rustc_wrapper_fallback(),
                 Vec::new(),
             ));


src/lib.rs
```diff
@@ -2948,7 +2948,12 @@ impl Build {
...
     fn env_tool(&self, name: &str) -> Option<(PathBuf, Option<String>, Vec<String>)> {
         let tool = match self.getenv_with_target_prefixes(name) {
-            Ok(tool) => tool,
+            Ok(tool) => {
+                if tool.is_empty() {
+                    return None;
+                }
+                tool
+            }
             Err(_) => return None,
```
NobodyXu, line 2957: You can change this to:

```
    Ok(tool) if !tool.is_empty() => tool,
    _ => return None,
```

NobodyXu (changes requested):
Thanks, one nit suggestion:

cathaysia:
Is anything blocking this?

src/lib.rs
```diff
@@ -2948,8 +2948,8 @@ impl Build {
     /// Returns compiler path, optional modifier name from whitelist, and arguments vec
     fn env_tool(&self, name: &str) -> Option<(PathBuf, Option<String>, Vec<String>)> {
         let tool = match self.getenv_with_target_prefixes(name) {
-            Ok(tool) => tool,
-            Err(_) => return None,
+            Ok(tool) if !tool.trim().is_empty() => tool,
```
NobodyXu, line 2951:
```suggestion
            Ok(tool) if !tool.trim().is_empty() => tool.trim(),
```

NobodyXu (changes requested):
Just another suggestion:

Since you have use `trim()` before checking `.is_empty()`, I'd like the result to also be `trim()`.

NobodyXu:
> Is anything blocking this?

I was working on something else, but generally I will response within a few hours.

Your PR generally looks good to me, just have a little bit to improve on, before merging it in.

src/lib.rs
cathaysia: trim before is_empty was used for avoid  empty value(such the value is space). maybe we should combine two

NobodyXu: Yeah that makes sense

NobodyXu (approved):
Thanks, sorry I somehow didn't get a notification from github

PR #1027, opened 2024-04-08
Fix linking with llvm-lib

### LINKED ISSUE #1026: `1.0.91` update breaks `cargo-xwin`
related issue: https://github.com/rust-cross/cargo-xwin/issues/104

minimal reproducible example: https://github.com/namse/cc-rs-1.0.91-issue

Would you please check this problem?

NobodyXu:
Fix #1026


russelltg:
This makes sense--i didn't consider the use of llvm-lib. Maybe I should be using that in my projects instead..... 🤷‍♀️

PR #1029, opened 2024-04-10
Add visionOS support

madsmtm:
I haven't added this to CI, as the Rust builds themselves are still somewhat broken until https://github.com/rust-lang/libc/pull/3568 lands.

CC @BlackHoleFox

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1892,6 +1892,7 @@ impl Build {
                 if !target.contains("apple-ios")
                     && !target.contains("apple-watchos")
                     && !target.contains("apple-tvos")
+                    && !target.contains("apple-visionos")
                 {
                     cmd.push_cc_arg("-ffunction-sections".into());
                     cmd.push_cc_arg("-fdata-sections".into());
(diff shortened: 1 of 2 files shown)


QuentinPerez:
We were definitely working on the same thing; I made almost the same changes. The only differences are the XROS_DEPLOYMENT_TARGET and this additional check on lines 2658 and 2674
```rust
                // there is no -m{}os-version-min for watchos, tvos, visionos
                if matches!(os, AppleOs::Ios | AppleOs::MacOs) {
                    cmd.args.push(
                        format!("-m{}os-version-min={}", sdk_details.sdk_prefix, min_version)
                            .into(),
                    );
                }

```

madsmtm:
> there is no -m{}os-version-min for watchos, tvos, visionos

There is, see [Clang's docs for the `availability` attribute](https://clang.llvm.org/docs/AttributeReference.html#availability) and [Clang's docs for the flag itself](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mtvos-version-min).

However, you right that it does not exist for visionOS, I've opened https://github.com/llvm/llvm-project/issues/88271 for that.

madsmtm:
Using just `--target` is [recommended by Clang Driver developers](https://github.com/llvm/llvm-project/issues/88271#issuecomment-2050425068), which we already do, so I think that is a non-issue.

Have opened https://github.com/rust-lang/cc-rs/issues/1030 to track removing the other flags from Apple targets.

BlackHoleFox (approved):
Nice :D

NobodyXu:
Thank you!

Issue #1020, opened 2024-03-29
Add ability to have include_directories before CFLAGS


WesleyRosenblum:
Currently , the contents of any `CFLAGS` or `CXXFLAGS` environment variables are added to the build cmd prior to `include_directories` that had been added using `fn include`:

https://github.com/rust-lang/cc-rs/blob/be62f4a9027a30ccbfffa494ce8ee3487de8cb9b/src/lib.rs#L1766-L1770

This can result in conflicts if the specified includes need to come prior to the `CFLAGS` includes. 

We've worked around this by manually prepending includes to `CFLAGS` (see https://github.com/aws/s2n-tls/pull/4338), but it would be preferable for this to be configurable in `cc`. 

Thanks!

NobodyXu:
That's definitely doable.

I would accept any PR for this.

Issue #376, opened 2019-02-26
Insert user-provided CFLAGS at the end of the command line


mephi42:
I'm trying to build `libsqlite3-sys`, which uses `cc-rs` on Windows, and I am running into a gcc internal compiler error. Since it happens in the DWARF area of the compiler, I would like to switch off the debug info generation. In order to do that, I set `CFLAGS=-g0`, however, this results in the following command line:

```
running: "gcc.exe" "-O3" "-ffunction-sections" "-fdata-sections" "-g0" "-g" "-fno-omit-frame-pointer" "-m64" "-DSQLITE_CORE" "-DSQLITE_DEFAULT_FOREIGN_KEYS=1" "-DSQLITE_ENABLE_API_ARMOR" "-DSQLITE_ENABLE_COLUMN_METADATA" "-DSQLITE_ENABLE_DBSTAT_VTAB" "-DSQLITE_ENABLE_FTS3" "-DSQLITE_ENABLE_FTS3_PARENTHESIS" "-DSQLITE_ENABLE_FTS5" "-DSQLITE_ENABLE_JSON1" "-DSQLITE_ENABLE_LOAD_EXTENSION=1" "-DSQLITE_ENABLE_MEMORY_MANAGEMENT" "-DSQLITE_ENABLE_RTREE" "-DSQLITE_ENABLE_STAT2" "-DSQLITE_ENABLE_STAT4" "-DSQLITE_HAVE_ISNAN" "-DSQLITE_SOUNDEX" "-DSQLITE_THREADSAFE=1" "-DSQLITE_USE_URI" "-DHAVE_USLEEP=1" "-o" "<redacted>/sqlite3.o" "-c" "sqlite3/sqlite3.c"
```

Note that my `-g0` comes before the automatically inserted `-g` - this leads to it being completely ignored. To me, one of the points of having `CFLAGS` is to be able to insert overrides like this one. Would it be possible to insert user-provided `CFLAGS` at the end of the command line?

GeorgeHahn:
This should be fixed in version 1.0.30 - `libsqlite3-sys` should pull this version in automatically.

alexcrichton:
Closing since this should be fixed!

PR #1043, opened 2024-04-23
Use specified include directories prior to CFLAGS

justsmth:
## Issues
See: 
* #1020
* #376

## Problem
The header files in the directories provided via `CFLAGS` can conflict with the specific headers needed for the project being built.  The include files specified by the project should take precedence over those provided via `CFLAGS`.

## Proposed change
A simple re-ordering of the arguments to add the specified include directories prior to those in `CFLAGS`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1773,15 +1773,15 @@ impl Build {
             cmd.push_cc_arg(format!("-std{}{}", separator, std).into());
         }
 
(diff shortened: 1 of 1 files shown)


NobodyXu (approved):
Thanks, it LGTM, I don't think it would break use of CFLAGS/CXXFLAGS

PR #1046, opened 2024-04-30
Drop support for Visual Studio 12 (2013)

dpaoliello:
Visual Studio 2013 went out of support on Apr 9, 2024: <https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2013>

Additionally, the Rust Compiler has required Visual Studio 2017 since 1.66: <https://github.com/rust-lang/rust/pull/102577/commits/9b3db34072551680c12da4192ad528a01579ce61>

### DIFF
--- a/src/windows/find_tools.rs
+++ b/src/windows/find_tools.rs
@@ -256,7 +256,7 @@ mod impl_ {
     impl LibraryHandle {
         fn new(name: &[u8]) -> Option<Self> {
             let handle = unsafe { LoadLibraryA(name.as_ptr() as _) };
-            (!handle.is_null()).then(|| Self(handle))
+            (!handle.is_null()).then_some(Self(handle))
         }
 
         /// Get a function pointer to a function in the library.
@@ -1178,7 +1178,7 @@ mod impl_ {
     }
 
     #[inline(always)]
-    pub(super) fn has_msbuild_version(version: &str, _: &dyn EnvGetter) -> bool {
+    pub(super) fn has_msbuild_version(_version: &str, _: &dyn EnvGetter) -> bool {
         false
     }
 }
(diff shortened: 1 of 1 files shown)


NobodyXu:
It seems reasonable to drop vs 12, but I will wait for others' review in case they are against this change.

src/windows/find_tools.rs
```diff
@@ -87,15 +87,12 @@ pub fn find_tool(target: &str, tool: &str) -> Option<Tool> {
...
         .or_else(|| impl_::find_msvc_14(tool, target))
-        .or_else(|| impl_::find_msvc_12(tool, target))
 }
 
 /// A version of Visual Studio
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 #[non_exhaustive]
 pub enum VsVers {
-    /// Visual Studio 12 (2013)
-    Vs12,
```
thomcc, line 98: This is a public struct, so removing this would be a major breaking change. I don't want `cc` to need 2.0 for this.

thomcc (commented):
Eventually it probably makes sense to remove, but we still support Rust pre-1.66 (we support back to 1.63). We should keep the enum member though, to avoid needing to cut cc 2.0.

src/windows/find_tools.rs
cuviper: The variant could be `#[deprecated]` instead, especially since users already need a wildcard pattern for `#[non_exhaustive]` if they're matching this.

dpaoliello: Ah, sorry, I didn't realize that this was exported from the crate!
I've added back `Vs12` and marked it as deprecated.

NobodyXu:
As per zulip discussion, we decided to wait until msrv is bumped to 1.66

since the current msrv of cc still supports vs 12, merging it now might break some users.

Once msrv is bumped to 1.66, we will merge this.

NobodyXu:
Discussion is in t-libs/crates, request review for cc-rs

NobodyXu (approved):

PR #1047, opened 2024-05-02
Don't use the value of SDKROOT if it doesn't match the target.

jfgoog:
The SDKROOT environment variable can be problematic on Mac. In cases where you are compiling for multiple targets, SDKROOT cannot be right for all of them. Furthermore, the system Python interpreter sets SDKROOT to the MacOSX SDK if not already set.

So, if you are using a Python script to run a build, and you need to build for multiple targets, the logic in the cc crate doesn't work. This is precisely what is happening with rustc itself, and so we can't upgrade the version of the cc crate used by the bootstrap code.

(Unsetting SDKROOT doesn't work either because the custom clang build that rustc uses for CI depends on it being set)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3656,7 +3656,7 @@ impl Build {
         // Code copied from rustc's compiler/rustc_codegen_ssa/src/back/link.rs.
         if let Some(sdkroot) = env::var_os("SDKROOT") {
             let p = PathBuf::from(sdkroot);
-            let sdkroot = sdkroot.to_string_lossy();
+            let sdkroot = p.to_string_lossy();
             match sdk {
                 // Ignore `SDKROOT` if it's clearly set for the wrong platform.
                 "appletvos"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3654,7 +3654,7 @@ impl Build {
 
     fn apple_sdk_root(&self, sdk: &str) -> Result<OsString, Error> {
         // Code copied from rustc's compiler/rustc_codegen_ssa/src/back/link.rs.
-        if let Ok(sdkroot) = env::var("SDKROOT") {
+        if let Some(sdkroot) = env::var_os("SDKROOT") {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -3653,8 +3653,42 @@ impl Build {
     }
 
     fn apple_sdk_root(&self, sdk: &str) -> Result<OsString, Error> {
-        if let Some(sdkroot) = env::var_os("SDKROOT") {
-            return Ok(sdkroot);
+        // Code copied from rustc's compiler/rustc_codegen_ssa/src/back/link.rs.
+        if let Ok(sdkroot) = env::var("SDKROOT") {
```
NobodyXu, line 3657: Let's use var_os here to handle non-utf8 path:

```suggestion
        if let Some(sdkroot) = env::var_os("SDKROOT") {
```

NobodyXu (changes requested):

src/lib.rs
jfgoog: We don't have `contains()` in OsString, which we need. An alternative that would avoid converting it to UTF-8 would be to parse it into a Path

madsmtm (approved):
I think this makes sense to do

src/lib.rs
```diff
@@ -3653,8 +3653,42 @@ impl Build {
     }
 
     fn apple_sdk_root(&self, sdk: &str) -> Result<OsString, Error> {
+        // Code copied from rustc's compiler/rustc_codegen_ssa/src/back/link.rs.
         if let Some(sdkroot) = env::var_os("SDKROOT") {
-            return Ok(sdkroot);
+            let p = Path::new(&sdkroot);
```
NobodyXu, line 3658:
```suggestion
            let p = Path::new(&sdkroot);
            let sdkroot = sdkroot.to_string_lossy();
```

NobodyXu (approved):
Thank you!

PR #1050, opened 2024-05-03
Avoid recursive calls into try_get_compiler when items are in the cache

dpaoliello:
Have `try_get_compiler` provide the compiler path to `is_flag_supported`, and use `get_base_compiler` if called externally.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -601,26 +601,17 @@ impl Build {
     /// `known_flag_support` field. If `is_flag_supported(flag)`
     /// is called again, the result will be read from the hash table.
     pub fn is_flag_supported(&self, flag: &str) -> Result<bool, Error> {
-        let target = self.get_target()?;
+        self.is_flag_supported_inner(flag, self.get_base_compiler()?.path(), &self.get_target()?)
+    }
 
-        let mut compiler = {
-            let mut cfg = Build::new();
-            cfg.flag(flag)
-                .cargo_metadata(self.cargo_output.metadata)
-                .target(&target)
-                .opt_level(0)
-                .host(&self.get_host()?)
-                .debug(false)
-                .cpp(self.cpp)
-                .cuda(self.cuda);
-            if let Some(ref c) = self.compiler {
-                cfg.compiler(c.clone());
-            }
-            cfg.try_get_compiler()?
-        };
-
+    fn is_flag_supported_inner(
+        &self,
+        flag: &str,
+        compiler_path: &Path,
+        target: &str,
+    ) -> Result<bool, Error> {
         let compiler_flag = CompilerFlag {
-            compiler: compiler.path.clone().into(),
+            compiler: compiler_path.into(),
             flag: flag.into(),
         };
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -601,26 +601,14 @@ impl Build {
...
-                .debug(false)
-                .cpp(self.cpp)
-                .cuda(self.cuda);
-            if let Some(ref c) = self.compiler {
-                cfg.compiler(c.clone());
-            }
-            cfg.try_get_compiler()?
-        };
+    fn is_flag_supported_inner(&self, flag: &str, compiler_path: &Path) -> Result<bool, Error> {
+        let target = self.get_target()?;
```
NobodyXu, line 608: let's also move this to line 629

NobodyXu: Thinking about this again - maybe we can move this into try_get_compiler?

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -637,6 +623,21 @@ impl Build {
         let out_dir = self.get_out_dir()?;
         let src = self.ensure_check_file()?;
         let obj = out_dir.join("flag_check");
+        let target = self.get_target()?;
```
NobodyXu, line 626: Let's move these calls into try_get_compiler and is_flag_supported, that would save a few calls in try_get_compiler

src/lib.rs
```diff
@@ -1805,7 +1806,10 @@ impl Build {
...
 
         for flag in self.flags_supported.iter() {
-            if self.is_flag_supported(flag).unwrap_or(false) {
+            if self
+                .is_flag_supported_inner(flag, &cmd.path)
+                .unwrap_or(false)
+            {
                 cmd.push_cc_arg((**flag).into());
             }
         }
```
NobodyXu, line 1815: A while ago, I was thinking of parallelising this to speed it up.

If you'd like to do this, I recommend creating a new fn is_flags_supported.

You can take a look at compile_objects, if you are interested in doing so

NobodyXu (changes requested):

src/lib.rs
dpaoliello: Except that these are post cache check, so moving them earlier hurts the fast path (even if we can amortize them by moving them out of the loop in `try_get_compiler`). An alternative is to make them `&mut Option<_>` so that they can be remembered if they are set in the loop, but I'm not sure that's worth the complexity it introduces.

src/lib.rs
dpaoliello: That's an interesting idea, but I'd rather do it in a different PR.

One complication: I'd want to make sure that we don't set up any async stuff if everything is already cached...

src/lib.rs
NobodyXu: That's true, they are probably cheap enough to not be an issue.

src/lib.rs
NobodyXu: We could iterate over all flags, and then create two vec `supported_flags: Vec<&str>` and `flags_to_check: Vec<(Command, &str)>`.

Doing so would also avoid the duplicate calls I mentioned in the other code review.

NobodyXu: > That's an interesting idea, but I'd rather do it in a different PR.

That makes sense, I can create a separate issue for it.

NobodyXu: Opened #1052 for this

src/lib.rs
```diff
@@ -601,26 +601,17 @@ impl Build {
...
-                .host(&self.get_host()?)
-                .debug(false)
-                .cpp(self.cpp)
-                .cuda(self.cuda);
-            if let Some(ref c) = self.compiler {
-                cfg.compiler(c.clone());
-            }
-            cfg.try_get_compiler()?
-        };
+        self.is_flag_supported_inner(flag, self.get_base_compiler()?.path(), &self.get_target()?)
```
NobodyXu, line 604: I thought we were going to leave this in the inner function.

NobodyXu (changes requested):

NobodyXu (approved):
Thank you!

src/lib.rs
NobodyXu: While it is a bit different from what I expected, I think this change is ok

PR #1051, opened 2024-05-04
Replace println with print_warning in Build::try_get_compiler

NobodyXu:
A println would not be seen by user unless -v is passed to cargo, using print_warning here makes more sense as user can see the message by default and also disable it if they don't want to see it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1762,7 +1762,7 @@ impl Build {
         if !no_defaults {
             self.add_default_flags(&mut cmd, &target, &opt_level)?;
         } else {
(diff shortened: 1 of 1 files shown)


thomcc:
Hmm, isn't this a setting they requested? Why does it make sense to warn on things the user asks for?

NobodyXu:
> Hmm, isn't this a setting they requested? Why does it make sense to warn on things the user asks for?

Hmmm yeah...Maybe it makes more sense to remove this completely

NobodyXu:
@thomcc I opened #1063 to remove the println

PR #1054, opened 2024-05-04
fix: emscripten build on windows

atty303:
Since family detection is performed for `cmd.exe`, it recognizes `msvc` instead of `clang` and passes the wrong compilation options (`-nologo`).


NobodyXu:
cc @thomcc Could you review this please?

thomcc (approved):
This makes sense to me.

PR #1068, opened 2024-05-09
Support compiling on wasm targets

jozanza:
In this PR, I added `target_family = "wasm"` to the `compile_error!` exceptions list in a few modules. 

For more context, in a wasm game engine SDK I maintain ([`turbo`](https://crates.io/crates/turbo-genesis-sdk)), I'm pulling in a crate ([`solana-sdk`](https://crates.io/crates/solana-sdk)). However, `cc-rs` is deeply buried in dependencies of dependencies of that crate. This wouldn't be a terrible problem since `cc-rs` (especially the parallel feature stuff) is far from any code path that runs in wasm and optimization can get rid of a unused code, however, the `compile_error!` macro prevents `cc-rs` prevents builds from ever completing in the first place.

So the only option in my case (where `cc-rs` is buried deep in the dependency tree and I'm targeting wasm) currently is to fork and patch. This gets brittle fairly quickly since other dependencies can have version conflicts and force me to create more branches with version bumps and other minor tweaks over time.

I'm hoping we can land this patch and carve out exceptions for wasm with `compile_error!` moving forward 🙏🏽

### DIFF
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -130,6 +130,8 @@ impl StderrForwarder {
                             self.bytes_available_failed = true;
                             MIN_BUFFER_CAPACITY
                         }
+                        #[cfg(target_family = "wasm")]
+                        Err(_) => unimplemented!(),
                         Ok(bytes_available) => MIN_BUFFER_CAPACITY.max(bytes_available),
                     }
                 } else {
--- a/src/parallel/stderr.rs
+++ b/src/parallel/stderr.rs
@@ -3,7 +3,7 @@ use std::{convert::TryInto, process::ChildStderr};
 
 use crate::{Error, ErrorKind};
 
-#[cfg(all(not(unix), not(windows)))]
+#[cfg(all(not(unix), not(windows), not(target_family = "wasm"))]
 compile_error!("Only unix and windows support non-blocking pipes! For other OSes, disable the parallel feature.");
 
 #[cfg(unix)]
--- a/src/parallel/stderr.rs
+++ b/src/parallel/stderr.rs
@@ -3,7 +3,7 @@ use std::{convert::TryInto, process::ChildStderr};
 
 use crate::{Error, ErrorKind};
 
-#[cfg(all(not(unix), not(windows), not(target_family = "wasm"))]
+#[cfg(all(not(unix), not(windows), not(target_family = "wasm")))]
 compile_error!("Only unix and windows support non-blocking pipes! For other OSes, disable the parallel feature.");
 
 #[cfg(unix)]
--- a/src/tempfile.rs
+++ b/src/tempfile.rs
@@ -6,7 +6,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
-#[cfg(not(any(unix, target_os = "wasi", windows)))]
+#[cfg(not(any(unix, target_family = "wasm", windows)))]
 compile_error!("Your system is not supported since cc cannot create named tempfile");
 
 fn rand() -> u64 {


madsmtm:
Wait... These configs are for the _host_ machine - why are you compiling `cc` _itself_ to run on WASM?

Anyhow, I'd rather suggest that you fix the dependencies of `solana-sdk` that rely on `cc`, namely `blake3` and `solana-program`, of which the last one [is already cfg-guarded, so it should just work on WASM](https://github.com/solana-labs/solana/blob/v1.18.12/sdk/program/Cargo.toml#L90-L91).

NobodyXu (approved):
Thanks!

Can you update our CI to run `cargo check` for wasm target?

NobodyXu (changes requested):
(Sorry I'm bit sleepy today.)

Is cc-rs used in the build-script?
If so then it would be build for your native OS instead of wasm.

jozanza:
> Wait... These configs are for the _host_ machine - why are you compiling `cc` _itself_ to run on WASM?
> 
> 
> 
> Anyhow, I'd rather suggest that you fix the dependencies of `solana-sdk` that rely on `cc`, namely `blake3` and `solana-program`, of which the last one [is already cfg-guarded, so it should just work on WASM](https://github.com/solana-labs/solana/blob/v1.18.12/sdk/program/Cargo.toml#L90-L91).

Yeah that's a great suggestion, and it is fixed in the latest version of solana-sdk as you noticed. The problem was originally caused by that dependency not being cfg-guarded in older versions. Unfortunately, those older versions are often still in use and will always need a patch.

#It's very easy for library authors to not cfg-guard their deps with wasm exclusions (or even realize they should) or to include a dependency that fails to do so. So if another dependency of a dependency makes the same mistake (not cfg-guarding cc) even after such a fix, we'll wind up with the same problem in the future, and once again need to fork/patch. So that's my train of thought here. I'd also suggest that perhaps the compile error being thrown may not be particularly useful for the wasm platform target in practice.

NobodyXu:
@jozanza cc is used as a build-dep right?

If it is a build-dep, then I can't imagine it causing your wasm build to fail, because your host OS is still unix/windows.

Is cc used as a normal dependency in your crate?

jozanza:
You're right @NobodyXu. The issue is it's not my crate that directly uses cc. If it was, using build-deps or cfg-guarding would work just fine. In my situation, cc is deep in the dependency tree, so I can't control these things by any other means than patching.

NobodyXu:
Could you do a `cargo tree -i cc`?

I wonder what is their use case for this, if it's a valid use case (not a mistake) I'm willing to merge this PR.

Usually nobody uses cc on wasm, because wasm does not support spawning any process, which means you can't really use cc to compile anything on wasm.

NobodyXu:
Hello, now that I think about it again, we actually should support wasm/wasi.

There's ongoing effort to support sandboxing (likely using wasi), so we should plan ahead.

cc @thomcc @the8472

src/command_helpers.rs
```diff
@@ -130,6 +130,8 @@ impl StderrForwarder {
                             self.bytes_available_failed = true;
                             MIN_BUFFER_CAPACITY
                         }
+                        #[cfg(target_family = "wasm")]
+                        Err(_) => unimplemented!(),
```
NobodyXu, line 134: Is this matching necessary?

Can't we return it directly?

NobodyXu:
cc @jozanza sorry for the delay, I now think it makes sense to accept this PR, just a minor question.

jozanza:
Oh great to hear @NobodyXu! What's the question?

src/command_helpers.rs
jozanza: Hey sorry could you clarify what "it" is in this case? If I remember correctly, I think the main issue here is that on wasm platforms, the match statement ends up missing an `Err` case and won't compile as a result. To be honest, I wasn't too sure about the logic behind this match statement since it seems to conditionally return a `usize` or a `bool` depending on the platform and whether the parallel feature is enabled. But my naive assessment was that some sort of panic would make sense here for the error case when targeting wasm 😅

NobodyXu:
> Oh great to hear @NobodyXu! What's the question?

There's a comment on the command_helper, just a minor one

src/command_helpers.rs
NobodyXu: I think we should not call this function on wasm.

The `bytes_available` function should also be put behind a `cfg(not(target_family = "wasm"))`

NobodyXu:
Can you also update CI to run `cargo check` for wasm targets?

Thanks

src/command_helpers.rs
NobodyXu: Took another look and it seems ok to panic, because the `bytes_available` always return `Ok()`.

```suggestion
                        #[cfg(target_family = "wasm")]
                        Err(_) => panic!("bytes_available should always succeed on wasm"),
```

NobodyXu:
Thanks, I've opened #1160 which runs the wasm in CI

jozanza:
Awesome! Thanks @NobodyXu

PR #1076, opened 2024-05-20
Don't assume GCC toolchain supports the same dwarf format rust uses

### LINKED ISSUE #1075: `--gdwarf=4` is not compatible with some newer mainstream gcc/as toolchains
I ran into a situation where I'm unable to compile C code via `cc` if `$CC` is `gcc-11` or `gcc-13`, though it works if I use `env CC=gcc-9` and, of course, with `CC` set to any variant of `clang`. All three gcc toolchains were installed via `apt` under Ubuntu 23.10.

The actual command executed by the compiler:

```
[fish 0.1.0] error occurred: Command "sccache" "gcc-13" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-4" "-fno-omit-frame-pointer" "-m64" "-I" "/mnt/d/rand/fish/target/debug/build/fish-1dce6c150ef9f00d/out" "-o" "/mnt/d/rand/fish/target/debug/build/fish-1dce6c150ef9f00d/out/2e40c9e35e9506f4-libc.o" "-c" "src/libc.c" with args gcc-13 did not execute successfully (status code exit status: 1).
[fish 0.1.0]
[fish 0.1.0]
The following warnings were emitted during compilation:

warning: fish@0.1.0: as: unrecognized option '--gdwarf-4'
```

The interesting thing is that `gcc-9 -dumpspecs | grep gdwarf` prints only gdwarf2 support, but both `gcc-11` and `gcc-13` show support for up to gdwarf5:

```shell
> gcc-9 -dumpspecs | grep dwarf
%{%:debug-level-gt(0):%{gstabs*:--gstabs}%{!gstabs*:%{g*:--gdwarf2}}} %{fdebug-prefix-map=*:--debug-prefix-map %*}
%{gsplit-dwarf:
```

```shell
> gcc-11 -dumpspecs | grep dwarf
%{%:debug-level-gt(0):%{!gstabs*:%{g*:%{%:dwarf-version-gt(4):--gdwarf-5 ;%:dwarf-version-gt(3):--gdwarf-4 ;%:dwarf-version-gt(2):--gdwarf-3 ;:--gdwarf2 }}}}
%{gsplit-dwarf:
```

so if anything, I would have expected `gcc-9` to be the one that doesn't support it, but the problem is with `as`:

```shell
> as --gdwarf-4
as: unrecognized option '--gdwarf-4'
```

`-gdwarf-4` was introduced in 0fff25ae9aebfc0635e8b2ac719fda4047b4e252 (cc 1.0.77+), with the rationale that this would cause the compiler to match the compiler format *rust* is using, but I think this doesn't account for the fact that rust could be using an entirely different assembler/toolchain than what C code executed by `build.rs` uses.

mqudsi:
While it may be a fairly safe assumption that we can use the same dwarf version rust uses (0fff25a) if we're compiling with clang, it might not be a valid assumption for the Gnu toolchain.

I tried to pull the logic up to `add_default_flags()`, but that would require running `is_flag_supported()` in the call to `add_default_flags()` which, while possible, might not be the expected or desired behavior. The alternate implementation would look like this, but I think this PR as submitted is probably the better way to go.

<details>
<pre lang="diff">
index 9b2816f..69b7e28 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1921,7 +1921,15 @@ impl Build {
                 cmd.args.push("-G".into());
             }
             let family = cmd.family;
-            family.add_debug_flags(cmd, self.get_dwarf_version());
+            let (flag, opt_flag) = family.get_debug_flags(self.get_dwarf_version());
+            if let Some(flag) = flag {
+                cmd.args.push(flag.into());
+            }
+            if let Some(flag) = opt_flag {
+                if self.is_flag_supported(&flag).unwrap_or(false) {
+                    cmd.args.push(flag.into());
+                }
+            }
         }
 <br>
         if self.get_force_frame_pointer() {
diff --git a/src/tool.rs b/src/tool.rs
index ea0f36f..ff6f517 100644
--- a/src/tool.rs
+++ b/src/tool.rs
@@ -416,19 +416,21 @@ pub enum ToolFamily {
 }
 <br>
 impl ToolFamily {
-    /// What the flag to request debug info for this family of tools look like
-    pub(crate) fn add_debug_flags(&self, cmd: &mut Tool, dwarf_version: Option<u32>) {
+    /// What the flag to request debug info for this family of tools look like.
+    ///
+    /// Return value is a tuple of `(Option<Flag>, Option<FlagIfSupported>)` where `Flag` may
+    /// be set unconditionally but `FlagIfSupported` should be tested for support first.
+    pub(crate) fn get_debug_flags(
+        &self,
+        dwarf_version: Option<u32>,
+    ) -> (Option<String>, Option<String>) {
         match *self {
-            ToolFamily::Msvc { .. } => {
-                cmd.push_cc_arg("-Z7".into());
-            }
-            ToolFamily::Gnu | ToolFamily::Clang { .. } => {
-                cmd.push_cc_arg(
-                    dwarf_version
-                        .map_or_else(|| "-g".into(), |v| format!("-gdwarf-{}", v))
-                        .into(),
-                );
-            }
+            ToolFamily::Msvc { .. } => (Some("-Z7".into()), None),
+            ToolFamily::Clang { .. } => match dwarf_version {
+                Some(v) => (Some(format!("-gdwarf-{v}")), None),
+                None => (Some("-g".into()), None),
+            },
+            ToolFamily::Gnu => (Some("-g".into()), Some(format!("-g").into())),
         }
     }
</pre>
</details>

Closes #1075.

### DIFF
--- a/src/tool.rs
+++ b/src/tool.rs
@@ -422,12 +422,12 @@ impl ToolFamily {
             ToolFamily::Msvc { .. } => {
                 cmd.push_cc_arg("-Z7".into());
             }
+            ToolFamily::Clang { .. } if dwarf_version.is_some() => {
+                cmd.push_cc_arg(format!("-gdwarf-{}", dwarf_version.unwrap()).into());
+            }
+            // Don't assume gcc supports the same dwarf version rust/llvm does (#1075)
             ToolFamily::Gnu | ToolFamily::Clang { .. } => {
-                cmd.push_cc_arg(
-                    dwarf_version
-                        .map_or_else(|| "-g".into(), |v| format!("-gdwarf-{}", v))
-                        .into(),
-                );
+                cmd.push_cc_arg("-g".into());
             }
         }
     }


mqudsi:
I'm not sure what to do about the tests, they're hard-coded to expect `-gdwarf-4` or `-gdwarf-2` depending on the target but the tests don't surface gcc vs clang so it's hard to make them dynamic.

PR #1078, opened 2024-06-02
Deprecate cc::Build::static_flag

NobodyXu:
cc only compiles file and does not do any linking at all, so `-static` doesn't make sense

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -766,7 +766,7 @@ impl Build {
     /// ```
     #[deprecated(
         since = "1.0.99",
-        note = "cc itself always generates a archive. If you use cc::Tool::to_command(), then you shall manually add -static to command"
+        note = "cc itself always generates an archive. If you really need `-static`, use `cc::Tool::to_command()` and add it to the `Command` arguments."
     )]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -764,6 +764,10 @@ impl Build {
     ///     .static_flag(true)
     ///     .compile("foo");
     /// ```
+    #[deprecated(
+        since = "1.0.99",
+        note = "cc itself always generates a archive. If you use cc::Tool::to_command(), then you shall manually add -static to command"
```
thomcc, line 769:
```suggestion
        note = "cc itself always generates an archive. If you really need `-static`, use `cc::Tool::to_command()` and add it to the `Command` arguments."
```

thomcc (commented):
On the one hand, yeah this is (IIUC) basically a linker flag and we don't invoke the linker. On the other hand, I'm 80% sure I've had to use it before.

Is there something that brought about wanting to make this change?

NobodyXu:
It's the same motivation with #1077 , cc doesn't do linking by itself, users using cc for linking would need to convert to Tools to command and manually do that

PR #1079, opened 2024-06-02
Fix panic safety issue in StderrForwarder

### LINKED ISSUE #1036: Possible panic safety issue in StderrForwarder
When auditing changes to `cc`, I found a theoretical panic safety issue at https://github.com/rust-lang/cc-rs/commit/227b770f1d52d3e538f0bdfbe3636acb58ccb87d#diff-fbc116db4f047e6e29dac1fc3c5a5f1f759060e9558aff5707624841b57c5258R123-R128 . As an optimization, the code extends the length of the buffer to its capacity (which will cause it to include uninitialized data), and relies on it being set back to contain only initialized data in all code paths. If `stderr.read` panics and that panic is caught by a caller of `forward_available`, then the inner buffer of `StderrForwarder` will still contain uninitialized data. I don't think it is likely that this can be triggered maliciously, hence the public issue.

NobodyXu:
Fixed #1036

### DIFF
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -104,7 +104,7 @@ impl StderrForwarder {
                 let to_reserve = if self.is_non_blocking && !self.bytes_available_failed {
                     match crate::parallel::stderr::bytes_available(stderr) {
                         #[cfg(windows)]
-                        Ok(0) => return false,
+                        Ok(0) => break false,
                         #[cfg(unix)]
                         Ok(0) => {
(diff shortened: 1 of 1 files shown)


ChrisDenton (approved):
Looks good to me. My only nit is that I don't think this is allow is necessary any more:

https://github.com/rust-lang/cc-rs/blob/ca1f5f9d48c0d917ab27901cd54ed665793d0ddf/src/command_helpers.rs#L93

workingjubilee:
thanks for patching this!

PR #1080, opened 2024-06-03
add as_excutable option

ds1sqe:
add new option on build
( as_executable(flag:bool) )
which create executable output and 
do not compile as lib

### DIFF
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -301,6 +301,25 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
     Ok(objects)
 }
 
+/// Create executable path from given input.
+pub(crate) fn executable_path_from_file(
+    dst: &Path,
+    executable_name: &str,
+) -> Result<Arc<Path>, Error> {
+    let out = dst.join(executable_name);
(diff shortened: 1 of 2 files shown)


NobodyXu (changes requested):
Hello, thanks for the PR, but I am hesitant to merge it.

cc is designed for compiling to static archive instead of executable, it would be a lot more work in supporting executable.

ds1sqe:
But it's will be great if can compile into executable, and i think we can do that by a little efforts.
So why do not add feature with unstable feature?
and with my PR, we can compile as executable (if the "files" have unique main) on gcc and clang and msvc, if the language is C or C++.
with existing "out_dir" and compile(`path`) option.

ds1sqe:
And also there are some projects which have to compile some c / cpp project and configure root rust project with compiled c / cpp executable. that's why i upload this PR

NobodyXu:
I think you can create another crate cc-executable which uses cc to generate an archive first, then compile it to an executable.

The code in this PR breaks parallel compilation, whereas using cc to generate an archive first would support parallel compilation and easier to implement.

cc @thomcc  @ds1sqe I do think we can support generating object files instead of archive though, to support compiling to executable in another crate

PR #1083, opened 2024-06-05
Fix WASI compilation for C++

antaalt:
It seems that currently cc-rs cannot build c++ to WASI correctly, it is using clang instead of clang++, not declaring some unsupported features such as no-exceptions (WASI does not support exceptions), and not linking libc++ aswell that is coming from WASI SDK.
With this PR, I am hoping this could fix this and avoid some extra code such as .compiler("clang++") to fix it

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1912,6 +1912,10 @@ impl Build {
                         cmd.push_cc_arg("-fno-plt".into());
                     }
                 }
+                if target.contains("wasm32") {
+                    // WASI does not support exceptions yet.
+                    cmd.push_cc_arg("-fno-exceptions".into());
+                }
             }
         }
 
@@ -2861,7 +2865,11 @@ impl Build {
                     || target == "wasm32-unknown-wasi"
                     || target == "wasm32-unknown-unknown"
                 {
-                    "clang".to_string()
+                    if self.cpp {
+                        "clang++".to_string()
+                    } else {
+                        "clang".to_string()
+                    }
                 } else if target.contains("vxworks") {
                     if self.cpp {
                         "wr-c++".to_string()
@@ -3099,6 +3107,7 @@ impl Build {
                         | target.contains("openbsd")
                         | target.contains("aix")
                         | target.contains("linux-ohos")
+                        | target.contains("wasm32")
                     {
                         Ok(Some("c++".to_string()))
                     } else if target.contains("android") {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1318,7 +1318,7 @@ impl Build {
                 let wasi_sysroot = self.wasi_sysroot()?;
                 self.cargo_output.print_metadata(&format_args!(
                     "cargo:rustc-flags=-L {}/lib/wasm32-wasi -lstatic=c++ -lstatic=c++abi",
-                    wasi_sysroot.display()
+                    wasi_sysroot
                 ));
             }
         }
@@ -1926,7 +1926,7 @@ impl Build {
                     cmd.push_cc_arg("-fno-exceptions".into());
                     // Link clang sysroot
                     let wasi_sysroot = self.wasi_sysroot()?;
-                    cmd.push_cc_arg(format!("--sysroot={}", wasi_sysroot.display()).into());
+                    cmd.push_cc_arg(format!("--sysroot={}", wasi_sysroot).into());
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1912,6 +1912,19 @@ impl Build {
                         cmd.push_cc_arg("-fno-plt".into());
                     }
                 }
+                if target.contains("wasm32-wasi") {
```
madsmtm, line 1915: Maybe just:

```suggestion
                if target.contains("-wasi") {
```

To support `wasm64-wasi` if that ever arrives?

src/lib.rs
```diff
@@ -1912,6 +1912,19 @@ impl Build {
...
+                    // WASI does not support exceptions yet.
+                    cmd.push_cc_arg("-fno-exceptions".into());
+                    let wasi_sdk = std::env::var("WASI_SDK").expect("Could not find WASI_SDK. Download it from github & setup environment variable WASI_SDK targetting the folder.");
```
NobodyXu, line 1918: Please use `var_os` since the path can contain non utf-8 characters:

```suggestion
                    let wasi_sdk = std::env::var_os("WASI_SDK").expect("Could not find WASI_SDK. Download it from github & setup environment variable WASI_SDK targetting the folder.");
```

NobodyXu (changes requested):

antaalt:
I will look into the changes requested, in the meantime, I think it should be better to use WASI sysroot and not rely on WASI SDK so that we can use standard clang and not wasi SDK bundled clang, I will try to look into this aswell

thomcc:
> not declaring some unsupported features such as no-exceptions (WASI does not support exceptions)

FWIW, I don't think it's libc's job to automatically pass a flag like `-fno-exceptions`, if that's what you mean.

antaalt:
You mean the flag should still be user provided ? As exceptions are not supported on the target and WASI only support clang as a compiler right now, I think it does make sense to provide it

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -3852,6 +3873,18 @@ impl Build {
         }
     }
 
+    fn wasi_sysroot(&self) -> Result<PathBuf, Error> {
```
NobodyXu, line 3876: `self.getenv` returns an `Arc`, so

```suggestion
    fn wasi_sysroot(&self) -> Result<Arc<Path>, Error> {
```

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -1912,6 +1920,14 @@ impl Build {
                         cmd.push_cc_arg("-fno-plt".into());
                     }
                 }
+                if target.contains("-wasi") {
+                    // WASI does not support exceptions yet.
+                    // https://github.com/WebAssembly/exception-handling
```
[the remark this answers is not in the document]
NobodyXu: cc @thomcc I am ok with this `-fno-exception` as long as we only enable it for the wasi thst does not support it.

IIRC rust adds new preview targets for WASI with new features, so this could be fine ifcc checked the target?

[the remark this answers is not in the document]
thomcc: Yeah, probably fine for `-wasi` (e.g. preview 1), but eventually one of these will support exceptions and it'll be weird that we turn it off.

thomcc (approved):
Seems fine I suppose, but I can't help but think this will cause problems in the future.

src/lib.rs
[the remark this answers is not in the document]
NobodyXu: Ok I've found it:

> Rust 1.78 will introduce new wasm32-wasip1 (tier 2) and wasm32-wasip2 (tier 3) targets. wasm32-wasip1 is an effective rename of the existing wasm32-wasi target, freeing the target name up for an eventual WASI 1.0 release.

https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html#:~:text=Rust%201.78%20will%20introduce%20new,an%20eventual%20WASI%201.0%20release.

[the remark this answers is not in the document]
NobodyXu: So the current wasm32-wasi will be renamed to wasm32-wasip1.

cc @antaalt can you change the target name please, to avoid breakage in the future?

cc @thomcc I think this is enough to prevent the breakage?

src/lib.rs
```diff
@@ -1912,6 +1920,14 @@ impl Build {
                         cmd.push_cc_arg("-fno-plt".into());
                     }
                 }
+                if target == "wasm32-wasi" {
```
NobodyXu, line 1923:
```suggestion
                if target == "wasm32-wasip1" {
```

src/lib.rs
```diff
@@ -1313,6 +1313,14 @@ impl Build {
                 self.cargo_output
                     .print_metadata(&format_args!("cargo:rustc-link-lib={}", stdlib));
             }
+            // Link c++ lib from WASI sysroot
+            if self.get_target()?.contains("wasi") {
+                let wasi_sysroot = self.wasi_sysroot()?;
+                self.cargo_output.print_metadata(&format_args!(
+                    "cargo:rustc-flags=-L {}/lib/wasm32-wasi -lstatic=c++ -lstatic=c++abi",
```
NobodyXu, line 1320: Took another look and realise it is hardcoded to wasi.

I don't like it, but I think it's ok given that the preview wasi is also wasi.

NobodyXu (changes requested):

NobodyXu (approved):
Thank you!

It LGTM now!

PR #1088, opened 2024-06-07
Fix clippy warnings on Windows

ChrisDenton:
This is likely best reviewed commit-by-commit. The first commit simply uses `cargo clippy --fix` to automatically fix warnings so there almost certainly aren't any behavioural changes. The other commits are manual fixes but should be fine too. The most tricky one is probably 397dd22 because it's rearranging boolean expressions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1956,8 +1956,8 @@ impl Build {
         // Target flags
         match cmd.family {
             ToolFamily::Clang { .. } => {
-                if !cmd.has_internal_target_arg
-                    && !(target.contains("android")
+                if !(cmd.has_internal_target_arg
+                    || target.contains("android")
                         && android_clang_compiler_uses_target_arg_internally(&cmd.path))
                 {
                     let (arch, rest) = target.split_once('-').ok_or_else(|| {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2590,10 +2590,7 @@ impl Build {
         } else {
             AppleOs::Ios
         };
-        let is_mac = match os {
-            AppleOs::MacOs => true,
-            _ => false,
-        };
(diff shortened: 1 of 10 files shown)


NobodyXu (changes requested):

src/lib.rs
```diff
@@ -1956,8 +1947,8 @@ impl Build {
         // Target flags
         match cmd.family {
             ToolFamily::Clang { .. } => {
-                if !cmd.has_internal_target_arg
-                    && !(target.contains("android")
+                if !(cmd.has_internal_target_arg
+                    || target.contains("android")
                         && android_clang_compiler_uses_target_arg_internally(&cmd.path))
```
NobodyXu, line 1952: Same here, I think we need to call the ()

src/lib.rs
```diff
@@ -3370,7 +3353,7 @@ impl Build {
         // CROSS_COMPILE is of the form: "arm-linux-gnueabi-"
         let cc_env = self.getenv("CROSS_COMPILE");
         let cross_compile = cc_env.as_ref().map(|s| s.trim_end_matches('-').to_owned());
-        cross_compile.or(linker_prefix).or(match &target[..] {
+        cross_compile.or(linker_prefix).or(match target {
```
NobodyXu, line 3356: IMO we'd need

```suggestion
        cross_compile.or(linker_prefix).or_else(|| match target {
```

NobodyXu (changes requested):

src/lib.rs
ChrisDenton: Done. Changing it does affect the formatting.

NobodyXu (approved):
Thanks!

PR #1089, opened 2024-06-07
Don't try to free DLL on drop

ChrisDenton:
Unloading DLLs is an extremely unsafe operation and should not be done without due caution. It's much safer to simply leak it unless we know for certain that it's safe to unload (which is hard to do without co-operations from the DLL itself).

Currently a mitigating factor is that we only use `LibraryHandle` for `kernel32.dll`, which is highly unlikely to ever be unloaded, so it's essentially a no-op anyway. However, it would become more of a problem if we ever use `LibraryHandle` for other DLLs.


NobodyXu (approved):
Thanks, that makes sense, IIRC musl also does the same to avoid the tricky global destructor in dynlib.

PR #1097, opened 2024-06-21
Add tests that execute `{TVOS,XROS}_DEPLOYMENT_VERSION` detecion logic.

briansmith:
This causes the version detection to be skipped and thus it isn't being tested. Remove it so that we test the version detection.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -130,9 +130,9 @@ jobs:
         uses: ilammy/msvc-dev-cmd@v1
         if: startsWith(matrix.build, 'windows-clang')
       - uses: Swatinem/rust-cache@v2
-      - run: cargo test ${{ matrix.no_run }} --workspace --target ${{ matrix.target }}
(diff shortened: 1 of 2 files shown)


briansmith:
The code has:
```
            AppleOs::TvOs => deployment_from_env("TVOS_DEPLOYMENT_TARGET")
                .or_else(default_deployment_from_sdk)
                .unwrap_or_else(|| "9.0".into()),

            AppleOs::VisionOS => deployment_from_env("XROS_DEPLOYMENT_TARGET")
                .or_else(default_deployment_from_sdk)
                .unwrap_or_else(|| "1.0".into()),
```
...so even if the detection from the SDK fails, shouldn't these tests still pass? It seems like there's something wrong here.

briansmith:
The issue is caused by the fact that GitHub Actions defaults to Xcode 15.0.x, not 15.2, and 15.2 is the first to support visionos.

PR #1099, opened 2024-06-21
Use `dep:` syntax to avoid implicit features.

waywardmonkeys:
The only real externally visible feature should be `parallel` and not the implicit features for the optional dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ once_cell = { version = "1.19", optional = true }
 libc = { version = "0.2.62", default-features = false, optional = true }
 
 [features]
-parallel = ["libc", "jobserver", "once_cell"]
+parallel = ["dep:libc", "dep:jobserver", "dep:once_cell"]
 
 [dev-dependencies]
 tempfile = "3"


waywardmonkeys:
With this, adding `cc` as a dependency results in:

```
      Adding cc (local) to build-dependencies
             Features:
             - parallel
```

rather than:

```
      Adding cc v1.0.99 to build-dependencies
             Features:
             - jobserver
             - libc
             - once_cell
             - parallel
```

NobodyXu (approved):
Thank you!

Nice improvements to feature documentation

rafales:
Guys, this broke a lot of code because solana-program has been using `jobserver` dependency for a long time.

NobodyXu:
> Guys, this broke a lot of code because solana-program has been using `jobserver` dependency for a long time.

In that case it's a no-op, enabling jobserver does not enable the parallel compilation, it's still compiled in sequential, just with one extra dependency doing nothing.

waywardmonkeys:
@NobodyXu It looks like they enabled both `parallel` and `jobserver`:

https://github.com/solana-labs/solana/blob/27eff8408b7223bb3c4ab70523f8a8dca3ca6645/sdk/program/Cargo.toml#L92

Now this would fail since there is no `jobserver` feature.

Maybe `cc` should add one that does nothing, but there's no means for deprecating a feature either. I could submit a PR that does one of:

```
jobserver = ["parallel"]
```

or

```
jobserver = []
```

And perhaps we could emit a warning when `cfg(feature = "jobserver")`

NobodyXu:
Right yeah that makes sense, I would accept a PR for it.

waywardmonkeys:
I'll do this ... but I notice that it has been 3 weeks and only apparently one thing is broken and I don't see that anyone in that community has reported an issue or filed a PR to fix their usage of `cc`.

PR #1100, opened 2024-06-22
Accpet `OsStr` instead of `str` for flags

### LINKED ISSUE #1018: Support arbitrary `OsStr`s as arguments to compiler
I want to pass `-fdebug-prefix-map=PATH1=PATH2` to the compiler. However, to be most portable, these should be `&Path`s and the overall flag should be an `OsString`. However, I don't see a way to pass custom os strings to the builder, unless I manually recreate the logic in `.compile(output)` with a bunch of `.to_command()`s. 

Could there be an easy way to pass arbitrary os strings as arguments?

NobodyXu:
Fixed #1018

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -267,8 +267,8 @@ pub struct Build {
     no_default_flags: bool,
     files: Vec<Arc<Path>>,
     cpp: bool,
-    cpp_link_stdlib: Option<Option<Arc<Path>>>,
-    cpp_set_stdlib: Option<Arc<Path>>,
+    cpp_link_stdlib: Option<Option<Arc<str>>>,
+    cpp_set_stdlib: Option<Arc<str>>,
     cuda: bool,
     cudart: Option<Arc<str>>,
     ccbin: bool,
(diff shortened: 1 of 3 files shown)


madsmtm (approved):
Definitely a good idea!

src/lib.rs
```diff
@@ -954,6 +965,43 @@ impl Build {
...
+    ///
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     .shared_flag(true)
+    ///     .cpp_link_stdlib("stdc++")
+    ///     .compile("libfoo.so");
+    /// ```
+    #[deprecated(since = "1.0.100", note = "please use `cpp_link_stdlib_path` instead")]
+    pub fn cpp_link_stdlib<'a, V: Into<Option<&'a str>>>(
```
ChrisDenton, line 980: Do we absolutely have to deprecate this rather than, I dunno, doing some special trait shenanigans so it works with str but OsStr also works? Maybe we do but it seems unfortunate.

ChrisDenton (commented):
I'm ok with this but I have just one query really. Not a blocker though.

src/lib.rs
NobodyXu: No I don't think we can, `Into< Option<&str>>` prevents this.

thomcc: Yeah, this is unfortunate. It's usually not a path, but a literal `"c++"` or `"stdc++"` so I don't know that deprecating this is worth it.

thomcc (commented):
Unsure about the deprecation.

src/lib.rs
NobodyXu: I will revert that

src/lib.rs
```diff
@@ -954,6 +965,43 @@ impl Build {
...
+    /// # Example
+    ///
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     .shared_flag(true)
+    ///     .cpp_link_stdlib("stdc++")
+    ///     .compile("libfoo.so");
+    /// ```
+    #[deprecated(since = "1.0.100", note = "please use `cpp_link_stdlib_path` instead")]
```
NobodyXu, line 979:
```suggestion
```

src/lib.rs
```diff
@@ -954,6 +965,43 @@ impl Build {
...
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     .cpp_set_stdlib("c++")
+    ///     .compile("libfoo.a");
+    /// ```
+    #[deprecated(since = "1.0.100", note = "please use `cpp_set_stdlib_path` instead")]
```
NobodyXu, line 997:
```suggestion
```

src/lib.rs
NobodyXu: Reverted

NobodyXu: I've reverted changes to `cpp_{set, link}_stdlib` according to review feedback of @thomcc

sagudev:
Just note that this is a breaking change: https://github.com/servo/servo/actions/runs/9672905431/job/26686002372?pr=32616#step:12:441

NobodyXu:
Oops looks like cargo-semver-check missed this one cc @obi1kenobi

```
678 |             build.flag(&confdefs_path.to_string_lossy());
    |                   ----  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::AsRef<std::ffi::OsStr>` is not implemented for `std::borrow::Cow<'_, str>`, which is required by `&std::borrow::Cow<'_, str>: std::convert::AsRef<std::ffi::OsStr>`
    |                   |
```

looks like we would have to revert the change to the interface and add a set of new interface to accept  OsStr

NobodyXu:
@sagudev It is unfortunate that AsRef<OsStr> does not work for Cow<'_, str>, however looking at the log, it seems that they are converting an OsStr to Cow<'_, str> to workaround a past limiation of our interface.

maybe the best way forward is to pass the OsStr directly?

sagudev:
Yes, indeed. I just wanted to warn that this is a breaking change, although trivial to fix on our side, there might be more crates affected by this.

NobodyXu:
Thank you, now that we have released 1.0.100, I'm not sure if we should yank and then revert its interface, that could break people on new releases.

On the other hand, most people are probably using old version and has yet update, so maybe we could still yank and revert?

thomcc:
It's probably worth yanking rather than breaking semver if we have a way to avoid breaking semver.

thomcc:
Eh, At this point we've probably missed the boat for this to be a good candidate for yanking. I may change my mind if more reports of bustage come in though.

PR #1105, opened 2024-06-24
Fix invalid wasi targets compatibility

antaalt:
Thanks for your help following [Fix WASI compilation for C++](https://github.com/rust-lang/cc-rs/pull/1083), I have tested it on my side from main and spotted some issues with incompatible targets, took some time to clean this aswell and ensure every targets are clearly defined to avoid potential issues with no-exceptions flags

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1288,7 +1288,8 @@ impl Build {
         self.compile_objects(&objects)?;
         self.assemble(lib_name, &dst.join(gnu_lib_name), &objects)?;
 
-        if self.get_target()?.contains("msvc") {
+        let target = self.get_target()?;
+        if target.contains("msvc") {
             let compiler = self.get_base_compiler()?;
             let atlmfc_lib = compiler
                 .env()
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3922,6 +3921,16 @@ impl Build {
...
     }
+    fn is_wasi_target(target: &str) -> bool {
+        const TARGETS: [&'static str; 5] = [
+            "wasm32-wasi",
+            "wasm32-wasip1",
+            "wasm32-wasip1-threads",
+            "wasm32-wasip2",
+            "wasm32-wasi-threads",
+        ];
+        return TARGETS.contains(&target);
```
madsmtm, line 3932: Why are we listing each target? Would `target.contains("wasi")` not be more forwards-compatible?

antaalt: Maybe, it was mostly to ensure if they had a target supporting exceptions or something like that to handle them separately, but maybe its overkill

src/lib.rs
```diff
@@ -1334,11 +1335,12 @@ impl Build {
                     .print_metadata(&format_args!("cargo:rustc-link-lib={}", stdlib.display()));
             }
             // Link c++ lib from WASI sysroot
-            if self.get_target()?.contains("wasi") {
+            if Build::is_wasi_target(target.as_ref()) {
                 let wasi_sysroot = self.wasi_sysroot()?;
                 self.cargo_output.print_metadata(&format_args!(
-                    "cargo:rustc-flags=-L {}/lib/wasm32-wasi -lstatic=c++ -lstatic=c++abi",
-                    Path::new(&wasi_sysroot).display()
+                    "cargo:rustc-flags=-L {}/lib/{} -lstatic=c++ -lstatic=c++abi",
```
NobodyXu, line 1341: Can you explain a bit about this change please?

In your previous PR it was hard coded, and now it changes to wasm32-wasi

NobodyXu (commented):
Thank you!

BTW, cc @antaalt  it'd be great if we can have some testing for wasi targets!

src/lib.rs
NobodyXu: Hmmm maybe I'm just thinking too much, it's probably because we expanded the targets supported.

antaalt: Yes, the lib folder structure in sysroot looks like this: image

antaalt:
> Thank you!
> 
> BTW, cc @antaalt it'd be great if we can have some testing for wasi targets!

Yes, I will look into it

PR #1114, opened 2024-06-30
Fix compilation for wasm: env WASI_SYSROOT should be optional

NobodyXu:
On ubuntu-latest github action runner, it used to work without the environment variable.

Fixed #1109 #1113

PR #1115, opened 2024-06-30
Add wasm32-unknown targets

### LINKED ISSUE #1113: v1.0.102 breaks wasm builds with trunk
I depend on cc through zstd-safe (zstd-safe -> zstd-sys -> cc). v1.0.102 causes wasm builds through trunk (ex: `trunk serve`) to display the following error when accessing the wasm webpage.

```
Uncaught TypeError: The specifier “env” was a bare specifier, but was not remapped to anything. Relative module specifiers must start with “./”, “../” or “/”.
```

I've tested manually reverting cc through `Cargo.lock` and a clean build. In my testing, cc v1.0.101 does not have this issue.

Possibly related to #1105 and building with `wasm32-unknown-unknown` target. 

Repo and build steps for reference (I've committed the cargo lock with v1.0.101 to mitigate this issue currently):
https://github.com/Kuuuube/hentaigana_input
https://github.com/Kuuuube/hentaigana_input/blob/main/.github/workflows/pages.yml

Kuuuube:
Mucked around with this and this seems to fix #1113 along with the changes made in #1114

Targeting the pr branch of #1114 so this can be pulled into that pr

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3942,12 +3942,14 @@ impl Build {
         }
     }
     fn is_wasi_target(target: &str) -> bool {
-        const TARGETS: [&'static str; 5] = [
+        const TARGETS: [&'static str; 7] = [
             "wasm32-wasi",
             "wasm32-wasip1",
             "wasm32-wasip1-threads",
             "wasm32-wasip2",
             "wasm32-wasi-threads",
+            "wasm32-unknown-wasi",
+            "wasm32-unknown-unknown",
         ];
         return TARGETS.contains(&target);
     }


NobodyXu (approved):
Thanks!

src/lib.rs
```diff
@@ -3942,12 +3942,14 @@ impl Build {
...
     fn is_wasi_target(target: &str) -> bool {
-        const TARGETS: [&'static str; 5] = [
+        const TARGETS: [&'static str; 7] = [
             "wasm32-wasi",
             "wasm32-wasip1",
             "wasm32-wasip1-threads",
             "wasm32-wasip2",
             "wasm32-wasi-threads",
+            "wasm32-unknown-wasi",
+            "wasm32-unknown-unknown",
```
NobodyXu, line 3952: With the additional of wasm32-unknown-unknown, I feel like I should rename this to `is_wasm_target` instead, but then this list here also doesn't cover all wasm targets.

PR #1118, opened 2024-07-01
Fixed link break about compile-time-requirements

mikoto2000:
It had already been removed from the README and could no longer be found through this link.

I considered linking to it below, but thought docs.rs would be a better place for documentation.

https://github.com/rust-lang/cc-rs/blob/243038b6e956615e04a98fc2ed44e155719ae20a/src/lib.rs#L129


NobodyXu (approved):
Thank you!

PR #1137, opened 2024-07-07
Use raw-dylib for windows-sys

NobodyXu:
To avoid having to declare `#[link(name = "...")]`

PR #1140, opened 2024-07-07
Bump msrv to 1.65

NobodyXu:
Required for #1137

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -201,13 +201,13 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install Rust
         run: |
-          rustup toolchain install 1.63.0 --no-self-update --profile minimal
+          rustup toolchain install 1.65.0 --no-self-update --profile minimal
           rustup toolchain install nightly --no-self-update --profile minimal
-          rustup default 1.63.0
+          rustup default 1.65.0
         shell: bash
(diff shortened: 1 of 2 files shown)


NobodyXu:
Unfortunately Debian stable is still on 1.63 https://packages.debian.org/search?keywords=rustc

ChrisDenton:
What is the actual MSRV policy currently? It's used to be roughly stable - 2. But then this crate languished for awhile so the practical MSRV slipped further through inaction. While I don't think this needs the MSRV to be as aggressive as it used to be, I'm also not sure there's much reason to have Debian LTS as a baseline, It appears they already ship a cc-rs package: https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=librust-cc

thomcc:
I'd be fine with something like stable-5. I know a lot of people have strong feelings about MSRV, but ancient MSRVs increase maintenance burden and we are low on those resources.

thomcc:
More concretely, 1.65 is from 2022 so it seems fine to update to that now.

thomcc (approved):

xtqqczze:
> More concretely, 1.65 is from 2022 so it seems fine to update to that now.

I think bumping the MSRV to 1.65 would be perfectly reasonable in 2026, see https://github.com/rust-lang/cc-rs/issues/1805.

PR #1143, opened 2024-07-08
Bump msrv to 1.67

NobodyXu:
For #1137

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -198,7 +198,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
     env:
-      MSRV: 1.66.0
+      MSRV: 1.67.0
     steps:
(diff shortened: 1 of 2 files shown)


NobodyXu:
We are discussing this on zulip https://rust-lang.zulipchat.com/#narrow/stream/351149-t-libs.2Fcrates/topic/cc-rs.20MSRV.20policy

NobodyXu:
> (breaking any crates which support building on Debian)

It seems that Debian already vendors the cc crate, they have an old version of cc

https://packages.debian.org/stable/rust/librust-cc+jobserver-dev

NobodyXu:
> For those who want x-lang LTO or to use a toolchain that is built on the same machines as the rest of their host crates pulled by cargo need to support 1.63.

Thanks, that's a scenario I didn't consider.

Is it possible to pin the cc version via Cargo.lock?

Issue #1147, opened 2024-07-12
`is_flag_supported` does not respect `emit_rerun_if_env_changed`


oconnor663:
In short, setting `emit_rerun_if_env_changed(false)` on  a `cc::Build` does not prevent the `is_flag_supported` method from printing `rerun-if-env-changed` directives.

Here's a quick hello world example:

**Cargo.toml**

```toml
[package]
name = "scratch"
version = "0.1.0"
edition = "2021"

[build-dependencies]
cc = "1.1.0"
```

**main.rs**

```rust
fn main() {}
```

**foo.c**

```c
void foo() {}
```

**build.rs**

```rust
fn main() {
    let mut build = cc::Build::new();
    build.file("foo.c");
    build.compile("foo");
}
```

If I run a verbose build and look at the `rerun-if-env-changed` directives, here's what I see:

```
$ touch build.rs && cargo build -vvv 2>&1 | grep rerun-if-env-changed
[scratch 0.1.0] cargo:rerun-if-env-changed=CC_x86_64-unknown-linux-gnu
[scratch 0.1.0] cargo:rerun-if-env-changed=CC_x86_64_unknown_linux_gnu
[scratch 0.1.0] cargo:rerun-if-env-changed=HOST_CC
[scratch 0.1.0] cargo:rerun-if-env-changed=CC
[scratch 0.1.0] cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
[scratch 0.1.0] cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
[scratch 0.1.0] cargo:rerun-if-env-changed=CFLAGS_x86_64-unknown-linux-gnu
[scratch 0.1.0] cargo:rerun-if-env-changed=CFLAGS_x86_64_unknown_linux_gnu
[scratch 0.1.0] cargo:rerun-if-env-changed=HOST_CFLAGS
[scratch 0.1.0] cargo:rerun-if-env-changed=CFLAGS
[scratch 0.1.0] cargo:rerun-if-env-changed=AR_x86_64-unknown-linux-gnu
[scratch 0.1.0] cargo:rerun-if-env-changed=AR_x86_64_unknown_linux_gnu
[scratch 0.1.0] cargo:rerun-if-env-changed=HOST_AR
[scratch 0.1.0] cargo:rerun-if-env-changed=AR
[...]

PR #1148, opened 2024-07-12
Fix is_flag_supported not respecting emit_rerun_if_env_changed (#1147)

oconnor663:
When I add a Cargo override to the example from #1147:

```toml
[patch.crates-io]
cc = { path = "/tmp/cc-rs" }
```

Then the extra `rerun-if-env-changed` directives go away (except for one that's maybe a separate small bug):

```
$ touch build.rs && cargo build -vvv 2>&1 | grep rerun-if-env-changed
[scratch 0.1.0] cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
[scratch 0.1.0] cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
```

I'm not sure how to add a unit test for this, since it doesn't look like `emit_rerun_if_env_changed` is currently tested.

It also seems like it's pretty easy to write more bugs like this, since anytime a feature is added to `Build` we need to remember to come to these lines and think about whether it should be inherited. I'm not sure how to make that better though. Thoughts?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -641,7 +641,8 @@ impl Build {
                 .host(&self.get_host()?)
                 .debug(false)
                 .cpp(self.cpp)
-                .cuda(self.cuda);
+                .cuda(self.cuda)
+                .emit_rerun_if_env_changed(self.emit_rerun_if_env_changed);
             cfg.try_get_compiler()?
         };
 


hintron:
> It also seems like it's pretty easy to write more bugs like this, since anytime a feature is added to `Build` we need to remember to come to these lines and think about whether it should be inherited. I'm not sure how to make that better though. Thoughts?

Is there some way to do a `self.clone()` instead of `Build::new()`, so all the fields are inherited and only the fields that deviate need to be explicitly changed?

NobodyXu:
> Is there some way to do a self.clone() instead of Build::new(), so all the fields are inherited and only the fields that deviate need to be explicitly changed?

That will end up with too many configurations inherited, better start afresh and assign what's needed.

NobodyXu (approved):
Thanks!

PR #1150, opened 2024-07-12
Add empty `jobserver` feature.

waywardmonkeys:
Since we removed implicit features, anyone that used this crate and specifically enabled `jobserver` would now be broken as this crate no longer had that feature (or a couple of others).

Add a placeholder `jobserver` feature that doesn't enable `parallel` (just as it didn't enable it before).

The other features removed weren't likely to be enabled by anyone.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,9 @@ libc = { version = "0.2.62", default-features = false, optional = true }
 
 [features]
 parallel = ["dep:libc", "dep:jobserver", "dep:once_cell"]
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
This should fix the problem reported in #1099 by @rafales.

The actual problem should also be reported to the @solana-labs crew.

waywardmonkeys:
I reported it and the issue was auto-closed as the whole thing has been forked. I looked in the fork and it no longer uses `cc` ... but since this was technically part of the public API, this change is still fine to go in, I guess.

NobodyXu (approved):
Thanks

PR #1154, opened 2024-07-13
Fix cyclic compilation: Use vendored once_cell

### LINKED ISSUE #1146: Cyclic package dependency with certain package features set
Create a simple cargo project with these dependencies in Cargo.toml

```
parking_lot = { version = "0.12.3", features = ["deadlock_detection"] }
once_cell = { version = "1.19.0", features = ["parking_lot"] }
cc = { version = "1.1.0", features = ["parallel"] }
```

Run `cargo check`, then cargo will emits error:
```
error: cyclic package dependency: package `backtrace v0.3.73` depends on itself. Cycle:
package `backtrace v0.3.73`
    ... which satisfies dependency `backtrace = "^0.3.60"` of package `parking_lot_core v0.9.10`
    ... which satisfies dependency `parking_lot_core = "^0.9.3"` of package `once_cell v1.19.0`
    ... which satisfies dependency `once_cell = "^1.19"` of package `cc v1.1.0`
    ... which satisfies dependency `cc = "^1.0.97"` of package `backtrace v0.3.73`
```

The dependencies formed a cycle after https://github.com/rust-lang/cc-rs/pull/1037 (which introduces the `once_cell` dependency)

NobodyXu:
Fix #1146

### DIFF
--- a/src/parallel/job_token.rs
+++ b/src/parallel/job_token.rs
@@ -2,7 +2,7 @@ use std::marker::PhantomData;
 
 use crate::Error;
 
-use super::once_cell::OnceCell;
+use super::once_lock::OnceLock;
 
 pub(crate) struct JobToken(PhantomData<()>);
 
@@ -37,7 +37,7 @@ impl JobTokenServer {
     ///    compilation.
     fn new() -> &'static Self {
         // TODO: Replace with a OnceLock once MSRV is 1.70
-        static JOBSERVER: OnceCell<JobTokenServer> = OnceCell::new();
+        static JOBSERVER: OnceLock<JobTokenServer> = OnceLock::new();
 
         JOBSERVER.get_or_init(|| {
             unsafe { inherited_jobserver::JobServer::from_env() }
@@ -71,7 +71,7 @@ impl ActiveJobTokenServer {
 }
 
 mod inherited_jobserver {
-    use super::{JobToken, OnceCell};
+    use super::{JobToken, OnceLock};
 
     use crate::{parallel::async_executor::YieldOnce, Error, ErrorKind};
 
@@ -137,7 +137,7 @@ mod inherited_jobserver {
         pub(super) fn enter_active(&self) -> ActiveJobServer<'_> {
             ActiveJobServer {
                 jobserver: self,
-                helper_thread: OnceCell::new(),
+                helper_thread: OnceLock::new(),
             }
         }
     }
@@ -163,7 +163,7 @@ mod inherited_jobserver {
 
     pub(crate) struct ActiveJobServer<'a> {
         jobserver: &'a JobServer,
-        helper_thread: OnceCell<Option<HelperThread>>,
+        helper_thread: OnceLock<Option<HelperThread>>,
     }
 
     impl<'a> ActiveJobServer<'a> {
--- a/src/parallel/job_token.rs
+++ b/src/parallel/job_token.rs
@@ -62,8 +62,8 @@ impl ActiveJobTokenServer {
         }
     }
 
-    pub(crate) async fn acquire(&self) -> Result<JobToken, Error> {
-        match &self {
+    pub(crate) async fn acquire(&mut self) -> Result<JobToken, Error> {
+        match self {
             Self::Inherited(jobserver) => jobserver.acquire().await,
             Self::InProcess(jobserver) => Ok(jobserver.acquire().await),
         }
@@ -71,7 +71,7 @@ impl ActiveJobTokenServer {
 }
 
 mod inherited_jobserver {
-    use super::{JobToken, OnceLock};
+    use super::JobToken;
 
     use crate::{parallel::async_executor::YieldOnce, Error, ErrorKind};
 
@@ -137,7 +137,7 @@ mod inherited_jobserver {
         pub(super) fn enter_active(&self) -> ActiveJobServer<'_> {
             ActiveJobServer {
                 jobserver: self,
-                helper_thread: OnceLock::new(),
+                helper_thread: None,
             }
         }
     }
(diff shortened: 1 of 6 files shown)


src/parallel/once_cell.rs
```diff
@@ -0,0 +1,47 @@
...
+        Self {
+            once: Once::new(),
+            value: UnsafeCell::new(MaybeUninit::uninit()),
+            _marker: PhantomData,
+        }
+    }
+
+    pub(crate) fn get_or_init(&self, f: impl FnOnce() -> T) -> &T {
+        self.once.call_once_force(|_| {
+            unsafe { &mut *self.value.get() }.write(f());
```
workingjubilee, line 26: In the unlikely event the Once is poisoned, I don't think we should try to continue, as cc-rs generally is used for run-once invocations. If we're vendoring everything we should just mimic the `get_or_try_init` API.

src/parallel/job_token.rs
```diff
@@ -183,10 +183,30 @@ mod inherited_jobserver {
...
+                        }
+
+                        let helper_thread = if let Some(thread) = helper_thread {
+                            thread
+                        } else {
+                            return Err(Error::new(
+                                ErrorKind::JobserverHelpThreadError,
+                                "Creation failed",
+                            ));
+                        };
```
workingjubilee, line 209: Why are we changing all this?

src/parallel/job_token.rs
```diff
@@ -163,7 +163,7 @@ mod inherited_jobserver {
 
     pub(crate) struct ActiveJobServer<'a> {
         jobserver: &'a JobServer,
-        helper_thread: OnceCell<HelperThread>,
+        helper_thread: OnceCell<Option<HelperThread>>,
```
workingjubilee, line 166: I think we're losing the plot here. `OnceLock` is already spiritually an `Option`, where the "Some or None?" discriminant is encoded using the `Once`'s internal "completed or not?" state. We shouldn't need all this.

workingjubilee (changes requested):
I would prefer to use the OnceLock terminology from std if we're going to use a Sync type, because I have already gotten used to the Cell vs. Lock distinction in my personal usage of these APIs.

src/parallel/job_token.rs
NobodyXu: Because the helper thread creation might fail, and since I can't implement `get_or_try_init` in a few lines, I put an `Option` there instead for the error.

src/parallel/job_token.rs
NobodyXu: Same as https://github.com/rust-lang/cc-rs/pull/1154#discussion_r1676984166

> Because the helper thread creation might fail, and since I can't implement get_or_try_init in a few lines, I put an Option there instead for the error.

workingjubilee: Then why not just put everything inline?

thomcc:
I'm requested as a reviewer, but I'll leave my review to @workingjubilee, feel free to land when she's satisfied.

src/parallel/job_token.rs
NobodyXu: > Then why not just put everything inline?

I'd like to make the `OnceLock` implementation as simple as possible (to make it easier to audit).

If I was to move this to `get_or_try_init`, then I'm not sure how to handle this case:
 - `get_or_try_init` error
 - `get_or_try_init` gets called again, but `Once` is already completed

I could put that `Option` inline, then I would have to return `Result<Option<T>, E>`, sounds a bit more complex than just `&T` in `get_or_init`

src/parallel/job_token.rs
workingjubilee: if "I can't implement `get_or_try_init` in a few lines" is a concern, we shouldn't have an abstraction at all: we should just put everything right here, no `OnceCell` or `OnceLock` generic type.

adding the type to avoid a cyclic compilation error but not being willing to maintain an equivalent API means that we really should just be using a dependency, whether it's the crate or the stdlib. having the type for a single usage invites adding more usages of it. but if we have more usages added to this crate, but aren't willing to either use a dependency, or maintain an equivalent API, then we are just courting incorrectness creeping in.

NobodyXu: The problem is we can't call `Once::pison` here.

The stdlib version can access that private method but we can't do it here, so the `OnceLock` impl here will have some paper cuts.

Alternatively, I can implement a `OnceLock::get_or_try_init` that panics, if you try to call it again after it fails.

workingjubilee: Right, I wouldn't expect us to actively poison it ourselves.

Sure, that sounds fine.

src/parallel/job_token.rs
workingjubilee: My view is simplicity is achieved by avoiding genericity over cases that we don't actually want to reason about or handle properly. We don't necessarily need OnceLock<HelperThread>, we need OnceHelperThread.

NobodyXu: Hmmm that sounds reasonable, I could add a `OnceHelperThread`

NobodyXu: Took another look, perhaps we could use an `Option` for helper thread, since it is created for each call to `Build::compile_objects`, so it can have exclusive access to the struct.

workingjubilee: Certainly!

NobodyXu: Updated, now it uses an `Option` instead of `OnceLock`, it's much simpler and has less overhead!

src/parallel/once_cell.rs
NobodyXu: Thanks, updated

src/parallel/job_token.rs
```diff
@@ -184,9 +184,12 @@ mod inherited_jobserver {
...
                         // Fallback to creating a help thread with blocking acquire
-                        let helper_thread = self
-                            .helper_thread
-                            .get_or_try_init(|| HelperThread::new(self.jobserver))?;
+                        let helper_thread = if let Some(thread) = self.helper_thread.as_ref() {
+                            thread
+                        } else {
+                            self.helper_thread
+                                .insert(HelperThread::new(self.jobserver)?)
+                        };
```
workingjubilee, line 192: Much nicer.

workingjubilee (approved):
Thanks, this looks much better.

PR #1173, opened 2024-08-06
Fix broken link in docs.rs

### LINKED ISSUE #1172: Issue/Docs: Dead links and domain hijacking?
The "Compile-time Requirements" section contains a few dead links. The Win-builds link now goes to someones blog "if im wrong my bad!" when I think it should lead to here https://github.com/alex85k/winbuilds/blob/master/README.md

Another issue is the "here" link for MSYS2, the section/header it refers to no longer exists `https://github.com/rust-lang/rust#building-on-windows`.
The following looks like a good replacement.
https://github.com/rust-lang/rust/blob/master/INSTALL.md#building-on-windows

NobodyXu:
Fixed #1172

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,8 +140,7 @@
 //!   set up by running the appropriate developer tools shell.
 //! * Windows platforms targeting MinGW (e.g. your target triple ends in `-gnu`)
 //!   require `cc` to be available in `PATH`. We recommend the
-//!   [MinGW-w64](https://www.mingw-w64.org/) distribution, which is using the
-//!   [Win-builds](https://github.com/alex85k/winbuilds/blob/master/README.md) installation system.
+//!   [MinGW-w64](https://www.mingw-w64.org/) distribution.
 //!   You may also acquire it via
 //!   [MSYS2](https://www.msys2.org/), as explained [here][msys2-help].  Make sure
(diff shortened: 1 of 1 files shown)


NobodyXu:
cc @ChrisDenton can you take a look at this please?

There're some broken links in our docs.rs, on how to build on windows.

src/lib.rs
```diff
@@ -141,14 +141,14 @@
 //! * Windows platforms targeting MinGW (e.g. your target triple ends in `-gnu`)
 //!   require `cc` to be available in `PATH`. We recommend the
 //!   [MinGW-w64](https://www.mingw-w64.org/) distribution, which is using the
-//!   [Win-builds](http://win-builds.org/) installation system.
+//!   [Win-builds](https://github.com/alex85k/winbuilds/blob/master/README.md) installation system.
```
ChrisDenton, line 144: Do we really need to link to win-builds? Why do users need to know what mingw-w64 is built on?

I'm asking because I think we should be very careful with third party links from official rust-lang documentation.

ChrisDenton: I think maybe we can just say we recommend `MinGW-w64` and leave it at that.

NobodyXu: Thanks, I've removed it

ChrisDenton (approved):

PR #1175, opened 2024-08-08
Add custom CC wrapper to the wrapper whitelist

hodrob84:
Some toolchains use custom (or at least not widely used) CC wrappers, and passes that through the CC env var before the compiler, like:

  CC = <custom_wrapper> <something>-gcc

Add support to CC_KNOWN_WRAPPER_CUSTOM env var, to be able to extend the 'known_wrappers' list with this <custom_wrapper>, and handle these special cases.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3120,7 +3120,11 @@ impl Build {
         //
         // It's true that everything here is a bit of a pain, but apparently if
         // you're not literally make or bash then you get a lot of bug reports.
-        let known_wrappers = ["ccache", "distcc", "sccache", "icecc", "cachepot"];
+        let mut known_wrappers = vec!["ccache", "distcc", "sccache", "icecc", "cachepot"];
+        let custom_wrapper = self.getenv("CC_KNOWN_WRAPPER_CUSTOM");
+        if custom_wrapper.is_some() {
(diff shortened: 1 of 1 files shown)


NobodyXu:
Thanks for the PR!

Can you please list some of the wrappers used?

If it is often used we can just add to our pre-defined list.

hodrob84:
This is 'ucfront-gcc' in my case. I just wanted to avoid to pollute it with a less commonly used one, and add something to make it easier to fit to one's purpose.

NobodyXu:
Thanks, is passing `CC=ucfront-gcc` not enough?

I.e., our compiler detection doesn't work?

If the compiler detection doesn't work for `ucfront-gcc`, then I think we should fix the detection logic instead.

AFAIK `ucfront-gcc` in debian usually means gcc for a specific target/libc or other config, so it should have the same kinds of behavior as `gcc`.

hodrob84:
Sorry, I wasn't clear. That's just the wrapper. The passed CC is:
`CC=ucfront-gcc <arch>-gcc`

NobodyXu:
In this case, I does think it makes sense.

Cc @thomcc does it LGTY?

thomcc:
I don't see a problem with this.

NobodyXu:
Thanks, I'll merge it in then

PR #1176, opened 2024-08-10
Remap Windows targets triples to their LLVM counterparts

mati865:
Fixes https://github.com/rust-lang/cc-rs/issues/1167

PR #1177, opened 2024-08-10
Replace vector of tuples with BTreeMap which already is sorted and free of duplicate

mati865:
I think this is more convenient to work with but feel free to close it if you don't agree.

Depends on: https://github.com/rust-lang/cc-rs/pull/1176


NobodyXu (commented):
Yeah using BTreeMap is indeed than Vec + sort_unstable

NobodyXu:
cc @mati865 would you like to rebase this?

NobodyXu (approved):
Thank you!

PR #1183, opened 2024-08-15
`CC_SHELL_ESCAPED_FLAGS`: improve docs

xbjfk:
Hello, after a little thinking I realise I wasn't too happy with the docs I wrote and think they are verbose - and decided to make it more concise and in plain English: 

 - `CC_SHELL_ESCAPED_FLAGS` - if set, `*FLAGS` will be parsed as if they were shell arguments (similar to `make` and `cmake`) rather than splitting them on each space. For example, with `CFLAGS='a "b c"'`, the compiler will be invoked with 2 arguments - `a` and `b c` - rather than 3: `a`, `"b` and `c"`.

Thanks :)


NobodyXu (approved):
Thanks, the new doc is simpler to read.

PR #1186, opened 2024-08-16
Fix detect family: should detect emscripten as clang, closes #1185

### LINKED ISSUE #1185: `emcc`/`em++` are detected as `ToolFamily::Msvc`
introduced by: https://github.com/rust-lang/cc-rs/pull/1015
https://github.com/rust-lang/cc-rs/blob/main/src/tool.rs#L173-L174 

For Msvc family, it will add `-nologo` as default flags., but emcc or em++ does not accept this.

meowtec:
Fixed #1185

### DIFF
--- a/src/detect_compiler_family.c
+++ b/src/detect_compiler_family.c
@@ -5,3 +5,8 @@
 #ifdef __GNUC__
 #pragma message "gcc"
 #endif
+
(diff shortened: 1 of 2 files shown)


NobodyXu (approved):
Thanks, LGTM

PR #1198, opened 2024-09-04
Treat VxWorks wr-cc as a Gnu compiler

### LINKED ISSUE #1197: VxWorks wr-cc is detected as MSVC compiler, cc adds flags "no-logo", "W4" and "--" which causes compilation issues.
More discussion at Zulip: https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Optimizing.20Compiler.20builtins.20causes.20build.20failure.20on.20VxWorks

Follow up of issue: https://github.com/rust-lang/compiler-builtins/issues/679

biabbas:
Fixes #1197 
Solution: 
1. Remove ci style arguments from command for wr-cc.
2. Remove --target=value from command as this caused wr-cc to look for wrong target config file.

### DIFF
--- a/src/detect_compiler_family.c
+++ b/src/detect_compiler_family.c
@@ -10,3 +10,6 @@
 #pragma message "emscripten"
 #endif
 
+#ifdef __VXWORKS__
+#pragma message "VxWorks"
+#endif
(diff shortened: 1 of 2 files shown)


NobodyXu (approved):
Thank you!

PR #1212, opened 2024-09-17
disable pic for targets that end in `-none`

### LINKED ISSUE #1211: Target strings that end in -none (such as aarch64-unknown-none) should not default to -fPIC
According to the [docs](https://docs.rs/cc/latest/cc/struct.Build.html#method.pic)

> This [pic] option defaults to false for windows-gnu and bare metal targets and to true for all other targets.

This means the bare metal target `aarch64-unknown-none` should default to false, but the current logic only looks for `-none-`, and not `-none`.

schultetwin1:
This fixes #1211

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1983,6 +1983,7 @@ impl Build {
                 if self.pic.unwrap_or(
                     !target.contains("windows")
                         && !target.contains("-none-")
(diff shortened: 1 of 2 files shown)


NobodyXu (approved):
Thank you!

PR #1215, opened 2024-09-27
Don't rerun if PATH changes

ChrisDenton:
Prevents spurious rebuilds on Windows (https://github.com/rust-lang/cc-rs/pull/1103/files#r1674907945) as suggested by @NobodyXu.

> It looks like this change causes crates to rebuild on Windows any time the `PATH` changes. Anecdotally, that seems to have a pretty high false positive rate, and since some `PATH` changes are associated with tools like rust-analyer that run constantly in the background, this can make it so that every build is a complete rebuild in some setups / with some dependencies. See for example https://github.com/BLAKE3-team/BLAKE3/issues/324. I might file this as an issue shortly, but I wanted to mention it real quick before I forget.

_Originally posted by oconnor663_

> Perhaps we can special-case PATH here?
>
> For example, Update the getenv to not emit anything for PATH

_Originally posted by NobodyXu_

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3718,7 +3718,7 @@ impl Build {
         if let Some(val) = self.env_cache.read().unwrap().get(v).cloned() {
             return val;
         }
-        if self.emit_rerun_if_env_changed && !provided_by_cargo(v) {
+        if self.emit_rerun_if_env_changed && !provided_by_cargo(v) && v != "PATH" {
             self.cargo_output
                 .print_metadata(&format_args!("cargo:rerun-if-env-changed={}", v));
         }


NobodyXu (approved):
(duplicate)

NobodyXu (approved):
I wonder if we should do something more?

I.e. when executing binaries, get the path of binary and ask cargo to watch it.

Unfortunately there isn't any API in stdlib to do so and it sounds a hit too hard for now, but I think we should think about this.

ChrisDenton:
Yeah, that might be nice but would definitely be a more complex feature as we'd need to account for different platforms.

NobodyXu:
> Yeah, that might be nice but would definitely be a more complex feature as we'd need to account for different platforms.

And it'd still be half broken: If user add another compiler in /usr/local/bin/cc while we ask cargo to detect change at /usr/bin/cc, it'd still miss that.

I think if we want to do this **correctly**, by making sure we always recompiles if the compiler gets swapped, we will have to do it the same way sccache detects change.

We will have to open the compiler binary, hashed the binary itself, and all the dynamic libraries it linked in (libclang.so, libllvm.so).

We'd need a new `cargo:rerun-if` primitives for such thing, because cargo decides if the build-script gets run.

NobodyXu:
And even that may not work, if the compiler is wrapped by a bash script.

Honestly I feel like this is something incredibly hard to get right, we probably want to document this in our crate-level doc.

ChrisDenton:
That's probably a better approach. Honestly, users know their build better than we can possibly infer and are therefore better able to do this. Or at least there's no way we can reasonably cover all cases well.

src/lib.rs
```diff
@@ -3718,7 +3718,7 @@ impl Build {
         if let Some(val) = self.env_cache.read().unwrap().get(v).cloned() {
             return val;
         }
-        if self.emit_rerun_if_env_changed && !provided_by_cargo(v) {
+        if self.emit_rerun_if_env_changed && !provided_by_cargo(v) && v != "PATH" {
```
RalfJung, line 3721: Someone looking at this code in the future will be very confused about why PATH got special-cased here. Would have been good to add a comment. :)

NobodyXu: Yeah I could add one there

NobodyXu: Opened #1247

PR #1221, opened 2024-09-30
Fix wasm32-wasip1-threads:  shared-memory disallowed due to not compiled with 'atomics' or 'bulk-memory' features

### LINKED ISSUE #1217: Support for wasm32-wasip1-threads
The -pthread is required as a flag.
For -D_WASI_EMULATED_MMAN, it is not necessary yet because the Rust compiler does not support it!

NobodyXu:
Fixed #1217


NobodyXu:
cc @oligamiq opened this PR which should fix your compilation error

Issue #1219, opened 2024-09-29
Add `Build::from_cargo` or similar to make it very explicit when in a build script


madsmtm:
`cc`'s code is littered with brittle checks centered on `rustc` target triple. A lot of this is unnecessary though, since Cargo sets `cfg` values for the target as environment variables, see [the docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) for more examples. `cc` can't actually use those in most cases though, since it doesn't know whether it's being run inside a Cargo build script, or outside of it.

A prominent example is that `CARGO_CFG_TARGET_ABI` would be very useful for detecting Apple's simulator targets (doing it based on the target name is [misery](https://github.com/rust-lang/cc-rs/blob/5f5e1c8c18b53aa1524f720ce21b5a7986a0bf6b/src/lib.rs#L2740-L2741), and I think we're currently doing it wrong some places).

So I got to thinking, would it make sense to "split" `cc`'s entry point into two, and deprecated `cc::Build::new`? Something like:
```rust
// build.rs
fn main() {
    // Emits all the rerun-if-changed flags, `cargo_metadata` defaults to true, etc.
    // Reads from `CARGO_CFG_TARGET_*` to initialize target data.
    cc::Build::from_cargo()
        .file("foo.c")
        .file("bar.c")
        .compile("foo");
}

// main.rs
fn main() {
    // Does not print needless Cargo metadata.
    // Using cc from outside build.rs, so has to specify target.
    // The target is parsed eagerly into some internal structure that represents it,
    // maybe using `target-lexicon`, and behind a feature flag? In any case, we
    // can say that this use-case is less supported, and the support burden will
    // perhaps be lessened?
    cc::Build::from_target("aarch64-unknown-linux-gnu")
        .file("foo.c")
        .file("bar.c")
        .compile("foo");
}
```

madsmtm:
Actually, I think my biggest issue is the ad-hoc parsing of the target triple; ideally, that would be done first thing inside `Build::try_compile` as a completely separate step, and the rest of the code would never touch the unparsed target triple.

Would you accept a PR changing the target parsing to be up-front? And would you accept a private dependency on [`target-lexicon`](https://docs.rs/target-lexicon/) (also used by the Cranelift and GCC rustc backends)? Or would it need to be vendored (possibly with a script) ~to maintain the illusion of low compile times just because you don't have any dependencies~?

[...]

PR #1224, opened 2024-09-30
Reduce the need for the host target triple

madsmtm:
It seems unnecessary to me that the user needs to specify the host target triple when using `cc` outside a build script, `cc` _ought_ to be able to figure it out from context. This is harder than it seems though, for example `armv6-unknown-freebsd` cross compiling to `armv7-unknown-freebsd` (however unlikely) would be impossible to determine using just `target_*` cfgs.

For now, though, I've removed the internal `get_host` helper, and replaced it with `get_is_cross_compile`, and then used `cfg!(...)` in the places where we were reading from the host variable before.

Tangentially related to https://github.com/rust-lang/cc-rs/issues/1219.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -283,6 +283,9 @@ pub struct Build {
     ccbin: bool,
     std: Option<Arc<str>>,
     target: Option<Arc<str>>,
+    /// The host compiler.
+    ///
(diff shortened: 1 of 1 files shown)


NobodyXu (approved):
Thanks LGTM

Issue #1126, opened 2024-07-04
Wasm32-unknown-unknown does not need SYS_ROOT override


NobodyXu:
#1114 conflates wasm with wasi targets. While wasi targets should change the sysroot to the wasi-sdk-provided one, the non-wasi target `wasm32-unknown-unknown` should probably keep the default sysroot.

madsmtm:
Things also changed a bit in https://github.com/rust-lang/cc-rs/pull/1225, we should be using `target.os == "wasi"` or `target.arch == "wasm32" || target.arch == "wasm64"` depending on what the desired condition is.

madsmtm:
Reviewing all the places where `wasm` or `wasi` is mentioned:
1. Uses WASI_SYSROOT, seems like it is correct to use `target.os`:
    https://github.com/rust-lang/cc-rs/blob/2050013e69bc07e275a65583dd26e8905dfdff73/src/lib.rs#L1406-L1415
2. Might need `target.arch`, 
    https://github.com/rust-lang/cc-rs/blob/2050013e69bc07e275a65583dd26e8905dfdff73/src/lib.rs#L1996-L2011
3. Might need `target.arch` for `-fno-exceptions` and `-pthread`?
    https://github.com/rust-lang/cc-rs/blob/2050013e69bc07e275a65583dd26e8905dfdff73/src/lib.rs#L2012-L2027
4. Preferring Clang as the compiler might need to be done for `target.arch` instead?
    https://github.com/rust-lang/cc-rs/blob/2050013e69bc07e275a65583dd26e8905dfdff73/src/lib.rs#L2707-L2712
5. The standard C++ library might need `target.arch`? Though might also be correct to use `target.os`
    https://github.com/rust-lang/cc-rs/blob/2050013e69bc07e275a65583dd26e8905dfdff73/src/lib.rs#L2936-L2971
6. Probably need to also check `wasm64` when finding archiver?
    https://github.com/rust-lang/cc-rs/blob/2050013e69bc07e275a65583dd26e8905dfdff73/src/lib.rs#L3097-L3114

But ultimately, I am unsure of all of these, which is why I haven't pursued it myself.

@daxpedda, do you know which of these places it would be correct to use `target.os == "wasi"`, and which ones it'd be correct to use `matches!(target.arch, "wasm32" | "wasm64")`?

daxpedda:
[...]

PR #1225, opened 2024-09-30
Use Cargo's target information when possible

madsmtm:
rustc's target triples generally only have a vague resemblance to each other and to the information needed by `cc`. Let's instead prefer `CARGO_CFG_*` variables when available, since these contain the information directly from the compiler itself.

In the cases where it isn't available (i.e. when running outside of a build script), we fall back to parsing the target triple, but instead of doing it in an ad-hoc fashion with string manipulation, we do it in a more structured fashion up front.

Fixes https://github.com/rust-lang/cc-rs/issues/1219 (at least my main gripe with the current implementation).

Builds upon https://github.com/rust-lang/cc-rs/pull/1224.


src/target.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        Ok(Target {
+            full_arch: full_arch.into(),
+            arch: arch.into(),
+            vendor: vendor.into(),
+            os: os.into(),
+            env: env.into(),
+            abi: abi.into(),
+        })
+    }
```
madsmtm, line 310: So, I spent some time on actually implementing the target parsing, and as you can see with the test below, it _works_ now for all targets that `rustc` supports, but I wonder: Is this even the desired approach?

Another solution would be to run `rustc --print all-target-specs-json` at build time, and generate a static lookup table from target name to the target information. Such a table would need updating every time `rustc` adds a new target (when running outside of a build script), but that's honestly usually the case in `cc` anyhow, and at least this would be something automatic instead (and bug fixes in `rustc` to the target info would automatically get transfered to `cc`).

@NobodyXu wdyt?

src/target.rs
```diff
@@ -0,0 +1,415 @@
...
+            "thumbv7-unknown-linux-gnueabihf",
+            "thumbv7-unknown-linux-musleabihf",
+            "x86_64-rumprun-netbsd",
+        ];
+
+        for target in targets {
+            // Check that it parses
+            let _ = Target::from_str(target).unwrap();
+        }
+    }
```
madsmtm, line 414: These were a few targets I found lying around in the source code. I'm not sure how good `cc`'s support is for targets that require a custom target json file (i.e. most embedded targets)? We should probably extend this table to ensure we also support those.

dev-tools/gen-target-info/src/main.rs
```diff
@@ -1,41 +1,11 @@
...
-            let rust_target_parts = target.splitn(4, '-').collect::<Vec<_>>();
-            let os = *rust_target_parts.get(2)?;
-            (os.contains("windows") && target != &*target_spec.llvm_target)
-                .then_some((&**target, &*target_spec.llvm_target))
-        })
-        .collect::<BTreeMap<_, _>>();
-    write_target_tuple_mapping(f, "WINDOWS_TRIPLE_MAPPING", &windows_target_mapping);
-}
-
 fn main() {
```
madsmtm, line 8: The generator is useless now, though I'll wait with removing it until we decide if we should be generating stuff in https://github.com/rust-lang/cc-rs/pull/1225/files#r1781935441.

src/target.rs
NobodyXu: Doing it at build time is a bit hard:
 - IIRC `rustc --print all-target-specs-json` is only available on nightly
 - cc has msrv 1.63
 - cc can be used outside of build-script, and it's not guaranteed to have `rustc`

> Such a table would need updating every time rustc adds a new target (when running outside of a build script)

Yes we have a CI that run the generation and create PR if there's any update.

NobodyXu: Doing it at build time is a bit hard:
 - IIRC `rustc --print all-target-specs-json` is only available on nightly
 - cc has msrv 1.63
 - cc can be used outside of build-script, and it's not guaranteed to have `rustc`

> Such a table would need updating every time rustc adds a new target (when running outside of a build script)

Yes we have a CI that run the generation and create PR if there's any update.

src/lib.rs
```diff
@@ -1966,14 +1975,15 @@ impl Build {
...
                 }
 
-                if cmd.is_like_clang() && target.contains("windows") {
+                if cmd.is_like_clang() && target.os == "windows" {
                     // Disambiguate mingw and msvc on Windows. Problem is that
                     // depending on the origin clang can default to a mismatchig
                     // run-time.
-                    cmd.push_cc_arg(format!("--target={}", target).into());
+                    // FIXME: Convert rustc target to Clang target.
+                    cmd.push_cc_arg(format!("--target={raw_target}").into());
```
NobodyXu, line 1983: this won't work on cc's msrv 1.63

src/target.rs
```diff
@@ -0,0 +1,418 @@
...
+//! See the `target-lexicon` crate for a more principled approach to this.
+
+use std::{env, str::FromStr};
+
+use crate::{Error, ErrorKind};
+
+/// The parts of `rustc`'s target triple.
+///
+/// See <https://doc.rust-lang.org/cargo/appendix/glossary.html#target>.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
NobodyXu, line 12: Do we really need all that derive?

src/target.rs
```diff
@@ -0,0 +1,418 @@
...
+        let arch = getenv("CARGO_CFG_TARGET_ARCH")?;
+        let vendor = getenv("CARGO_CFG_TARGET_VENDOR")?;
+        let os = getenv("CARGO_CFG_TARGET_OS")?;
+        let env = getenv("CARGO_CFG_TARGET_ENV")?;
+        // `target_abi` was stabilized in Rust 1.78, so may not always be available.
+        let abi = if let Ok(abi) = getenv("CARGO_CFG_TARGET_ABI") {
+            abi
+        } else {
+            Self::from_str(&target)?.abi
+        };
```
NobodyXu, line 70: Can we fallback to `TARGET` if these `CARGO_*` env is not set please?

There might be someone depending on that behavior

src/target.rs
```diff
@@ -0,0 +1,418 @@
...
+    fn from_str(target: &str) -> Result<Self, Error> {
+        let mut components = target.split('-');
+
+        // Insist that the triple start with a valid architecture.
+        let full_arch = components.next().ok_or(Error::new(
+            ErrorKind::InvalidTarget,
+            format!("target `{target}` was empty"),
+        ))?;
+
+        let arch = match full_arch {
```
NobodyXu, line 95: Is it possible for any hard coded literals in this file to be retrieved from rustc?

src/lib.rs
```diff
@@ -4249,31 +4182,28 @@ fn autodetect_android_compiler(target: &str, host: &str, gnu: &str, clang: &str)
...
-        Some("arm64")
-    } else if target.contains("i686") {
-        Some("i386")
-    } else if target.contains("powerpc") {
-        Some("ppc")
-    } else if target.contains("powerpc64") {
-        Some("ppc64")
-    } else {
-        None
+fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str {
```
NobodyXu, line 4193: I think this should be in `target.rs`

I also wonder if rustc provides these info?
It looks like something that could be available in there.

NobodyXu (changes requested):

src/lib.rs
madsmtm: Yes it will? This has been possible since 1.58: https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings

NobodyXu: Oh sorry my memory is wrong 😅

NobodyXu: Oh sorry my memory is wrong 😅

src/target.rs
madsmtm: I'd _argue_ that the documentation for `fn target` says that the `TARGET` parsing only applies to build scripts, reproduced below (emphasis mine).

> Configures the target this configuration will be compiling for.
>
> This option is automatically scraped from the `TARGET` environment variable **by build scripts**, so it's not required to call this function.

But I digress, you're right that `cc` is widely used and someone probably relies on it somewhere.

src/lib.rs
madsmtm: If we do generation, then we might be able to extract it from `llvm_target`.

src/target.rs
madsmtm: Okay, I'll try to implement that.

It does carry some problems, I've identified at least the following special cases:
1. A target triple is unsupported by `rustc` (e.g. because it is a custom / bare metal target, or because it is in the process of being upstreamed).
2. An outdated `cc` version that contains outdated data / the user is using an outdated version of `rustc` compared to `cc`'s data.

---

To handle the first case, we can read `CARGO_CFG_*` when in a build script, but we'll have to error when outside.

To solve this, it might make sense for `cc` to "vendor" target information for common bare metal targets, for example [the ones for the Arduino](https://github.com/Rahix/avr-hal/tree/main/avr-specs)? Though we could also declare it as not supported, or say that the user should use the `rustc` target, in the case of Arduino, as `.target("avr-unknown-gnu-atmega328")`?

(Actually, why _isn't_ this target information just upstreamed into `rustc`? That'd both be easier for us, `rustc`, the maintainers of `avr-hal`, and for the user?)

---

The second case is trickier, but I _think_ we want to use `CARGO_CFG_*` when available, and fall back to the static data in `cc` when not? That is, we prefer `rustc`'s view of the world, even though it may be outdated and incorrect.

As an example, `armv7a-kmc-solid_asp3-eabihf` has `eabihf` in the name, but does not specify `target_abi = "eabihf"`. Let's assume that it should (I'll open a PR later to `rustc` to figure out if that's true), and that it gets fixed in Rust 1.84.

Next, let's assume two versions of `cc` had been generated, `v1.1.25` with `rustc +1.81`, and `v1.1.26` with `rustc +1.84`. The first version would assume `target_abi = ""`, and the second would assume `target_abi = "eabihf"`.

This would make `cc v1.1.25` have the wrong behaviour when using `rustc +1.84`.

If we instead read the value from `CARGO_CFG_TARGET_ABI`, then we'd think the same as `rustc`, i.e. that it is `target_abi = ""` when using `rustc +1.81`, and `target_abi = "eabihf"` when using `rustc +1.84`.

This is obviously still incorrect, e.g. the user could be using `cc v1.1.26` and still get the wrong `target_abi = ""`, but at least the user doesn't get differing behaviour between things generated by `rustc`, and things generated by the C compiler that `cc` calls?

Honestly not sure here.

madsmtm:
I've now gone through my own FIXMEs, and opened PRs against `rustc` to fix the discrepancies I saw when trying to parse the target information.
- https://github.com/rust-lang/rust/issues/131172
- https://github.com/rust-lang/rust/issues/131165
- The `win7` I didn't do, it's clearly a hack that they do intentionally.
- https://github.com/rust-lang/rust/pull/131166
- https://github.com/rust-lang/rust/pull/131168
- https://github.com/rust-lang/rust/pull/131169
- https://github.com/rust-lang/rust/pull/131170
- https://github.com/rust-lang/rust/pull/131171
- armv[6|7]-unknown-freebsd seem to set `gnu` explicitly, so imma assume it's intentional.
- https://github.com/rust-lang/rust/pull/131173
- https://github.com/rust-lang/rust/pull/131174

src/lib.rs
NobodyXu: That'd be nice, I think cc has too many hardcoded information in it, and it gets hard to maintain or read very quickly.

src/target.rs
NobodyXu: My thoughts is that:
 - respect CARGO_CFG env if set, since it should be correct and up to date. If it is not, user can update their cargo (if people want cc to ignore it, we could expose an env/fn to ignore it, but I won't do it unless any user actually wants it).
 - otherwise fallback to data extracted from rustc at build time. If it is too old, then user should either upgrade cc or rustc/cargo
 - for non-official targets, using CARGO_CFG is preferred, and we can add popular non-official targets to gen-target-info (it's easier to update than directly hardcoding it in cc).

madsmtm:
I've implemented the approach discussed in https://github.com/rust-lang/cc-rs/pull/1225#discussion_r1781935441 now.

src/lib.rs
madsmtm: I think it makes sense to punt on this for later, as we're not guaranteed to be able to know `llvm_target` (e.g. custom target JSON and/or unknown targets), and so we'll have to implement some sort of conversion anyhow.

dev-tools/gen-target-info/src/main.rs
```diff
@@ -1,52 +1,73 @@
...
-        .collect::<BTreeMap<_, _>>();
-    write_target_tuple_mapping(f, "RISCV_ARCH_MAPPING", &riscv_target_mapping);
-}
+fn generate_target_mapping(f: &mut File, target_specs: &RustcTargetSpecs) -> std::io::Result<()> {
+    writeln!(f, "use super::Target;")?;
+    writeln!(f)?;
+    writeln!(
+        f,
+        "pub(crate) fn get(target_triple: &str) -> Option<Target> {{"
+    )?;
```
NobodyXu, line 19: I think write_str is more readable, since you can use `{` instead of `{{`, which is really hard to read

```suggestion
    f.write_str("pub(crate) fn get(target_triple: &str) -> Option<Target> {"
    )?;
```

dev-tools/gen-target-info/src/main.rs
```diff
@@ -1,52 +1,73 @@
...
+        let abi = spec.abi.as_deref().unwrap_or("");
+
+        writeln!(f, "        {triple:?} => Target {{")?;
+        writeln!(f, "            full_arch: {full_arch:?}.into(),")?;
+        writeln!(f, "            arch: {arch:?}.into(),")?;
+        writeln!(f, "            vendor: {vendor:?}.into(),")?;
+        writeln!(f, "            os: {os:?}.into(),")?;
+        writeln!(f, "            env: {env:?}.into(),")?;
+        writeln!(f, "            abi: {abi:?}.into(),")?;
+        writeln!(f, "        }},")?;
```
NobodyXu, line 37: Would it possible to have an array of `[(&'static str, Target); N]`?

I think writing a large match here is less readable than having a sorted array.

NobodyXu: I also suspect that a match would generate more code than a linear/binary search in sorted array, thus taking longer to compile.

Matching would basically be lots of `if`s, the compiler might also optimize the `if` into a jump table, but that would still mean a lot of `Target` creation at runtime with OOM panic handling blocks.

src/target.rs
```diff
@@ -0,0 +1,180 @@
...
+/// The parts of `rustc`'s target triple.
+///
+/// See <https://doc.rust-lang.org/cargo/appendix/glossary.html#target>.
+#[derive(Debug, PartialEq)]
+pub(crate) struct Target {
+    /// The full architecture, including the subarchitecture.
+    ///
+    /// This differs from `cfg!(target_arch)`, which only specifies the
+    /// overall architecture, which is too coarse for certain cases.
+    pub full_arch: Cow<'static, str>,
```
NobodyXu, line 20: I think we can just use `&'a str` here, it'd be simpler and a smaller struct.

For `from_cargo_environment_variables`, we can cache the environment variables in `cc::Build` so that we can return a `Target<'_>`, we already do env caching so this is not a problem.

We can use the [homebrew `OnceLock` impl](https://github.com/rust-lang/cc-rs/blob/main/src/parallel/once_lock.rs) within our codebase, so that we could avoid lifetime problems with `MutexGuard`/`RwLockGuard`.

NobodyXu: P.S. I'm not worried/obsessed about the allocation as it doesn't matter much, I just prefer a simpler structure.

Caching the env var in `Build` would also be consistent with how other env var currently works:

Once we read the env var from system into `Build`, it's never changed in that object and any objects created by `Build::clone`.

NobodyXu (changes requested):

dev-tools/gen-target-info/src/main.rs
madsmtm: Huh, I seemed to remember `Cow` not really being usable in `const`s because they have drop glue, but apparently I remembered wrong, and I think it's doable.

src/target.rs
madsmtm: Hmm, I don't see how `OnceLock` would help with avoiding a guard? Do you mean to store cached environment variables in a global instead?

NobodyXu: You can put `OnceLock` inside `cc::Build` struct.

For example:

```rust
struct Build {
    target_triple: OnceLock<String>,
    // ...
}
```

store individual fields of `Target` directly in it, and return `Target<'_>` with a lifetime

dev-tools/gen-target-info/src/main.rs
madsmtm: I have decided to leave it in for `Target {{`, since I think it makes it clearer to read that the formatting lines up.

src/lib.rs
madsmtm: Have opened https://github.com/rust-lang/cc-rs/issues/1249 to track it there

src/target.rs
madsmtm: Hmm, I'm still not _entirely_ sure I get what you want me to do, and I feel like it'd make the code more convoluted?

Would you mind doing it?

src/lib.rs
```diff
@@ -4045,18 +3984,6 @@ impl Build {
...
-            "wasm32-wasi",
-            "wasm32-wasip1",
-            "wasm32-wasip1-threads",
-            "wasm32-wasip2",
-            "wasm32-wasi-threads",
-            "wasm32-unknown-wasi",
-            "wasm32-unknown-unknown",
-        ];
-        TARGETS.contains(&target)
-    }
```
madsmtm, line 4059: Cross-referencing https://github.com/rust-lang/cc-rs/issues/1126, I've replaced usage of `is_wasi_target` with `target.os == "wasi"` in this PR, but that of course won't catch `wasm32-unknown-unknown` - so there might now be places where instead of `target.os == "wasi"`, the correct check would be `target.arch == "wasm32" || target.arch == "wasm64"`

src/target.rs
NobodyXu: Sure leave it to me

NobodyXu: Sure leave it to me

NobodyXu: I think we could merge it in as-is and then I can open a separate PR to optimize it.

NobodyXu (approved):
Thank you for all the hard work!

src/lib.rs
NobodyXu: Well...I think we need to do this, this change might break some use cases.

PR #1226, opened 2024-10-02
Allow disabling `cc`'s ability to run commands via the `CC_FORCE_DISABLE` environment variable

thomcc:
This came up in RustConf as a desire from users/maintainers of 3rd-party build systems, such as Buck2, Bazel, and their ilk.

The idea is that they should be in full control over building all native dependencies, and currently it's (apparently) very annoying and time-consuming to ensure that no `build.rs` is compiling/assembling/linking/etc native libraries on it's own. By setting this flag, they can get an error on any crate that attempts this.

In terms of the implementation, this was more invasive than I'd like, and I'm open to suggestions about alternative ways of doing this. I failed to locate a single choke-point where returning a `Result::Err` would sufficiently neuter our functionality.

I think the approach I've taken is awkward, and would be borderline-unmaintainable were it not for `clippy::disallowed_methods`. That said, we do have access to that clippy lint, which I think pushes this over the edge.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,5 +1,7 @@
 disallowed-methods = [
     { path = "std::env::var_os", reason = "Please use Build::getenv" },
     { path = "std::env::var", reason = "Please use Build::getenv" },
+    { path = "std::process::Command::new", reason = "Please use crate::command_new()" },
+    { path = "cc::tool::Tool::to_command", reason = "Bypasses `is_disabled()` check. Use try_to_command() instead." },
 ]
 doc-valid-idents = ["AppleClang", "OpenBSD", ".."]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,6 +89,13 @@
 //!   For example, with `CFLAGS='a "b c"'`, the compiler will be invoked with 2 arguments -
 //!   `a` and `b c` - rather than 3: `a`, `"b` and `c"`.
 //! * `CXX...` - see [C++ Support](#c-support).
+//! * `CC_FORCE_DISABLE` - If set, `cc` will never run any [`Command`]s, and methods that
+//!   would return an [`Error`]. This is intended for use by third-party build systems
+//!   which want to be absolutely sure that they are in control of building all
+//!   dependencies. Note that operations that return [`Tool`]s such as
+//!   [`Build::get_compiler`] may produce less accurate results as in some cases `cc` runs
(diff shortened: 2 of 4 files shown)


clippy.toml
```diff
@@ -1,5 +1,7 @@
 disallowed-methods = [
     { path = "std::env::var_os", reason = "Please use Build::getenv" },
     { path = "std::env::var", reason = "Please use Build::getenv" },
+    { path = "std::process::Command::new", reason = "Please use crate::command_new()" },
+    { path = "cc::tool::Tool::to_command", reason = "Bypasses `is_disabled()` check. Use try_to_command() instead." },
```
thomcc, line 5: This is just for internals, obviously users can continue to use `Tool::to_command` as much as they want.

src/tool.rs
```diff
@@ -213,15 +214,15 @@ impl Tool {
                 Some(fname) if fname.contains("clang") => match clang_driver {
                     Some("cl") => ToolFamily::Msvc { clang_cl: true },
                     _ => ToolFamily::Clang {
-                        zig_cc: is_zig_cc(&path, cargo_output),
+                        zig_cc: is_zig_cc(&path, cargo_output).unwrap_or(false),
```
thomcc, line 217: `unwrap_or(false)` here is a bit janky. I don't think it makes much of a difference what we report when `cc` is disabled, though.

src/lib.rs
```diff
@@ -4220,7 +4244,60 @@ fn android_clang_compiler_uses_target_arg_internally(clang_path: &Path) -> bool
     false
 }
 
-fn autodetect_android_compiler(target: &str, gnu: &str, clang: &str) -> String {
+/// Returns true if `cc` has been disabled by `CC_FORCE_DISABLE`.
+pub(crate) fn is_disabled() -> bool {
```
thomcc, line 4248: It's unfortunate that this is toplevel and not on `Build`, but changing everywhere that produces `Command`s to take a `&Build` was far too invasive a change. This already is pushing it.

NobodyXu (changes requested):
How about adding checks to `Build::compile_objects`?

It is the function responsible for doing the actual compilation, all other functions compile/try_compile/compile_intermediates/try_compile_intermediates all delegate to it.

ChrisDenton:
Tbh this does seem a more general problem with the architecture of Cargo's build scripts rather than cc per se. But if using cc as a proxy helps as a mitigation then I'm not against it.

PR #1228, opened 2024-10-04
Remove incorrect "lib" prefixes in CXXSTDLIB doc comments

graydon:
This fixes two incorrect doc comments that currently describe possible values of the `CXXSTDLIB` in terms of strings that include a `lib` prefix. The correct values should all _exclude_ that prefix (as they do in the actual code being described).


NobodyXu (approved):
Thank you!

Issue #1205, opened 2024-09-07
Support `arm64e-apple-darwin`


madsmtm:
See [the platform support docs](https://doc.rust-lang.org/rustc/platform-support/arm64e-apple-darwin.html) and https://github.com/rust-lang/rust/issues/73628.

We currently error with:
```
error occurred: Unknown architecture for macOS target
```

This is affecting bootstrap, i.e. you cannot do `./x build --target=arm64e-apple-darwin`.

(Note that the target [may be renamed in the future](https://rust-lang.zulipchat.com/#narrow/stream/233931-t-compiler.2Fmajor-changes/topic/Promote.20tier.203.20arm64e-apple-darwin.20target.E2.80.A6.20compiler-team.23717/near/420641916), but we should support it with its current name).

CC target maintainer @arttet.

Issue #542, opened 2020-08-13
 The target passed to clang is wrong for aarch64-apple-darwin


glandium:
cc-rs passes the rust target unchanged, but unfortunately, in the case of aarch64-apple-darwin, the clang target is different, and clang doesn't recognize aarch64-apple-darwin at all. The right target to use is arm64-apple-darwin.

SebDanielsson:
Anyone working on this?

lmtr0:
Hello there, any Status on this?

dot-asm:
I for one have no problem compiling for M1. I don't see any unrecognized flags passed to cc. I do observe additional --target flag if I attempt to compile with `CC` environment variable set to `clang`, but it's correct arm64-apple-darwin. In other words this looks resolved...

lmtr0:
No, it isn't `arm64-apple-darwin` isn't a valid target
```
[root@52bd9e36c736 be]# cargo build --target arm64-apple-darwin
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `rustc - --crate-name ___ --print=file-names --target arm64-apple-darwin --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit status: 1)
  --- stderr
  error: Error loading target specification: Could not find specification for target "arm64-apple-darwin". Run `rustc --print target-list` for a list of built-in targets

```

dot-asm:
> No, it isn't, `arm64-apple-darwin` isn't a valid target

What I meant was that if I run `env CC=clang cargo build` on a M1 machine **or** `env CC=clang cargo build --target=aarch64-apple-darwin` on an x86 machine, cc-rs passes `--target=arm64-apple-darwin` to clang. This is with cc-rs 1.0.73.

lmtr0:
Ah, got it, sorry I misunderstood. Btw I used o64-clang (for amd64 builds) instead of o64h-clang (for aarch64 builds) and somehow, it worked.
```
[root@52bd9e36c736 be]# CC=o64-clang RUSTFLAGS="-C linker=o64-clang -C link-args=-D__ARM_ARCH__" cargo build --target aarch64-apple-darwin
   Compiling untrusted v0.7.1
   Compiling ring v0.16.20
   Compiling be v0.1.0 (/be)
    Finished dev [unoptimized + debuginfo] target(s) in 3.27s
[...]

Issue #1249, opened 2024-10-31
Use more information from `rustc`'s target spec


madsmtm:
There's a lot more useful information exposed in `rustc`'s target spec that `cc` could use after https://github.com/rust-lang/cc-rs/pull/1225. Opening this issue to track the progress on using those.

Specifically:
- [ ] `llvm-target` and `llvm-abiname`
- [ ] `target-endian`
- [ ] `linker`
- [ ] `pre-link-objects`/`post-link-objects`/`pre-link-objects-fallback`/`crt-static-allows-dylibs`/etc. for better `-Clink-self-contained`?
- [ ] `link-env`/`link-env-remove`
- [ ] `cpu`?
- [ ] `dynamic-linking` (for diagnostics?)
- [ ] `frame-pointer`
- [ ] `default-dwarf-version`
- [ ] `position-independent-executables`/`static-position-independent-executables`
- [ ] `panic-strategy` (for `-fno-exceptions`)
- [ ] `eh-frame-header`?

Note that we still have to implement fallbacks for each of these for custom target specs.

Issue #1005, opened 2024-03-09
Problems with `est-idf-sys` on Mac m1 while building rust template example


samutigro:
I had problems with `est-idf-sys` on my Mac m1 while building rust template example, every time I tried to build the project it was failing with this error:

`error: failed to run custom build command for `esp-idf-sys v0.34.1`

Caused by:
  process didn't exit successfully: `/Users/samueleauteri/Desktop/Rust/template/target/debug/build/esp-idf-sys-5ad39a82e35dc1cd/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=ESP_IDF_TOOLS_INSTALL_DIR
  cargo:rerun-if-env-changed=ESP_IDF_SDKCONFIG
  cargo:rerun-if-env-changed=ESP_IDF_SDKCONFIG_DEFAULTS
  cargo:rerun-if-env-changed=MCU
  cargo:rerun-if-env-changed=ESP_IDF_SYS_ROOT_CRATE
  cargo:rerun-if-env-changed=ESP_IDF_VERSION
  cargo:rerun-if-env-changed=ESP_IDF_REPOSITORY
  cargo:rerun-if-env-changed=ESP_IDF_CMAKE_GENERATOR
  cargo:rerun-if-env-changed=IDF_PATH
  cargo:rerun-if-env-changed=EXTRA-COMPONENTS
  cargo:rerun-if-env-changed=ESP_IDF_COMPONENTS
  cargo:rerun-if-env-changed=ESP_IDF_COMPONENT_MANAGER
  cargo:rerun-if-changed=/Users/samueleauteri/Desktop/Rust/template/sdkconfig.defaults
  CMAKE_PREFIX_PATH_xtensa-esp32-espidf = None
  CMAKE_PREFIX_PATH_xtensa_esp32_espidf = None
  TARGET_CMAKE_PREFIX_PATH = None
  CMAKE_PREFIX_PATH = None
  CMAKE_xtensa-esp32-espidf = None
  CMAKE_xtensa_esp32_espidf = None
  TARGET_CMAKE = None
  CMAKE = None
[...]

PR #1252, opened 2024-11-01
Use `rustc`'s knowledge of LLVM/Clang target triples

madsmtm:
Part of https://github.com/rust-lang/cc-rs/issues/1249.

I _tried_ to keep it separate from cleaning up the Apple stuff, but things were quite intertwined, so I ended up having to do some of it. Still, this PR aims to have as few functional changes as possible (the only effect should be that we now use the correct Clang target).

Fixes https://github.com/rust-lang/cc-rs/issues/542 (was fixed before, but now we do it properly by passing `arm64-apple-macosx` instead of `arm64-apple-darwin`).
Fixes https://github.com/rust-lang/cc-rs/issues/1255.
Fixes https://github.com/rust-lang/cc-rs/issues/1205.
Might improve the situation in https://github.com/rust-lang/cc-rs/issues/1005.

This affects (checking local test script...) 128 targets, though I feel pretty confident that the change is correct, and if it's not, then it's likely an issue in `rustc` as well.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -234,7 +234,7 @@ use shlex::Shlex;
 mod parallel;
 mod target;
 mod windows;
-use target::Target;
+use self::target::TargetInfo;
 // Regardless of whether this should be in this crate's public API,
 // it has been since 2015, so don't break it.
 pub use windows::find_tools as windows_registry;
@@ -634,7 +634,7 @@ impl Build {
         &self,
         flag: &OsStr,
         compiler_path: &Path,
-        target: &Target,
+        target: &TargetInfo,
     ) -> Result<bool, Error> {
         let compiler_flag = CompilerFlag {
             compiler: compiler_path.into(),
@@ -1932,7 +1932,7 @@ impl Build {
     fn add_default_flags(
         &self,
         cmd: &mut Tool,
-        target: &Target,
+        target: &TargetInfo,
         opt_level: &str,
     ) -> Result<(), Error> {
         let raw_target = self.get_raw_target()?;
@@ -3579,10 +3579,10 @@ impl Build {
             .or_else(|| prefixes.first().copied())
     }
 
-    fn get_target(&self) -> Result<Target, Error> {
+    fn get_target(&self) -> Result<TargetInfo, Error> {
         match &self.target {
             Some(t) => t.parse(),
-            None => Target::from_cargo_environment_variables(),
+            None => TargetInfo::from_cargo_environment_variables(),
         }
     }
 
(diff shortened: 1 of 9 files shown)


NobodyXu (changes requested):

src/lib.rs
```diff
@@ -2181,18 +2087,17 @@ impl Build {
...
-                        // FIXME: Convert rustc target to Clang target.
-                        cmd.push_cc_arg(format!("--target={}", raw_target).into());
-                    } else if target.os == "windows" {
-                        cmd.args.push(
-                            format!("--target={}-pc-windows-{}", target.full_arch, target.env)
-                                .into(),
-                        )
+                    // Add version information to the target.
+                    let llvm_target = if target.vendor == "apple" {
+                        let deployment_target = self.apple_deployment_target(&target);
```
NobodyXu, line 2094: Clippy error:

```suggestion
                        let deployment_target = self.apple_deployment_target(target);
```

NobodyXu (changes requested):

src/target/llvm.rs
```diff
@@ -0,0 +1,91 @@
...
+    };
+    let os = match os {
+        "darwin" => "macosx",
+        "visionos" => "xros",
+        "uefi" => "windows",
+        os => os,
+    };
+    let env = match env {
+        "newlib" | "nto70" | "nto71" | "ohos" | "p1" | "p2" | "relibc" | "sgx" | "uclibc" => "",
+        abi => abi,
```
NobodyXu, line 52: Nit naming convention:

```suggestion
        env => env,
```

src/target/llvm.rs
```diff
@@ -0,0 +1,91 @@
...
+        "sim" => "simulator",
+        "llvm" | "softfloat" | "uwp" | "vec-extabi" => "",
+        "ilp32" => "_ilp32",
+        abi => abi,
+    };
+    match (env, abi) {
+        ("", "") => format!("{arch}-{vendor}-{os}"),
+        (env, "") => format!("{arch}-{vendor}-{os}-{env}"),
+        ("", abi) => format!("{arch}-{vendor}-{os}-{abi}"),
+        (env, abi) => format!("{arch}-{vendor}-{os}-{env}{abi}"),
```
NobodyXu, line 64:
```suggestion
        (env, abi) => format!("{arch}-{vendor}-{os}-{env}{abi}"),
```

NobodyXu (changes requested):

NobodyXu (approved):
THanks, lGTM!

src/target/apple.rs
```diff
@@ -0,0 +1,19 @@
+use super::TargetInfo;
+
+impl TargetInfo {
+    pub(crate) fn apple_sdk_name(&self) -> &'static str {
+        match (&*self.os, &*self.abi) {
+            ("macos", "") => "macosx",
+            ("ios", "") => "iphoneos",
```
glandium, line 7: Fun fact: x86_64-apple-ios is actually iphonesimulator.

NobodyXu: oops 😅 looks like we have to update this

NobodyXu: https://github.com/rust-lang/cc-rs/blob/7a786c5e3dc149f32e565360dd8c3e43748592f5/src/target/apple.rs#L8

Is checking for `abi = "sim"` enough?

madsmtm: Yes, it is, `rustc` [sets the correct ABI](https://github.com/rust-lang/cc-rs/blob/7a786c5e3dc149f32e565360dd8c3e43748592f5/src/target/generated.rs#L2811).

Issue #1030, opened 2024-04-11
Don't specify `-arch` and `-m{}os-version-min=` on Apple platforms


madsmtm:
As recommended by Clang developers in https://github.com/llvm/llvm-project/issues/88271#issuecomment-2050425068, we already specify `--target`, and that should be enough.

Related: https://github.com/rust-lang/cc-rs/pull/1029, the `-mxros-version-min`/`-mvisionos-version-min` option does not exist.

CC @BlackHoleFox.

NobodyXu:
sounds good to me, IIRC specifying version-min might be related to errors in other projects using cc, maybe even rustc bootstrap, which failed to compile llvm.

BlackHoleFox:
Yeah ditching it seems like a fine option. Only thought is maybe compatibility with old (whatever matches 10.12) XCode versions but that might not be a real issue since there is no reason to build iOS (and derivative) apps with anything except the last N-2 XCode releases for app store reasons.

Would the idea be to remove use of `-m{}os-version-min` and `-arch` everywhere except macOS?

QuentinPerez:
Maybe  `platform_version`, it seems to support visionos.

```
     -platform_version platform min_version sdk_version
             This is set to indicate the platform, oldest supported version of that platform that output is to be used on, and the SDK that the output was
             built against.  platform is a numeric value as defined in <mach-o/loader.h>, or it may be one of the following strings:
             • macos
             • ios
             • tvos
             • watchos
             • bridgeos
             • visionos
             • xros
             • mac-catalyst
             • ios-simulator
             • tvos-simulator
             • watchos-simulator
             • visionos-simulator
             • xros-simulator
             • driverkit
             Specifying a newer min or SDK version enables the linker to assume features of that OS or SDK in the output file. The format of min_version
             and sdk_version is a version number such as 10.13 or 10.14

```

PR #1256, opened 2024-11-02
Apple: Remove redundant flags

madsmtm:
Clang's `--target` argument is enough to specify the architecture, OS and deployment target, so let's remove the redundant `-arch`, `-m32`/`-m64` and `-m*-version-min` flags.

The first commit is a non-functional refactor, the others do the actual removal. Proper GCC support is still broken, since we specify `-arch` and not `-march`, will fix that separately.

Fixes https://github.com/rust-lang/cc-rs/issues/1030.


NobodyXu:
cc @BlackHoleFox can you take a look at this PR please?

madsmtm:
I decided to go a different direction with the PR, have updated the title and description to reflect that.

src/lib.rs
```diff
@@ -2528,113 +2526,38 @@ impl Build {
 
     fn apple_flags(&self, cmd: &mut Tool) -> Result<(), Error> {
         let target = self.get_target()?;
-        let arch_str = &*target.full_arch;
 
-        let arch = if target.os == "macos" {
-            match arch_str {
-                "i686" => AppleArchSpec::Device("-m32"),
```
NobodyXu, line 2535: It seems that the `-m32`/`-m64` flag is also removed, I think it might affect correctness?

NobodyXu (changes requested):
Thanks, there's a merge conflict and one review feedback.

src/lib.rs
madsmtm: Sorry, this was intentional, updated the issue description to match.

It won't affect correctness because:
- On Clang, the `--target` flag is enough.
- On GCC, [we already set `-m32`/`-m64`](https://github.com/rust-lang/cc-rs/blob/d2822d93afc90f9727cfe775f63701bbc793f68a/src/lib.rs#L2047-L2055), the flag was duplicated.

NobodyXu: I think the m32/m64 settings there does not detect arm64?

madsmtm: `-m64` is wrong on Aarch64, the target does not share compiler with ARM (unlike Sparc, Mips, x86, etc.), so `-m64` isn't needed (and isn't [documented](https://gcc.gnu.org/onlinedocs/gcc/AArch64-Options.html) either).

See also https://stackoverflow.com/questions/58244095/gcc-7-error-unrecognized-command-line-option-m64.

I recall reading somewhere that it works nowadays, but only because it's deliberately ignored.

(But good that you're asking the question!)

NobodyXu (approved):
Thanks LGTM, I will cut a release and hopefully there's no regression.

src/lib.rs
NobodyXu: Thanks

BlackHoleFox:
> cc @BlackHoleFox can you take a look at this PR please?

Suppose I'm 4 hours late but this looks fine to me 🎉

PR #1257, opened 2024-11-02
Reduce size of `cc::Build`  and size of generated targets

NobodyXu:


src/target.rs
```diff
@@ -2,129 +2,52 @@
...
+mod parser;
+
+pub(crate) use parser::TargetInfoParser;
 
 /// Information specific to a `rustc` target.
 ///
 /// See <https://doc.rust-lang.org/cargo/appendix/glossary.html#target>.
 #[derive(Debug, PartialEq, Clone)]
-pub(crate) struct TargetInfo {
+pub(crate) struct TargetInfo<'a> {
```
NobodyXu, line 20: cc @madsmtm This is the optimization I was proposing, can you take a look and confirm this looks good please?

src/target/parser.rs
```diff
@@ -0,0 +1,130 @@
+use std::{env, str::FromStr};
+
+use crate::{
+    target::{llvm, TargetInfo},
+    utilities::OnceLock,
+    Error, ErrorKind,
+};
+
+#[derive(Debug)]
+struct TargetInfoParserInner {
```
madsmtm, line 10: Nit:
```suggestion
struct ParsedTargetInfo {
```

(Or `OwnedTargetInfo`)

madsmtm (commented):
Don't have a strong preference for either version (in favour is that matching on `&str` is nicer, against is that we have to maintain the list of struct fields).

src/target/parser.rs
NobodyXu: Hmmmm I think `Inner` suffix is ok, if we want to rename it we probably want to rename the other one as well.

NobodyXu:
> against is that we have to maintain the list of struct fields

True we need that extra field maintenance, but that would make the generated binary smaller.

AFAIK `std::borrow::Cow<'static, str>` is 4 pointers long, and `&str` is only 2 pointers long.

Given that we use `Cow` a lot it's almost 2x smaller constant data, so I think it should be worth the extra effort.

NobodyXu:
P.S. I think that some crates use cc-rs at runtime as a normal dep, so this should reduce their binary size.

PR #1264, opened 2024-11-03
Use -windows-gnu for all UEFI targets

nicholasbishop:
This restores the behavior prior to
https://github.com/rust-lang/cc-rs/pull/1252, when the UEFI targets were hardcoded in src/lib.rs.


NobodyXu (commented):
cc @madsmtm can you take a look at this please?

nicholasbishop:
That makes sense; unfortunately I don't have answers to most of those questions yet. The specific motivation for this PR was trying to build https://github.com/Amanieu/minicov for `x86_64-unknown-uefi`; that used to work, but stopped working with cc-v1.1.32. When the LLVM target is `x86_64-unknown-windows`, I get an error from a missing `malloc.h`.

Possibly the Rust UEFI aarch64/x86_64 targets should be switched to -windows-gnu, as i686 already is.

madsmtm:
Fair enough, then perhaps a comment that we don't know why this hack is needed, and link to this PR ;)

Alternatively, if you're feeling venturous, I found https://krinkinmu.github.io/2020/10/11/efi-getting-started.html#bulding, which seems to suggest that the _actual_ fix is to pass the `-ffreestanding` flag?

madsmtm:
Actually, seems like LLVM should have support for UEFI nowadays: https://discourse.llvm.org/t/rfc-uefi-driver-support-uefi-target/73261.

So maybe pass those targets (`*-unknown-uefi`) to Clang instead? (Though I wouldn't block a quickfix on that).

nicholasbishop:
Thanks for opening that PR! I wasn't aware clang had added these targets.

I've updated this PR to add a comment describing the history, the current state, and a TODO for the future.

Also tweaked the way the override is set, so that it won't immediately change to follow rustc if the rustc targets are changed, since I think we would want to be intentional about changing cc-rs to match.

madsmtm (approved):
Looks good to me now, thanks for writing things out!

Issue #1254, opened 2024-11-01
Mechanism for automatically passing flags used by rustc


mrkajetanp:
When cc-rs is running in a build script, it currently does not check which flags were passed to rustc.

There are some flags which in order to be fully effective need to be passed to both rustc and cc, a good example being AArch64 branch protection - if the Rust code is built with BTI but the C component is not, that will disable BTI for the whole binary.

Would some mechanism for checking which flags were passed to rustc and determining their corresponding cc flags be desirable to have? This could either be in the form of adding them in automatically, a separate function such as `inherit_rustc_flags()` or even just a warning to the user that some flags that should match don't match.

I'm thinking about implementing something along the lines of what I described above, but I wanted to ask for some thoughts from people who work on this crate. Is this desirable in the first place? If so, roughly what direction should I go in with this?

NobodyXu:
Thanks, we definitely want more data to be fetcher from cargoz

Can you provide more details on what rustc/cargo flags it is?

mrkajetanp:
The specific case I'm coming at this with is `-Z branch-protection`, e.g. `-Z branch-protection=pac-ret,bti`.
The corresponding cc flag would be `-mbranch-protection=pac-ret+bti`.
I imagine this would apply to some other flags too and in the end we could just have a `rustc => cc` flag match statement that could be expanded over time.
We can read the flags through `CARGO_ENCODED_RUSTFLAGS` and parse those from there.

NobodyXu:
Thanks that's a good idea, I would accept a PR for it.

mrkajetanp:
Sounds good, will write up a PR then.
Should this be an opt-in thing with its own builder function? E.g. like
```
cc::Build::new().file("foo.c").inherit_rustc_flags().compile("foo");
```
Or should we have this done by default with some opt-out?

NobodyXu:
I think opt-out makes more sense, or maybe we don't even need an opt-out mechanism.

AFAIK the current cargo env usage does not have any opt-out mechanism.

PR #1266, opened 2024-11-03
Fall back to `rustc`'s default value for `+crt-static`/`fn static_flag`

madsmtm:
And parse more information from `rustc`, to prepare for doing more of https://github.com/rust-lang/cc-rs/issues/1249.

The net fix is fairly small, it's that 37 targets (the full list is all targets with `crt-static` in their features table) now have `-static` passed also when compiling outside a build script (it was already passed when compiling in a build script).


madsmtm:
Opened https://github.com/rust-lang/cargo/issues/14778 for the issue of Cargo not passing this in `CARGO_CFG_TARGET_FEATURE`. There is not really a way that `cc` can detect whether this is the case, apart from maybe [reading `CARGO_ENCODED_RUSTFLAGS`](https://github.com/rust-lang/cc-rs/issues/1254)?

dev-tools/gen-target-info/src/target_specs.rs
```diff
@@ -32,3 +35,48 @@ pub struct RustcTargetSpecs(
...
+        for cfg in cfgs {
+            let (name, value) = cfg
+                .split_once('=')
+                .map(|(n, v)| (n.trim(), Some(v.trim().trim_matches('"'))))
+                .unwrap_or((cfg.trim(), None));
+
+            match (name, value) {
+                ("target_feature", Some(value)) => target_features.push(value.to_string()),
+                ("target_family", Some(value)) => target_families.push(value.to_string()),
+                ("target_endian", Some(value)) => target_endian = Some(value.to_string()),
```
NobodyXu, line 67: Nit:

Shall we have a check for duplicate value, just in case rustc gives us some invalid output?

src/target/generated.rs
```diff
@@ -14,6 +14,7 @@ pub(crate) const LIST: &[(&str, TargetInfo<'static>)] = &[
             env: "",
             abi: "",
             unversioned_llvm_target: "arm64-apple-macosx",
+            features: "aes,crc,dit,dotprod,dpb,dpb2,fcma,fhm,flagm,flagm2,fp16,frintts,jsconv,lor,lse,lse2,neon,paca,pacg,pan,pmuv3,ras,rcpc,rcpc2,rdm,sb,sha2,sha3,ssbs,v8.1a,v8.2a,v8.3a,v8.4a,vh",
```
NobodyXu, line 17: The generated file is getting larger and I'm slightly worried about the compile-time/binary size.

It probably would be fine for now and we should not hesitate adding more, if it eases maintenance burden and reduce hardcoded values.

NobodyXu (commented):
Thanks this mostly LGTM

Since there's a cargo bug, I think we should wait until upstream provides a clear answer to the issue.

src/target/generated.rs
madsmtm: I suspect that the features here are going to be fairly important, but yeah, I do understand your worry.

I guess if we wanted to decrease binary size, we could parse them into a `&[Feature]` beforehand?

dev-tools/gen-target-info/src/target_specs.rs
madsmtm: Good idea

src/target/generated.rs
NobodyXu: On the other hand, we could replace all this features with a boolean since all we need to know is if `cry-static` is included in the feature or not.

If we need more features later, then we can introduce them later, and probably use an enum which is more efficient than `&str`

NobodyXu: If we want to reduce the binary size/compilation time, then I could think of two ideas:

First, split the target and `TargetInfo` into two arrays, this would help simplify the array, and might make binary search faster as more target can be in the cache.

For `TargetInfo`, we can store the string inline.

We can internalise the string into a one big `&str` and refer to it via index, we can simply use a `u16` for base pointer.

The simplest impl is to use `HashSet` in gen-target-info to dedup them, and then concat them into one giant string, each terminated by `\0` and assign them each a `u16` based on the length within the giant string.

That will compress 16B `&'static str` on 64-bit platform to 2B, and should improve compile-time as well.

Though it looks like something out of scope for this ticket.

NobodyXu: So for this PR, cc @madsmtm I just want you to replace `features: &str` with a `crt_static: boolean`, that would be good enough.

The rest of the idea should be discussed and completed in a different PR.

mrkajetanp:
> shall we read from CARGO_ENCODED_RUSTFLAGS instead?

I should be able to post my PR for #1254 this week which introduces reading CARGO_ENCODED_RUSTFLAGS and adding cc flags based on that so it could come in handy for this?
Though in the general case I don't think it's a good idea to pass target features from there, instead target features should be read from CARGO_CFG_TARGET_FEATURE and managed separately. But if there's a bug like in this case, a stopgap solution could go there just fine.

polarathene:
> given that the upstream probably won't fix `CARGO_CFG_TARGET_FEATURE` very soon, shall we read from `CARGO_ENCODED_RUSTFLAGS` instead?

The same issue occurs there, the implicit usage (musl target) does not show up unless it's been explicitly set.

polarathene:
Not entirely familiar with the discussion on targets related to this PR, but this might be of interest?:
- [`build-target` crate](https://github.com/OpenByteDev/build-target) enums/methods related to build target info. Fields are queried on demand via the `CARGO_CFG_*` ENVs.
- [`platforms` crate](https://crates.io/crates/platforms) similar but more generic lookup of rust platform/"triples" rather than only the active build target, all this data is static. There is an `ALL` const and [individual consts](https://github.com/rustsec/rustsec/blob/13604e1909456e25752784681f0367b13a21e47b/platforms/src/platform/platforms.rs#L3224-L3233) of each platform with common fields to query, the targets can be parsed through serde, etc.

So if there's concern about size/maintenance, perhaps that's a bit helpful to delegate to a specialized crate?

PR #153, opened 2017-04-16
Support for the FRC toolchain

connorworley:
This PR adds support for the [toolchain](http://first.wpi.edu/FRC/roborio/toolchains/) used in the FIRST Robotics Competition.

PR #1268, opened 2024-11-04
Remove support for FRC

madsmtm:
FRC was added in https://github.com/rust-lang/cc-rs/pull/153, but since https://github.com/rust-lang/rust/pull/42063 never landed (as a custom target turned out to be unnecessary), this is code that we're never going to hit.

So let's remove support for it in `cc` too.

CC @connorworley

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2203,15 +2203,6 @@ impl Build {
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


NobodyXu (approved):
Thanks!

PR #1270, opened 2024-11-05
Use relative directory for obj files hash

nmattia:
When producing `.o` files, the absolute path of each source file is hashed and used as a prefix for the output object files. While this avoids clashes between different source files with the same basenames, this causes determinism issues in the build as the full system path influences the build output.

This change strips the out_dir parent from the source file paths, assuming that the source files are in the out_dir as well.

### DIFF
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -312,14 +312,14 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
         // object files with the same filename in different subfolders.
         let mut hasher = hash_map::DefaultHasher::new();
 
-        // Make the dirname relative to avoid full system paths influencing the sha and
-        // making the output system-dependent
-        let prefix = dst.parent().expect("Could not get parent of '{dst}'");
-        let prefix: &str = &prefix.to_string_lossy();
-        let err = format!("Could not strip prefix '{prefix}' from '{dirname}' to make '{dirname}' relative");
-        let dirname = dirname.strip_prefix(prefix).expect(&err);
+        // Make the dirname relative (if possible) to avoid full system paths influencing the sha
+        // and making the output system-dependent
+        let mut dirname = dirname.to_string();
+        if let Ok(root) = std::env::var("CARGO_MANIFEST_DIR") {
+            dirname = dirname.strip_prefix(&root).unwrap_or(&dirname).to_string();
+        }
 
-        hasher.write(dirname.to_string().as_bytes());
+        hasher.write(dirname.as_bytes());
         let obj = dst
             .join(format!("{:016x}-{}", hasher.finish(), basename))
             .with_extension("o");
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -1,6 +1,7 @@
 //! Miscellaneous helpers for running commands
 
 use std::{
+    borrow::Cow,
     collections::hash_map,
     ffi::OsString,
     fmt::Display,
@@ -320,7 +321,7 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
         #[allow(clippy::disallowed_methods)]
         let dirname = if let Some(root) = std::env::var_os("CARGO_MANIFEST_DIR") {
             let root = root.to_string_lossy();
-            crate::Cow::Borrowed(dirname.strip_prefix(&*root).unwrap_or(&dirname))
+            Cow::Borrowed(dirname.strip_prefix(&*root).unwrap_or(&dirname))
         } else {
             dirname
         };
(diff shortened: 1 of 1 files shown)


src/command_helpers.rs
```diff
@@ -311,6 +311,14 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
         // Hash the dirname. This should prevent conflicts if we have multiple
         // object files with the same filename in different subfolders.
         let mut hasher = hash_map::DefaultHasher::new();
+
+        // Make the dirname relative to avoid full system paths influencing the sha and
```
NobodyXu, line 315: I'm ok with making it relative to cargo workspace root

Alternatively maybe hashing the entire file is also a good idea, but:
 - it would then be opt-in (because extra dep)
 - attackers can change the file while we are calculating hashsum/compiling

nmattia: > hashing the entire file is also a good idea

I like that actually. Up to you!

NobodyXu: > I like that actually. Up to you!

In that case, I would like you to add an optional feature `sha1-file-hashing` which enables dep [`sha1-checked`](https://lib.rs/crates/sha1-checked).

nmattia: @NobodyXu actually upon further reflection, I think it might be nice if things work out of the box, i.e. without needed to enable a feature. I've updated the change to make it relative to the workspace root:

> I'm ok with making it relative to cargo workspace root

PTAL!

nmattia:
@NobodyXu I missing some context re. `Build::getenv` vs `std::env::var`, can you advise here?

https://github.com/rust-lang/cc-rs/pull/1103

NobodyXu:
I'd recommend using `Build::getenv`

NobodyXu:
Actually, it's ok to use env::var here, because CARGO_* is a cargo env that always triggers rebuild.

You just need to add a clippy ignore and a comment emplaning why the clippy lint is ignored

src/command_helpers.rs
```diff
@@ -311,7 +311,19 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
         // Hash the dirname. This should prevent conflicts if we have multiple
         // object files with the same filename in different subfolders.
         let mut hasher = hash_map::DefaultHasher::new();
-        hasher.write(dirname.to_string().as_bytes());
+
+        // Make the dirname relative (if possible) to avoid full system paths influencing the sha
+        // and making the output system-dependent
+        let mut dirname = dirname.to_string();
```
NobodyXu, line 317:
```suggestion
        let mut dirname = Cow::Borrowed(dirname);
```

NobodyXu: I'd prefer having a

```
let dirname = if let Ok(root) = env::var(...) { ... } else { Cow::Borrowed(dirname) };
```

then having a mutable variable.

src/command_helpers.rs
```diff
@@ -311,7 +311,19 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
...
+
+        // Here we allow using std::env::var (instead of Build::getenv) because
+        // CARGO_* variables always trigger a rebuild when changed
+        #[allow(clippy::disallowed_methods)]
+        if let Ok(root) = std::env::var("CARGO_MANIFEST_DIR") {
```
NobodyXu, line 322: Please use `var_os` and `to_string_lossy` to support non utf-8 characters:
```suggestion
        if let Ok(root) = std::env::var_os("CARGO_MANIFEST_DIR") {
```

NobodyXu (changes requested):

NobodyXu:
Thanks, just a few change requests

src/command_helpers.rs
```diff
@@ -311,7 +311,21 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
...
+
+        // Make the dirname relative (if possible) to avoid full system paths influencing the sha
+        // and making the output system-dependent
+        //
+        // NOTE: Here we allow using std::env::var (instead of Build::getenv) because
+        // CARGO_* variables always trigger a rebuild when changed
+        #[allow(clippy::disallowed_methods)]
+        let dirname = if let Some(root) = std::env::var_os("CARGO_MANIFEST_DIR") {
+            let root = root.to_string_lossy();
+            crate::Cow::Borrowed(dirname.strip_prefix(&*root).unwrap_or(&dirname))
```
NobodyXu, line 323:
```suggestion
            Cow::Borrowed(dirname.strip_prefix(&*root).unwrap_or(&dirname))
```

Lets import `std::borrow::Cow` here

NobodyXu (changes requested):
Thanks, just one final comment:

NobodyXu (approved):
Thank you!

PR #1277, opened 2024-11-07
Fix for diffrent hashes between diffrent build paths

harishsadineni1:
when the build paths are diffrent for rust sources, the objects files genarted are having diffrent hash values for each build.

- Updated the hashing mechanism to ensure consistent hash values across different build paths.
- Addressed issues with reproducibility by normalizing directory paths before hashing, preventing discrepancies caused by varying build environments.

### DIFF
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -288,6 +288,7 @@ fn wait_on_child(
 /// and store them in the output Object.
 pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<Object>, Error> {
     let mut objects = Vec::with_capacity(files.len());
+    let target_substring = ["rustc"];
     for file in files {
         let basename = file
             .file_name()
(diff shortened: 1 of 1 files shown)


src/command_helpers.rs
```diff
@@ -308,10 +309,29 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
...
+                     //If a target is found and pos is None, set it
+                         if pos.is_none() || index < pos.unwrap() {
+                             pos = Some(index);
+                         }
+                }
+            }
+            pos
+        }
+
+        let filtered_dirname = if let Some(pos) = find_target_position(&dirname, &target_substring) {
```
NobodyXu, line 326: IMHO stripping `rustc` isn't very reliable, user could have `rustc` somewhere in the path.

I think stripping `CARGO_MANIFEST_DIR` is a better solution.

NobodyXu (changes requested):

NobodyXu:
Thanks, I think #1270 already implements this

PR #1279, opened 2024-11-07
build: Inherit flags from rustc

### LINKED ISSUE #1254: Mechanism for automatically passing flags used by rustc

mrkajetanp:
Where applicable, detect which RUSTFLAGS were set for rustc and convert them into their corresponding cc flags in order to ensure consistent codegen across Rust and non-Rust modules.

Resolves #1254

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -1,0 +1,482 @@
+use crate::target::TargetInfo;
+use crate::{Build, Error, ErrorKind, ToolFamily};
+use std::borrow::Cow;
+use std::ffi::OsString;
+use std::path::Path;
+
+#[derive(Debug, PartialEq, Default)]
+pub(crate) struct RustcCodegenFlags {
+    branch_protection: Option<String>,
+    code_model: Option<String>,
+    no_vectorize_loops: bool,
+    no_vectorize_slp: bool,
+    profile_generate: Option<String>,
+    profile_use: Option<String>,
+    control_flow_guard: Option<String>,
+    lto: Option<String>,
+    relocation_model: Option<String>,
+    embed_bitcode: Option<bool>,
+    force_frame_pointers: Option<bool>,
+    link_dead_code: Option<bool>,
+    no_redzone: Option<bool>,
+    soft_float: Option<bool>,
+}
+
+impl RustcCodegenFlags {
+    // Parse flags obtained from CARGO_ENCODED_RUSTFLAGS
+    pub(crate) fn parse(rustflags_env: &str) -> Result<RustcCodegenFlags, Error> {
+        fn is_flag_prefix(flag: &str) -> bool {
+            [
+                "-Z",
+                "-C",
+                "--codegen",
+                "-L",
+                "-l",
+                "-o",
+                "-W",
+                "--warn",
+                "-A",
+                "--allow",
+                "-D",
+                "--deny",
+                "-F",
+                "--forbid",
+            ]
+            .contains(&flag)
+        }
+
+        fn handle_flag_prefix<'a>(prev: &'a str, curr: &'a str) -> Cow<'a, str> {
+            match prev {
+                "--codegen" | "-C" => Cow::from(format!("-C{}", curr)),
+                // Handle flags passed like --codegen=code-model=small
+                _ if curr.starts_with("--codegen=") => Cow::from(format!("-C{}", &curr[10..])),
+                "-Z" => Cow::from(format!("-Z{}", curr)),
+                "-L" | "-l" | "-o" => Cow::from(format!("{}{}", prev, curr)),
+                // Handle lint flags
+                "-W" | "--warn" => Cow::from(format!("-W{}", curr)),
+                "-A" | "--allow" => Cow::from(format!("-A{}", curr)),
+                "-D" | "--deny" => Cow::from(format!("-D{}", curr)),
+                "-F" | "--forbid" => Cow::from(format!("-F{}", curr)),
+                _ => Cow::from(curr),
+            }
+        }
+
+        let mut codegen_flags = Self::default();
+
+        let mut prev_prefix = None;
+        for curr in rustflags_env.split("\u{1f}") {
+            let prev = prev_prefix.take().unwrap_or("");
+            if prev.is_empty() && is_flag_prefix(curr) {
+                prev_prefix = Some(curr);
+                continue;
+            }
+
+            let rustc_flag = handle_flag_prefix(prev, curr);
+            codegen_flags.set_rustc_flag(&rustc_flag)?;
+        }
+
+        Ok(codegen_flags)
+    }
+
+    fn set_rustc_flag(&mut self, flag: &str) -> Result<(), Error> {
+        // Convert a textual representation of a bool-like rustc flag argument into an actual bool
+        fn arg_to_bool(arg: impl AsRef<str>) -> Option<bool> {
+            match arg.as_ref() {
+                "y" | "yes" | "on" | "true" => Some(true),
+                "n" | "no" | "off" | "false" => Some(false),
+                _ => None,
+            }
+        }
+
+        let (flag, value) = if let Some((flag, value)) = flag.split_once('=') {
+            (flag, Some(value.to_owned()))
+        } else {
+            (flag, None)
+        };
+
+        fn flag_ok_or(flag: Option<String>, msg: &'static str) -> Result<String, Error> {
+            flag.ok_or(Error::new(ErrorKind::InvalidFlag, msg))
+        }
+
+        match flag {
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#code-model
+            "-Ccode-model" => {
+                self.code_model = Some(flag_ok_or(value, "-Ccode-model must have a value")?);
+            }
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#no-vectorize-loops
+            "-Cno-vectorize-loops" => self.no_vectorize_loops = true,
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#no-vectorize-slp
+            "-Cno-vectorize-slp" => self.no_vectorize_slp = true,
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#profile-generate
+            "-Cprofile-generate" => {
+                self.profile_generate =
+                    Some(flag_ok_or(value, "-Cprofile-generate must have a value")?);
+            }
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#profile-use
+            "-Cprofile-use" => {
+                self.profile_use = Some(flag_ok_or(value, "-Cprofile-use must have a value")?);
+            }
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#control-flow-guard
+            "-Ccontrol-flow-guard" => self.control_flow_guard = value.or(Some("true".into())),
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#lto
+            "-Clto" => self.lto = value.or(Some("true".into())),
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#relocation-model
+            "-Crelocation-model" => {
+                self.relocation_model =
+                    Some(flag_ok_or(value, "-Crelocation-model must have a value")?);
+            }
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#embed-bitcode
+            "-Cembed-bitcode" => self.embed_bitcode = value.map_or(Some(true), arg_to_bool),
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#force-frame-pointers
+            "-Cforce-frame-pointers" => {
+                self.force_frame_pointers = value.map_or(Some(true), arg_to_bool)
+            }
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#link-dead-code
+            "-Clink-dead-code" => self.link_dead_code = value.map_or(Some(true), arg_to_bool),
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#no-redzone
+            "-Cno-redzone" => self.no_redzone = value.map_or(Some(true), arg_to_bool),
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#soft-float
+            // Note: This flag is now deprecated in rustc.
+            "-Csoft-float" => self.soft_float = value.map_or(Some(true), arg_to_bool),
+            // https://doc.rust-lang.org/beta/unstable-book/compiler-flags/branch-protection.html
+            // FIXME: Drop the -Z variant and update the doc link once the option is stabilised
+            "-Zbranch-protection" | "-Cbranch-protection" => {
+                self.branch_protection =
+                    Some(flag_ok_or(value, "-Zbranch-protection must have a value")?);
+            }
+            _ => {}
+        }
+        Ok(())
+    }
+
+    // Rust and clang/cc don't agree on what equivalent flags should look like.
+    pub(crate) fn cc_flags(
+        &self,
+        build: &Build,
+        path: &Path,
+        family: ToolFamily,
+        target: &TargetInfo,
+        flags: &mut Vec<OsString>,
+    ) {
+        // Push `flag` to `flags` if it is supported by the currently used CC
+        let mut push_if_supported = |flag: OsString| {
+            if build
+                .is_flag_supported_inner(&flag, path, target)
+                .unwrap_or(false)
+            {
+                flags.push(flag);
+            } else {
+                build.cargo_output.print_warning(&format!(
+                    "Inherited flag {:?} is not supported by the currently used CC",
+                    flag
+                ));
+            }
+        };
+
+        match family {
+            ToolFamily::Clang { .. } | ToolFamily::Gnu => {
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mbranch-protection
+                if let Some(value) = &self.branch_protection {
+                    push_if_supported(
+                        format!("-mbranch-protection={}", value.replace(",", "+")).into(),
+                    );
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mcmodel
+                if let Some(value) = &self.code_model {
+                    push_if_supported(format!("-mcmodel={value}").into());
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fno-vectorize
+                if self.no_vectorize_loops {
+                    push_if_supported("-fno-vectorize".into());
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fno-slp-vectorize
+                if self.no_vectorize_slp {
+                    push_if_supported("-fno-slp-vectorize".into());
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fprofile-generate
+                if let Some(value) = &self.profile_generate {
+                    push_if_supported(format!("-fprofile-generate={value}").into());
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fprofile-use
+                if let Some(value) = &self.profile_use {
+                    push_if_supported(format!("-fprofile-use={value}").into());
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mguard
+                if let Some(value) = &self.control_flow_guard {
+                    let cc_val = match value.as_str() {
+                        "y" | "yes" | "on" | "true" | "checks" => Some("cf"),
+                        "nochecks" => Some("cf-nochecks"),
+                        "n" | "no" | "off" | "false" => Some("none"),
+                        _ => None,
+                    };
+                    if let Some(cc_val) = cc_val {
+                        push_if_supported(format!("-mguard={cc_val}").into());
+                    }
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-flto
+                if let Some(value) = &self.lto {
+                    let cc_val = match value.as_str() {
+                        "y" | "yes" | "on" | "true" | "fat" => Some("full"),
+                        "thin" => Some("thin"),
+                        _ => None,
+                    };
+                    if let Some(cc_val) = cc_val {
+                        push_if_supported(format!("-flto={cc_val}").into());
+                    }
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fPIC
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fPIE
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mdynamic-no-pic
+                if let Some(value) = &self.relocation_model {
+                    let cc_flag = match value.as_str() {
+                        "pic" => Some("-fPIC"),
+                        "pie" => Some("-fPIE"),
+                        "dynamic-no-pic" => Some("-mdynamic-no-pic"),
+                        _ => None,
+                    };
+                    if let Some(cc_flag) = cc_flag {
+                        push_if_supported(cc_flag.into());
+                    }
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fembed-bitcode
+                if let Some(value) = &self.embed_bitcode {
+                    let cc_val = if *value { "all" } else { "off" };
+                    push_if_supported(format!("-fembed-bitcode={cc_val}").into());
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fno-omit-frame-pointer
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fomit-frame-pointer
+                if let Some(value) = &self.force_frame_pointers {
+                    let cc_flag = if *value {
+                        "-fno-omit-frame-pointer"
+                    } else {
+                        "-fomit-frame-pointer"
+                    };
+                    push_if_supported(cc_flag.into());
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-dead_strip
+                if let Some(value) = &self.link_dead_code {
+                    if !value {
+                        push_if_supported("-dead_strip".into());
+                    }
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mno-red-zone
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mred-zone
+                if let Some(value) = &self.no_redzone {
+                    let cc_flag = if *value {
+                        "-mno-red-zone"
+                    } else {
+                        "-mred-zone"
+                    };
+                    push_if_supported(cc_flag.into());
+                }
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-msoft-float
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mno-soft-float
+                if let Some(value) = &self.soft_float {
+                    let cc_flag = if *value {
+                        "-msoft-float"
+                    } else {
+                        "-mno-soft-float"
+                    };
+                    push_if_supported(cc_flag.into());
+                }
+            }
+            ToolFamily::Msvc { .. } => {
+                // https://learn.microsoft.com/en-us/cpp/build/reference/guard-enable-control-flow-guard
+                if let Some(value) = &self.control_flow_guard {
+                    let cc_val = match value.as_str() {
+                        "y" | "yes" | "on" | "true" | "checks" => Some("cf"),
+                        "n" | "no" | "off" | "false" => Some("cf-"),
+                        _ => None,
+                    };
+                    if let Some(cc_val) = cc_val {
+                        push_if_supported(format!("/guard:{cc_val}").into());
+                    }
+                }
+                // https://learn.microsoft.com/en-us/cpp/build/reference/oy-frame-pointer-omission
+                if let Some(value) = &self.force_frame_pointers {
+                    let cc_flag = if *value { "/Oy-" } else { "/Oy" };
+                    push_if_supported(cc_flag.into());
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[track_caller]
+    fn check(env: &str, expected: &RustcCodegenFlags) {
+        let actual = RustcCodegenFlags::parse(env).unwrap();
+        assert_eq!(actual, *expected);
+    }
+
+    #[test]
+    fn codegen_type() {
+        let expected = RustcCodegenFlags {
+            code_model: Some("tiny".into()),
+            ..RustcCodegenFlags::default()
+        };
+        check("-Ccode-model=tiny", &expected);
+        check("-C\u{1f}code-model=tiny", &expected);
+        check("--codegen\u{1f}code-model=tiny", &expected);
+        check("--codegen=code-model=tiny", &expected);
+    }
+
+    #[test]
+    fn precedence() {
+        check(
+            "-ccode-model=tiny\u{1f}-Ccode-model=small",
+            &RustcCodegenFlags {
+                code_model: Some("small".into()),
+                ..RustcCodegenFlags::default()
+            },
+        );
+    }
+
+    #[test]
+    fn two_valid_prefixes() {
+        let expected = RustcCodegenFlags::default();
+        check("-L\u{1f}-Clto", &expected);
+    }
+
+    #[test]
+    fn three_valid_prefixes() {
+        let expected = RustcCodegenFlags {
+            lto: Some("true".into()),
+            ..RustcCodegenFlags::default()
+        };
+        check("-L\u{1f}-L\u{1f}-Clto", &expected);
+    }
+
+    #[test]
+    fn all_rustc_flags() {
+        // Throw all possible flags at the parser to catch false positives
+        let flags = [
+            // Set all the flags we recognise first
+            "-Ccode-model=tiny",
+            "-Ccontrol-flow-guard=yes",
+            "-Cembed-bitcode=no",
+            "-Cforce-frame-pointers=yes",
+            "-Clto=false",
+            "-Clink-dead-code=yes",
+            "-Cno-redzone=yes",
+            "-Cno-vectorize-loops",
+            "-Cno-vectorize-slp",
+            "-Cprofile-generate=fooprofile",
+            "-Cprofile-use=fooprofile",
+            "-Crelocation-model=pic",
+            "-Csoft-float=yes",
+            "-Zbranch-protection=bti,pac-ret,leaf",
+            // Set flags we don't recognise but rustc supports next
+            // rustc flags
+            "--cfg",
+            "a",
+            "--check-cfg 'cfg(verbose)",
+            "-L",
+            "/usr/lib/foo",
+            "-l",
+            "static:+whole-archive=mylib",
+            "--crate-type=dylib",
+            "--crate-name=foo",
+            "--edition=2021",
+            "--emit=asm",
+            "--print=crate-name",
+            "-g",
+            "-O",
+            "-o",
+            "foooutput",
+            "--out-dir",
+            "foooutdir",
+            "--target",
+            "aarch64-unknown-linux-gnu",
+            "-W",
+            "missing-docs",
+            "-D",
+            "unused-variables",
+            "--force-warn",
+            "dead-code",
+            "-A",
+            "unused",
+            "-F",
+            "unused",
+            "--cap-lints",
+            "warn",
+            "--version",
+            "--verbose",
+            "-v",
+            "--extern",
+            "foocrate",
+            "--sysroot",
+            "fooroot",
+            "--error-format",
+            "human",
+            "--color",
+            "auto",
+            "--diagnostic-width",
+            "80",
+            "--remap-path-prefix",
+            "foo=bar",
+            "--json=artifact",
+            // Codegen flags
+            "-Car",
+            "-Ccodegen-units=1",
+            "-Ccollapse-macro-debuginfo=yes",
+            "-Cdebug-assertions=yes",
+            "-Cdebuginfo=1",
+            "-Cdefault-linker-libraries=yes",
+            "-Cdlltool=foo",
+            "-Cextra-filename=foo",
+            "-Cforce-unwind-tables=yes",
+            "-Cincremental=foodir",
+            "-Cinline-threshold=6",
+            "-Cinstrument-coverage",
+            "-Clink-arg=-foo",
+            "-Clink-args=-foo",
+            "-Clink-self-contained=yes",
+            "-Clinker=lld",
+            "-Clinker-flavor=ld.lld",
+            "-Clinker-plugin-lto=yes",
+            "-Cllvm-args=foo",
+            "-Cmetadata=foo",
+            "-Cno-prepopulate-passes",
+            "-Cno-stack-check",
+            "-Copt-level=3",
+            "-Coverflow-checks=yes",
+            "-Cpanic=abort",
+            "-Cpasses=foopass",
+            "-Cprefer-dynamic=yes",
+            "-Crelro-level=partial",
+            "-Cremark=all",
+            "-Crpath=yes",
+            "-Csave-temps=yes",
+            "-Csplit-debuginfo=packed",
+            "-Cstrip=symbols",
+            "-Csymbol-mangling-version=v0",
+            "-Ctarget-cpu=native",
+            "-Ctarget-feature=+sve",
+            // Unstable options
+            "-Ztune-cpu=machine",
+        ];
+        check(
+            &flags.join("\u{1f}"),
+            &RustcCodegenFlags {
+                code_model: Some("tiny".into()),
+                control_flow_guard: Some("yes".into()),
+                embed_bitcode: Some(false),
+                force_frame_pointers: Some(true),
+                link_dead_code: Some(true),
+                lto: Some("false".into()),
+                no_redzone: Some(true),
+                no_vectorize_loops: true,
+                no_vectorize_slp: true,
+                profile_generate: Some("fooprofile".into()),
+                profile_use: Some("fooprofile".into()),
+                relocation_model: Some("pic".into()),
+                soft_float: Some(true),
+                branch_protection: Some("bti,pac-ret,leaf".into()),
+            },
+        );
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -261,6 +261,9 @@ mod tempfile;
 mod utilities;
 use utilities::*;
 
+mod flags;
+use flags::*;
+
 #[derive(Debug, Eq, PartialEq, Hash)]
 struct CompilerFlag {
     compiler: Box<Path>,
@@ -328,6 +331,7 @@ pub struct Build {
     emit_rerun_if_env_changed: bool,
     shell_escaped_flags: Option<bool>,
     build_cache: Arc<BuildCache>,
+    inherit_rustflags: bool,
 }
 
 /// Represents the types of errors that may occur while using cc-rs.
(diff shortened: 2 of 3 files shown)


mrkajetanp:
It's not fully ready yet, so far I'm looking for some feedback.
The main question is that as is the implementation doesn't check whether CC supports a flag before passing it. I tried using the`is_flag_supported(_inner)` functions to filter them out but the build script was infinite looping so I thought it could be good to just ask here first.

src/lib.rs
```diff
@@ -2439,6 +2455,32 @@ impl Build {
...
+                } else if flag.starts_with("-Z") || flag.starts_with("-C") {
+                    Some(&flag[2..])
+                } else {
+                    Some(flag)
+                }
+            })
+            .flat_map(|flag| rustc_to_cc_flag(flag, cmd.family))
+            .collect();
+
+        cmd.args.append(&mut cc_flags);
```
NobodyXu, line 2480:
```suggestion
        cmd.args.extend(
            env_os
                .to_string_lossy()
                .split("\u{1f}")
                // Strip prefixes from rustc flags
                .flat_map(|flag| {
                    if flag == "-Z" || flag == "-C" {
                        None
                    } else if flag.starts_with("-Z") || flag.starts_with("-C") {
                        Some(&flag[2..])
                    } else {
                        Some(flag)
                    }
                })
                .flat_map(|flag| rustc_to_cc_flag(flag, cmd.family))
         );
```

NobodyXu: It is very unfortunate that we have to convert it to a utf-8 string for processing...

On unix/wasm we can at least convert it to bytes.

It should also use shlex, if the feature/dep is enabled.

shlex supports both bytes and str, so it can be used in both unix/wasm and windows.

src/lib.rs
```diff
@@ -4221,6 +4263,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+        ToolFamily::Clang { .. } | ToolFamily::Gnu => match flag {
+            _ if flag.starts_with("branch-protection=") => {
+                Some(format!("-m{}", flag.replace(",", "+")).into())
+            }
+            _ if flag.starts_with("code-model=") => {
+                Some(flag.replace("code-model", "-mcmodel").into())
+            }
+            _ if flag.starts_with("no-vectorize-loops") => Some("-fno-vectorize".into()),
+            _ if flag.starts_with("no-vectorize-slp") => Some("-fno-slp-vectorize".into()),
+            _ if flag.starts_with("profile-generate") => Some(format!("-f{}", flag).into()),
```
NobodyXu, line 4278: I'd prefer inline arg if our msrv supports it:

```suggestion
            _ if flag.starts_with("profile-generate") => Some(format!("-f{flag}").into()),
```

src/lib.rs
```diff
@@ -4221,6 +4263,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+            _ if flag.starts_with("profile-use") => Some(format!("-f{}", flag).into()),
```
NobodyXu, line 4279: Same

```suggestion
            _ if flag.starts_with("profile-use") => Some(format!("-f{flag}").into()),
```

src/lib.rs
```diff
@@ -4221,6 +4263,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+                let cc_val = match rustc_val {
+                    "y" | "yes" | "on" | "true" => "all",
+                    "n" | "no" | "off" | "false" => "off",
+                    _ => return None,
+                };
+                Some(format!("-fembed-bitcode={cc_val}").into())
+            }
+            _ if flag.starts_with("force-frame-pointers") => {
+                let force_frame_pointers = if flag.contains("=") {
+                    let (_, rustc_val) = flag.split_once("=").unwrap();
```
NobodyXu, line 4301:
```suggestion
                let force_frame_pointers = if Some((_, rustc_val)) = flag.split_once("=") {
```

src/lib.rs
```diff
@@ -4221,6 +4263,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+            _ if flag.starts_with("link-dead-code") => {
+                if flag.contains("=") {
+                    let (_, rustc_val) = flag.split_once("=").unwrap();
+                    match rustc_val {
+                        "n" | "no" | "off" | "false" => Some("-dead_strip".into()),
+                        _ => None,
+                    }
+                } else {
+                    None
+                }
```
NobodyXu, line 4326:
```suggestion
                match flag.split_once("=") {
                    Some((_, "n" | "no" | "off" | "false")) => Some("-dead_strip".into()),
                    _ => None,
                }
```

src/lib.rs
```diff
@@ -4221,6 +4263,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+            }
+            _ if flag.starts_with("lto") => {
+                let lto_mode = if flag.contains("=") {
+                    let (_, rustc_val) = flag.split_once("=").unwrap();
```
NobodyXu, line 4330:
```suggestion
                let lto_mode = if Some((_, rustc_val)) = flag.split_once("=") {
```

src/lib.rs
```diff
@@ -4221,6 +4263,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+                    }
+                } else {
+                    "full"
+                };
+
+                Some(format!("-flto={}", lto_mode).into())
+            }
+            _ if flag.starts_with("no-redzone") => {
+                let no_redzone = if flag.contains("=") {
+                    let (_, rustc_val) = flag.split_once("=").unwrap();
```
NobodyXu, line 4344:
```suggestion
                let no_redzone = if Some((_, rustc_val)) = flag.split_once("=") {
```

src/lib.rs
```diff
@@ -4221,6 +4263,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+                    "pic" => "-fPIC",
+                    "pie" => "-fPIE",
+                    "dynamic-no-pic" => "-mdynamic-no-pic",
+                    _ => return None,
+                };
+                Some(cc_flag.into())
+            }
+            _ if flag.starts_with("soft-float") => {
+                let soft_float = if flag.contains("=") {
+                    let (_, rustc_val) = flag.split_once("=").unwrap();
```
NobodyXu, line 4373:
```suggestion
                let soft_float = if Some((_, rustc_val)) = flag.split_once("=") {
```

src/lib.rs
```diff
@@ -4221,6 +4263,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
     }
 }
 
+// Rust and clang/cc don't agree on what equivalent flags should look like either.
+fn rustc_to_cc_flag(flag: &str, family: ToolFamily) -> Option<OsString> {
+    match family {
+        ToolFamily::Clang { .. } | ToolFamily::Gnu => match flag {
```
NobodyXu, line 4269: Do we really want it to be a match given that every branch just use `if flag.starts_with`?

src/lib.rs
```diff
@@ -4221,6 +4263,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+                let cc_val = match rustc_val {
+                    "y" | "yes" | "on" | "true" | "checks" => "cf",
+                    "n" | "no" | "off" | "false" => "cf-",
+                    _ => return None,
+                };
+                Some(format!("/guard:{cc_val}").into())
+            }
+            _ if flag.starts_with("force-frame-pointers") => {
+                let force_frame_pointers = if flag.contains("=") {
+                    let (_, rustc_val) = flag.split_once("=").unwrap();
```
NobodyXu, line 4404:
```suggestion
                let force_frame_pointers = if Some((_, rustc_val)) = flag.split_once("=") {
```

NobodyXu (changes requested):
Thanks, I have some feedback with the code style/lint:

NobodyXu:
> The main question is that as is the implementation doesn't check whether CC supports a flag before passing it. I tried using the`is_flag_supported(_inner)` functions to filter them out but the build script was infinite looping so I thought it could be good to just ask here first.


It's because in `is_flag_supported` we create a new `cc::Build` object and use it for detecting support.

https://docs.rs/cc/latest/src/cc/lib.rs.html#660

Because the rust flag is turned on by default, it causes infinite looping.

To fix this, you could use `self.flags_supported`, which is checked just before compilation without creating a new `cc::Build` `

https://docs.rs/cc/latest/src/cc/lib.rs.html#1909

mrkajetanp:
> To fix this, you could use `self.flags_supported`, which is checked just before compilation without creating a new `cc::Build` `

Ah interesting, I can't push to `self.flags_supported` because that'd require a `&mut self`, so I could only do that from an explicit builder function, from `Build::new()` or from `compile()`. 

I'm calling `add_inherited_rustflags` just before the snippet you linked where `is_flag_supported_inner` is called, should I not be able to just call that function myself directly instead of pushing to the Vec?

src/lib.rs
mrkajetanp: > It should also use shlex

So first pass it through shlex then split like in the snippet above? Shlex doesn't split on `\u{1f}` so running CARGO_ENCODED_RUSTFLAGS through it just ends up with a Vec of size 1 with the only element being all of CARGO_ENCODED_RUSTFLAGS.

Also, the suggested change above does not compile. If I use extend I need to mutably borrow cmd for it, so I then can't immutably borrow cmd to get cmd.family when calling rustc_to_cc_flag

src/lib.rs
mrkajetanp: Up to you really, I don't think it matters too much. I prefer the match style over a ladder of ifs but it'll compile to the same thing I'm pretty sure.

NobodyXu:
> I'm calling add_inherited_rustflags just before the snippet you linked where is_flag_supported_inner is called, should I not be able to just call that function myself directly instead of pushing to the Vec?

Yes in that case you can call `is_flag_supported_inner` directly

NobodyXu:
> and then inside the function even doing just this makes it infinite loop on creating the Tool for compiler inside is_flag_supported_inner

Sorry, can you clarify on what function is `is_flag_supported_inner` being called in that trigger infinite loop?

mrkajetanp:
> Sorry, can you clarify on what function is is_flag_supported_inner being called in that trigger infinite loop?

Yep, I'm calling `is_flag_supported_inner` inside `add_inherited_rustflags` - that's the second snippet. `add_inherited_rustflags` is being called inside `try_get_compiler` - that's the first snippet.

NobodyXu:
Sorri I made a mistake 

[`is_flag_suppirted_inner`](https://docs.rs/cc/latest/src/cc/lib.rs.html#660) actually creates another `cc::Build`

So calling it there would trigger infinite loop.

To fix that, you need to disable inherited flags detection in it:

https://docs.rs/cc/latest/src/cc/lib.rs.html#668

src/lib.rs
NobodyXu: > Shlex doesn't split on `\u{1f}` so running CARGO_ENCODED_RUSTFLAGS through it just ends up with a Vec of size 1 with the only element being all of CARGO_ENCODED_RUSTFLAGS.

Thanks, I thought it's the same as other flags, turns out I'm wrony.

> If I use extend I need to mutably borrow cmd for it, so I then can't immutably borrow cmd to get cmd.family when calling rustc_to_cc_flag

Sorry I didn't notice that.

`cmd.family` is copyable, so you can copy it out to a variable and use it.

src/lib.rs
NobodyXu: Thanks, in that case, I'm ok with the existing style.

mrkajetanp:
Ah indeed you're right, now it makes sense - many thanks! :)

src/lib.rs
```diff
@@ -2439,6 +2456,38 @@ impl Build {
...
+            })
+            .flat_map(|flag| rustc_to_cc_flag(flag, cmd.family))
+            // Filter out flags not supported by the currently used CC
+            .filter(|flag| {
+                self.is_flag_supported_inner(flag, &cmd.path, &target)
+                    .unwrap_or(false)
+            })
+            .collect();
+
+        cmd.args.append(&mut cc_flags);
```
NobodyXu, line 2487: I believe that using rust destruction will avoid the heap allocation and the lifetime error: 

```suggestion
        let Tool { family, path, args } = cmd;

        args.extend(env_os
            .to_string_lossy()
            .split("\u{1f}")
            // Strip prefixes from rustc flags
            .flat_map(|flag| {
                if flag == "-Z" || flag == "-C" {
                    None
                } else if flag.starts_with("-Z") || flag.starts_with("-C") {
                    Some(&flag[2..])
                } else {
                    Some(flag)
                }
            })
            .flat_map(|flag| rustc_to_cc_flag(flag, family))
            // Filter out flags not supported by the currently used CC
            .filter(|flag| {
                self.is_flag_supported_inner(flag, path, &target)
                    .unwrap_or(false)
            })
        );
```

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -2439,6 +2456,38 @@ impl Build {
...
+                    Some(flag)
+                }
+            })
+            .flat_map(|flag| rustc_to_cc_flag(flag, cmd.family))
+            // Filter out flags not supported by the currently used CC
+            .filter(|flag| {
+                self.is_flag_supported_inner(flag, &cmd.path, &target)
+                    .unwrap_or(false)
+            })
+            .collect();
```
madsmtm, line 2485: It would be nice to have at least a docs link here, it's unclear to me that we're correctly parsing everything?

And if we can't guarantee that we correctly parse everything, it'd be nice to have some reasoning why that is fine? And if it's not fine, then we should probably panic if we found an option that we don't understand?

madsmtm: For example, I was worried that this wouldn't catch something like `rustc "-C panic=abort"`, but it seems like, so that's a non-issue since `rustc` doesn't parse it either.

madsmtm: Another example that is _actually_ an issue: This doesn't parse `rustc --codegen panic=abort`.

But again, I'm not so worried about that specific case here, that can (and should) be fixed, I'm worried about the correctness / soundness at a high level.

src/lib.rs
```diff
@@ -2439,6 +2456,38 @@ impl Build {
...
+            .split("\u{1f}")
+            // Strip prefixes from rustc flags
+            .flat_map(|flag| {
+                if flag == "-Z" || flag == "-C" {
+                    None
+                } else if flag.starts_with("-Z") || flag.starts_with("-C") {
+                    Some(&flag[2..])
+                } else {
+                    Some(flag)
+                }
```
madsmtm, line 2477: I'm inclined to not support `-Z` flags? Or at least make it clear which ones are unstable

src/lib.rs
```diff
@@ -4221,6 +4270,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
 }
 
+// Rust and clang/cc don't agree on what equivalent flags should look like either.
+fn rustc_to_cc_flag(flag: &str, family: ToolFamily) -> Option<OsString> {
+    match family {
+        ToolFamily::Clang { .. } | ToolFamily::Gnu => match flag {
+            _ if flag.starts_with("branch-protection") => {
+                Some(format!("-m{}", flag.replace(",", "+")).into())
+            }
+            _ if flag.starts_with("code-model") => {
```
madsmtm, line 4280: I'd prefer `flag.starts_with("code-model=")` or similar, to ensure that we don't match here if a new flag like `-Ccode-model-` was added.

Actually, I'd rather prefer that the code did `flag.split_once('=')` all the way at the start.

src/lib.rs
```diff
@@ -2439,6 +2456,38 @@ impl Build {
...
+            .flat_map(|flag| {
+                if flag == "-Z" || flag == "-C" {
+                    None
+                } else if flag.starts_with("-Z") || flag.starts_with("-C") {
+                    Some(&flag[2..])
+                } else {
+                    Some(flag)
+                }
+            })
+            .flat_map(|flag| rustc_to_cc_flag(flag, cmd.family))
```
madsmtm, line 2479: I see the appeal in the simplicity of this, but I think this approach of ignoring where a flag came from is wrong, we should be much more strict.

For example, `rustc --cfg branch-protection="yes"` would translate to the flag `-mbranch-protection="yes"`, which is completely bogus.

src/lib.rs
```diff
@@ -4221,6 +4270,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
     }
 }
 
+// Rust and clang/cc don't agree on what equivalent flags should look like either.
+fn rustc_to_cc_flag(flag: &str, family: ToolFamily) -> Option<OsString> {
```
madsmtm, line 4274: Nit: I think there might be value in splitting this into two steps; a "parsing" step, and a . Something like:

```rust
struct CodegenFlags {
    code_model: &'static str,
    vectorize_loops: bool,
    // ...
}

impl CodegenFlags {
    fn parse(rustc_flags: &str) -> CodegenFlags {
        // ...
    }
    
    fn cc_flags(&self, family: ToolFamily) -> Vec<OsString> {
        // ...
    }
}
```

This would make it easier to ensure that we parse the flags equally regardless of the ToolFamily, and might make it easier to combine this information with stuff from `rustc`.

src/lib.rs
```diff
@@ -2439,6 +2456,38 @@ impl Build {
...
+                } else {
+                    Some(flag)
+                }
+            })
+            .flat_map(|flag| rustc_to_cc_flag(flag, cmd.family))
+            // Filter out flags not supported by the currently used CC
+            .filter(|flag| {
+                self.is_flag_supported_inner(flag, &cmd.path, &target)
+                    .unwrap_or(false)
+            })
```
madsmtm, line 2484: I think we may want to avoid this for several reasons:
1. Performance; I'm not sure how many flags Cargo sets by default (?), but each is an extra compiler invocation that could probably have been avoided.
2. Correctness; Setting `-Csoft-float=yes` changes the ABI, so it is (from my understanding?) _required_ for soundness that the corresponding `-msoft-float` flag is passed. If the C compiler doesn't support it, it should abort compilation, not generate something that won't work.

madsmtm (changes requested):
Thanks for starting on this 🎉!

src/lib.rs
mrkajetanp: > And if we can't guarantee that we correctly parse everything, it'd be nice to have some reasoning why that is fine?

The way I've been thinking about this is that we can pretty much only do best-effort here. There are several compilers here, all with their own changing flags so trying to 100% support everything looks like a losing battle.
At the same time, this is mainly intended to be a convenience feature. Users writing their build scripts in the long run should probably be putting the flags they want themselves explicitly either way.

The important concern is that we don't accidentally break anything. Not passing a flag is always fine since that's the status quo and the user can just do `.flag("foo")` to fix it. If we pass a flag that breaks the build that's obviously an issue so we should avoid that.

That's why I've taken the approach of not panicking and not showing any errors, I see this just as a mechanism for the obvious stuff like "-C lto needs the C code to be built with -flto". This approach also eases the maintenance burden a lot, peoples' build script will not suddenly start crashing due to unexpected or changed flags.

src/lib.rs
mrkajetanp: The main motivation for this was aarch64 branch protection, because the way BTI for instance works is that if you link with any object that doesn't have BTI then it'll get disabled for the whole binary. The branch protection flags are still unstable in upstream Rust. We could just show a warning for those?

mrkajetanp: The main motivation for this was aarch64 branch protection, because the way BTI for instance works is that if you link with any object that doesn't have BTI then it'll get disabled for the whole binary. The branch protection flags are still unstable in upstream Rust. We could just show a warning for those?

src/lib.rs
mrkajetanp: > I'd prefer flag.starts_with("code-model=") or similar, to ensure that we don't match here if a new flag like -Ccode-model- was added.

Is that really a concern we have? If so we can match with the '=' for sure, but the example you have would go against how all the rustc flags currently work so I'd be pretty surprised if someone added it.

> Actually, I'd rather prefer that the code did flag.split_once('=') all the way at the start.

Well we can't do that because some of the codegen flags don't contain a '='. Either they just don't take it or the lack of it implies a certain option.

src/lib.rs
mrkajetanp: I did also consider an approach where we merge the flag prefixes with flags and then match on flags like `-Clto` and `-Zbranch-protection`. The downside there is that e.g. with branch protection, once the flag gets stabilised it'll stop matching. I guess we could get around it by matching it on either -Z or -C. Should I switch back to the more strict approach then?

src/lib.rs
madsmtm: Fair, maybe I'm just worried about these flags being treated the same by `cc` regardless of whether they're stable or not. Having different functions that parse them like `rustc_unstable_codegen_to_cflag` and `rustc_codegen_to_cflag` would appease a lot of my worries here.

src/lib.rs
mrkajetanp: good idea, sure that works :)

src/lib.rs
mrkajetanp: > I'm not sure how many flags Cargo sets by default

`CARGO_ENCODED_RUSTFLAGS` is empty by default so those will all be user-provided flags, meaning not too many in practice. I think the benefit of knowing whether the used compiler supports the flag outweighs the slight performance overhead here?

> If the C compiler doesn't support it, it should abort compilation, not generate something that won't work.

I see the point, yeah. Maybe we could just generate a warning if we find out a flag isn't supported? Usually these kinds of flags come to clang and gcc before they come to Rust so this filtering here is mainly here so that the code isn't too picky with compiler versions. Otherwise I could refactor to check if the flag is supported within the big match statement so that we can decide whether to skip, warn or abort on a per-flag basis.

src/lib.rs
mrkajetanp: Right, so just do two passes where one matches `-C` flags and the other one matches `-Z` flags?

src/lib.rs
mrkajetanp: Hm that's true, yeah I can make the matching more strict, sure :)

src/lib.rs
```diff
@@ -4221,6 +4270,161 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+                        "pie" => "-fPIE",
+                        "dynamic-no-pic" => "-mdynamic-no-pic",
+                        _ => return None,
+                    };
+                    Some(cc_flag.into())
+                } else {
+                    None
+                }
+            }
+            _ if flag.starts_with("soft-float") => {
```
madsmtm, line 4377: This was recently deprecated btw:
https://github.com/rust-lang/rust/issues/129893

src/lib.rs
madsmtm: I think I'd prefer that, yes

src/lib.rs
mrkajetanp: Ah yes you're right. Good question, on one hand it's deprecated but on the other hand if someone does pass it to rustc then they *really* need to make sure the C code is compiled with the equivalent flag too so there might be value in keeping it. I don't really mind either way so it's up to you, should I drop it?

src/lib.rs
mrkajetanp: At a second glance, all fields of this struct would need to be Options because we need to distinguish between e.g. true/false/unset for a given flag. It's probably not a huge deal though, should I still do it?

src/lib.rs
NobodyXu: IMO we should support both the new `-Ctarget-feature=+soft-float` and this deprecated one, otherwise the binary generated will be broken and it'd be very very hard to track it down.

src/lib.rs
NobodyXu: Yeah I think that's ok

src/lib.rs
madsmtm: Yeah, let's keep it, but perhaps with a comment that it's deprecated (and why we still choose to keep it).

src/lib.rs
```diff
@@ -4221,6 +4255,291 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
     }
 }
 
+#[derive(Debug)]
+struct RustcCodegenFlags {
```
madsmtm, line 4259: Nit: Put this in a separate file (`lib.rs` is far too big already IMO)

src/lib.rs
```diff
@@ -4221,6 +4255,291 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+        codegen_flags
+    }
+
+    fn set_rustc_flag(&mut self, flag: &str) {
+        if flag.starts_with("-Z") {
+            self.set_unstable_rustc_flag(flag);
+        } else {
+            self.set_stable_rustc_flag(flag);
+        }
+    }
```
madsmtm, line 4338: Nit: I'd be fine with combining this into one function now that `-C` and `-Z` are part of the flag prefix.

src/lib.rs
mrkajetanp: If someone uses the flag they'll already get the deprecation warning from rustc, but we can add a comment in the code as well I suppose

src/lib.rs
```diff
@@ -4221,6 +4255,291 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+    embed_bitcode: Option<bool>,
+    force_frame_pointers: Option<bool>,
+    link_dead_code: Option<bool>,
+    no_redzone: Option<bool>,
+    soft_float: Option<bool>,
+}
+
+impl RustcCodegenFlags {
+    // Parse flags obtained from CARGO_ENCODED_RUSTFLAGS
+    fn parse(rustflags_env: &str) -> RustcCodegenFlags {
```
madsmtm, line 4278: Nit: A few unit tests of this would be nice. Something like:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[track_caller]
    fn check(env: &str, expected: &RustcCodegenFlags) {
        let actual = RustcCodegenFlags::parse(env);
        assert_eq!(actual, *expected);
    }

    #[test]
    fn codegen_type() {
        let expected = RustcCodegenFlags {
            code_model: Some("tiny".into()),
            ..RustcCodegenFlags::default()
        };
        check("-Ccode-model=tiny", &expected);
        check("-C\u{f1}code-model=tiny", &expected);
        check("--codegen\u{f1}code-model=tiny", &expected);
        check("--codegen=code-model=tiny", &expected);
    }

    #[test]
    fn precedence() {
        check("-Ccode-model=tiny\u{1f}-Ccode-model=small", RustcCodegenFlags {
            code_model: Some("small".into()),
            ..RustcCodegenFlags::default()
        });
    }

    // ...
}
```

src/lib.rs
```diff
@@ -4221,6 +4255,291 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+            "-Cno-vectorize-slp" => self.no_vectorize_slp = true,
+            "-Cprofile-generate" => self.profile_generate = value,
+            "-Cprofile-use" => self.profile_use = value,
+            "-Ccontrol-flow-guard" => self.control_flow_guard = value.or(Some("true".into())),
+            "-Clto" => self.lto = value.or(Some("true".into())),
+            "-Crelocation-model" => self.relocation_model = value,
+            "-Cembed-bitcode" => {
+                self.embed_bitcode = value.map_or(Some(true), |val| match val.as_str() {
+                    "y" | "yes" | "on" | "true" => Some(true),
+                    "n" | "no" | "off" | "false" => Some(false),
```
madsmtm, line 4359: Nit: Perhaps factor this into a separate function (or something)?

src/lib.rs
```diff
@@ -4221,6 +4255,291 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+
+    fn set_stable_rustc_flag(&mut self, flag: &str) {
+        let (flag, value) = if let Some((flag, value)) = flag.split_once('=') {
+            (flag, Some(value.to_owned()))
+        } else {
+            (flag, None)
+        };
+
+        match flag {
+            "-Ccode-model" => self.code_model = value,
```
madsmtm, line 4348: Nit: I'd prefer a link to the relevant option at each line. Something like:
```suggestion
            // https://doc.rust-lang.org/rustc/codegen-options/#code-model
            "-Ccode-model" => self.code_model = value,
```

src/lib.rs
```diff
@@ -4221,6 +4255,291 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+
+        let mut flags: Vec<OsString> = vec![];
+
+        match family {
+            ToolFamily::Clang { .. } | ToolFamily::Gnu => {
+                if let Some(value) = &self.branch_protection {
+                    push_if_supported(
+                        &mut flags,
+                        format!("-mbranch-protection={}", value.replace(",", "+")).into(),
+                    );
```
madsmtm, line 4438: Nice to have all of these! Now we can (in a follow-up to this PR) selectively disable it when we know that all the compilers we care about support the flag.

src/lib.rs
```diff
@@ -4221,6 +4255,291 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+            (flag, Some(value.to_owned()))
+        } else {
+            (flag, None)
+        };
+
+        match flag {
+            "-Ccode-model" => self.code_model = value,
+            "-Cno-vectorize-loops" => self.no_vectorize_loops = true,
+            "-Cno-vectorize-slp" => self.no_vectorize_slp = true,
+            "-Cprofile-generate" => self.profile_generate = value,
```
madsmtm, line 4351: Nit: I think I'd prefer a style more like:
```suggestion
            "-Cprofile-generate" => self.profile_generate = Some(value.expect("-Cprofile-generate must have a value")),
```

That way, we know that a bare `-Cprofile-generate` wasn't specified (yes, this wouldn't be accepted by `rustc`, but we'd enter a wrong state if it was accepted in the future).

madsmtm (approved):
Looks a lot better to me, thanks for doing the work! Only nits from my side now.

src/lib.rs
madsmtm: Maybe also a check for every single flag that `rustc` has:
```rust
let flags = [
    "--cfg",
    "a",
    "-L",
    "/usr/lib/foo",
    "-l",
    "static:+whole-archive=mylib",
    "--crate-type=dylib",
    // ...
    "-g",
    "-O",
    // ...
    "-Car",
    "-Ccode-model=tiny",
    "-Ccodegen-units=1",
    "-Ccollapse-macro-debuginfo=yes",
    // ...
];
check(flags.join('\u{f1}'), ...);
```

mrkajetanp: Okay that's a hilarious edge-case, nice one. I think if we add -L as a prefix then this example would get joined to `-L-Clto` and then ignored because it doesn't match anything which would get the behaviour we want. How does that sound?

src/lib.rs
```diff
@@ -4221,6 +4255,291 @@ fn map_darwin_target_from_rust_to_compiler_architecture(target: &Target) -> &str
...
+
+    fn set_unstable_rustc_flag(&mut self, flag: &str) {
+        let (flag, value) = if let Some((flag, value)) = flag.split_once('=') {
+            (flag, Some(value.to_owned()))
+        } else {
+            (flag, None)
+        };
+
+        match flag {
+            "-Zbranch-protection" => self.branch_protection = value,
```
mrkajetanp, line 4403: Actually, how about we match this one on `"-Zbranch-protection" | "-Cbranch-protection"` straight away? That way the crate won't need to be updated when the option gets stabilised

NobodyXu: Yeah if you merge the unstable/stable flags, then I think matching it like this makes sense

src/lib.rs
madsmtm: Great idea!

I'd also prefer it for a lot of the other short flags (`-l`, `-o`, `-W`, `-A`, `-D`, `-F`), just be sure.

mrkajetanp:
Should be good now I think :)

src/flags.rs
```diff
@@ -0,0 +1,478 @@
...
+    }
+
+    // Rust and clang/cc don't agree on what equivalent flags should look like.
+    pub fn cc_flags(
+        &self,
+        build: &Build,
+        path: &PathBuf,
+        family: &ToolFamily,
+        target: &Target,
+    ) -> Vec<OsString> {
```
NobodyXu, line 165:
```suggestion
    pub fn cc_flags(
        &self,
        build: &Build,
        path: &PathBuf,
        family: ToolFamily,
        target: &Target,
        args: &mut Vec<OsString>,
    ) {
```

How about passing args directly so that we can push directly into it, and pass family as value since it's copyable and very small (much smaller than pointer)?

Alternatively, we can pass the `tool: &mut Tool` here, whichever is more convenient.

src/flags.rs
```diff
@@ -0,0 +1,478 @@
...
+        let push_if_supported = |flags: &mut Vec<OsString>, flag: OsString| {
```
NobodyXu, line 166:
```suggestion
        let push_if_supported = |flags: &mut Vec<OsString>, flag: String| {
```


Given that all usages of this function creates `String`, it would be more ergonomic to pass it to the function than doing additional `.into()`

NobodyXu (changes requested):
Thanks!

I am happy about the direction and general code, just a few questions regarding panic, ergonomic and some simple optimization

src/flags.rs
```diff
@@ -0,0 +1,478 @@
...
+        let (flag, value) = if let Some((flag, value)) = flag.split_once('=') {
+            (flag, Some(value.to_owned()))
+        } else {
+            (flag, None)
+        };
+
+        match flag {
+            // https://doc.rust-lang.org/rustc/codegen-options/index.html#code-model
+            "-Ccode-model" => {
+                self.code_model = Some(value.expect("-Ccode-model must have a value"));
```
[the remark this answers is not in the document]
mrkajetanp: @madsmtm specifically asked me add this so maybe he should weigh in here? I think CARGO_ENCODED_RUSTFLAGS is only supposed to be set by cargo so this should be fine. If a user manually sets this env var to something that wouldn't be accepted through RUSTFLAGS then they're breaking the assumptions cargo makes anyway.

src/flags.rs
```diff
@@ -0,0 +1,478 @@
...
+
+    fn set_rustc_flag(&mut self, flag: &str) {
+        // Convert a textual representation of a bool-like rustc flag argument into an actual bool
+        fn arg_to_bool(arg: &str) -> Option<bool> {
+            match arg {
+                "y" | "yes" | "on" | "true" => Some(true),
+                "n" | "no" | "off" | "false" => Some(false),
+                _ => None,
+            }
+        }
```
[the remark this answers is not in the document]
mrkajetanp: How about just this then? Not much point making it into two functions since it's only used here to avoid code duplication
```
        fn arg_to_bool(arg: impl AsRef<str>) -> Option<bool> {
            match arg.as_ref() {
                "y" | "yes" | "on" | "true" => Some(true),
                "n" | "no" | "off" | "false" => Some(false),
                _ => None,
            }
        }
```

src/flags.rs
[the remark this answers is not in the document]
madsmtm: Yeah, I requested it in https://github.com/rust-lang/cc-rs/pull/1279#discussion_r1847166403, copying my reasoning there:

> This way, we know that a bare `-Ccode-model` wasn't specified (yes, this wouldn't be accepted by rustc, but we'd enter a wrong state if it was accepted in the future).

I'd be fine with a warning, ignoring the flag or returning an error, I just don't really want us to assume that it's something it's not.

src/flags.rs
mrkajetanp: The into() will be done anyway because it needs to be an OsString for `is_flag_supported` and to actually push it to the vector. Should I just move the into()s inside the helper function then?

mrkajetanp: It'd have to be something along the lines of
```
        let mut push_if_supported = |flag: String| {
            let flag = OsString::from(flag);
            if build
                .is_flag_supported_inner(&flag, path, target)
                .unwrap_or(false)
            {
                flags.push(flag);
            } else {
                build.cargo_output.print_warning(&format!(
                    "Inherited flag {:?} is not supported by the currently used CC",
                    flag
                ));
            }
        };
```

madsmtm (approved):

src/flags.rs
```diff
@@ -0,0 +1,478 @@
...
+            let curr = env_flags[i];
+            let prev = if i > 0 { env_flags[i - 1] } else { "" };
+
+            // Do not process prefixes on their own
+            if !is_flag_prefix(curr) {
+                // Concat flags preceded by a prefix
+                let rustc_flag = handle_flag_prefix(prev, curr);
+                codegen_flags.set_rustc_flag(&rustc_flag);
+            }
+        }
```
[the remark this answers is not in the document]
mrkajetanp: Still parsed correctly :) It'll go like:
`prev: -L, curr: -L, flag: -L-L => ignore`
`curr: -L, curr: -Clto, flag: -L-Clto => ignore`
`prev: -Clto, curr: <foo>, flag: <foo>`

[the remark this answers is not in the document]
mrkajetanp: Oh no wait, is the correct behaviour here supposed to be that we do accept -Clto?

src/flags.rs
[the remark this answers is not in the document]
mrkajetanp: Returning an error could work, how about this?
```
        fn flag_ok_or(flag: Option<String>, msg: &'static str) -> Result<String, Error> {
            flag.ok_or(Error::new(ErrorKind::InvalidArgument, msg))
        }

        match flag {
            // https://doc.rust-lang.org/rustc/codegen-options/index.html#code-model
            "-Ccode-model" => {
                self.code_model = Some(flag_ok_or(value, "-Ccode-model must have a value")?);
            }
```
I can add a new ErrorKind too if we want to distinguish them

src/flags.rs
mrkajetanp: Or just make push_if_supported into an inner function instead of a closure

src/flags.rs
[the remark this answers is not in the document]
mrkajetanp: Pain, oh well your suggestion fixes it, will do. Thanks!

src/flags.rs
mrkajetanp: Some of the flags are &str as well so if we start taking String it'll be &str -> String -> OsString for those instead of &str -> OsString directly as it is now. For the ones that use format!() and do create a String it'd just be moving the call converting them to OsString. I can do it if you'd like me to but I think this change adds some overhead for certain flags.

src/flags.rs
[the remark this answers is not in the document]
NobodyXu: Yeah I think a new ErrorKind makes sense

src/flags.rs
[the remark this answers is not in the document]
NobodyXu: Yes that's fine to me, the reason I create two function is to avoid duplicate code monomorphisation, but it probably won't make much difference in real life so go for the simpler one please.

src/flags.rs
NobodyXu: I'm ok with both honestly, one potential problem with removing the param, is that you can't push to the `args` directly and use the function at the same time.

If that doesn't happen, then yeah removing the param seems to be cleaner.

src/flags.rs
NobodyXu: Maybe we can use `impl Into<OsString>` instead?

That'd work with `&str`, `String` or `OsString`, and no extra allocation.

mrkajetanp:
```
rustc E0562: `impl Trait` is now allowed in closure parameters
```
Well that's annoying. If we make `push_if_supported` into a normal function to get around this then we'd need to explicitly pass everything it needs and the calls would end up looking like
```
push_if_supported("-fno-vectorize", flags, build, path, target)
```
Which is even less ergonomic than the `.into()s` we have there at the moment.

NobodyXu: E0562 seems to be about returning impl Traits from closure?

https://doc.rust-lang.org/error_codes/E0562.html

I was thinking about

```rust
        let mut push_if_supported = |flag: impl Into<OsString>| {
            let flag = OsString::from(flag);
            if build
                .is_flag_supported_inner(&flag, path, target)
                .unwrap_or(false)
            {
                flags.push(flag);
            } else {
                build.cargo_output.print_warning(&format!(
                    "Inherited flag {:?} is not supported by the currently used CC",
                    flag
                ));
            }
        };
```

NobodyXu: Thanks, in that case maybe we should keep the existing code.

I know cursed way to work around this (create a struct and put the fn as associated fn), but it's simply not worth the effort here.

mrkajetanp: yeah I agree - all the other options are unfortunately even more messy.

src/lib.rs
```diff
@@ -2365,6 +2386,25 @@ impl Build {
...
+
+        let Tool {
+            family, path, args, ..
+        } = cmd;
+
+        let env = env_os.to_string_lossy();
+        let codegen_flags = RustcCodegenFlags::parse(&env)?;
+        let mut cc_flags_args: Vec<OsString> = vec![];
+        codegen_flags.cc_flags(self, path, *family, target, &mut cc_flags_args);
+        args.extend(cc_flags_args);
```
NobodyXu, line 2404: Hmmm is there any lifetime issue you've encountered ?

```suggestion
        codegen_flags.cc_flags(self, path, *family, target, &mut args);
```

NobodyXu (changes requested):
Thank you!

This is the last review pass from me, and it mostly contains code suggestions that can be applied easily.

Once resolved I will merge it.

src/flags.rs
```diff
@@ -0,0 +1,487 @@
...
+                    | "-o"
+                    | "-W"
+                    | "--warn"
+                    | "-A"
+                    | "--allow"
+                    | "-D"
+                    | "--deny"
+                    | "-F"
+                    | "--forbid"
+            )
```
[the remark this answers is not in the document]
mrkajetanp: The macro just expands to a match statement, is array lookup preferable to that? This was a direct clippy suggestion.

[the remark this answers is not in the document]
NobodyXu: Hmmm, I think using an array is cleaner than match.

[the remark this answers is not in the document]
mrkajetanp: Hmm sure, I'm fine with it either way I suppose.

src/flags.rs
```diff
@@ -0,0 +1,487 @@
...
+                "-D" | "--deny" => Cow::from(format!("-D{}", curr)),
+                "-F" | "--forbid" => Cow::from(format!("-F{}", curr)),
+                _ => Cow::from(curr),
+            }
+        }
+
+        let mut codegen_flags = Self::default();
+
+        let mut prev_prefix = None;
+        for curr in rustflags_env.split("\u{1f}") {
```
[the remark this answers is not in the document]
mrkajetanp: By the time this code runs the flags have already been validated by rustc and rustc doesn't accept "empty" flags afaik

src/lib.rs
mrkajetanp: ah yes this can now be shortened, sure why not

src/flags.rs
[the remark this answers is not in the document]
NobodyXu: Thanks I am happy with this then

src/flags.rs
```diff
@@ -0,0 +1,482 @@
...
+                "-o",
+                "-W",
+                "--warn",
+                "-A",
+                "--allow",
+                "-D",
+                "--deny",
+                "-F",
+                "--forbid",
+            ].contains(flag)
```
NobodyXu, line 44:
```suggestion
            ].contains(&flag)
```

src/lib.rs
```diff
@@ -2365,6 +2386,23 @@ impl Build {
...
+            None => return Ok(()),
+        };
+
+        let Tool {
+            family, path, args, ..
+        } = cmd;
+
+        let env = env_os.to_string_lossy();
+        let codegen_flags = RustcCodegenFlags::parse(&env)?;
+        codegen_flags.cc_flags(self, path, *family, target, &mut args);
```
NobodyXu, line 2402:
```suggestion
        codegen_flags.cc_flags(self, path, *family, target, args);
```

NobodyXu (changes requested):
This should fix the pipeline

NobodyXu (approved):
Thank you!

Issue #1275, opened 2024-11-06
Regression when updating from version `1.1.31` to newer when compiling `ring` to `wasm32-unknown-unknown`


matheus23:
Hi, when updating from cc version 1.1.31 to cc version 1.1.32, I fail to build ring for wasm32-unknown-unknown in github actions ([see this PR](https://github.com/n0-computer/quinn/pull/18) with me testing different `cc` versions).

Specifically, ring generates "env" imports in its Wasm file:
```scheme
  (import "env" "ring_core_0_17_8_x25519_fe_neg" (func $ring_core_0_17_8_x25519_fe_neg (;31;) (type 16)))
  (import "env" "ring_core_0_17_8_x25519_ge_scalarmult_base" (func $ring_core_0_17_8_x25519_ge_scalarmult_base (;32;) (type 3)))
  (import "env" "ring_core_0_17_8_x25519_ge_frombytes_vartime" (func $ring_core_0_17_8_x25519_ge_frombytes_vartime (;33;) (type 4)))
  (import "env" "ring_core_0_17_8_x25519_fe_invert" (func $ring_core_0_17_8_x25519_fe_invert (;34;) (type 2)))
  (import "env" "ring_core_0_17_8_x25519_fe_mul_ttt" (func $ring_core_0_17_8_x25519_fe_mul_ttt (;35;) (type 3)))
```
This indicates that there's some `extern "C"` in `ring` that didn't get picked up by `wasm-bindgen` for some reason and made it to the final `.wasm`.

I'd expect there to be no such imports at all. Having these imports means it's essentially not executable by Wasm hosts such as nodejs or browsers (because they don't provide an "env" module).

Downgrading `cc` from `1.1.32` to `1.1.31` fixes these issues.

Of course, this sounds a lot like a `ring`, issue, but given that this occurs when updating `cc`s *patch* version, I would've expected that `ring` doesn't need to change anything to keep working, right?

I'm also out of my depth here. I'd love to help debug this, but it works on my machine, just not on github actions.

NobodyXu:
cc @madsmtm could that due to the wasm32-unknown-unknown change in the recent PR?

The one where we have to replace `target.os` with abi

madsmtm:
I don't know what's the correct thing to do, see https://github.com/rust-lang/cc-rs/issues/1126#issuecomment-2460394021

PR #1284, opened 2024-11-10
Fix WASM vs. WASI options

madsmtm:
I went through all the places where we mention WASI and WASM, and tried to make sense of which places need which options. See also https://github.com/rust-lang/cc-rs/issues/1126#issuecomment-2460394021. Still with the caveat that I have no idea if any of what I'm doing here is correct.

The effect to the options for `wasm32-unknown-unknown` is that we no longer pass `-fPIC`, and that we now pass `-fno-exceptions`.

Fixes https://github.com/rust-lang/cc-rs/issues/1126.
Fixes https://github.com/rust-lang/cc-rs/issues/1275.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3103,7 +3103,7 @@ impl Build {
                         name = format!("em{}", tool).into();
                         Some(self.cmd(&name))
                     }
-                } else if target.arch == "wasm32" {
+                } else if target.arch == "wasm32" || target.arch == "wasm64" {
                     // Formally speaking one should be able to use this approach,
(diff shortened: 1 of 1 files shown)


matheus23:
Yeah, can confirm this [fixes the compilation](https://github.com/n0-computer/quinn/actions/runs/11774644144/job/32793580927) compared to [latest released `cc` version](https://github.com/n0-computer/quinn/actions/runs/11774613909/job/32793494962).

NobodyXu (approved):
Thanks!

Issue #1265, opened 2024-11-03
windows_registry: add api which accepts just the target arch


bjorn3:
Currently rustc has to pass the actual value of the `--target` argument to `cc::windows_registry::find_tool`, making it one of the only places in rustc where the name of the target matters rather than the contents of the target. This also for example means that if you take the target spec of x86_64-pc-windows-msvc and put it in a file called aarch64-foo.json, rustc will incorrectly try to find a linker for aarch64. Or more realistically if you name it foo.json, rustc will not know which target to find a linker for even though the target spec itself contains the architecture to compile for.

madsmtm:
Perhaps we could also change it to take _either_ just the arch, or a full target triple? And then deprecate the former?

NobodyXu:
Maybe we should split that part into a separate crate?

I feel like it is not exactly part of cc, just put it in cc for convenience, and sometimes we might want to change its API.

Issue #1281, opened 2024-11-08
Regression in 1.1.32 in `cc::windows_registry::find_tool`


slonopotamus:
Way to reproduce

1. Windows
2. Installed Visual Studio 2022 with C++ workload
3. Attempt to find path to compiler executable using `let tool = cc::windows_registry::find_tool("x86_64-msvc", "cl.exe").unwrap();`

cc-1.1.31 and older work.
cc-1.1.32 and newer panic because `find_tool` returns `None`.

Can both repro on my own machine and on GitHub Actions.

Maybe I am using `find_tool` the wrong way and need to pass different arguments?

NobodyXu:
x86_64-msvc is not a valid rust target.

I'd recommend changing it to x86_64-pc-windows-msvc

slonopotamus:
Thanks, using `x86_64-pc-windows-msvc` fixed things for me

PR #1285, opened 2024-11-11
windows_registry: Allow only specifying the architecture

madsmtm:
Fixes https://github.com/rust-lang/cc-rs/issues/1265 by allowing either an architecture or a full target name to be specified.

Would have avoided the regression from https://github.com/rust-lang/cc-rs/pull/1225 in https://github.com/rust-lang/cc-rs/issues/1281.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3874,7 +3874,11 @@ impl Build {
             }
         }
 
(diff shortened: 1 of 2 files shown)


NobodyXu (approved):
Thanks LGTM

PR #1292, opened 2024-11-14
Allow disabling cc's ability to compile via env var CC_FORCE_DISABLE

NobodyXu:
Replacement for #1226

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,6 +89,13 @@
 //!   For example, with `CFLAGS='a "b c"'`, the compiler will be invoked with 2 arguments -
 //!   `a` and `b c` - rather than 3: `a`, `"b` and `c"`.
 //! * `CXX...` - see [C++ Support](#c-support).
+//! * `CC_FORCE_DISABLE` - If set, `cc` will never run any [`Command`]s, and methods that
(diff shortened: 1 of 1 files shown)


thomcc (approved):

src/lib.rs
```diff
@@ -4023,6 +4039,51 @@ impl AsmFileExt {
...
+    let val = CACHE.load(Relaxed);
+    // We manually cache the environment var, since we need it in some places
+    // where we don't have access to a `Build` instance.
+    #[allow(clippy::disallowed_methods)]
+    fn compute_is_disabled() -> bool {
+        match std::env::var_os("CC_FORCE_DISABLE") {
+            // Not set? Not disabled.
+            None => false,
+            // Respect `CC_FORCE_DISABLE=0` and some simple synonyms.
+            Some(v) if &*v != "0" && &*v != "false" && &*v != "no" => false,
```
saleemrashid, line 4055: This check appears to be the wrong way around (i.e. `CC_FORCE_DISABLE=0` returns `true` and `CC_FORCE_DISABLE=1` returns `false`).

madsmtm: Thanks, I've opened https://github.com/rust-lang/cc-rs/pull/1371

Issue #1297, opened 2024-11-21
Cross-compilation to Yocto breaks after #1225 (1.1.32)


tronical:
I noticed that after #1225 as part of 1.1.32, cross-compilation of the Rust Skia bindings fails when targeting a Yocto  environment (say Poky, but could be any distro name / vendor name).

Slightly more concretely, the skia build uses cc to compile some `.cpp` bindings code and calls `cc_build.target(target_str)` eventually, where `target_str` is `arm-poky-linux-gnueabi`. That's technically the correct target for the gcc Yocto configured/built. The rust target on the other hand is `armv7-unknown-linux-gnueabihf`.

Should the rust skia bindings build avoid calling `set_target()`, now that cc also tries to figure out the target from the cargo environment? I'm slightly hesitant here as the rust target is not the same as the one gcc is configured for. Or should cc be extended to parse this triplet?

madsmtm:
[`Build::target`](https://docs.rs/cc/latest/cc/struct.Build.html#method.target) is the `rustc` target triple (and always has been), so if that's `armv7-unknown-linux-gnueabihf`, then that's what you should set (or yes, ideally leave it out if in a build script)

I'm not familiar with neither Yocto nor Poky, but if the above is not enough to get it to work, then we can add special support for this target (or, ideally, add upstream support in `rustc`).

nekopsykose:
that doesn't really seem correct though? the comment says:  

```
    /// This will fail when using a custom target triple unknown to `rustc`.
    fn from_str(target_triple: &str) -> Result<Self, Error> {
        if let Ok(index) =
            generated::LIST.binary_search_by_key(&target_triple, |(target_triple, _)| target_triple)
        {
            let (_, info) = &generated::LIST[index];
            Ok(info.clone())
        } else {
            Err(Error::new(
                ErrorKind::InvalidTarget,
                format!("unknown target `{target_triple}`"),
            ))
        }
    }
```

in my case:

```
$ rustc --print target-list
...
x86_64-chimera-linux-musl
...
```

i.e. it is definitely a target rustc knows (the distro rust is aware of the custom vendor triples), but yet:

```
$ hx -g build
Building 206 grammars


error occurred: unknown target `x86_64-chimera-linux-musl`
```

(that calls `.host(BUILD_TARGET)` which is set from `cargo:rustc-env=BUILD_TARGET=std::env::var("TARGET")` in `build.rs`)

[...]

PR #1299, opened 2024-11-22
Fix: Fetch target info from Cargo even if `Build::target` is manually set

madsmtm:
In the first commit, I've documented more closely what `Build::target` does, which should avoid confusion like in https://github.com/rust-lang/cc-rs/issues/1297.

And then it occurs to me that some users might be doing `build.target(&std::env::var("TARGET").unwrap())` (even though `Build::target` has been available since `cc v1.0.0`), which would give different behaviour than if they didn't do that (since we no longer source from Cargo). So I've added a hacky fix for this case.

We might want to take it even further, and warn if both `TARGET` and `Build::target` are set?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1088,10 +1088,16 @@ impl Build {
         self
     }
 
-    /// Configures the target this configuration will be compiling for.
+    /// Configures the `rustc` target this configuration will be compiling
+    /// for.
     ///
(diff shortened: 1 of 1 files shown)


NobodyXu (approved):
Thanks LGTM!

> We might want to take it even further, and warn if both `TARGET` and `Build::target` are set?

Hmm maybe some users do have a valid use case for this, so let's skip it for now.

PR #1302, opened 2024-11-22
CI: Avoid running scheduled jobs on forks

mati865:
This replicates https://github.com/rust-lang/rust/pull/113029

Such jobs are going to fail anyway: https://github.com/mati865/cc-rs/actions/runs/11965764607/job/33360331758


NobodyXu (approved):
Thank you!

PR #1303, opened 2024-11-24
Fix msvc stdout not shown on error

### LINKED ISSUE #1260: How to get a more friendly error message in cc?
I'm looking for a way to get more user-friendly error messages to help diagnose issues in my C/C++ code or the bridging code between Rust and C. How can I achieve this?

Here's a simple example of what I'm facing:

Suppose you have a C file intended for use in Rust. You would typically compile it in build.rs using the cc crate. If the C file is correct, everything works fine. However, in the example below, the error message is quite confusing.
Image
Image

You can see that cc simply reports "error occurred" without providing further details. In contrast, compiling the same file with gcc highlights the specific error line.Is there a way to configure the cc crate to provide detailed gcc or msvc error messages when there are issues in my C/C++ code? How can I set this up?

NobodyXu:
Fixed #1260

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1785,8 +1785,6 @@ impl Build {
             self.fix_env_for_apple_os(&mut cmd)?;
         }
 
-        disable_localization(&mut cmd);
-
         Ok((cmd, name))
     }
 
@@ -1951,6 +1949,8 @@ impl Build {
             let warnings_to_errors_flag = cmd.family.warnings_to_errors_flag().into();
             cmd.push_cc_arg(warnings_to_errors_flag);
         }
+
+       disable_localization(&mut cmd);
 
         Ok(cmd)
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1784,9 +1784,8 @@ impl Build {
         if cfg!(target_os = "macos") {
             self.fix_env_for_apple_os(&mut cmd)?;
         }
-        if msvc {
-            disable_localization(&mut cmd);
-        }
+
+        disable_localization(&mut cmd);
 
         Ok((cmd, name))
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1958,7 +1958,7 @@ impl Build {
         // This should be acceptable because other messages from rustc are in English anyway,
         // and may also be desirable to improve searchability of the linker diagnostics.
         if matches!(cmd.family, ToolFamily::Msvc { clang_cl: false }) {
-            cmd.env.push(("VSLANG".into(), "1033".into())); 
+            cmd.env.push(("VSLANG".into(), "1033".into()));
         } else {
             cmd.env.push(("LC_ALL".into(), "C".into()));
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1957,14 +1957,10 @@ impl Build {
         // and prevent inspection of msvc output in case of errors, which we occasionally do.
         // This should be acceptable because other messages from rustc are in English anyway,
         // and may also be desirable to improve searchability of the linker diagnostics.
-       
-        // No harm in setting both env vars simultaneously.
-        // Unix-style linkers.
         if matches!(cmd.family, ToolFamily::Msvc { clang_cl: false }) {
-            cmd.env.push(("LC_ALL".into(), "C".into()));
+            cmd.env.push(("VSLANG".into(), "1033".into())); 
         } else {
-            // MSVC's `link.exe`.
-            cmd.env.push(("VSLANG".into(), "1033".into()));
+            cmd.env.push(("LC_ALL".into(), "C".into()));
         }
 
         Ok(cmd)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1784,6 +1784,9 @@ impl Build {
         if cfg!(target_os = "macos") {
             self.fix_env_for_apple_os(&mut cmd)?;
         }
+        if msvc {
+            disable_localization(&mut cmd);
+        }
 
         Ok((cmd, name))
     }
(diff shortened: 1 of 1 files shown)


madsmtm (commented):
Why do this in `create_compile_object_cmd` (and not more generally in `try_get_compiler`)?

NobodyXu:
Hmmm I didn't think of this at all!

Looking at it closer, I think there are two reasons not to do that:
 - `try_get_compiler` is a public function, changing it might break someone else's workflow, i.e. what if user actually expects non-English output?
 - `try_get_compiler` only adds args not env, while `create_compile_object_cmds` does add env (for apple only though)

madsmtm:
> what if user actually expects non-English output

If we consider changing the shown language a breaking change, then changing it in `create_compile_object_cmd` is also a breaking change, no?

> `try_get_compiler` only adds args not env, while `create_compile_object_cmds` does add env (for apple only though)

I doubt there's a thought-out reason for that? In any case, I'm pretty sure that workaround can be removed nowadays (the problem was using the general `*-apple-darwin` target instead of the specific `*-apple-macosx` target).

> Upon further checking `try_get_compiler` is used in `try_expand` and `is_flag_supported_inner`, which checks for exit code and stderr output of the compiler https://docs.rs/cc/latest/src/cc/lib.rs.html#713 , add env that changes stdout/stderr might break it

Those all seem positive to force being English (provided we decide to force English at all)?

src/lib.rs
```diff
@@ -1784,6 +1784,9 @@ impl Build {
         if cfg!(target_os = "macos") {
             self.fix_env_for_apple_os(&mut cmd)?;
         }
+        if msvc {
```
madsmtm, line 1787: Why only when linking for MSVC? Seems like we'd want to do this everywhere, if we wanted to do it in the first place?

NobodyXu: Thanks updated.

NobodyXu:
> If we consider changing the shown language a breaking change, then changing it in `create_compile_object_cmd` is also a breaking change, no?

yes that's true.

> Those all seem positive to force being English (provided we decide to force English at all)?

Yeah, thanks, I've changed it to be set in try_get_compiler`

src/lib.rs
```diff
@@ -4094,6 +4096,21 @@ fn check_disabled() -> Result<(), Error> {
...
+/// Such messages may cause issues with text encoding on Windows
+/// and prevent inspection of msvc output in case of errors, which we occasionally do.
+/// This should be acceptable because other messages from rustc are in English anyway,
+/// and may also be desirable to improve searchability of the linker diagnostics.
+fn disable_localization(cmd: &mut Command) {
+    // No harm in setting both env vars simultaneously.
+    // Unix-style linkers.
+    cmd.env("LC_ALL", "C");
+    // MSVC's `link.exe`.
+    cmd.env("VSLANG", "1033");
```
madsmtm, line 4111: True that there's no _harm_, but it does clutter up the output with unnecessary env vars. How about this instead?
```suggestion
    if cmd.family == Msvc { clang_cl: false } {
        cmd.env("VSLANG", "1033");
    } else {
        cmd.env("LC_ALL", "C");
    }
```

src/lib.rs
```diff
@@ -1950,6 +1950,8 @@ impl Build {
             cmd.push_cc_arg(warnings_to_errors_flag);
         }
 
+       disable_localization(&mut cmd);
```
madsmtm, line 1953: Just so that we've considered it: Should this be a "default" flag (should it be added in `add_default_flags`)?

My own opinion is no, since `CRATE_CC_NO_DEFAULTS` seems more targeted towards making cross-compilation work.

NobodyXu: I think since it doesn't have any impact on the artifact generated, it shouldn't be considered as a flag

src/lib.rs
NobodyXu: Thanks, updated

src/lib.rs
```diff
@@ -1950,7 +1950,22 @@ impl Build {
...
+        // This should be acceptable because other messages from rustc are in English anyway,
+        // and may also be desirable to improve searchability of the linker diagnostics.
+       
+        // No harm in setting both env vars simultaneously.
+        // Unix-style linkers.
+        if matches!(cmd.family, ToolFamily::Msvc { clang_cl: false }) {
+            cmd.env.push(("LC_ALL".into(), "C".into()));
+        } else {
+            // MSVC's `link.exe`.
+            cmd.env.push(("VSLANG".into(), "1033".into()));
```
madsmtm: Swapped
```suggestion
            cmd.env.push(("VSLANG".into(), "1033".into()));
        } else {
            cmd.env.push(("LC_ALL".into(), "C".into()));
```

NobodyXu: Hmm I think you are looking at an old commit?

src/lib.rs
```diff
@@ -1950,6 +1950,19 @@ impl Build {
             cmd.push_cc_arg(warnings_to_errors_flag);
         }
 
+        // Copied from <https://github.com/rust-lang/rust/blob/5db81020006d2920fc9c62ffc0f4322f90bffa04/compiler/rustc_codegen_ssa/src/back/linker.rs#L27-L38>
+        //
+        // Disables non-English messages from localized linkers.
+        // Such messages may cause issues with text encoding on Windows
+        // and prevent inspection of msvc output in case of errors, which we occasionally do.
+        // This should be acceptable because other messages from rustc are in English anyway,
+        // and may also be desirable to improve searchability of the linker diagnostics.
```
madsmtm, line 1959: Nit: Update this to refer to "compilers" instead of "linkers"

madsmtm (approved):

src/lib.rs
NobodyXu: Thanks updated

PR #1304, opened 2024-11-27
Add support for using sccache wrapper with cuda/nvcc

flaub:
This allows any specified `RUSTC_WRAPPER` to be invoked when using nvcc.


NobodyXu (changes requested):
Thanks!

NobodyXu (approved):

PR #1307, opened 2024-11-29
[AIX] use -maix64 instead of -m64

mustartt:
On AIX, `gcc -m64` is not a valid flag. We use the corresponding `-maix64` for GCC and the regular `-m64` for Clang.


NobodyXu (approved):
Thanks LGTM, I have read through the code before but forgot to approve and merge it

PR #1310, opened 2024-12-02
Improve detection of msvc environment when compiling from msvc

### LINKED ISSUE #1308: Compiling from `msbuild` incorrectly uses target build tools for build scripts
The problem arises when running `cargo build` from within a VS project and targeting something other than the native arch (e.g., using `msbuild.exe myproj.sln /p:Platform=otherarch`). When cargo builds a build script, it uses the target linker (added to `PATH` by Visual Studio) instead of the host linker, which naturally fails.

This is a problem as of Rust 1.83 due to the following fix: https://github.com/rust-lang/cc-rs/pull/1201. That PR also contains some discussion about this issue.

dlon:
Fix #1308. This makes path/env detection a bit pickier by not using `PATH` if the compiler has an unexpected target. Previously this caused issues for build scripts since rustc always needs to target the native arch.

### DIFF
--- a/src/windows/find_tools.rs
+++ b/src/windows/find_tools.rs
@@ -353,20 +353,69 @@ mod impl_ {
         }
     }
 
+    /// Checks to see if the target's arch matches the VS environment. Returns `None` if the
+    /// environment is unknown.
+    fn is_vscmd_target(target: TargetArch<'_>, env_getter: &dyn EnvGetter) -> Option<bool> {
+        is_vscmd_target_env(target, env_getter).or_else(|| is_vscmd_target_cl(target, env_getter))
+    }
+
     /// Checks to see if the `VSCMD_ARG_TGT_ARCH` environment variable matches the
     /// given target's arch. Returns `None` if the variable does not exist.
-    fn is_vscmd_target(target: TargetArch<'_>, env_getter: &dyn EnvGetter) -> Option<bool> {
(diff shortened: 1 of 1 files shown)


ChrisDenton:
Hmm, I did have to consider this a bit more. On balance, I think I'd rather keep the non-msbuild case the same as it was to avoid unintended side-effects.

So ideally this PR would only affect MSBuild.

ChrisDenton (commented):
Ok, I've been testing this and I'm happy that it solves the issue. I do have one small suggestion.

src/windows/find_tools.rs
```diff
@@ -353,20 +353,69 @@ mod impl_ {
...
+        cl.stderr(std::process::Stdio::piped())
+            .stdout(std::process::Stdio::null());
+
+        let out = cl.output().ok()?;
+        let cl_arch = out
+            .stderr
+            .split(|&b| b == b'\n' || b == b'\r')
+            .next()?
+            .split(|&b| b == b' ')
+            .last()?;
```
ChrisDenton, line 410:
```suggestion
            .rsplit(|&b| b == b' ')
            .next()?
```

dlon: Makes a lot more sense, thanks!

NobodyXu (approved):
Thanks!

PR #1312, opened 2024-12-03
Support new NTO QNX targets

flba-eb:
This change:
- Adds support for two new targets
- Adds support for a new target_env value
- Gives users a bit more information where an error message comes from (e.g. if target is unknown)

This depends on rust-lang/rust/pull/133631

CC: @jonathanpallant @japaric @gh-tr @AkhilTThomas

### DIFF
--- a/src/target/llvm.rs
+++ b/src/target/llvm.rs
@@ -42,7 +42,8 @@ pub(crate) fn guess_llvm_target_triple(
         os => os,
     };
     let env = match env {
-        "newlib" | "nto70" | "nto71" | "ohos" | "p1" | "p2" | "relibc" | "sgx" | "uclibc" => "",
+        "newlib" | "nto70" | "nto71" | "nto71_iosock" | "ohos" | "p1" | "p2" | "relibc" | "sgx"
+        | "uclibc" => "",
         env => env,
     };
     let abi = match abi {


madsmtm:
Thanks for the PR!

The `generated.rs` shouldn't be updated manually, it will be updated by a CI job within a week of the Rust PR merging.

Also, please add the better error message in a separate PR.

---

I think this is the first time that a Rust target is added since https://github.com/rust-lang/cc-rs/pull/1225, I'd like to know what you think we can do better (apart from improving the error message) to make that work smoother in the future?

flba-eb:
> Thanks for the PR!
> 
> The `generated.rs` shouldn't be updated manually, it will be updated by a CI job the next Friday after the Rust PR merges.
> 
> I think this is the first time that a Rust target is added since #1225, I'd like to know what you think we can do better (apart from improving the error message) to make such work smoother in the future?

Thanks for the feedback, I've reverted the changes of `generated.rs`.

Something that would have helped me (and I think @thejpster as well) is:
- A note in https://rustc-dev-guide.rust-lang.org/building/new-target.html about this issue. It does mention some required changes in e.g. `Cargo.toml` to allow targets not yet supported by nightly Rust, but nothing about the cc crate.
- A more clear error message of the cc crate. This PR tries to improve it a bit (so that the user at least knows where the error message is coming from), but e.g. a link to some additional documentation (like the dev guide of previous bullet point) could be even better.

NobodyXu:
> The generated.rs shouldn't be updated manually, it will be updated by a CI job the next Friday after the Rust PR merges.

You can modify and run `cargo run` inside the dev-tools/gen-target-info https://github.com/rust-lang/cc-rs/tree/main/dev-tools/gen-target-info

It will generate this src/target/generated

NobodyXu (changes requested):
Thanks, LGTM!

cc @madsmtm can you double check this please ?

NobodyXu (approved):

madsmtm (approved):

PR #1313, opened 2024-12-04
Better error message when failing on unknown targets

madsmtm:
Based on feedback in https://github.com/rust-lang/cc-rs/pull/1312.

This assumes that the dev guide will be updated as well, haven't submitted a PR for that yet though.

CC @flba-eb, wdyt?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3979,7 +3979,7 @@ impl Default for Build {
 }
 
 fn fail(s: &str) -> ! {
-    eprintln!("\n\nerror occurred: {}\n\n", s);
+    eprintln!("\n\nerror occurred in cc-rs: {}\n\n", s);
     std::process::exit(1);
 }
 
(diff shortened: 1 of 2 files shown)


madsmtm:
Hmm, actually, I'm not sure what to update the `rustc` dev guide with? It already [has a section on patching crates](https://rustc-dev-guide.rust-lang.org/building/new-target.html#patching-crates), and that section mentions both `libc` and `cc`?

What would you suggest? Just make it more prominent? Or does it not really matter if we have this error message in `cc-rs`?

jonathanpallant:
The proposed error text looks great, and would have bypassed the ~90 minutes of debugging I had to do working on a new target.

NobodyXu (approved):
Thanks, the new error message is much better!

flba-eb:
This is fantastic, thank you very much! ❤️

PR #1319, opened 2024-12-12
Add support for C/C++ compiler for Neutrino QNX: `qcc`

flba-eb:
This adds support for using the C/C++ compiler `qcc` from Neutrino QNX, with the target parameters required.

So far, it was required to set various environment variables.

CC: @jonathanpallant @japaric @gh-tr @AkhilTThomas


NobodyXu (changes requested):
Thanks, is it possible to setup CI and a new test to actually use qcc for testing

flba-eb:
Unfortunately this is not possible because `qcc` requires a paid license.

jonathanpallant:
Is this going to let us drop all the `CFLAGS_<target>` environment variables from the QNX Neutrino libstd builds?

flba-eb:
> Is this going to let us drop all the `CFLAGS_<target>` environment variables from the QNX Neutrino libstd builds?

Yes, see also https://github.com/rust-lang/rust/pull/134211 😄

src/lib.rs
```diff
@@ -2092,6 +2092,16 @@ impl Build {
                         cmd.push_cc_arg("-pthread".into());
                     }
                 }
+
+                if target.os == "nto" {
+                    let arg = match target.arch {
+                        "i586" => "-Vgcc_ntox86_cxx",
+                        "aarch64" => "-Vgcc_ntoaarch64le_cxx",
+                        "x86_64" => "-Vgcc_ntox86_64_cxx",
+                        _ => panic!("Unknown architecture for Neutrino QNX: {}", target.arch),
```
NobodyXu, line 2101:
```suggestion
                        _ => return Err(Error::new(ErrorKind::InvalidTarget, format!("Unknown architecture for Neutrino QNX: {}", target.arch)),
```

Let's not use panic, the code snippet would need cargo-fmt though

NobodyXu (changes requested):

NobodyXu:
Not very familiar with Neutrino QNX, cc @madsmtm can you have a look at this PR please?

src/lib.rs
flba-eb: Good suggestion, thanks. Done.

src/lib.rs
```diff
@@ -2803,6 +2818,8 @@ impl Build {
                     format!("arm-kmc-eabi-{}", gnu)
                 } else if target.arch == "aarch64" && target.vendor == "kmc" {
                     format!("aarch64-kmc-elf-{}", gnu)
+                } else if target.os == "nto" {
+                    "qcc".into()
```
madsmtm, line 2822: Should this also support `q++`?

src/lib.rs
```diff
@@ -2092,6 +2092,21 @@ impl Build {
                         cmd.push_cc_arg("-pthread".into());
                     }
                 }
+
+                if target.os == "nto" {
+                    let arg = match target.arch {
```
madsmtm, line 2097: Could we link to the docs for this? Something like:

```suggestion
                    // Select the target with `-V`. Only supported by qcc:
                    // https://www.qnx.com/developers/docs/7.0.0/index.html#com.qnx.doc.neutrino.utilities/topic/q/qcc.html
                    let arg = match target.arch {
```

src/lib.rs
madsmtm, line 2822: It sounds like this target _only_ supports compiling with `qcc`, is that true?

If so, @NobodyXu is this the right place to set this up? This would only select the _default_ compiler, right? What if the user passes `cc::Build::new().compiler("clang")`? Is the expected behaviour to still try to use Clang, even though we know that'll fail?

madsmtm:
(Also not familiar with QNX, I'm fine with deferring to the target maintainers here)

src/lib.rs
flba-eb: My main concern was to compile the Rust toolchain without having to set a bunch of environment variables (which only builds C files), but I think it makes sense to add support for `q++`. Good point, thanks.

flba-eb: > It sounds like this target _only_ supports compiling with `qcc`, is that true?

I think/guess you have to use `qcc` if you want to include system header files due to license reasons. In other words, usually only `qcc` is used.
If someone wants to use a very non-default compiler, they can set it via environment variables.

madsmtm: My concern is that if we add the `-V` flag, then you won't be able to compile with other compilers that don't support that flag

src/lib.rs
flba-eb: Done

src/lib.rs
flba-eb: I get your point. Thinking about my previous statement: It's not just header files, you also cannot link to create executables or shared objects without license violations. I don't think we need to support non-qcc compilers.

flba-eb: Should be resolved.

madsmtm: Cool, thanks for the clarification.

Gonna keep this open pending answer wrt. default vs. required compiler.

madsmtm (approved):
Thanks!

Final few questions from me:
- Do we need to do something else to support the `io-pkt`/`io-sock` divide? Do we need to pass a special flag or something?
- Is there a compiler flag for choosing between `7.0`, `7.1` and `8.0`? E.g. do we actually need to pass e.g. `-V7.0,gcc_ntoaarch64le_cxx`, or am I misunderstanding that?

flba-eb:
> Thanks!
> 
> Final few questions from me:
> 
> * Do we need to do something else to support the `io-pkt`/`io-sock` divide? Do we need to pass a special flag or something?

Only linking requires an additional library search path which is added by rustc when rustc links. As far as I understand cc, it never calls a linker itself. If cc can be used to link shared objects / executables, then it would make sense to investigate.

> * Is there a compiler flag for choosing between `7.0`, `7.1` and `8.0`? E.g. do we actually need to pass e.g. `-V7.0,gcc_ntoaarch64le_cxx`, or am I misunderstanding that?

No, I'm only adding -Vgcc_ntoaarch64le_cxx. The version number is optional. Also, this version relates to the used compiler version (e.g. "gcc 8.3"), not the QNX version.

jonathanpallant:
The correct compiler is added to your PATH by running a shell script / batch file supplied with the QNX Software Development Platform, so you can't have both QNX SDP 8.0 and QNX SDP 7.1 (for example) versions of `qcc` in your PATH at the same time.

jonathanpallant:
(and the OS is called QNX Neutrino, not Neutrino QNX, fwiw)

src/lib.rs
jonathanpallant: I don't think anyone is compiling for QNX Neutrino and not using `qcc`. If you pay this much for your RTOS, you colour between the lines, not outside.

madsmtm: I meant, I wanted @NobodyXu to confirm that this was the best place to set it up ;)

flba-eb:
> > The correct compiler is added to your PATH by running a shell script / batch file supplied with the QNX Software Development Platform, so you can't have both QNX SDP 8.0 and QNX SDP 7.1 (for example) versions of `qcc` in your PATH at the same time.
> 
> I'm just worried that users might have e.g. QNX SDP 7.1 configured while building the `aarch64-unknown-nto-qnx700` target, is there a check against such incompatibilities anywhere?

The toolchain does not reveal the QNX version, unfortunately. You would need to start parsing header files, like I've tried with https://github.com/rust-lang/libc/pull/4192 ... but as also written there:
Even if you are having QNX 7.1 on the target and a QNX 7.1 toolchain, it does not mean they are compatible. I've worked in a project where the toolchain was updated every 4 weeks and was always incompatible -- its part of the QNX experience to recompile everything all the time "just in case". And then you try to resolve a coredump just to figure out that things still does not fit together. If you want to be sure, you really need to hash the toolchain, compiler, header files and always compare it.

src/lib.rs
```diff
@@ -2803,6 +2821,12 @@ impl Build {
                     format!("arm-kmc-eabi-{}", gnu)
                 } else if target.arch == "aarch64" && target.vendor == "kmc" {
                     format!("aarch64-kmc-elf-{}", gnu)
+                } else if target.os == "nto" {
+                    if self.cpp {
+                        "q++".to_string()
+                    } else {
+                        "qcc".to_string()
+                    }
```
madsmtm, line 2829: Nit: Would be nice with a note here saying that `qcc` is the only supported compiler (and possibly linking to this PR for context).

flba-eb: I've added only a short remark here as I feel it helps more in `fn add_default_flags` where we assume that qcc/q++ is used (when adding the special arguments).

NobodyXu (approved):
Thank you!

PR #1322, opened 2024-12-17
Check linking when testing if compiler flags are supported

### LINKED ISSUE #1321: `build: Inherit flags from rustc` causes rustc build failures
https://github.com/rust-lang/rust/pull/134425#issuecomment-2549090540

Bisected to #1279

Looks like we get something like
```
/usr/bin/ld: /rustroot/bin/../lib/LLVMgold.so: error loading plugin: /rustroot/bin/../lib/LLVMgold.so: cannot open sha
red object file: No such file or directory
```
while running `tikv-jemalloc-sys`'s `configure`, as well as some
```
Inherited flag "-fembed-bitcode=all" is not supported by the currently used CC
```
warnings

clubby789:
Fixes #1321

Several of the inherited flags affect the linker, so we *do* need to go all the way to linking to verify them

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -725,9 +725,6 @@ impl Build {
             },
         );
 
-        // Checking for compiler flags does not require linking
-        cmd.arg("-c");
-
         cmd.arg(&src);
 
         let output = cmd.output()?;


thomcc (approved):

NobodyXu (approved):
Thanks!

faern:
Can we get this published as a patch release, or are there plans around when to cut the next release? Asking since this currently blocks https://github.com/rust-lang/rust/pull/134419, which in turn blocks https://github.com/rust-lang/rust/issues/133794

NobodyXu:
cc will create one release PR every Saturday, and then I can merge and release it.

I could release this early if it is urgent

faern:
Cool. I did not know you had that weekly cadence. I think it would be nice if we can make the bump in `rustc` before January 3rd, so the MSBuild bug goes away with Rust 1.85 at least.

NobodyXu:
Given that it's a bug affecting rustc and might affect rustc's next release, I will cut a release now to ensure you'd have that in.

NobodyXu:
cc @clubby789 @faern new patch release containing the fix has released

PR #1326, opened 2024-12-22
Support target mips-zkm-zkvm-elf

eigmax:
A general-purpose zkVM based on MIPS32V2 has been implemented and we are trying to add a new target for Rust.

### DIFF
--- a/src/target/generated.rs
+++ b/src/target/generated.rs
@@ -1517,6 +1517,18 @@ pub(crate) const LIST: &[(&str, TargetInfo<'static>)] = &[
         },
     ),
     (
+        "mips-zkm-zkvm-elf",
+        TargetInfo {
+            full_arch: "mips",
+            arch: "mips",
+            vendor: "zkm",
+            os: "zkvm",
+            env: "",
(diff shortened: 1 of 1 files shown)


src/target/generated.rs
```diff
@@ -1516,6 +1516,18 @@ pub(crate) const LIST: &[(&str, TargetInfo<'static>)] = &[
             unversioned_llvm_target: "mips-unknown-linux-uclibc",
         },
     ),
+    (
+        "mips-zkm-zkvm-elf",
```
NobodyXu, line 1520: Thank you!


Please do not edit this file directly

This file is generated by https://github.com/rust-lang/cc-rs/tree/main/dev-tools/gen-target-info/src

So when new targets are added in rustc, this file should be automatically regenerated in CI.

NobodyXu (changes requested):

NobodyXu:
Sorry but it seems that the generated file is still manually edited?

NobodyXu:
It seems that the generated file is still manually edited?

PR #1328, opened 2024-12-22
Fix compiler family detection issue with clang-cl on macOS

### LINKED ISSUE #1327: Compiler family detection issue with `clang-cl` on macOS
```
warning: crfsuite-sys@0.3.1: clang-cl: warning: '/Users/messense/Projects/crfsuite-rs/target/x86_64-pc-windows-msvc/debug/build/crfsuite-sys-a73da6d4e4774cd8/out/14314275059944146007detect_compiler_family.c' treated as the '/U' option [-Wslash-u-filename]
warning: crfsuite-sys@0.3.1: clang-cl: note: use '--' to treat subsequent arguments as filenames
warning: crfsuite-sys@0.3.1: clang-cl: error: no input files
warning: crfsuite-sys@0.3.1: Compiler family detection failed due to error: ToolExecError: Command "clang-cl" "-E" "/Users/messense/Projects/crfsuite-rs/target/x86_64-pc-windows-msvc/debug/build/crfsuite-sys-a73da6d4e4774cd8/out/14314275059944146007detect_compiler_family.c" with args clang-cl did not execute successfully (status code exit status: 1).
```

It seems like the `/Users/...` absolute path on macOS does not work well with `clang-cl`.

To reproduce:

```bash
$ touch empty.c
$ ln -s $(xcrun -f clang) clang-cl
$ ./clang-cl -E $(realpath ./empty.c)
clang-cl: warning: '/Users/messense/Projects/cargo-xwin/empty.c' treated as the '/U' option [-Wslash-u-filename]
clang-cl: note: use '--' to treat subsequent arguments as filenames
clang-cl: error: no input files

# this works fine
$ ./clang-cl -E ./empty.c 
# 1 "./empty.c"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 375 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "./empty.c" 2
```

This is handled in https://github.com/rust-lang/cc-rs/blob/9138052fc3e6de612ec99a49c6a411fabd570599/src/lib.rs#L1791-L1797 but does not seem to be handled when doing compiler family detection: https://github.com/rust-lang/cc-rs/blob/9138052fc3e6de612ec99a49c6a411fabd570599/src/tool.rs#L136-L145

NobodyXu:
Fixed #1327

### DIFF
--- a/src/tool.rs
+++ b/src/tool.rs
@@ -105,6 +105,46 @@ impl Tool {
             .unwrap_or_default()
         }
 
+        fn guess_family_from_stdout(
+            stdout: &str,
+            path: &Path,
+            cargo_output: &CargoOutput,
+        ) -> Result<ToolFamily, Error> {
+            cargo_output.print_debug(&stdout);
+
+            // https://gitlab.kitware.com/cmake/cmake/-/blob/69a2eeb9dff5b60f2f1e5b425002a0fd45b7cadb/Modules/CMakeDetermineCompilerId.cmake#L267-271
+            // stdin is set to null to ensure that the help output is never paginated.
+            let accepts_cl_style_flags =
+                run(Command::new(path).arg("-?").stdin(Stdio::null()), path, &{
+                    // the errors are not errors!
+                    let mut cargo_output = cargo_output.clone();
+                    cargo_output.warnings = cargo_output.debug;
+                    cargo_output.output = OutputKind::Discard;
+                    cargo_output
+                })
+                .is_ok();
+
+            let clang = stdout.contains(r#""clang""#);
+            let gcc = stdout.contains(r#""gcc""#);
+            let emscripten = stdout.contains(r#""emscripten""#);
+            let vxworks = stdout.contains(r#""VxWorks""#);
+
+            match (clang, accepts_cl_style_flags, gcc, emscripten, vxworks) {
+                (clang_cl, true, _, false, false) => Ok(ToolFamily::Msvc { clang_cl }),
+                (true, _, _, _, false) | (_, _, _, true, false) => Ok(ToolFamily::Clang {
+                    zig_cc: is_zig_cc(path, cargo_output),
+                }),
+                (false, false, true, _, false) | (_, _, _, _, true) => Ok(ToolFamily::Gnu),
+                (false, false, false, false, false) => {
+                    cargo_output.print_warning(&"Compiler family detection failed since it does not define `__clang__`, `__GNUC__`, `__EMSCRIPTEN__` or `__VXWORKS__`, also does not accept cl style flag `-?`, fallback to treating it as GNU");
+                    Err(Error::new(
+                        ErrorKind::ToolFamilyMacroNotFound,
+                        "Expects macro `__clang__`, `__GNUC__` or `__EMSCRIPTEN__`, `__VXWORKS__` or accepts cl style flag `-?`, but found none",
+                    ))
+                }
+            }
+        }
+
         fn detect_family_inner(
             path: &Path,
             cargo_output: &CargoOutput,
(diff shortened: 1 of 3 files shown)


NobodyXu:
I can't know for sure if the compiler is clang-cl, so can't apply the workaround reliably.

Alternatively, we could combine `cfg(macos)` with accept cl flags check.

if both are true then it's likely clang-cl on macOS, which needs workaroud.

madsmtm (changes requested):
It is not clear to me why we even support `clang-cl` on macOS and Linux? `cl.exe` is not something you'd run on those platforms, so why would you ever run `clang-cl`?

> `cfg(macos)`

That's a bad idea IMO, since this problem is also present on Linux and Windows hosts, just less prominently since their root paths usually start with a lowercase letter (on Linux), or prefixed with the drive letter (on Windows), but that's not a requirement!

---

Actually, the `--` separator seems to be supported by normal Clang too, so why don't we just do this workaround everywhere?

This would also avoid issues if the user did e.g. `.file("-l")`.

NobodyXu:
> It is not clear to me why we even support `clang-cl` on macOS and Linux? `cl.exe` is not something you'd run on those platforms, so why would you ever run `clang-cl`?

Based on the original issue, I believe they are using cargo-xwin to cross compile to msvc windows from a unix environment.

> Actually, the `--` separator seems to be supported by normal Clang too, so why don't we just do this workaround everywhere?

Does gcc and other compiler support it?

If they mostly do, we can add `--` to compiler detection.

> This would also avoid issues if the user did e.g. `.file("-l")`.

Good idea

madsmtm:
> Does gcc and other compiler support it?

```console
$ touch a.c
$ gcc -- a.c
gcc: error: unrecognized command-line option '--'
```

Doesn't seem like it?

src/tool.rs
```diff
@@ -105,6 +105,46 @@ impl Tool {
...
+        fn guess_family_from_stdout(
+            stdout: &str,
+            path: &Path,
+            cargo_output: &CargoOutput,
+        ) -> Result<ToolFamily, Error> {
+            cargo_output.print_debug(&stdout);
+
+            // https://gitlab.kitware.com/cmake/cmake/-/blob/69a2eeb9dff5b60f2f1e5b425002a0fd45b7cadb/Modules/CMakeDetermineCompilerId.cmake#L267-271
+            // stdin is set to null to ensure that the help output is never paginated.
+            let accepts_cl_style_flags =
```
madsmtm, line 117: Hmm, I'm wondering, can't we do this check _first_, and then use that to figure out whether to do `--` or not based on that?

Or, does `cl.exe` not support `--` either?

src/tool.rs
```diff
@@ -140,53 +180,30 @@ impl Tool {
...
-                }),
-                (false, false, true, _, false) | (_, _, _, _, true) => Ok(ToolFamily::Gnu),
-                (false, false, false, false, false) => {
-                    cargo_output.print_warning(&"Compiler family detection failed since it does not define `__clang__`, `__GNUC__`, `__EMSCRIPTEN__` or `__VXWORKS__`, also does not accept cl style flag `-?`, fallback to treating it as GNU");
-                    Err(Error::new(
-                        ErrorKind::ToolFamilyMacroNotFound,
-                        "Expects macro `__clang__`, `__GNUC__` or `__EMSCRIPTEN__`, `__VXWORKS__` or accepts cl style flag `-?`, but found none",
-                    ))
-                }
+            if stdout.contains("-Wslash-u-filename") {
```
madsmtm, line 197: This... Seems brittle :/.

Not that I have a better solution (yet).

src/tool.rs
madsmtm: Ah, yeah, it doesn't.

src/tool.rs
```diff
@@ -140,53 +180,30 @@ impl Tool {
...
 
+            // When expanding the file, the compiler prints a lot of information to stderr
+            // that it is not an error, but related to expanding itself.
+            //
+            // cc would have to disable warning here to prevent generation of too many warnings.
+            let mut compiler_detect_output = cargo_output.clone();
+            compiler_detect_output.warnings = compiler_detect_output.debug;
+
             let stdout = run_output(
                 Command::new(path).arg("-E").arg(tmp.path()),
```
madsmtm, line 191: Actually, I wonder, why don't we just detect the compiler by running `$path -v`?

Output `clang -v` on macOS:
```console
Apple clang version 16.0.0 (clang-1600.0.26.6)
Target: arm64-apple-darwin23.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Output `gcc -v` on Ubuntu:
```console
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/aarch64-linux-gnu/11/lto-wrapper
Target: aarch64-linux-gnu
Configured with: [...]
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04)
```

Output `cl.exe -v` on Windows (taking advantage of `cl` always outputting version information):
```console
Microsoft (R) C/C++ Optimizing Compiler Version 19.42.34435 for x86
Copyright (C) Microsoft Coporation.  All rights reserved

cl : Command line warning D9002 : ignoring unknown option '-v'
cl : Command line error D8003 : missing source filename

```

madsmtm: And then `$path -?` to differentiate `clang-cl` from `clang`.

NobodyXu: I think this is what we are previously doing, but get removed due to not reliable enough.

src/tool.rs
NobodyXu: I don't like it either, but I really can't think of an alternative solution that is better

src/tool.rs
madsmtm: Hmm, I suspected so. Found https://github.com/rust-lang/cc-rs/pull/1000.

src/tool.rs
madsmtm: What does CMake do?

NobodyXu: Not sure , I didn't check about cmake.

I did remember the `-?` solution is copied from it, so perhaps it uses that as a hint?

madsmtm (approved):
I don't feel I know enough to really gauge whether this is the right approach or not, so I'll leave that up to you.

I don't think GitHub allows me to make a neutral review after having requested changes, so just so that won't block you, I've approved it.

NobodyXu:
Thanks, let me merge it and try shipping it, if it doesn't work we have to revert

messense:
Unfortunately it seems to broke `cargo-zigbuild` badly: https://github.com/rust-cross/cargo-zigbuild/actions/runs/12723439506/job/35468553100

> error occurred in cc-rs: Command LC_ALL="C" "/home/runner/.cache/cargo-zigbuild/0.19.7/zigcc-aarch64-unknown-linux-gnu-6363.sh" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-4" "-fno-omit-frame-pointer" "--target=aarch64-unknown-linux-gnu" "-I" "zstd/lib/" "-I" "zstd/lib/common" "-I" "zstd/lib/legacy" "-fvisibility=hidden" "-DZSTD_LIB_DEPRECATED=0" "-DXXH_PRIVATE_API=" "-DZSTDLIB_VISIBILITY=" "-DZDICTLIB_VISIBILITY=" "-DZSTDERRORLIB_VISIBILITY=" "-DZSTD_LEGACY_SUPPORT=1" "-o" "/home/runner/work/cargo-zigbuild/cargo-zigbuild/tests/zstd-rs/target/aarch64-unknown-linux-gnu/debug/build/zstd-sys-c42cdd11ab2b8d1c/out/7faed3f8272f2313-huf_decompress_amd64.o" "-c" "--" "zstd/lib/decompress/huf_decompress_amd64.S" with args zigcc-aarch64-unknown-linux-gnu-6363.sh did not execute successfully (status code exit status: 1).

messense:
And it does not seem to resolve the clang-cl issue either:

> warning: libz-ng-sys@1.1.16: Compiler family detection failed due to error: ToolExecError: Command "clang-cl" "-E" "/Users/messense/Projects/uv/target/aarch64-pc-windows-msvc/debug/build/libz-ng-sys-e348506bb081d40e/out/442624705732975273detect_compiler_family.c" with args clang-cl did not execute successfully (status code exit status: 1).

tokatoka:
I think this will break if people uses `cc` with their own compiler wrappers 

For our case, we use a compiler wrapper based on clang and in our wrapper we didn't handle argument `--` .
This fix automatically adds "--" automatically, and we had to handle it else it broke.

NobodyXu:
@messense @tokatoka I've opened #1359, can you try it out please?

tbetcke:
@NobodyXu I have the same issue with `mpicc` as compiler wrapper (e.g. used by [rsmpi](https://github.com/rsmpi/rsmpi). Version 1.2.7 of `cc-rs` works fine. Version `1.2.8` fails with the `--` issue.

PR #1335, opened 2024-12-29
Allow using Visual Studio target names in `find_tool`

ChrisDenton:
Make `find_tool` understand VS arch names. This is admittedly partly motivated by the fact that rustc internally uses the name `x86` to refer to either `i586` and `i686` where the distinction doesn't matter.

I also centralized parsing of the user-provided arch name in one place, rather than being repeated in multiple places. This uncovered a bug in `vs15plus_vc_paths` where we were testing the wrong `target` variable due to shadowing. I fixed it and renamed the two variables `target_arch` and `target_dir` to be clearer.

### DIFF
--- a/src/windows/find_tools.rs
+++ b/src/windows/find_tools.rs
@@ -496,7 +496,7 @@ mod impl_ {
                     return None;
                 }
                 let mut tool = Tool::with_family(path, MSVC_FAMILY);
-                if target == TargetArch::X86 {
+                if target == TargetArch::X64 {
                     tool.env.push(("Platform".into(), "X64".into()));
                 }
                 if matches!(target, TargetArch::Arm64 | TargetArch::Arm64ec) {
@@ -704,7 +704,7 @@ mod impl_ {
             }
             _ => return None,
         };
-        let target_dir = lib_subdir(target_arch)?;
+        let target_dir = target_arch.as_vs_arch();
         // The directory layout here is MSVC/bin/Host$host/$target/
         let path = instance_path.join(r"VC\Tools\MSVC").join(version);
         // We use the first available host architecture that can build for the target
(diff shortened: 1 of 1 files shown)


src/windows/find_tools.rs
```diff
@@ -493,18 +496,18 @@ mod impl_ {
                     return None;
                 }
                 let mut tool = Tool::with_family(path, MSVC_FAMILY);
-                if target == "x86_64" {
+                if target == TargetArch::X86 {
```
NobodyXu, line 499: Given that it originally matches against `x86_64`, I think it should be `X64`

```suggestion
                if target == TargetArch::X64 {
```

src/windows/find_tools.rs
```diff
@@ -1009,36 +1005,29 @@ mod impl_ {
...
-        match target.into() {
-            "i586" | "i686" => Some("x86"),
-            "x86_64" => Some("x64"),
-            "arm" | "thumbv7a" => Some("arm"),
-            "aarch64" | "arm64ec" => Some("arm64"),
-            _ => None,
-        }
+    fn lib_subdir(target: TargetArch) -> Option<&'static str> {
+        Some(target.as_vs_arch())
     }
```
NobodyXu, line 1022: Perhaps we can remove this function?

NobodyXu (changes requested):
Thank you!

Just find one mistake, and another nit code refactoring advice

src/windows/find_tools.rs
ChrisDenton: Whoops, nice catch!

src/windows/find_tools.rs
ChrisDenton: Good idea.

NobodyXu (approved):
Thank you

PR #1336, opened 2024-12-31
Fix is_flag_supported on msvc

### LINKED ISSUE #1324: bug: Build::is_flag_supported `/std:c++17` returns false on 1.2.1 ->1.2.5 transition (on x86_64-pc-windows-msvc)
This is suppossedly a bug of https://docs.rs/cc/1.2.5/cc/struct.Build.html#method.is_flag_supported

on x86_64-pc-windows-msvc. 

After applying diff https://github.com/near/cargo-near/pull/281/commits/4a4a77263360b51fa6c46509ba2317582dd9a767
, running build script in https://github.com/brson/wasm-opt-rs/blob/8854bd8a58d7393c7bacd950884aea521ce8e19f/components/wasm-opt-sys/build.rs#L365 
fails https://github.com/near/cargo-near/actions/runs/12417889777/job/34669855813

Image
Image

Pinning `cc` to 1.2.1 in https://github.com/near/cargo-near/pull/282 helps (build succeeds on the identical image)

ChrisDenton:
Fixes #1324 by skipping linking the CRT when testing for supported compiler flags on Windows. We use the `_LINK_` environment variable to avoid disturbing the command line.

The more complex alternative would be to get the library paths and add them to the `LIB` environment variable.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -727,6 +727,13 @@ impl Build {
 
         cmd.arg(&src);
 
+        // On MSVC skip the CRT by setting the entry point to `main`.
+        // This way we don't need to add the default library paths.
+        if compiler.is_like_msvc() {
+            // Flags from _LINK_ are appended to the linker arguments.
+            cmd.env("_LINK_", "-entry:main");
+        }
+
         let output = cmd.output()?;
         let is_supported = output.status.success() && output.stderr.is_empty();
 


NobodyXu (approved):
Thank you for putting up a PR so quick!

Usually we cut a release every Friday/Sat, but if this is urgent enough I can cut a release tomorrow

PR #1337, opened 2024-12-31
Make `is_flag_supported_inner` take an `&Tool` instead of a `&Path`

ChrisDenton:
By passing in a `Tool` we allow other properties to be inspected by `is_flag_supported_inner` if necessary. I originally made this change to support grabbing an environment variable from the `Tool` but ended up going in a different direction. Still I think having the stronger type is useful on its own merit. And in the process of doing this I also ended up simplifying the arguments to `cc_flags`.


NobodyXu (approved):
Thanks for the refactoring!

PR #1338, opened 2025-01-01
Document new supported architecture names in `windows_registry::find`

madsmtm:
These were introduced #1335, but weren't documented.

CC @ChrisDenton.


ChrisDenton (approved):
Oops sorry, I didn't realise we had a list

PR #1339, opened 2025-01-01
Pass deployment target with `-m*-version-min=`

madsmtm:
Clang's supports many ways of passing the deployment target, [and prefers](https://github.com/llvm/llvm-project/issues/88271#issuecomment-2050425068) that you use the `-target` flag for doing so.

This works poorly work with configuration files though, which Homebrew's LLVM uses, so we use the `-mmacosx-version-min=`, `-miphoneos-version-min=`, `-mtargetos=` etc. flags to pass the deployment target instead. See also discussion in https://github.com/Homebrew/homebrew-core/issues/197532.

I will be submitting a similar PR to `rustc` at some point to fix this in there too.

Fixes https://github.com/rust-lang/cc-rs/issues/1278 (together with https://github.com/Homebrew/homebrew-core/issues/197278).

### DIFF
--- a/dev-tools/gen-target-info/src/main.rs
+++ b/dev-tools/gen-target-info/src/main.rs
@@ -26,17 +26,9 @@ fn generate_target_mapping(f: &mut File, target_specs: &RustcTargetSpecs) -> std
         let env = spec.env.as_deref().unwrap_or("");
         let abi = spec.abi.as_deref().unwrap_or("");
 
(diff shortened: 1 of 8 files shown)


madsmtm:
I have opened https://github.com/rust-lang/rust/pull/136333 for doing the same in `rustc` now, so I believe this is ready to land.

NobodyXu (approved):
Thanks, LGTM

briansmith:
This seems to have broken my builds in CI: filed #1383.

PR #1340, opened 2025-01-02
CI: Add "test-pass" job

NobodyXu:
As a shortcut for checking all CI jobs have succeeded.


madsmtm (commented):
Hmm, I'm not sure I understand? You want to use this for the "Require status checks to pass before merging" value? Is it not enough to just set it up like this:

require-status-checks

NobodyXu:
No that's not enough, you have to add the CI jobs there otherwise it would do nothing

PR #1347, opened 2025-01-07
Update `windows-bindgen` dependency

kennykerr:
Updates the `cc` crate to use the latest version of the `windows-bindgen` crate. 
https://github.com/microsoft/windows-rs/releases/tag/0.61.0

A few notes worth mentioning about this change:

* The `bindgen` arguments have been simplified and I've updated them to produce the same form of output.
* The `--etc` option can be used to avoid having to parse the filter file yourself. 
* The filters no longer need to include the fully qualified type/function names.
* The output is now sorted more consistently but is otherwise largely the same. 
* There are a handful of additional types in the output. Those reflect a more complete representation of the requested filters and their dependencies.

Here's more detailed information on the latest version of the `windows-bindgen` crate:
https://github.com/microsoft/windows-rs/pull/3359

Adding @ChrisDenton as he originally came up with this model for the Rust project. I figure you'll be updating the bindings for the Rust compiler/library yourself so I thought I'd take a stab at this smaller project.


src/windows/windows_sys.rs
```diff
@@ -69,12 +70,24 @@ impl GUID {
...
+#[repr(C)]
+pub struct IUnknown_Vtbl {
+    pub QueryInterface: unsafe extern "system" fn(
+        this: *mut core::ffi::c_void,
+        iid: *const GUID,
+        interface: *mut *mut core::ffi::c_void,
+    ) -> HRESULT,
+    pub AddRef: unsafe extern "system" fn(this: *mut core::ffi::c_void) -> u32,
+    pub Release: unsafe extern "system" fn(this: *mut core::ffi::c_void) -> u32,
+}
```
madsmtm, line 90: Nit: I wonder why these are added?

madsmtm (approved):

src/windows/windows_sys.rs
kennykerr: The metadata used to described `HMODULE` - required by `FreeLibrary` and friends - actually depends on `HINSTANCE`. 

image

Similarly, `CoCreateInstance` depends on `IUnknown`.

This is what I meant by the following in the PR description:

> There are a handful of additional types in the output. Those reflect a more complete representation of the requested filters and their dependencies.

The `bindgen` crate doesn't know whether you actually need these dependencies but it includes them as there are cases where they are definitely required.

madsmtm: Ah, thanks for the detailed explanation!

PR #1357, opened 2025-01-10
tool: Add `is_like_clang_cl()` getter

MarijnS95:
In `ring` we need to check if the compiler is `clang-cl` to prevent overwriting it with `clang` (see all details in https://github.com/briansmith/ring/issues/2215), but `cc-rs` does not currently expose this despite already tracking it.  By adding this getter we can steer that logic to not overwrite the compiler when it is `clang-cl`.

Perhaps, since `ToolFamily` is `pub` (but not yet reexported from `lib.rs`), could we have a public `Tool::family()` getter to not have to extend these `is_xxx()` getters whenever a downstream crate wants to know something new?

### DIFF
--- a/src/tool.rs
+++ b/src/tool.rs
@@ -421,6 +421,11 @@ impl Tool {
         matches!(self.family, ToolFamily::Msvc { .. })
     }
 
+    /// Whether the tool is `clang-cl`-based MSVC-like.
+    pub fn is_like_clang_cl(&self) -> bool {
+        matches!(self.family, ToolFamily::Msvc { clang_cl: true })
+    }
+
     /// Supports using `--` delimiter to separate arguments and path to source files.
     pub(crate) fn supports_path_delimiter(&self) -> bool {
         matches!(


NobodyXu (approved):
Thanks!

> could we have a public Tool::family() getter

Yes that'd makes sense with non_exhaustive enum

NobodyXu:
cc usually has a new release every Friday (and we already have one today).

If this is very important, then I can cut another release today/tomorrow

MarijnS95:
> Yes that'd makes sense with non_exhaustive enum

Let's see if I can make a second PR.

> If this is very important, then I can cut another release today/tomorrow

Unsure, let's see how quickly `ring` wants to accept this. I don't personally have a big hurry with this, we can just `[patch]` away :)

MarijnS95:
> cc usually has a new release every Friday (and we already have one today).

It looks like it failed: https://github.com/rust-lang/cc-rs/pull/1355

Meaning you could perhaps trivially roll this in :)

MarijnS95:
@NobodyXu perhaps we don't need to release with it if we release with #1358 instead :)

Issue #882, opened 2023-10-03
Add "prefer to use clang instead of MSVC" API that can automatically find Microsoft-provided clang


briansmith:
I would love to be able in my build.rs to instead just say `b.prefer_clang()` to get cc-rs to automatically discover Microsoft-provided clang's and use it if/when it is available, when it would otherwise choose MSVC.

Some projects, such as [*ring*](https://github.com/briansmith/ring), *require* clang when building for Windows AArch64.
Some projects, such as *ring* would *prefer* to build with clang if it were available, when building for Windows X86-64, because the C code is optimized for clang in various ways (because Google uses clang for everything, and the C code we have comes from Google, and Google's preferences are unlikely to change any time soon).

Currently I am recommending people to do this:
```
$env:Path += ";C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\Llvm\x64\bin"
```
to get *ring* to build for aarch64-pc-windows-msvc using the clang that Microsoft provides with the build tools. Unfortunately, this advice is problematic:
* It is one extra step that shouldn't be necessary.
* "2022" varies depending on the version of MS build tools that is released
* "BuildTools" varies depending on whether the user installed the "Build Tools," "Community," "Enterprise," etc. edition of Visual Studio.
* "x64" varies depending on the host: ARM64 vs. x64, IIRC.

ChrisDenton:
This sounds like a job for `vswhere.exe`. If Visual Studio is installed then it's always located at `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe` and can discover paths.

thomcc:
I'd take patches to support this, but have no strong feeling on the implementation strategy (calling out to something like vswhere is certainly acceptable if it works well though).

The API name `prefer_clang_over_msvc()` is a little unwieldy but I don't think it's likely to be commonly needed so it probably doesn't matter.

briansmith:
> I'd take patches to support this, but have no strong feeling on the implementation strategy (calling out to something like vswhere is certainly acceptable if it works well though).

How does using vswhere.exe differ from what cc-rs currently does to find MSVC itself? I wonder it might be better for cc-rs to adapt what it currently does for finding MSVC to find clang directly without shelling out to vswhere?

ChrisDenton:
[...]

PR #1358, opened 2025-01-10
tool: Expose `family()` in favour of incomplete `is_like_xxx()` wrappers

MarijnS95:
Instead of adding a new `is_like_xxx()` function every time new information is needed, or new enumeration variants would be added, provide the (now `non_exhaustive`) `enum` directly to callers to match what they're interested in.

Also removes `is_like_clang_cl()` again from #1357 since that did not yet make it into a release, and would only cause unnecessary duplication in our API patterns.

### DIFF
--- a/src/tool.rs
+++ b/src/tool.rs
@@ -281,11 +281,11 @@ impl Tool {
         let mut chars = flag.chars();
 
         // Only duplicate check compiler flags
-        if self.is_like_msvc() {
-            if chars.next() != Some('/') {
-                return false;
-            }
-        } else if (self.is_like_gnu() || self.is_like_clang()) && chars.next() != Some('-') {
+        let flag_start = match self.family {
+            ToolFamily::Msvc { .. } => '/',
+            ToolFamily::Gnu | ToolFamily::Clang { .. } => '-',
+        };
+        if chars.next() != Some(flag_start) {
             return false;
         }
 
@@ -395,6 +395,11 @@ impl Tool {
         flags
     }
 
+    /// The family of this tool, representing convention of arguments etc.
+    pub fn family(&self) -> ToolFamily {
+        self.family
+    }
+
     /// Whether the tool is GNU Compiler Collection-like.
     pub fn is_like_gnu(&self) -> bool {
         self.family == ToolFamily::Gnu
@@ -421,11 +426,6 @@ impl Tool {
         matches!(self.family, ToolFamily::Msvc { .. })
     }
 
-    /// Whether the tool is `clang-cl`-based MSVC-like.
-    pub fn is_like_clang_cl(&self) -> bool {
-        matches!(self.family, ToolFamily::Msvc { clang_cl: true })
-    }
-
     /// Supports using `--` delimiter to separate arguments and path to source files.
     pub(crate) fn supports_path_delimiter(&self) -> bool {
         matches!(
@@ -441,14 +441,24 @@ impl Tool {
 ///
 /// Detection of a family is done on best-effort basis and may not accurately reflect the tool.
 #[derive(Copy, Clone, Debug, PartialEq)]
+#[non_exhaustive]
 pub enum ToolFamily {
     /// Tool is GNU Compiler Collection-like.
+    #[non_exhaustive]
     Gnu,
     /// Tool is Clang-like. It differs from the GCC in a sense that it accepts superset of flags
     /// and its cross-compilation approach is different.
-    Clang { zig_cc: bool },
-    /// Tool is the MSVC cl.exe.
-    Msvc { clang_cl: bool },
+    #[non_exhaustive]
+    Clang {
+        /// Tool provided by zig
+        zig_cc: bool,
+    },
+    /// Tool is the MSVC `cl.exe`.
+    #[non_exhaustive]
+    Msvc {
+        /// Whether this is `clang-cl` provided by LLVM
+        clang_cl: bool,
+    },
 }
 
 impl ToolFamily {
--- a/dev-tools/cc-test/build.rs
+++ b/dev-tools/cc-test/build.rs
@@ -23,7 +23,10 @@ fn main() {
     run_forked_capture_output(&out, "metadata-off");
 
     run_forked_capture_output(&out, "warnings-off");
-    if cc::Build::new().get_compiler().is_like_msvc() {
+    if matches!(
+        cc::Build::new().get_compiler().family(),
+        cc::ToolFamily::Msvc { .. }
+    ) {
         // MSVC doesn't output warnings to stderr, so we can't capture them.
         // the test will use this env var to know whether to run the test.
         println!("cargo:rustc-env=TEST_WARNINGS_ON=0");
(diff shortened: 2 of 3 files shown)


src/tool.rs
```diff
@@ -281,12 +281,17 @@ impl Tool {
         let mut chars = flag.chars();
 
         // Only duplicate check compiler flags
-        if self.is_like_msvc() {
-            if chars.next() != Some('/') {
-                return false;
+        match self.family {
+            ToolFamily::Msvc { .. } => {
+                if chars.next() != Some('/') {
```
NobodyXu, line 286:
```suggestion
            ToolFamily::Msvc { .. }  if chars.next() != Some('/') => {
```

src/tool.rs
```diff
@@ -441,14 +455,24 @@ impl Tool {
 ///
 /// Detection of a family is done on best-effort basis and may not accurately reflect the tool.
 #[derive(Copy, Clone, Debug, PartialEq)]
+#[non_exhaustive]
 pub enum ToolFamily {
     /// Tool is GNU Compiler Collection-like.
+    #[non_exhaustive]
```
NobodyXu, line 461: I don't think you need to put this ahead of each variant.

src/tool.rs
```diff
@@ -395,12 +400,19 @@ impl Tool {
         flags
     }
 
+    /// The family of this tool, representing convention of arguments etc.
+    pub fn family(&self) -> ToolFamily {
+        self.family
+    }
+
     /// Whether the tool is GNU Compiler Collection-like.
+    #[deprecated = "Consider matching against the ToolFamily returned by family() instead"]
```
NobodyXu, line 409: Not sure if we should deprecate this now cc @madsmtm what do you think?

src/tool.rs
MarijnS95: We do, otherwise _fields on_ variants aren't `non_exhaustive`.

src/tool.rs
MarijnS95: Wasn't sure either, might be a bit harsh but otherwise we could forget about it.

MarijnS95: At the same time matching against `self.family()` for single variants is a bit ugly as you've seen.

MarijnS95:
> Have a few suggestions, and I will cut release tomorrow as it's too late for me.

If we incorporate this PR, shall I just go ahead and remove `is_like_clang_cl()` again since it's only extra duplication?

NobodyXu:
> If we incorporate this PR, shall I just go ahead and remove `is_like_clang_cl()` again since it's only extra duplication?

Yes i think so

briansmith:
I suggest we hold off on doing this until the situation gets out of hand. Deprecating the current methods will cause a lot of unneeded churn, especially for projects that don't want to increase their minimum cc-rs version.

MarijnS95:
@briansmith yeah makes sense, `1.x.x` is likely to stay around for a long time. I've removed it so that this PR is ready for the proposed merge and release whenever @NobodyXu is available again (and I'm likely asleep).

src/tool.rs
NobodyXu: Oh yeah...thanks

NobodyXu (approved):

NobodyXu:
> I wonder if it is a good idea to expose `family()` publicly like this at all, though?

Regarding API I think it's ok.

But since it's new API, cc @madsmtm can you take a look at this please?

> I would expect to have more thorough tests for it, first, if so.

I remember we do have some tests for compiler family detection

madsmtm (changes requested):
Hmm, I don't particularly like exposing `ToolFamily`, for several reasons:
- I'm not certain that `Msvc { clang_cl }` is the right nesting, it could also be e.g. `Clang { cl }`.
- I'm unsure if these three families are really the right separation, e.g. we might want `ToolFamily::VxWorks` or `ToolFamily::Emscripten` one day, but if we expose `ToolFamily` we're forced to put those under `ToolFamily::Gnu { vxworks }` or `ToolFamily::Clang { emscripten }`.
    Basically, I'm trying to say that even with `#[non_exhaustive]` on the enum, it'd still be a semantic breaking change to add a new variant.
- Methods allows us to do the lookup more lazily. E.g. right now we don't use the `zig_cc` variable internally, so we might actually want to call `is_zig_cc` lazily instead when the user asks, but that would be harder if we exposed `family`.

NobodyXu:
That's indeed true, we might want to reorder `ToolFamily` later...So perhaps we should close this PR for now and instead has more `is_xxx` functions?

madsmtm:
> So perhaps we should close this PR for now and instead has more `is_xxx` functions?

I'd prefer that, yes.

NobodyXu:
Sorry @MarijnS95, thank you for your work but we are still too worried about the potential API breakage/subpar API, we will close this PR for now and I will cut a new release.

MarijnS95:
Ugh, thanks for nothing. If I had known this wouldn't be merged - closed without proper discussion even - I would have at the very least reverted #1357 again. Now we're stuck with `is_like_clang_cl()` for eternity, which has more vague semantics than any of the counter-arguments that were given against `ToolFamily` :disappointed:.

In the end in `ring` we don't care about this function at all anymore.  Rather, we want to know if the compiler is specifically `cl.exe`.  For now our only way to check that is `is_like_msvc() && !is_like_clang_cl()` but from the above it's completely uncertain if those semantics will hold forever.

Any of the `is_like_*()` functions don't clarify in what _context_ they consider like-ness.  Support for arguments?  The way they compile (i.e. using MSVC or LLVM stack)?  `ToolFamily` carries this load.

---

> * I'm not certain that `Msvc { clang_cl }` is the right nesting, it could also be e.g. `Clang { cl }`.

Sure, we could have debated that before jumping the gun.

> * I'm unsure if these three families are really the right separation, e.g. we might want `ToolFamily::VxWorks` or `ToolFamily::Emscripten` one day, but if we expose `ToolFamily` we're forced to put those under `ToolFamily::Gnu { vxworks }` or `ToolFamily::Clang { emscripten }`.

Can you explain why?  Do they currently resolve to `Gnu`/`Clang` variants?

>   Basically, I'm trying to say that even with `#[non_exhaustive]` on the enum, it'd still be a semantic breaking change to add a new variant.

If the wrong thing currently resolves to the wrong variant, sure...  I didn't expect to have to fix/clean up this popular library though.

> * Methods allows us to do the lookup more lazily. E.g. right now we don't use the `zig_cc` variable internally, so we might actually want to call `is_zig_cc` lazily instead when the user asks, but that would be harder if we exposed `family`.

In `ring` there was already a concern about this, and it further signifies the point that `is_like_*()` is vague already and its semantics might change at any time.  Likewise strange to have a `zig_cc` bool that's private but never used; likely a remnant from a previous workaround?

NobodyXu:
> Rather, we want to know if the compiler is specifically cl.exe. For now our only way to check that is is_like_msvc() && !is_like_clang_cl() but from the above it's completely uncertain if those semantics will hold forever.

Hmmm maybe we can add another `is_cl_exe` function?

NobodyXu:
Sorry about the closing, I thought that we just don't want to expose it now and didn't think more of this as it was late for me.

Perhaps I should wait a bit longer for more feedback instead of rushing

briansmith:
> Hmmm maybe we can add another `is_cl_exe` function?

It would be awesome to implement #882, to provide "use MSVC-provided clang-cl instead of cl.exe" and "use MSVC-provided clang-cl instead of cl.exe" APIs, which would also do the job of discovering where clang.exe/clang-cl.exe live so they don't have to be in `$PATH`.

NobodyXu:
Let me have a look, I think that's possible.

MarijnS95:
I wasn't aware of #882, moving the burden to switch compilers (and all complication that comes with detecting it, rather than pushing it out to users via documentation that encourages them to modify `PATH`) into `cc-rs` sounds like an ideal solution to me. Keeping in mind compatibility between `clang-cl`/`cl.exe` that cannot be achieved with `clang` - think of it like switching between the MSVC and LLVM toolchain while using the correct/compatible tool (to not have to modify users' `CFLAGS` etc).

Let's investigate that before adding `is_cl_exe()` etc.

Issue #1360, opened 2025-01-11
Error with cargo zigbuild


phoepsilonix:
Error with cargo zigbuild(cc-rs 1.2.8)

Example
- Add mimalloc-rust to Hello World to satisfy the dependency with cc-rs.
- Add the --target option to cargo zigbuild and compile.

```sh
RUSTFLAGS="" cargo zigbuild -r --message-format=json-render-diagnostics --target aarch64-unknown-linux-gnu
```
```sh
RUSTFLAGS="" cargo zigbuild -r --message-format=json-render-diagnostics --target x86_64-unknown-linux-gnu
```

The following error occurs.
```sh
  cargo:warning=error: Unknown Clang option: '--'

  --- stderr


  error occurred in cc-rs: Command LC_ALL="C" "/home/phoepsilonix/.cache/cargo-zigbuild/0.19.7/zigcc-x86_64-unknown-linux-gnu-7d27.sh" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "--target=x86_64-unknown-linux-gnu" "-I" "./mimalloc/include" "-I" "./mimalloc/src" "-DMI_BUILD_SHARED=0" "-DMI_DEBUG_FUL=0" "-DMI_DEBUG=0" "-o" "/home/phoepsilonix/work/myrust/target/x86_64-unknown-linux-gnu/release/build/mimalloc-rust-sys-752dde8e6a6b38c9/out/96c74dbd1b51cfc5-static.o" "-c" "--" "./mimalloc/src/static.c" with args zigcc-x86_64-unknown-linux-gnu-7d27.sh did not execute successfully (status code exit status: 1).
```

I tried to fix this by adjusting the zig_cc rule. #1361

polarathene:
Resolved via https://github.com/rust-lang/cc-rs/pull/1361 ?

NobodyXu:
Thanks, yes, I believe it's resolved by that PR, feel free to reopen/open another issue if it still occurs

PR #1361, opened 2025-01-11
Adjusted zig cc judgment and avoided zigbuild errors(#1360)

phoepsilonix:
Adjusted is_zig_cc judgment and adjusted how supports_path_delimiter is handled.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1800,7 +1800,10 @@ impl Build {
             cmd.args(self.asm_flags.iter().map(std::ops::Deref::deref));
         }
 
-        if compiler.supports_path_delimiter() && !is_assembler_msvc {
+        if compiler.supports_path_delimiter()
+            && !is_assembler_msvc
+            && matches!(compiler.family, ToolFamily::Clang { zig_cc: false })
+        {
             // #513: For `clang-cl`, separate flags/options from the input file.
             // When cross-compiling macOS -> Windows, this avoids interpreting
             // common `/Users/...` paths as the `/U` flag and triggering
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1800,13 +1800,7 @@ impl Build {
             cmd.args(self.asm_flags.iter().map(std::ops::Deref::deref));
         }
 
-        if compiler.supports_path_delimiter()
-            && !is_assembler_msvc
-            && matches!(
-                compiler.family,
-                ToolFamily::Clang { zig_cc: false } | ToolFamily::Msvc { clang_cl: true }
-            )
-        {
+        if compiler.supports_path_delimiter() && !is_assembler_msvc {
             // #513: For `clang-cl`, separate flags/options from the input file.
             // When cross-compiling macOS -> Windows, this avoids interpreting
             // common `/Users/...` paths as the `/U` flag and triggering
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1802,7 +1802,7 @@ impl Build {
 
         if compiler.supports_path_delimiter()
             && !is_assembler_msvc
-            && matches!(compiler.family, ToolFamily::Clang { zig_cc: false })
+            && matches!(compiler.family, ToolFamily::Clang { zig_cc: false } | ToolFamilt::Msvc { clang_cl: true })
         {
             // #513: For `clang-cl`, separate flags/options from the input file.
             // When cross-compiling macOS -> Windows, this avoids interpreting
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1802,7 +1802,10 @@ impl Build {
 
         if compiler.supports_path_delimiter()
             && !is_assembler_msvc
-            && matches!(compiler.family, ToolFamily::Clang { zig_cc: false } | ToolFamilt::Msvc { clang_cl: true })
+            && matches!(
+                compiler.family,
+                ToolFamily::Clang { zig_cc: false } | ToolFamily::Msvc { clang_cl: true }
+            )
         {
             // #513: For `clang-cl`, separate flags/options from the input file.
             // When cross-compiling macOS -> Windows, this avoids interpreting
--- a/src/tool.rs
+++ b/src/tool.rs
@@ -103,6 +103,12 @@ impl Tool {
             )
             .map(|o| String::from_utf8_lossy(&o).contains("ziglang"))
             .unwrap_or_default()
+                || {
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -1800,7 +1800,10 @@ impl Build {
             cmd.args(self.asm_flags.iter().map(std::ops::Deref::deref));
         }
 
-        if compiler.supports_path_delimiter() && !is_assembler_msvc {
+        if compiler.supports_path_delimiter()
+            && !is_assembler_msvc
+            && matches!(compiler.family, ToolFamily::Clang { zig_cc: false })
```
NobodyXu, line 1805:
```suggestion
            && matches!(compiler.family, ToolFamily::Clang { zig_cc: false } | ToolFamilt::Msvc { clang_cl: true })
```

NobodyXu (changes requested):
Thank you!

Just one requested change

NobodyXu (approved):

src/lib.rs
phoepsilonix: Thank you!

src/tool.rs
```diff
@@ -103,6 +103,12 @@ impl Tool {
             )
             .map(|o| String::from_utf8_lossy(&o).contains("ziglang"))
             .unwrap_or_default()
+                || {
+                    match path.file_name().map(OsStr::to_string_lossy) {
+                        Some(fname) => fname.contains("zig"),
+                        _ => false,
+                    }
+                }
```
NobodyXu, line 111: I'm bit confused by this, it looks like this function is unused?

NobodyXu (changes requested):

NobodyXu:
It seems that even clang itself may not support `--`.

I would open another PR reverting changes to use that for clang

src/tool.rs
phoepsilonix: Previously, is_zig_cc would tend to judge a path as simply a clang environment, not a zig environment. 
And since the judgement of whether zig was included in a path like `zigcc-x86_64-unknown-linux-gnu-7d27.sh` was only possible if `detect_family` could not be used, there were some environments that were not judged as zig even though they were zig environments.

NobodyXu: I get this, but it seems that the closure is unused to me?

I don't see if used anywhere and not sure how it passes the test, which is why I'm confused

phoepsilonix: In my environment, the arch-derived Manjaro Linux (x86_64), without this fix, the result is Clang { zig_cc: false}, and the same error as #1360 continues to occur.

phoepsilonix: Up until now, the check was made based on whether the output of $CC --version included ziglang. I also added a check to see if the path included zig, but is something wrong?
By making this change, the check results for the cargo zigbuild command in my environment change to zig_cc: true.

yutannihilation: @NobodyXu 
> the closure is unused

Just in case your eyes are confused by `|| { }`, this isn't a closure. This is a `||` (or) and a `{ }` (block).

NobodyXu: Oh damn thanks... I am definitely confused, it just looks so much like a closure, IDK how rustc is able to parse that

src/lib.rs
```diff
@@ -1800,7 +1800,13 @@ impl Build {
             cmd.args(self.asm_flags.iter().map(std::ops::Deref::deref));
         }
 
-        if compiler.supports_path_delimiter() && !is_assembler_msvc {
+        if compiler.supports_path_delimiter()
+            && !is_assembler_msvc
+            && matches!(
+                compiler.family,
+                ToolFamily::Clang { zig_cc: false } | ToolFamily::Msvc { clang_cl: true }
+            )
```
NobodyXu, line 1808: I have fixed this in #1364 so this change is no longer necessary and can be reverted

NobodyXu (changes requested):

NobodyXu:
One last change requested and once resolved I will merge this PR!

src/tool.rs
phoepsilonix: I intend to do the same process as this.
```diff
diff --git a/src/tool.rs b/src/tool.rs
index 8793257..3aaeffc 100644
--- a/src/tool.rs
+++ b/src/tool.rs
@@ -94,7 +94,7 @@ impl Tool {
         cargo_output: &CargoOutput,
         out_dir: Option<&Path>,
     ) -> Self {
-        fn is_zig_cc(path: &Path, cargo_output: &CargoOutput) -> bool {
+        fn is_zig_cc_from_stdout(path: &Path, cargo_output: &CargoOutput) -> bool {
             run_output(
                 Command::new(path).arg("--version"),
                 path,
@@ -103,12 +103,17 @@ impl Tool {
             )
             .map(|o| String::from_utf8_lossy(&o).contains("ziglang"))
             .unwrap_or_default()
-                || {
-                    match path.file_name().map(OsStr::to_string_lossy) {
-                        Some(fname) => fname.contains("zig"),
-                        _ => false,
-                    }
-                }
+        }
+
+        fn is_zig_cc_from_path(path: &Path) -> bool {
+            match path.file_name().map(OsStr::to_string_lossy) {
+                Some(fname) => fname.contains("zig"),
+                _ => false,
+            }
+        }
+
+        fn is_zig_cc(path: &Path, cargo_output: &CargoOutput) -> bool {
+            is_zig_cc_from_stdout(path, cargo_output) || is_zig_cc_from_path(path)
         }
 
         fn guess_family_from_stdout(
```

NobodyXu: Thanks that's much clearer for me

src/lib.rs
phoepsilonix: With your fix, this check is no longer necessary, right? I reverted and pushed again.

NobodyXu: Yes, all change to this file can be reverted, not just my suggestion

NobodyXu: cc @phoepsilonix can you revert all changes made to that file please?

phoepsilonix: @NobodyXu 
I was able to confirm that it works here as well.
Thanks for the fix.

phoepsilonix: All that remains is to fix the zig cc checkpoint. Maybe it would be better to change the commit log.
Is this okay? I'll leave the rest to you. :smile:

NobodyXu (approved):

src/lib.rs
NobodyXu: Thanks, I will merge and cut a new release

PR #1363, opened 2025-01-12
build: Don't pass inherited PGO flags to GNU compilers

### LINKED ISSUE #1354: Propagated coverage flags cause `gcov` link errors
When compiling code like the [example](https://docs.rs/cc/1.2.7/cc/#example) and `RUSTFLAGS=-Cprofile-generate=$PWD`, it fails to link with errors like this:

    foo.c:1:(.text._sub_I_00100_0+0xf): undefined reference to `__gcov_init'

(and many other `__gcov_*` symbols)

That's with a default `CC` that is GCC; it's okay with Clang because that uses the same profiling library as Rust.

Maybe `-Cprofile-generate` (and `-Cprofile-use`) should not be translated for `ToolFamily::Gnu` at all? It seems unlikely that someone would use this and try to juggle both LLVM `profraw` and GCC `gcov` files, even if the link error is fixed.

mrkajetanp:
Clang and GNU use different PGO libraries, so PGO flags inherited from rustc should not automatically be passed to GNU compilers.

Fixes #1354

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -174,26 +174,109 @@ impl<'this> RustcCodegenFlags<'this> {
             }
         };
 
+        let clang_or_gnu =
+            matches!(family, ToolFamily::Clang { .. }) || matches!(family, ToolFamily::Gnu { .. });
+
+        // Flags shared between clang and gnu
+        if clang_or_gnu {
+            // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mbranch-protection
+            if let Some(value) = self.branch_protection {
+                push_if_supported(
+                    format!("-mbranch-protection={}", value.replace(",", "+")).into(),
+                );
+            }
+            // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mcmodel
+            if let Some(value) = self.code_model {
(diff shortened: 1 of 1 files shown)


NobodyXu (approved):
Thank you LGTM


cc @madsmtm can you double check this please?

src/flags.rs
```diff
@@ -174,26 +174,106 @@ impl<'this> RustcCodegenFlags<'this> {
...
+                } else {
+                    "-mno-soft-float"
+                };
+                push_if_supported(cc_flag.into());
+            }
+        }
+
+        // Compiler-exclusive flags
+        match family {
+            ToolFamily::Clang { .. } => {
```
madsmtm, line 276: Please document here that while GCC supports the `-fprofile-generate=` and `-fprofile-use=` flags, we don't use them because they'll likely be incompatible with the Rust flag (while both Clang and Rust use LLVM under the hood, so they'll likely be compatible).

madsmtm (approved):
Approve of the approach, only concern is code comment (which is non-blocking if we're in a hurry).

src/flags.rs
mrkajetanp: Ah I did think about that, sure I'll add it in

mrkajetanp: there you go

PR #1367, opened 2025-01-15
Add prefer_clang_cl_over_msvc

NobodyXu:


src/lib.rs
```diff
@@ -2748,7 +2765,7 @@ impl Build {
             traditional
         };
 
-        let cl_exe = self.windows_registry_find_tool(&target, "cl.exe");
+        let cl_exe = self.windows_registry_find_tool(&target, msvc);
```
NobodyXu, line 2768: cc @ChrisDenton I would need some help on how to find `clang-cl`, here I just pass `clang-cl.exe` to `windows_registry::find_tool` if user prefers clang-cl over msvc, hoping it would find it, not sure if it is correct.

ChrisDenton: That might find it if it's installed with Visual Studio but it could be installed separately. We should more strongly prefer the environment for `clang-cl` (e.g. checking `PATH` as well).

NobodyXu: Should add I a dedicated `PATH` checking logic for `clang-cl.exe`?

Or should I add that to `windows_registry::find_tool`?

ChrisDenton: I think adding dedicated `PATH` checking logic is the safest option.

It'd be nice if `find_tool` where consistent but it's possible people may use it to specifically find the VS installed version. So unless someone complains about the `find_tool` behaviour, I'm not sure if it's worth the (admittedly small) risk of breakage.

briansmith: > That might find it if it's installed with Visual Studio but it could be installed separately. We should more strongly prefer the environment for clang-cl (e.g. checking PATH as well).

If it is installed separately then the user should specify `CC`/`TARGET_CC` when building. Otherwise, I think the behavior should be "use the clang-cl provided with the version of MSVC that we were about to use."

madsmtm:
> cc @madsmtm I am really busy recently and I don't think I can work on this for the next week, if you have time, can you take up this please?

I'd rather not, I don't feel familiar enough with MSVC to feel confident in anything relating to that.

NobodyXu:
cc @ChrisDenton hello, can you take over this PR please?

I don't have time recently to continue push this.

ChrisDenton:
Sure, I've added it to my queue.

NobodyXu:
Thank you

MarijnS95:
Is this possibly missing [`Fixes #882`](https://github.com/rust-lang/cc-rs/issues/882)? 🙏

briansmith:
(a) f the user has set `TARGET_CC`/`CC` (whichever is relevant) to something other than MSVC, then calling `prefer_clang_cl_over_msvc` should do nothing. For example, if they've chosen to use clang or (for some reason) GCC, or clang-cl, then prefer_clang_cl_over_msvc shouldn't override that choice.

(b) IDK what to do when the user set `TARGET_CC`/`CC` explicitly to MSVC. For my use case the best thing to do would be to choose the clang-cl that's bundled with that install of MSVC.

If the user set `TARGET_CC`/`CC` to some version of clang-cl, then I guess we should use that version of clang-cl. But this is handled by (a).

NobodyXu:
Closing in favor of #1516

PR #1371, opened 2025-01-17
Fix CC_FORCE_DISABLE=0 evaluating to true

madsmtm:
Found by @saleemrashid in https://github.com/rust-lang/cc-rs/pull/1292#pullrequestreview-2557590639.

Tested by running `CC_FORCE_DISABLE=0 cargo test`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4157,10 +4157,9 @@ fn is_disabled() -> bool {
         match std::env::var_os("CC_FORCE_DISABLE") {
             // Not set? Not disabled.
             None => false,
-            // Respect `CC_FORCE_DISABLE=0` and some simple synonyms.
-            Some(v) if &*v != "0" && &*v != "false" && &*v != "no" => false,
-            // Otherwise, we're disabled. This intentionally includes `CC_FORCE_DISABLE=""`
-            Some(_) => true,
+            // Respect `CC_FORCE_DISABLE=0` and some simple synonyms, otherwise
+            // we're disabled. This intentionally includes `CC_FORCE_DISABLE=""`
+            Some(v) => &*v != "0" && &*v != "false" && &*v != "no",
         }
     }
     match val {


src/lib.rs
```diff
@@ -4158,7 +4158,7 @@ fn is_disabled() -> bool {
             // Not set? Not disabled.
             None => false,
             // Respect `CC_FORCE_DISABLE=0` and some simple synonyms.
-            Some(v) if &*v != "0" && &*v != "false" && &*v != "no" => false,
+            Some(v) if &*v == "0" || &*v == "false" || &*v == "no" => false,
```
ChrisDenton, line 4161: As a quick fix this looks fine but is there a better way to write this? It does feel a bit like symbol soup and obviously the logic error was missed the first time so it'd be great if it could be made clearer.

madsmtm: I tried `matches!(v, "0" | "false" | "no")`, but that doesn't work because this is an `OsString`. Can't think of anything else.

ChrisDenton: Hm, an `as_deref()` would help avoid the `&*` at least by giving us an `OsStr`. But maybe we can do a bit better at the cost of some verbosity?

```rust
fn compute_is_disabled() -> bool {
    match std::env::var_os("CC_FORCE_DISABLE") {
        // Not set? Not disabled.
        None => false,
        Some(cc_force_disable) => match cc_force_disable.to_str() {
            // Respect `CC_FORCE_DISABLE=0` and some simple synonyms.
            Some("0" | "false" | "no") => false,
            // Otherwise, we're disabled. This intentionally includes `CC_FORCE_DISABLE=""`
            _ => true,
        },
    }
}
```

NobodyXu (changes requested):

src/lib.rs
madsmtm: I'll leave that decision to @NobodyXu.

src/lib.rs
```diff
@@ -4157,10 +4157,9 @@ fn is_disabled() -> bool {
...
             // Not set? Not disabled.
             None => false,
-            // Respect `CC_FORCE_DISABLE=0` and some simple synonyms.
-            Some(v) if &*v != "0" && &*v != "false" && &*v != "no" => false,
-            // Otherwise, we're disabled. This intentionally includes `CC_FORCE_DISABLE=""`
-            Some(_) => true,
+            // Respect `CC_FORCE_DISABLE=0` and some simple synonyms, otherwise
+            // we're disabled. This intentionally includes `CC_FORCE_DISABLE=""`
+            Some(v) => &*v != "0" && &*v != "false" && &*v != "no",
         }
```
madsmtm, line 4163: This now matches `Build::getenv_boolean`

NobodyXu (approved):
Thank you!

src/lib.rs
NobodyXu: I think it looks ok now, will merge and release

PR #1373, opened 2025-01-17
CC support for `wasm32-wali-linux-musl` Tier-3 Linux target

arjunr2:
To support the new Tier-3 target `wasm32-wali-linux-musl` within the rust compiler: rust-lang/rust#135651

Minimally, this just supports providing appropriate musl library linking and flags for the target

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1459,12 +1459,20 @@ impl Build {
                     .print_metadata(&format_args!("cargo:rustc-link-lib={}", stdlib.display()));
             }
             // Link c++ lib from WASI sysroot
-            if target.os == "wasi" {
-                if let Ok(wasi_sysroot) = self.wasi_sysroot() {
+            if target.arch == "wasm32" {
+                if target.os == "wasi" {
+                    if let Ok(wasi_sysroot) = self.wasi_sysroot() {
+                        self.cargo_output.print_metadata(&format_args!(
+                            "cargo:rustc-flags=-L {}/lib/{} -lstatic=c++ -lstatic=c++abi",
+                            Path::new(&wasi_sysroot).display(),
+                            self.get_raw_target()?
+                        ));
+                    }
+                } else if target.os == "linux" {
+                    let musl_sysroot = self.wasm_musl_sysroot().unwrap();
                     self.cargo_output.print_metadata(&format_args!(
-                        "cargo:rustc-flags=-L {}/lib/{} -lstatic=c++ -lstatic=c++abi",
-                        Path::new(&wasi_sysroot).display(),
-                        self.get_raw_target()?
+                        "cargo:rustc-flags=-L {}/lib -lstatic=c++ -lstatic=c++abi",
+                        Path::new(&musl_sysroot).display(),
                     ));
                 }
             }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1363,6 +1363,12 @@ impl Build {
                         target
                     ));
                 }
+            } else if target.contains("wasm32") && target.contains("linux") {
```
NobodyXu, line 1366: Would it be better to do an exact match?

```suggestion
            } else if target == "wasm32-wali-linux-musl" {
```

src/lib.rs
```diff
@@ -2178,6 +2184,25 @@ impl Build {
                         }
 
                         cmd.push_cc_arg(format!("--target={}", target).into());
+                    } else if target.contains("wasm32") && target.contains("linux") {
```
NobodyXu, line 2187: Same here

src/lib.rs
```diff
@@ -2178,6 +2184,25 @@ impl Build {
...
+                        ] {
+                            cmd.push_cc_arg(format!("-m{}", x).into());
+                        }
+                        for x in &["wasm-exceptions", "declspec"] {
+                            cmd.push_cc_arg(format!("-f{}", x).into());
+                        }
+                        let musl_sysroot = self.wasm_musl_sysroot().unwrap();
+                        cmd.push_cc_arg(
+                            format!("--sysroot={}", Path::new(&musl_sysroot).display()).into(),
+                        );
```
NobodyXu, line 2205: cc @madsmtm do you think we can use gen-target-info to auto generate this information?

NobodyXu (changes requested):

arjunr2:
Having these conditions rely on just `wasm32` and `linux` seem to make more sense to me. WALI/musl is just a specific ABI/env, and future non-WALI/non-musl targets can then reuse them. Is there a reason for why using the full target is better?

madsmtm:
I'd also really rather prefer not using the full target if we can.

src/lib.rs
madsmtm: Hmm, the `--target` at least definitely looks odd?

src/lib.rs
madsmtm: I'd prefer to avoid `.contains`, and use `target.arch == "wasm32" && target.os == "linux"` (or similar)

src/lib.rs
```diff
@@ -2178,6 +2184,25 @@ impl Build {
                         }
 
                         cmd.push_cc_arg(format!("--target={}", target).into());
+                    } else if target.contains("wasm32") && target.contains("linux") {
+                        // wasm32-linux currently uses the LLVM-18 wasi threads target
+                        cmd.push_cc_arg("--target=wasm32-wasi-threads".into());
```
madsmtm, line 2189: The use of `--target` seems odd? It is not clear why we are passing a different target here than the `llvm_target` in `rustc`'s `Target`, which is set to `wasm32-wasi`?

arjunr2: This configuration is similar to `wasm32-wasip1-threads`, which uses the  `wasm32-wasi` llvm target. I am not sure using llvm_target as `wasm32-wasi-threads` breaks any pattern matching within rustc, so I stuck to what the former currently does.

arjunr2:
@NobodyXu @madsmtm There seems to be failing checks here due to some `rustix` logic on arm unrelated to this PR. Can you confirm?

src/lib.rs
```diff
@@ -2179,8 +2187,25 @@ impl Build {
...
                         }
+                    } else if target.arch == "wasm32" && target.os == "linux" {
+                        for x in &[
+                            "atomics",
+                            "bulk-memory",
+                            "mutable-globals",
+                            "sign-ext",
+                            "exception-handling",
+                        ] {
+                            cmd.push_cc_arg(format!("-m{}", x).into());
```
NobodyXu, line 2198:
```suggestion
                            cmd.push_cc_arg(format!("-m{x}").into());
```

src/lib.rs
```diff
@@ -2179,8 +2187,25 @@ impl Build {
...
+                        }
+                        for x in &["wasm-exceptions", "declspec"] {
+                            cmd.push_cc_arg(format!("-f{}", x).into());
```
NobodyXu, line 2201:
```suggestion
                            cmd.push_cc_arg(format!("-f{x}").into());
```

NobodyXu (changes requested):
Thanks just two minor fmt changes

madsmtm:
> Probably worth lodging a ticket upstream

Filed https://github.com/bytecodealliance/rustix/pull/1407

NobodyXu (approved):
Thanks LGTM

Retried CI with rustix 1.0.2 and it has passed it

NobodyXu:
Thank you @madsmtm for fixing the rustix error so quick!

PR #455, opened 2019-11-05
Detect clang cl driver mode

kvark:
Fixes #454

PR #1378, opened 2025-01-31
Include wrapper args. in `stdout` family heuristics to restore classifying `clang --driver-mode=cl` as `Msvc { clang_cl: true }`

ErichDonGubler:
I'm not sure which version exactly broke this, but between 1.0.89 and latest releases, this case had broken for the primary tool family heuristic. This forces Firefox/[`mozilla-central`](https://hg.mozilla.org/mozilla-central) to work around breakage in a `cc` upgrade a la [D236305](https://phabricator.services.mozilla.com/D236305). We introduced it in https://github.com/rust-lang/cc-rs/pull/455, so we'd like to keep it working!

### DIFF
--- a/src/tool.rs
+++ b/src/tool.rs
@@ -16,6 +16,8 @@ use crate::{
     Error, ErrorKind, OutputKind,
 };
 
+pub(crate) type CompilerFamilyLookupCache = HashMap<Box<Path>, ToolFamily>;
+
 /// Configuration used to represent an invocation of a C compiler.
 ///
 /// This can be used to figure out what compiler is in use, what the arguments
@@ -40,7 +42,7 @@ pub struct Tool {
 impl Tool {
     pub(crate) fn new(
         path: PathBuf,
-        cached_compiler_family: &RwLock<HashMap<Box<Path>, ToolFamily>>,
+        cached_compiler_family: &RwLock<CompilerFamilyLookupCache>,
         cargo_output: &CargoOutput,
         out_dir: Option<&Path>,
     ) -> Self {
@@ -57,7 +59,7 @@ impl Tool {
     pub(crate) fn with_args(
         path: PathBuf,
         args: Vec<String>,
-        cached_compiler_family: &RwLock<HashMap<Box<Path>, ToolFamily>>,
+        cached_compiler_family: &RwLock<CompilerFamilyLookupCache>,
         cargo_output: &CargoOutput,
         out_dir: Option<&Path>,
     ) -> Self {
@@ -90,7 +92,7 @@ impl Tool {
         path: PathBuf,
         args: Vec<String>,
         cuda: bool,
-        cached_compiler_family: &RwLock<HashMap<Box<Path>, ToolFamily>>,
+        cached_compiler_family: &RwLock<CompilerFamilyLookupCache>,
         cargo_output: &CargoOutput,
         out_dir: Option<&Path>,
     ) -> Self {
@@ -238,7 +240,7 @@ impl Tool {
                 Some(fname) if fname.contains("clang") => {
                     let is_clang_cl = args
                         .iter()
-                        .find_map(|a| a.strip_prefix("--driver-mode=") == Some("cl"))
+                        .any(|a| a.strip_prefix("--driver-mode=") == Some("cl"));
                     if is_clang_cl {
                         ToolFamily::Msvc { clang_cl: true }
                     } else {
--- a/src/tool.rs
+++ b/src/tool.rs
@@ -46,7 +46,7 @@ impl Tool {
     ) -> Self {
         Self::with_features(
             path,
-            None,
+            vec![],
             false,
             cached_compiler_family,
             cargo_output,
@@ -54,16 +54,16 @@ impl Tool {
         )
     }
 
-    pub(crate) fn with_clang_driver(
+    pub(crate) fn with_args(
         path: PathBuf,
-        clang_driver: Option<&str>,
+        args: Vec<String>,
         cached_compiler_family: &RwLock<HashMap<Box<Path>, ToolFamily>>,
         cargo_output: &CargoOutput,
         out_dir: Option<&Path>,
     ) -> Self {
         Self::with_features(
             path,
-            clang_driver,
+            args,
             false,
             cached_compiler_family,
             cargo_output,
@@ -88,7 +88,7 @@ impl Tool {
 
     pub(crate) fn with_features(
         path: PathBuf,
-        clang_driver: Option<&str>,
+        args: Vec<String>,
         cuda: bool,
         cached_compiler_family: &RwLock<HashMap<Box<Path>, ToolFamily>>,
         cargo_output: &CargoOutput,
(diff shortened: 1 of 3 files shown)


madsmtm (commented):
Not too familiar with our compiler detection, but fine with it if @NobodyXu is.

I'm sorry that it broke, if you'd really like it to not break in the future, I'd suggest you add a test ;)

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Include compiler wrapper arguments (i.e., `CC=compiler arg1 arg2`) in the classification of compilers' tool family. This fixes cases like `clang --driver-mode=cl` being incorrectly detected as needing `clang`-like arguments, rather than `cl`-like arguments. [#????](https://github.com/rust-lang/cc-rs/????).
+
```
NobodyXu, line 10:
```suggestion
```

This is automatically generated/overwritten in release PR, created by release-plz

src/tool.rs
```diff
@@ -16,6 +16,8 @@ use crate::{
     Error, ErrorKind, OutputKind,
 };
 
+pub(crate) type CompilerFamilyLookupCache = HashMap<(Box<Path>, Vec<String>), ToolFamily>;
```
NobodyXu, line 19: I prefer to store program and args together in one box to reduce size of key and avoid repeated allocation:

```suggestion
pub(crate) type CompilerFamilyLookupCache = HashMap<Box<[OsStr]>, ToolFamily>;
```

src/tool.rs
```diff
@@ -207,25 +214,29 @@ impl Tool {
...
+            let cache_key = (Box::from(path), args.to_vec());
+            if let Some(family) = cached_compiler_family.read().unwrap().get(&cache_key) {
                 return Ok(*family);
             }
 
-            let family = detect_family_inner(path, cargo_output, out_dir)?;
+            let family = {
+                let (path, args) = &cache_key;
+                detect_family_inner(path, args, cargo_output, out_dir)?
+            };
```
NobodyXu, line 231: I think we already have `path` and `args` as arguments, and since they are reference they should be still useable?

```suggestion
            let family = detect_family_inner(path, args, cargo_output, out_dir)?;
```

NobodyXu (changes requested):
Thank you!

I have some feedbacks for the code itself, I think putting args in caching is the right approach

src/tool.rs
ErichDonGubler: Fixed as suggested with the latest force push; interdiff link: https://github.com/rust-lang/cc-rs/compare/c5c1a91ad6532e70081ead71ffc81485df116932..ffe0e4164b4d6ac31c62aed7888651d3d15f0c4f

src/tool.rs
ErichDonGubler: thought: That suggestion does lose some type-level guarantees about the arguments being `String`s. I don't see a concrete problem with it, but I'll give it a shot!

Implemented with latest force push. Interdiff link: https://github.com/rust-lang/cc-rs/compare/c5c1a91ad6532e70081ead71ffc81485df116932..ffe0e4164b4d6ac31c62aed7888651d3d15f0c4f

CHANGELOG.md
ErichDonGubler: Removed with a force-push; interdiff link: https://github.com/rust-lang/cc-rs/compare/ffe0e4164b4d6ac31c62aed7888651d3d15f0c4f..17c805a543f09e60f8c25296f322c9dd12c319bd

ErichDonGubler:
> Not too familiar with our compiler detection, but fine with it if @NobodyXu is.
> 
> I'm sorry that it broke, if you'd really like it to not break in the future, I'd suggest you add a test ;)

❤️ I think this is a pitch-perfect response. I'm not familiar with how `cc` does testing, so I scoped my efforts towards just fixing things, but I definitely am interested in avoiding future breakage too, so I'll roll up my sleeves and see if I can write a meaningful test.

src/tool.rs
```diff
@@ -207,25 +214,30 @@ impl Tool {
...
+                .iter()
+                .map(|p| p.as_os_str().to_owned())
+                .chain(args.iter().map(|a| OsStr::new(a).to_owned()))
+                .collect();
+            if let Some(family) = cached_compiler_family.read().unwrap().get(&cache_key) {
                 return Ok(*family);
             }
 
-            let family = detect_family_inner(path, cargo_output, out_dir)?;
+            let family = { detect_family_inner(path, args, cargo_output, out_dir)? };
```
NobodyXu, line 232: I don't think this parenthese is necessary:

```suggestion
            let family = detect_family_inner(path, args, cargo_output, out_dir)?;
```

src/tool.rs
```diff
@@ -16,6 +16,8 @@ use crate::{
     Error, ErrorKind, OutputKind,
 };
 
+pub(crate) type CompilerFamilyLookupCache = HashMap<Box<[OsString]>, ToolFamily>;
```
NobodyXu, line 19:
```suggestion
pub(crate) type CompilerFamilyLookupCache = HashMap<Box<[Box<OsStr>]>, ToolFamily>;
```

src/tool.rs
```diff
@@ -207,25 +214,30 @@ impl Tool {
...
             }
         }
-        let detect_family = |path: &Path| -> Result<ToolFamily, Error> {
-            if let Some(family) = cached_compiler_family.read().unwrap().get(path) {
+        let detect_family = |path: &Path, args: &[String]| -> Result<ToolFamily, Error> {
+            let cache_key = [path]
+                .iter()
+                .map(|p| p.as_os_str().to_owned())
+                .chain(args.iter().map(|a| OsStr::new(a).to_owned()))
+                .collect();
```
NobodyXu, line 227:
```suggestion
            let cache_key = [path.as_os_str()]
                .iter()
                .chain(args.iter().map(OsStr::new)
                .map(|s| s.to_owned().into_boxed_slice())
                .collect();
```

NobodyXu (changes requested):

NobodyXu:
> I'm actually not sure how to test this without having a fully functional `clang --driver-mode=cl` in CI

Maybe you can try adding a new shim here?

https://github.com/rust-lang/cc-rs/blob/65b4d9a263edd0484d45edaaccb48724de23c881/tests/support/mod.rs#L69

We currently has a gcc-shim binary in cc-rs https://github.com/rust-lang/cc-rs/blob/main/src/bin/gcc-shim.rs

Maybe we could have another clang-shim that does the testing?

src/tool.rs
ErichDonGubler: D'oh, you're right. Interdiff link: https://github.com/rust-lang/cc-rs/compare/17c805a543f09e60f8c25296f322c9dd12c319bd..5d2908124311708917ceee5a59a4dfe01f35d5ff

src/tool.rs
ErichDonGubler: Implemented as suggested; interdiff link: https://github.com/rust-lang/cc-rs/compare/17c805a543f09e60f8c25296f322c9dd12c319bd..5d2908124311708917ceee5a59a4dfe01f35d5ff

src/tool.rs
ErichDonGubler: 🤔 This suggestion doesn't compile, but I was able to make this compile:

```suggestion
            let cache_key = [path.as_os_str()]
                .iter()
                .cloned()
                .chain(args.iter().map(osstr::new))
                .map(into::into)
                .collect();
```

Pushed; interdiff link: https://github.com/rust-lang/cc-rs/compare/17c805a543f09e60f8c25296f322c9dd12c319bd..5d2908124311708917ceee5a59a4dfe01f35d5ff

NobodyXu (approved):
Thanks, LGTM!

For the test, you can add it via a separate PR

ErichDonGubler:
> Thanks, LGTM!
> 
> If you are still working on the test, you can open a separate PR for it.

Yes, that would be good! I'm doing a bit more refactoring to accommodate a new shim in `Test` than I think is appropriate for this fix, but I _have_ written a test locally that fails before and succeeds after this PR. So, I think I'm on the right track for a good follow-up PR. Thanks for the coaching!

One question: When this lands, is there a rough timeline for a release to come out?

madsmtm:
Re release, see https://github.com/rust-lang/cc-rs/pull/1377. If not in that, then next week.

PR #1379, opened 2025-01-31
Constrain `-Clto` and `-Cembed-bitcode` flag inheritance to be `clang`-only

jieyouxu:
In https://github.com/rust-lang/rust/issues/136338 we noticed two unsupported inherited flag warnings when the underlying `cc` is `gcc` and not `clang`:

```
warning: rustc_llvm@0.0.0: Inherited flag "-flto=thin" is not supported by the currently used CC
warning: rustc_llvm@0.0.0: Inherited flag "-fembed-bitcode=all" is not supported by the currently used CC
```

This PR constraints `-Clto` and `-Cembed-bitcode` flag inheritance to be `clang`-only, compared the current behavior of inheriting the flags for both `clang` and `gcc`.

### Constrain `-Cembed-bitcode` flag inheritance to be `clang`-only

`embed-bitcode` refers to embedding LLVM bitcode, which is not a thing for `gcc`[^gimple].

This PR moves the `-Cembed-bitcode` flag inheritance from being shared by `gcc`/`clang` to be `clang`-*only*.

### Constrain `-Clto` flag inheritance to be `clang`-only

While `clang` supports `-flto={thin,fat}`, `gcc` does not support this form. Furthermore, there isn't really a 1-to-1 mapping between `clang`'s ThinLTO/FatLTO and `gcc`'s LTO.

We noticed this in https://github.com/rust-lang/rust/issues/136338 when trying to compile `rustc_llvm` but with the underlying `cc` being `gcc`:

```
warning: rustc_llvm@0.0.0: Inherited flag "-flto=thin" is not supported by the currently used CC
```

For instance, [`gcc` has *multiple* LTO-related / LTO-affecting flags][gcc-lto-flags]:

- [`-ffat-lto-objects` and `-fno-fat-lto-objects`][gcc-no-fat-lto-objects]. And its docs state:
    > `-fno-fat-lto-objects` improves compilation time over plain LTO, but requires the complete toolchain to be aware of LTO.
- [`-fwhole-program`][gcc-whole-program]
- [`-flto-partition`][gcc-lto-partition]
- [`-flto-compression-level`][gcc-lto-compression-level]
- ... and a few more.

[gcc-lto-flags]: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-flto
[gcc-no-fat-lto-objects]: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-ffat-lto-objects
[gcc-whole-program]: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-fwhole-program
[gcc-lto-partition]: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-flto-partition
[gcc-lto-compression-level]: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-flto-compression-level

This PR moves the `-Clto` flag inheritance from being shared by `gcc`/`clang` to be `clang`-*only*.

[^gimple]: I suppose if you squint super super hard, embedding GIMPLE bytecode somehow...?

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -223,11 +223,6 @@ impl<'this> RustcCodegenFlags<'this> {
                     push_if_supported(cc_flag.into());
                 }
             }
-            // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fembed-bitcode
-            if let Some(value) = self.embed_bitcode {
-                let cc_val = if value { "all" } else { "off" };
-                push_if_supported(format!("-fembed-bitcode={cc_val}").into());
-            }
             // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fno-omit-frame-pointer
             // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fomit-frame-pointer
             if let Some(value) = self.force_frame_pointers {
@@ -273,6 +268,12 @@ impl<'this> RustcCodegenFlags<'this> {
                 // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fprofile-use
                 if let Some(value) = self.profile_use {
                     push_if_supported(format!("-fprofile-use={value}").into());
+                }
+
+                // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fembed-bitcode
+                if let Some(value) = self.embed_bitcode {
+                    let cc_val = if value { "all" } else { "off" };
+                    push_if_supported(format!("-fembed-bitcode={cc_val}").into());
                 }
 
                 // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-flto
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -209,17 +209,6 @@ impl<'this> RustcCodegenFlags<'this> {
                     push_if_supported(format!("-mguard={cc_val}").into());
                 }
             }
-            // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-flto
-            if let Some(value) = self.lto {
-                let cc_val = match value {
-                    "y" | "yes" | "on" | "true" | "fat" => Some("full"),
-                    "thin" => Some("thin"),
-                    _ => None,
-                };
-                if let Some(cc_val) = cc_val {
-                    push_if_supported(format!("-flto={cc_val}").into());
-                }
-            }
             // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fPIC
             // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fPIE
             // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mdynamic-no-pic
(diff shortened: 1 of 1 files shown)


madsmtm (approved):
Looks good to me!

Similar in spirit to https://github.com/rust-lang/cc-rs/pull/1363, CC @mrkajetanp, why were these originally marked enabled on GCC too? Are there other options here that we are not considering that we should be?

madsmtm:
I'm gonna merge this immediately, because this PR is clearly correct. I'll look into the other options myself, and file a follow-up PR if needed.

jieyouxu:
> I'm gonna merge this immediately, because this PR is clearly correct. I'll look into the other options myself, and file a follow-up PR if needed.

There may also be cases that while a specific flag-form is supported for both `clang` and `gcc`, their exact semantics/behavior may not mirror each other. However, I haven't looked into that, just a remark.

madsmtm:
Filed https://github.com/rust-lang/cc-rs/pull/1380

mrkajetanp:
As originally written, clang and gcc flags were just all one group because if either of the compilers didn't support one you'd just get this warning but it wouldn't be an actual error. Somewhat useful for when someone was using a different compiler than they intended etc.
Now that it's been changed to be more granular it does make sense to move some of these to clang-only, I agree :)

src/flags.rs
```diff
@@ -285,6 +269,24 @@ impl<'this> RustcCodegenFlags<'this> {
...
+                if let Some(value) = self.lto {
+                    let cc_val = match value {
+                        "y" | "yes" | "on" | "true" | "fat" => Some("full"),
+                        "thin" => Some("thin"),
+                        _ => None,
+                    };
+                    if let Some(cc_val) = cc_val {
+                        push_if_supported(format!("-flto={cc_val}").into());
+                    }
+                }
```
NobodyXu, line 289: Maybe we can pass `-ffat-lto-objects` for gcc and `-fuse-linker-plugin` (or -fwhole-program`) @madsmtm

NobodyXu: Yeah but I think we should turn on LTO, if rustc has LTO enabled.

With fat-LTO it will generate regular assembly so it should be able to work with rustc linking process?

madsmtm: My understanding of LTO is as follows:
- You enable LTO in the compiler, and the compiler will emit extra metadata in all object files (for LLVM-based compilers, it will be the LLVM-IR, for GCC-based compilers, something else (GIMPLE?)).
- When later invoking the linker, it ends up calling back into compiler infrastructure (for LLVM via `libLTO.dylib`, dunno about GCC).

Since linkers that call into LLVM won't understand GIMPLE, and vice-versa for linkers that call GCC, cross-language LTO is only gonna work when using the same compiler backend.

(One day, once `rustc` has more stable support for the GCC codegen backends, it may make sense to tell GCC to enable LTO, but until then we shouldn't).

NobodyXu: I think `-ffat-lto-objects` would emit both regular assembly code and G IMPLE, so it should be compatible to rustc @madsmtm

madsmtm: But I doubt that LLVM's libLTO would read that? See [the design docs](https://llvm.org/docs/LinkTimeOptimization.html#design-philosophy)?

NobodyXu: I think rustc invokes c compiler which invokes the linker, so it should be able to handle that?

madsmtm: Preface: I actually can't speak for GNU's `ld`, might be different there, but I'm pretty confident that LLVM-based LTO won't work with object files produced by GCC.

> I think rustc invokes c compiler which invokes the linker

Yeah, but my point is, the linker can only handle it's native "type" of LTO data. LTO doesn't work by reading the assembly in the object file and figuring out the semantics from there (I thought it did at first too), it reads extra compiler-specific metadata emitted in the object file. And having mismatching compiler metadata here won't work (may work for different Clang/LLVM/`rustc` versions, but definitely won't work between Clang and GCC).

I found [this stackoverflow question](https://stackoverflow.com/questions/51259340/interoperability-between-clang-gcc-and-lto) as well.

NobodyXu: Thanks, now that I think about it it's pretty reasonable, linker won't support running two backend for LTO at the same time.

PR #1380, opened 2025-01-31
Fix more flag inheritance

madsmtm:
Follow-up to https://github.com/rust-lang/cc-rs/pull/1379 / https://github.com/rust-lang/cc-rs/pull/1363, CC @mrkajetanp. I went through all the flags shared between GCC/Clang, and fixed some more discrepancies, see each commit for details.

See also [this comment](https://github.com/rust-lang/cc-rs/pull/1379#issuecomment-2626572223), these flags _may_ not have the same semantics between Clang/GCC, though I'm fairly sure that the remaining ones do.

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -225,12 +225,15 @@ impl<'this> RustcCodegenFlags<'this> {
                 push_if_supported(cc_flag.into());
             }
             // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-msoft-float
-            // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mno-soft-float
+            // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mhard-float
+            // https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html (several archs, search for `-msoft-float`).
+            // https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html (several archs, search for `-mhard-float`).
             if let Some(value) = self.soft_float {
                 let cc_flag = if value {
                     "-msoft-float"
                 } else {
-                    "-mno-soft-float"
+                    // Do not use -mno-soft-float, that's basically just an alias for -mno-implicit-float.
+                    "-mhard-float"
                 };
                 push_if_supported(cc_flag.into());
             }
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -16,7 +16,6 @@ pub(crate) struct RustcCodegenFlags<'a> {
     relocation_model: Option<&'a str>,
     embed_bitcode: Option<bool>,
     force_frame_pointers: Option<bool>,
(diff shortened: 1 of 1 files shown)


mrkajetanp:
Looks good to me :)

src/flags.rs
```diff
@@ -180,42 +177,40 @@ impl<'this> RustcCodegenFlags<'this> {
...
+            // https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html#index-mbranch-protection-1 (ARM)
+            // FIXME(madsmtm): Parse the value, and make sure we pass it in the correct format.
             if let Some(value) = self.branch_protection {
                 push_if_supported(
                     format!("-mbranch-protection={}", value.replace(",", "+")).into(),
                 );
             }
             // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mcmodel
+            // https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html (several archs, search for `-mcmodel=`).
+            // FIXME(madsmtm): Parse the model, to make sure we pass the correct value (depending on arch).
```
NobodyXu, line 190: Same for this fixme

madsmtm: I'm not sure, but it seems like RISC-V and SPARC all have a few different:
https://gcc.gnu.org/onlinedocs/gcc/SPARC-Options.html#index-mcmodel_003d-6
https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Options.html#index-mcmodel_003d-4

`-mcmodel=medlow`, `-mcmodel=medmid`, `-mcmodel=medany`, `-mcmodel=embmedany`.

I'm not sure how these translate to/from the rustc options?

src/flags.rs
```diff
@@ -180,42 +177,40 @@ impl<'this> RustcCodegenFlags<'this> {
         // Flags shared between clang and gnu
         if clang_or_gnu {
             // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mbranch-protection
+            // https://gcc.gnu.org/onlinedocs/gcc/AArch64-Options.html#index-mbranch-protection (Aarch64)
+            // https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html#index-mbranch-protection-1 (ARM)
+            // FIXME(madsmtm): Parse the value, and make sure we pass it in the correct format.
```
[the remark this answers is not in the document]
madsmtm: No, I found a better piece of documentation that made it clearer how the options work

src/flags.rs
NobodyXu: Nan I am not aware that there's such option in rust.

This looks more like something for bios/kernel/bootloader than for regular processes, and it probably only takes effect at link time since it decides the final layout of executable, so maybe we can skip it?

madsmtm: I'm fine with ignoring it for now.

NobodyXu (approved):
Thanks LGTM, merging in

PR #1382, opened 2025-01-31
Split `impl Build` in two

madsmtm:
I think it's important to distinguish between pure builder methods, and methods that invoke the compiler.

I'm also not happy with `try_flags_from_environment`, since it's dependent on `target`, `host` and the current environment, but unsure what to do about it yet.

This should be a non-functional change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -410,6 +410,7 @@ impl Object {
     }
 }
 
+/// Configure the builder.
 impl Build {
     /// Construct a new instance of a blank set of configuration.
     ///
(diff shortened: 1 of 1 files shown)


NobodyXu (commented):
Shall we actually move some methods into a separate file?

I.e. a dedicated build/mod.rs with build/compile.rs?

madsmtm:
> Shall we actually move some methods into a separate file?

Maybe one day? I don't feel like I have the full picture, so I'm hesitant to do that yet.

NobodyXu (approved):
thanks, LGTM, feel feee to merge it whenever you feel ready, since I'm not sure if it would conflict with the other PR you have I would let you decide when to merge them

NobodyXu:
Since there's no merge conflict, I've merged this in

Issue #1383, opened 2025-02-01
Regression in v1.2.11: clang: error: cannot specify '-mtargetos=xros1.0' along with '--target=arm64-apple-xros'


briansmith:
```
running: LC_ALL="C" "clang" "-O3" "-fPIC" "--target=arm64-apple-xros" "-mtargetos=xros1.0" "-isysroot" "/Applications/Xcode_15.2.app/Contents/Developer/Platforms/XROS.platform/Developer/SDKs/XROS1.0.sdk" "-I" "/Users/runner/work/ring/ring/include" "-I" "/Users/runner/work/ring/ring/target/aarch64-apple-visionos/release/build/ring-45c369a6134f7c8c/out" "-Wall" "-Wextra" "-fvisibility=hidden" "-std=c1x" "-Wall" "-Wbad-function-cast" "-Wcast-align" "-Wcast-qual" "-Wconversion" "-Wmissing-field-initializers" "-Wmissing-include-dirs" "-Wnested-externs" "-Wredundant-decls" "-Wshadow" "-Wsign-compare" "-Wsign-conversion" "-Wstrict-prototypes" "-Wundef" "-Wuninitialized" "-gfull" "-DNDEBUG" "-Werror" "-o" "/Users/runner/work/ring/ring/target/aarch64-apple-visionos/release/build/ring-45c369a6134f7c8c/out/25ac62e5b3c53843-curve25519.o" "-c" "/Users/runner/work/ring/ring/crypto/curve25519/curve25519.c"
[ring 0.17.8] 
[ring 0.17.8] cargo:warning=clang: error: cannot specify '-mtargetos=xros1.0' along with '--target=arm64-apple-xros'
```

The command run was:
```
mk/cargo.sh +nightly test -vv --target=aarch64-apple-visionos --no-run -Z build-std --release]
```
Where `mk/cargo.sh` is a wrapper around `cargo`

GitHub Actions Runner version 2.322.0
Operating System
  macOS
  14.7.2
  23H311

It was working fine with the previous release, AFAICT.

briansmith:
Here is the quick & dirty workaround I used for my CI to work around this: https://github.com/briansmith/ring/pull/2313

NobodyXu:
https://gitlab.kitware.com/cmake/cmake/-/issues/25188

looks like we should use `--target=arm64-apple-xros1.0` instead

cc @BlackHoleFox @madsmtm

NobodyXu:
[...]

PR #1384, opened 2025-02-03
Don't specify both `-target` and `-mtargetos=` on Apple targets

madsmtm:
So... at this point, I feel like I've made every possible mistake with this, it _has_ to be correct now! ;)

Instead of specifying `-target`, on Apple platforms, we use `-arch` + `-mtargetos=` (or likewise). This should uniquely identify the target. I mostly chose this for consistency, instead of only passing `-target` on visionOS and Mac Catalyst (to make it clear, the issue only happens there because we're forced to use `-mtargetos=` instead of `-m*-version-min=*` that we can use elsewhere).

Fixes https://github.com/rust-lang/cc-rs/issues/1383. To avoid such mistakes in the future, I've expanded our CI to include more Apple targets.

Tested locally with:
```
# Added to CI as well.
cargo +nightly build --workspace --tests -Zbuild-std \
  --target aarch64-apple-darwin \
  --target x86_64-apple-darwin \
  --target aarch64-apple-ios \
  --target aarch64-apple-ios-macabi \
  --target aarch64-apple-ios-sim \
  --target x86_64-apple-ios \
  --target x86_64-apple-ios-macabi \
  --target aarch64-apple-tvos \
  --target aarch64-apple-tvos-sim \
  --target aarch64-apple-visionos \
  --target aarch64-apple-visionos-sim \
  --target aarch64-apple-watchos \
  --target aarch64-apple-watchos-sim \
  --target arm64_32-apple-watchos \
  --target x86_64-apple-tvos \
  --target x86_64-apple-watchos-sim \
  --target x86_64h-apple-darwin

# 32-bit iOS targets
IPHONEOS_DEPLOYMENT_TARGET=10.10 cargo +nightly build --workspace --tests -Zbuild-std --target i386-apple-ios --target armv7s-apple-ios

# 32-bit watchOS target
WATCHOS_DEPLOYMENT_TARGET=5.0 cargo +nightly build --workspace --tests -Zbuild-std --target armv7k-apple-watchos

# Run on Intel macOS 10.12 (not added to CI for that reason).
cargo +nightly test --workspace -Zbuild-std --target i686-apple-darwin
```

(This excludes the `arm64e` targets, as I [still can't compile those](https://github.com/rust-lang/rust/issues/130085)).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,9 +22,13 @@ jobs:
             beta,
             nightly,
             linux32,
-            macos,
             aarch64-macos,
+            x86_64-macos,
             aarch64-ios,
+            aarch64-ios-sim,
+            x86_64-ios-sim,
+            aarch64-ios-macabi,
+            x86_64-ios-macabi,
             win32,
             win64,
             mingw32,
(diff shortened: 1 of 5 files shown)


briansmith:
> To avoid such mistakes in the future, I've expanded our CI to include more Apple targets.

Love to see this!

NobodyXu (approved):
Thank you for the quick fix!

NobodyXu:
cc @briansmith do you need a release today, or would you like to wait until Friday?

grachyov:
@NobodyXu i’d love it today too — having some issues targeting xros as well

PR #1385, opened 2025-02-03
allow specifying `rustc` when generating target information

Qix-:
Hello!

Adding a new platform target to `rust-lang/rust` was resulting in some hard to track errors, but I finally figured out it had to do with the generated list here. Adding them manually to the generated output in a fork is of course possible but I also added a means of passing in a path to `rustc`, e.g. from one of the bootstrap stages, to automatically update the target list.

Might be useful for those who are trying to get a fork up and running without manually editing generated stuff.

### DIFF
--- a/dev-tools/gen-target-info/src/main.rs
+++ b/dev-tools/gen-target-info/src/main.rs
@@ -84,15 +84,22 @@ fn generate_target_mapping(f: &mut File, target_specs: &RustcTargetSpecs) -> std
 
 fn main() {
     // Primarily use information from nightly.
-    let mut target_specs = get_target_specs_from_json();
+    let mut target_specs = get_target_specs_from_json(std::env::var("RUSTC").ok());
     // Next, read from MSRV to support old, removed targets.
-    for target_triple in get_targets_msrv().lines() {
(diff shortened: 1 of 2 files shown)


madsmtm (commented):
I'm mostly indifferent on this, will let @NobodyXu decide.

I do think there's value in the code currently, since it forces users to have `+nightly` and `+$MSRV`, which makes it easier for them to see what they need to do to re-generate the same as CI does. So this might also just be solvable with better docs, and the addition of a `+stage1` override?

Qix-:
Having any sort of `+override` means that you'd have to explicitly link `rustup` profiles, right?

Thinking about things a bit more, I think the sort of "ideal" case is that, in cases where the compiler is bootstrapping, forking/patching isn't necessary at all, and the target list can be fetched given the bootstrapped compilers at each stage - though that's a drastically more complicated solution.

Either way, I just did this to make things a little easier and to remove human error - I know my use case is niche so it may not be the best way to do it, or even valuable to have. If the `+stage1` override or something similar is preferred just let me know and I can adapt :)

NobodyXu (approved):
Thanks LGTM

PR #1386, opened 2025-02-04
Update lib.rs

Thewsthews:
Use of a shorthand struct without explicitly specifying the return type


madsmtm:
Yeah, I _guess_ we could do this, but I feel there are many places where `Self` could be used instead of the type name, so I don't see the gain with this PR. Am going to close it, but thanks for the contribution anyhow!

Issue #1388, opened 2025-02-04
Error when cross-compiling to macOS with cc 1.2.12


tuxu:
I have cross-compilation from Linux to macOS set up for a project. [`cc` 1.2.12 omits the `--target` argument](https://github.com/rust-lang/cc-rs/pull/1384), which Clang (currently using 18.1.8) doesn't like.

Minimum reproducible example with a dummy project:
```bash
$ export CARGO_TARGET_AARCH64_APPLE_DARWIN_RUSTFLAGS="-C linker=clang -C link-arg=--target=aarch64-apple-darwin -C link-arg=-fuse-ld=lld"
$ cargo new hello
$ cd hello
$ cargo add bzip2
$ cargo build --target aarch64-apple-darwin
```

This fails with:
```bash
  cargo:warning=clang: error: unsupported option '-arch' for target 'aarch64-unknown-linux-gnu'
  cargo:warning=clang: warning: argument unused during compilation: '-mmacosx-version-min=11.0' [-Wunused-command-line-argument]

  --- stderr


  error occurred in cc-rs: Command LC_ALL="C" "clang" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-2" "-fno-omit-frame-pointer" "-arch" "arm64" "-mmacosx-version-min=11.0" "-I" "bzip2-1.0.8" "-D_FILE_OFFSET_BITS=64" "-DBZ_NO_STDIO" "-o" "/tmp/hello/target/aarch64-apple-darwin/debug/build/bzip2-sys-9c6e283515f24e27/out/lib/a9dc8ba631b1466a-blocksort.o" "-c" "bzip2-1.0.8/blocksort.c" with args clang did not execute successfully (status code exit status: 1).
```

<details>
<summary>Click for full output</summary>

```bash
root@ac6839d585fb:/tmp# curl -L "https://github.com/phracker/MacOSX-SDKs/releases/download/11.0-11.1/MacOSX11.0.sdk.tar.xz" | tar xJ -C /opt
root@ac6839d585fb:/tmp# export SDKROOT=/opt/MacOSX11.0.sdk
root@ac6839d585fb:/tmp# export CARGO_TARGET_AARCH64_APPLE_DARWIN_RUSTFLAGS="-C linker=clang -C link-arg=--target=aarch64-apple-darwin -C link-arg=-fuse-ld=lld"
root@ac6839d585fb:/tmp# rustc --version
[...]

PR #1389, opened 2025-02-05
Fix cross-compiling for Apple platforms

madsmtm:
(So, of course we weren't done here...)

Basically, the only difference between this and pre-https://github.com/rust-lang/cc-rs/pull/1384 is that on visionOS and Mac Catalyst, we pass a versioned `--target` instead of unversioned `--target` + `-mtargetos=`.
This is similar to the alternative I noted in that PR, and while it does run into https://github.com/rust-lang/cc-rs/issues/1278 (the bug that configuration files + versioned targets are badly supported on older Clang) on those targets, I think that's the better option.

Fixes https://github.com/rust-lang/cc-rs/issues/1388. To avoid this in the future, I have added a CI step to verify that cross-compilation works (fails in the first commit, succeeds in the second).

### Alternatives

I looked through Clang's parsing code, and from that I glean that an alternative would probably be to set `--target` + `*_DEPLOYMENT_TARGET`. This has a few problems though:
- I suspect a lot of users are just inspecting `Tool::args`/`.get_compiler().get_command().get_args()`, and forget to also look at `Tool::env` (I know at least Rust's `bootstrap` currently forgets to do this).
- This also doesn't force GCC users to use a compiler configured for Darwin (so we'd still have to pass `-mmacosx-version-min=` there) (probably acceptable to not do if we did https://github.com/rust-lang/cc-rs/issues/1391).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -86,6 +86,16 @@ jobs:
             rust: stable
             target: x86_64-apple-ios-macabi
             no_run: --no-run # FIXME(madsmtm): Fix running tests
+          - build: cross-macos-aarch64
+            os: ubuntu-latest
+            rust: stable
+            target: aarch64-apple-darwin
+            no_run: --no-run
+          - build: cross-ios-aarch64
+            os: ubuntu-latest
+            rust: stable
+            target: aarch64-apple-ios
+            no_run: --no-run
           - build: windows-aarch64
             os: windows-latest
             rust: stable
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -155,6 +165,31 @@ jobs:
...
+          wget https://github.com/xybp888/iOS-SDKs/releases/download/iOS18.1-SDKs/iPhoneOS18.1.sdk.zip
+          unzip iPhoneOS18.1.sdk.zip
+          echo "SDKROOT=$(pwd)/iPhoneOS18.1.sdk" >> $GITHUB_ENV
+      - name: Set up Apple cross-compilation
+        if: startsWith(matrix.build, 'cross-macos') || startsWith(matrix.build, 'cross-ios')
+        run: |
+          # Test with clang/llvm for now, has better cross-compilation support (GCC requires downloading a different toolchain)
+          echo "CC=clang" >> $GITHUB_ENV
+          echo "CXX=clang++" >> $GITHUB_ENV
+          echo "AR=llvm-ar" >> $GITHUB_ENV
```
madsmtm, line 189: It is unclear to me why `llvm-ar` is required here, it seems to not be on my local Ubuntu 24 VM? But if we don't include it, we fail with:
```console
rust-lld: error: /home/runner/work/cc-rs/cc-rs/target/aarch64-apple-darwin/debug/build/cc-test-5875314a6fda6e38/out/libOptLinkage.a: archive has no index; run ranlib to add one
rust-lld: error: undefined symbol: _answer
```

NobodyXu: No idea maybe there's some apple/clang stuff default ar not supporting?

tests/test.rs
```diff
@@ -616,7 +616,7 @@ fn clang_apple_tvos() {
         .file("foo.c")
         .compile("foo");
 
-    test.cmd(0).must_have_in_order("-arch", "arm64");
+    test.cmd(0).must_have("--target=arm64-apple-tvos");
```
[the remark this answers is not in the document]
madsmtm: Have added some more tests for this now.

clubby789:
I'm not sure if this is relevant to the last fix, but trying to bump cc in rust gives us https://github.com/rust-lang/rust/pull/136720#issuecomment-2644499614

```
clang-15: error: cannot specify '-mtargetos=ios14.2-macabi' along with '--target=arm64-apple-ios14.2-macabi'
```

madsmtm:
Hmm, that's what https://github.com/rust-lang/cc-rs/pull/1384 fixed.

Seems like it's not failing in bootstrap, but when building `compiler-builtins`. Maybe `library/Cargo.lock` was somehow updated to include the broken `cc-rs` version?

NobodyXu (approved):
Thank you!

PR #1395, opened 2025-02-08
Add support for setting `-gdwarf-{version}` based on RUSTFLAGS

wesleywiser:
Detect if `-Zdwarf-version` (which will probably be stabilized soon as `-Cdwarf-version`) was passed in RUSTFLAGS and set the corresponding Clang/GCC option to the same value.

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -18,6 +18,7 @@ pub(crate) struct RustcCodegenFlags<'a> {
     force_frame_pointers: Option<bool>,
     no_redzone: Option<bool>,
     soft_float: Option<bool>,
+    dwarf_version: Option<u32>,
 }
 
 impl<'this> RustcCodegenFlags<'this> {
@@ -86,6 +87,10 @@ impl<'this> RustcCodegenFlags<'this> {
             }
         }
 
+        fn arg_to_u32(arg: impl AsRef<str>) -> Option<u32> {
+            arg.as_ref().parse().ok()
+        }
+
         let (flag, value) = if let Some((flag, value)) = flag.split_once('=') {
             (flag, Some(value))
         } else {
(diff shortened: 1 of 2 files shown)


NobodyXu (approved):
Thanks LGTM!

cc @madsmtm for a double checkq

src/flags.rs
```diff
@@ -148,6 +149,11 @@ impl<'this> RustcCodegenFlags<'this> {
                 self.branch_protection =
                     Some(flag_ok_or(value, "-Zbranch-protection must have a value")?);
             }
+            // https://doc.rust-lang.org/beta/unstable-book/compiler-flags/dwarf-version.html
+            // FIXME: Drop the -Z variant and update the doc link once the option is stablized
+            "-Zdwarf-version" | "-Cdwarf-version" => {
+                self.dwarf_version = Some(flag_ok_or(value, "-Zdwarf-version must have a value")?);
```
madsmtm, line 155: I can imagine us wanting to do different stuff based on the actual version, so maybe we should parse it to an integer here?

madsmtm: Maybe add a helper `arg_to_int` or smth, similar to `arg_to_bool`.

madsmtm (commented):
Looks fine after parsing. I verified that the semantics are the same.

src/flags.rs
wesleywiser: Done!

madsmtm (approved):
Thanks!

For posterity, I'll note that I considered if it would make sense to restrict even further (e.g. only accept `2`, `3`, `4` or `5`), but then again, Clang and GCC support more versions than `rustc` (and that will probably be true for the foreseeable future), and the compiler driver will just error if it doesn't support the flag anyhow. So it's probably better to treat these as integers, and be "future compatible" with a hypothetical DWARF 6.

PR #1403, opened 2025-02-14
Change flag ordering

madsmtm:
To allow flags set on the builder with `Build::flag`/`Build::flag_if_supported` to override other flags, and to allow `CFLAGS` to override _all_ other flags.

Found while doing https://github.com/rust-lang/cc-rs/pull/1401.

Now the order of flags are as follows, with later flags overriding previous ones. I've also here shown who "controls" which flag, to make it clear why we want the ordering to be like this.
1. Default flags
    - Controlled by `cc-rs`.
2. `rustc`-inherited flags
    - Controlled by `rustc`.
3. Builder flags
    - Controlled by the developer using `cc-rs` in e.g. their `build.rs`.
4. Environment flags
    - Controlled by the end user.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1892,79 +1892,17 @@ impl Build {
 
         let mut cmd = self.get_base_compiler()?;
 
-        // Disable default flag generation via `no_default_flags` or environment variable
-        let no_defaults = self.no_default_flags || self.getenv_boolean("CRATE_CC_NO_DEFAULTS");
-
-        if !no_defaults {
-            self.add_default_flags(&mut cmd, &target, &opt_level)?;
-        }
-
-        if let Some(ref std) = self.std {
-            let separator = match cmd.family {
-                ToolFamily::Msvc { .. } => ':',
-                ToolFamily::Gnu | ToolFamily::Clang { .. } => '=',
-            };
-            cmd.push_cc_arg(format!("-std{}{}", separator, std).into());
-        }
-
-        for directory in self.include_directories.iter() {
-            cmd.args.push("-I".into());
-            cmd.args.push(directory.as_os_str().into());
-        }
-
-        let flags = self.envflags(if self.cpp { "CXXFLAGS" } else { "CFLAGS" })?;
-        if let Some(flags) = &flags {
-            for arg in flags {
-                cmd.push_cc_arg(arg.into());
-            }
-        }
-
-        // If warnings and/or extra_warnings haven't been explicitly set,
-        // then we set them only if the environment doesn't already have
-        // CFLAGS/CXXFLAGS, since those variables presumably already contain
-        // the desired set of warnings flags.
-
-        if self.warnings.unwrap_or(flags.is_none()) {
-            let wflags = cmd.family.warnings_flags().into();
-            cmd.push_cc_arg(wflags);
-        }
-
-        if self.extra_warnings.unwrap_or(flags.is_none()) {
-            if let Some(wflags) = cmd.family.extra_warnings_flags() {
-                cmd.push_cc_arg(wflags.into());
-            }
-        }
-
-        for flag in self.flags.iter() {
-            cmd.args.push((**flag).into());
-        }
-
-        // Add cc flags inherited from matching rustc flags
-        if self.inherit_rustflags {
-            self.add_inherited_rustflags(&mut cmd, &target)?;
-        }
-
-        for flag in self.flags_supported.iter() {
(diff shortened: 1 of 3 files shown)


tests/cflags.rs
```diff
@@ -19,21 +19,39 @@ fn gnu_no_warnings_if_cflags() {
...
-        .must_have_in_order("-arbitrary3", "-arbitrary4");
+        // .must_have_in_order("-static", "-gdwarf-5")
+        // .must_have_in_order("-gdwarf-5", "-Lbuilder-flag1")
+        .must_have_in_order("-static", "-Lbuilder-flag1")
+        .must_have_in_order("-Lbuilder-flag1", "-Lbuilder-flag2")
+        .must_have_in_order("-Lbuilder-flag2", "-Larbitrary1")
+        .must_have_in_order("-Larbitrary1", "-Larbitrary2")
+        .must_have_in_order("-Larbitrary1", "-Larbitrary2")
+        .must_have_in_order("-Larbitrary2", "-Larbitrary3")
+        .must_have_in_order("-Larbitrary3", "-Larbitrary4");
```
NobodyXu, line 56: Nit: maybe we could access `.args` directly?

https://github.com/rust-lang/cc-rs/blob/0dd683ca4c57fff1d8701d5a780749fcf09a5e75/tests/support/mod.rs#L20

```suggestion
        .args == [
            "-static",
            "-Lbuilder-flag1",
            "-Lbuilder-flag2",
            "-Larbitrary1",
            "-Larbitrary2",
            "-Larbitrary3",
            "-Larbitrary4",
        ]
```

tests/cflags.rs
```diff
@@ -19,21 +19,39 @@ fn gnu_no_warnings_if_cflags() {
...
+/// 3. Builder flags.
+/// 4. Environment flags.
 fn cflags_order() {
-    unsafe { env::set_var("CFLAGS", "-arbitrary1") };
-    unsafe { env::set_var("HOST_CFLAGS", "-arbitrary2") };
-    unsafe { env::set_var("TARGET_CFLAGS", "-arbitrary2") };
-    unsafe { env::set_var("CFLAGS_x86_64_unknown_none", "-arbitrary3") };
-    unsafe { env::set_var("CFLAGS_x86_64-unknown-none", "-arbitrary4") };
+    // FIXME(madsmtm): Re-enable once `is_flag_supported` works in CI regardless of `target`.
+    // unsafe { std::env::set_var("CARGO_ENCODED_RUSTFLAGS", "-Cdwarf-version=5") };
```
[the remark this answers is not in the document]
madsmtm: There is now: https://github.com/rust-lang/cc-rs/issues/1410

tests/cflags.rs
madsmtm: `.args` contains all the other default flags too, which we don't want to test for here (since they're platform-specific).

tests/cflags.rs
```diff
@@ -19,21 +19,39 @@ fn gnu_no_warnings_if_cflags() {
...
+        .flag("-Lbuilder-flag2")
         .file("foo.c")
         .compile("foo");
 
     test.cmd(0)
-        .must_have_in_order("-arbitrary1", "-arbitrary2")
-        .must_have_in_order("-arbitrary2", "-arbitrary3")
-        .must_have_in_order("-arbitrary3", "-arbitrary4");
+        // .must_have_in_order("-static", "-gdwarf-5")
+        // .must_have_in_order("-gdwarf-5", "-Lbuilder-flag1")
```
NobodyXu, line 49:
```suggestion
```

src/lib.rs
```diff
@@ -1892,58 +1892,80 @@ impl Build {
...
 
+        // Specify various flags that are not considered part of the default flags above.
+        // FIXME(madsmtm): Should these be considered part of the defaults? If no, why not?
         if let Some(ref std) = self.std {
             let separator = match cmd.family {
                 ToolFamily::Msvc { .. } => ':',
                 ToolFamily::Gnu | ToolFamily::Clang { .. } => '=',
             };
             cmd.push_cc_arg(format!("-std{}{}", separator, std).into());
         }
```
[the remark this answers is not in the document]
NobodyXu: Ohh that's true...

tests/cflags.rs
madsmtm: This is meant to be re-added once https://github.com/rust-lang/cc-rs/issues/1410 is resolved.

PR #1404, opened 2025-02-15
Fix archiver detection for musl cross compilation

### LINKED ISSUE #1399: Archiver for musl is not `ar` when in Docker image for a different architecture
When I use `cc` natively on Linux or macOS, [`cc::Build::get_archiver`](https://docs.rs/cc/1.2.13/cc/struct.Build.html#method.get_archiver) always returns `ar` when I'm building for either x86 musl or ARM musl. However, when I try to do the same build in a [`rust`](https://hub.docker.com/_/rust) Docker image, and the [`--platform`](https://docs.docker.com/reference/cli/docker/buildx/build/#platform) is not my native platform, it sometimes attempts other names for the archiver program (like `musl-ar` or `aarch64-linux-musl-ar`) which cause the build to fail.

This forces me to set a `TARGET_AR` variable in my `Dockerfile` if I depend on crates like [`wasmtime`](https://crates.io/crates/wasmtime) that transitively depend on crates like [`zstd-sys`](https://crates.io/crates/zstd-sys); see [this Stack Overflow question](https://stackoverflow.com/q/79425799/5044950) for such an example, which is itself boiled down from gradbench/gradbench#233 that was my real use case.

See [this GitHub repository](https://github.com/samestep/docker-cross-example/tree/1e8b9e974a33c265444347d23c562cdc1733062d) which includes a full code example to reproduce the issue. As also written in that `README.md`, here are the different values I see returned by `cc::Build::get_archiver` in various contexts:

- ARM macOS
  - native: `"ar"`
  - native targeting x86 musl: `"ar"`
  - native targeting ARM musl: `"ar"`
  - Docker targeting x86 musl: `"musl-ar"`
  - Docker targeting ARM musl: `"ar"`
- x86 Linux
  - native: `"ar"`
  - native targeting x86 musl: `"ar"`
  - native targeting ARM musl: `"ar"`
  - Docker targeting x86 musl: `"ar"`
  - Docker targeting ARM musl: `"aarch64-linux-musl-ar"`

NobodyXu:
Fixed #1399

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3298,8 +3298,13 @@ impl Build {
                                 .iter()
                                 .filter_map(|infix| {
                                     let target_p = format!("{prefix}{infix}-{tool}");
-                                    let status = Command::new(&target_p).status().ok()?;
-                                    status.success().then_some(target_p)         
+                                    let status = Command::new(&target_p)
+                                        .stdin(Stdio::null())
+                                        .stdout(Stdio::null())
+                                        .stderr(Stdio::null())
+                                        .status()
+                                        .ok()?;
+                                    status.success().then_some(target_p)
                                 })
                                 .next()
                                 .unwrap_or_else(|| tool.to_string());
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3295,7 +3295,7 @@ impl Build {
                             //
                             // The same applies to ranlib.
                             let chosen = ["", "-gcc"]
-                                .into_iter()
+                                .iter()
                                 .filter_map(|infix| {
                                     let target_p = format!("{prefix}{infix}-{tool}");
                                     let status = Command::new(&target_p).status().ok()?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -232,7 +232,7 @@ use std::io::{self, Write};
 use std::path::{Component, Path, PathBuf};
 #[cfg(feature = "parallel")]
 use std::process::Child;
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::sync::{
     atomic::{AtomicU8, Ordering::Relaxed},
     Arc, RwLock,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3299,6 +3299,7 @@ impl Build {
                                 .filter_map(|infix| {
                                     let target_p = format!("{prefix}{infix}-{tool}");
                                     let status = Command::new(&target_p)
+                                        .arg("--version")
                                         .stdin(Stdio::null())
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3287,32 +3286,38 @@ impl Build {
...
-                                if Command::new(&target_p).output().is_ok() {
-                                    chosen = target_p;
-                                    break;
-                                }
-                            }
+                            let chosen = ["", "-gcc"]
+                                .iter()
+                                .filter_map(|infix| {
+                                    let target_p = format!("{prefix}{infix}-{tool}");
+                                    let status = Command::new(&target_p)
```
NobodyXu, line 3301: Feels like I should pass `--version`

```suggestion
                                    let status = Command::new(&target_p)
                                        .arg("--version")
```

samestep:
Just double-checking @NobodyXu @madsmtm, you're not waiting on me for anything here, right?

NobodyXu:
No I was waiting for code review from @madsmtm 

I will DM Mads for review

src/lib.rs
```diff
@@ -3287,32 +3286,39 @@ impl Build {
...
+                                        .arg("--version")
```
madsmtm, line 3302: Nit: I assume we're fine with this not working for `cl.exe` that don't accept the `--version` flag? Seems like `prefix_for_target` is used for `gcc`-like compilers, but maybe it'd be nice to have a comment about this?

src/lib.rs
```diff
@@ -3287,32 +3286,39 @@ impl Build {
...
+                                        .stdin(Stdio::null())
+                                        .stdout(Stdio::null())
+                                        .stderr(Stdio::null())
+                                        .status()
```
madsmtm, line 3306: Is there a reason why we don't invoke `which $target_p` instead? It seems like that would have the same effect of determining if the command exists (and it would be easier to optimize/understand, since we could implement the `which` algorithm ourselves).

madsmtm (approved):
Sry for the delay.

I'm still baffled that this works while `.output` doesn't, but am fine with it either way. Left a few comments, but none of them blocking, feel free to merge this without resolving those.

src/lib.rs
madsmtm: Which also has quite a few deps unfortunately. The core logic would be re-implementable in `cc` though.

NobodyXu: > Which also has quite a few deps unfortunately. The core logic would be re-implementable in `cc` though.

Just slightly worried about having more code to be maintained.

PR #1417, opened 2025-02-26
MSVC: Add LIB var when compiling flag_check

### LINKED ISSUE #1416: bug: Build::is_flag_supported /guard:cf returns false on 1.2.4 ->1.2.5 transition (on x86_64-pc-windows-msvc)
Seems related to #1324 
The build output looks like that in 1.2.5, while these messages don't show up in 1.2.4:
Image

When manually running the flag check command I get the following output:
Image

This in practice disables control flow guard mitigation for projects compiled with `cc`.
It is possible that /guard:cf adds a link dependency on MSVCRT.lib and the fix for #1324 didn't take that into account.

ChrisDenton:
Fixes #1416

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1357,11 +1357,15 @@ impl Build {
 
         cmd.arg(&src);
 
-        // On MSVC skip the CRT by setting the entry point to `main`.
-        // This way we don't need to add the default library paths.
         if compiler.is_like_msvc() {
-            // Flags from _LINK_ are appended to the linker arguments.
-            cmd.env("_LINK_", "-entry:main");
+            // On MSVC we need to make sure the LIB directory is included
+            // so the CRT can be found.
+            for (key, value) in &tool.env {
+                if key == "LIB" {
+                    cmd.env("LIB", value);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1357,11 +1357,15 @@ impl Build {
 
         cmd.arg(&src);
 
-        // On MSVC skip the CRT by setting the entry point to `main`.
-        // This way we don't need to add the default library paths.
         if compiler.is_like_msvc() {
-            // Flags from _LINK_ are appended to the linker arguments.
-            cmd.env("_LINK_", "-entry:main");
```
NobodyXu, line 1364: Do we no longer need this?

NobodyXu (changes requested):

src/lib.rs
ChrisDenton: Yeah, we're including the CRT so the normal entry point will be there now.

NobodyXu (approved):
Thank you for the quick fix!

It will be released on Friday/Saturday

PR #1424, opened 2025-02-27
Fix wasm32-unknown-unknown by passing -c

### LINKED ISSUE #1423: `is_flag_supported` returns false negatives
Seems like https://github.com/rust-lang/cc-rs/issues/892 is back. I once submitted a patch in https://github.com/rust-lang/cc-rs/pull/839 and that worked for me at the time, but after an update to `cc` v1.2.15 I'm having the same issues again. I'm on `wasm32-unknown-unknown` and using `.flag_if_supported("-Wno-unused-parameter")` doesn't add the flag even though it's supported.

NobodyXu:
Fix #1423

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1351,6 +1351,8 @@ impl Build {
             },
         );
 
+        cmd.arg("-c");
+
         if compiler.supports_path_delimiter() {
             cmd.arg("--");
         }


NobodyXu:
I don't know why it's removed, cc @madsmtm do you remember why?

madsmtm:
See https://github.com/rust-lang/cc-rs/pull/1322.

madsmtm:
Perhaps we need to add an (internal) flag "needs_linker" to `is_flag_supported`? And set it to `true` for `-fembed-bitcode=...` (and maybe others)?

I generally agree that we should not attempt to link if we can avoid it, so we really _should_ pass `-c` unless absolutely necessary that we don't.

madsmtm:
Hmm, actually, I think https://github.com/rust-lang/cc-rs/pull/1322 is completely wrong, and the real solution was https://github.com/rust-lang/cc-rs/pull/1379.

CC @clubby789, can you confirm that this PR still works for you?

src/lib.rs
```diff
@@ -1351,6 +1351,8 @@ impl Build {
             },
         );
 
+        cmd.arg("-c");
```
madsmtm, line 1354: Nit: Please add a comment linking to previous issues.
```suggestion
        // Checking for compiler flags does not require linking (and we _must_
        // avoid making it do so, since it breaks cross-compilation when the C
        // compiler isn't configured to be able to link).
        // https://github.com/rust-lang/cc-rs/issues/1423
        cmd.arg("-c");
```

clubby789:
> Hmm, actually, I think 1322 is wrong, and the real solution was 1379.
> 
> CC @clubby789, can you check if this PR still works for you?

1379 does indeed fix the specific case, but 1322 is about preventing the general case (of linker-specific flags being incorrectly reported as supported). The `needs_linker` idea sounds like a sensible fix to me

madsmtm:
But `cc` generally doesn't link, and thus we shouldn't be passing flags that affect the linker in the first place?

dot-asm:
> But `cc` generally doesn't link, and thus we shouldn't be passing flags that affect the linker in the first place?

Nor does it have a say in which linker will be called. Since the controversy seems to revolve around -fembed-bitcode, one should recognize that in order to utilize the embedded bitcode user has to explicitly instruct rustc to link the binary with suitable clang command, sufficiently recent version, and with necessary supplemental flags. And the point is that it's external to cc-rs. Also note that it's on the user to instruct both cc-rs and rustc to use clang, preferably the same version. With this in mind there is a case to be made that probing for -fembed-bitcode **without** -c flag adds no value(\*) to the user. Such a probe won't help solving any related problems, probing **with** -c is more than sufficient.

(\*) As a matter of fact binary code compiled with -fembed-bitcode alone is the same as one compiled without. It takes additional flags to utilize the bitcode even when you use clang by itself.

dot-asm:
On a vaguely related note. cc-rs appears to parse RUSTFLAGS and map -Clto rustc option to clang's -flto=full. Note that the -Clto is required to be complemented with -Cembed-bitcode. Which is **not** mapped. However, if it was mapped, it wouldn't make difference, because -flto makes clang generate raw bitcode regardless. Since rustc's preferred object format is embedded bitcode, it makes more sense to map -Cembed-bitcode and let -Clto go unmapped. How is it related to the issue at hand? This is just an example of a potential problem that lack of -c in the probe won't help solving.

NobodyXu:
Thanks for explanation, in that case cc @madsmtm we probably should merge it?

NobodyXu:
> Since rustc's preferred object format is embedded bitcode, it makes more sense to map -Cembed-bitcode and let -Clto go unmapped.

Well IIRC it doesn't matter unless cross lang LTO is turned on, otherwise rustc just invokes linker for the rust and external C crate separately?

dot-asm:
> otherwise rustc just invokes linker for the rust and external C crate separately?

I don't understand what "separately" means in the question. Rust links the final binary module by calling a linker, which doesn't care where its inputs come from, Rust or C. You can tell compilers to complement the object modules with LTO data, but if you don't tell rustc to use the linker that can make use of it, then LTO simply doesn't happen.

NobodyXu:
> if you don't tell rustc to use the linker that can make use of it, then LTO simply doesn't happen.

Yeah this was what I was trying to say, I was pretty sure most people don't enable cross lang LTO

NobodyXu:
merged in as this long over-due and I think the issue on this is settled cc @madsmtm

Issue #1426, opened 2025-02-28
Tracking issue for changes to target triples


madsmtm:
Since https://github.com/rust-lang/cc-rs/pull/1413, `cc-rs` parses the `rustc` target when running outside build scripts. To ensure that this is kept up-to-date with changes to `rustc`, there's a CI step that runs every so often, and adds a comment here if the latest nightly failed.

Resolving this requires a code change (which is why we don't open a PR to do it automatically). Resolved comments will be hidden.

NobodyXu:
Opened #1490

NobodyXu:
Fixed in #1517

NobodyXu:
Fixed in #1661 , I've manually rerun the ci to confirm it

NobodyXu:
```
---- target::parser::tests::parse_rustc_targets stdout ----
failed comparing aarch64-oe-linux-gnu:
  expected: Ok(TargetInfo { full_arch: "aarch64", arch: "aarch64", vendor: "unknown", os: "linux", env: "gnu", abi: "" })
    actual: Ok(TargetInfo { full_arch: "aarch64", arch: "aarch64", vendor: "oe", os: "linux", env: "gnu", abi: "" })

failed comparing aarch64-unknown-linux-pauthtest:
  expected: Ok(TargetInfo { full_arch: "aarch64", arch: "aarch64", vendor: "unknown", os: "linux", env: "musl", abi: "pauthtest" })
    actual: Err(Error { kind: UnknownTarget, message: "unknown environment/ABI `pauthtest` in target `aarch64-unknown-linux-pauthtest`" })

failed comparing armv7-oe-linux-gnueabihf:
  expected: Ok(TargetInfo { full_arch: "armv7", arch: "arm", vendor: "unknown", os: "linux", env: "gnu", abi: "eabihf" })
    actual: Ok(TargetInfo { full_arch: "armv7", arch: "arm", vendor: "oe", os: "linux", env: "gnu", abi: "eabihf" })

failed comparing i686-oe-linux-gnu:
  expected: Ok(TargetInfo { full_arch: "i686", arch: "x86", vendor: "unknown", os: "linux", env: "gnu", abi: "" })
    actual: Ok(TargetInfo { full_arch: "i686", arch: "x86", vendor: "oe", os: "linux", env: "gnu", abi: "" })

failed comparing riscv64-oe-linux-gnu:
[...]

PR #1434, opened 2025-03-13
Fix `wasm32-wali-linux-musl` target parsing

madsmtm:
See https://github.com/rust-lang/cc-rs/issues/1426#issuecomment-2721475424.

Intentionally left out of https://github.com/rust-lang/cc-rs/pull/1413 to allow testing the CI.

(I think this is actually a `rustc` issue, but I will wait 'till after the resolution of https://github.com/rust-lang/compiler-team/issues/850 before I go and fix that).


NobodyXu (approved):
Thanks, LGTM

PR #1443, opened 2025-03-28
Support compiler prefix `x86_64-linux-musl`

justsmth:
### Motivation
* See: [aws/aws-lc-rs#736](https://github.com/aws/aws-lc-rs/issues/736#issuecomment-2758980258)

### Context
I found that in some (most?) environments, the musl compiler to use for targeting `x86_64-unknown-linux-gnu` is prefixed with `x86_64-linux-musl`. For example, on Ubuntu (AMD64) the [`musl-dev` package](https://packages.ubuntu.com/noble/amd64/musl-dev/filelist) contains `/usr/bin/x86_64-linux-musl-gcc` but it does not contain `/usr/bin/musl-gcc`, which is provided by the [`musl-tools` package](https://packages.ubuntu.com/noble/amd64/musl-tools/filelist). (The [`musl-tools` package](https://packages.ubuntu.com/noble/musl-tools) depend on `musl-dev`.) 

When `musl-tools` is installed, `/usr/bin/musl-gcc` is simply a symlink to the former:
```
❯ file /usr/bin/musl-gcc
/usr/bin/musl-gcc: symbolic link to x86_64-linux-musl-gcc
```

### Change
This adds another compiler prefix (`x86_64-linux-musl`) to check for when building for this target.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3510,9 +3510,7 @@ impl Build {
                     "x86_64-unknown-linux-gnu" => self.find_working_gnu_prefix(&[
                         "x86_64-linux-gnu", // rustfmt wrap
                     ]), // explicit None if not found, so caller knows to fall back
-                    "x86_64-unknown-linux-musl" => {
-                        self.find_working_gnu_prefix(&["musl", "x86_64-linux-musl"])
-                    }
+                    "x86_64-unknown-linux-musl" => Some("x86_64-linux-musl"),
                     "x86_64-unknown-netbsd" => Some("x86_64--netbsd"),
                     _ => None,
                 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3510,7 +3510,9 @@ impl Build {
                     "x86_64-unknown-linux-gnu" => self.find_working_gnu_prefix(&[
                         "x86_64-linux-gnu", // rustfmt wrap
                     ]), // explicit None if not found, so caller knows to fall back
-                    "x86_64-unknown-linux-musl" => Some("musl"),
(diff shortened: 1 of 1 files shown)


madsmtm (commented):
Thanks, this seems to make sense, although if `musl-gcc` is a fallback to `x86_64-linux-musl-gcc` in all situations, I'd prefer to not do fallback, and only attempt to use `x86_64-linux-musl-gcc` (just like we do for the similar `aarch64-linux-musl-gcc`).

justsmth:
> Thanks, this seems to make sense, although if musl-gcc is a fallback to x86_64-linux-musl-gcc in all situations, I'd prefer to not do fallback, and only attempt to use x86_64-linux-musl-gcc (just like we do for the similar aarch64-linux-musl-gcc).

I kept the inclusion of the `musl` prefix (and listed first) as I'm not certain whether Ubuntu's packages are necessarily representative of how other Linux distributions provide the MUSL binaries.  Since previously the only prefix checked was `musl`, keeping it helps avoid breaking consumers that (for whatever reason) only have a `musl-gcc` compiler in their path.

justsmth:
> My issue with that is that it seems wrong still? E.g. if cross-compiling from aarch64-unknown-linux-musl to x86_64-unknown-linux-musl, you'll likely also have musl-gcc in your path, but it'll point to the wrong thing.

Yeah, that's fair. In the latest commit, it only uses the `x86_64-linux-musl` prefix.

madsmtm (approved):
Excellent, thanks!

Issue #594, opened 2021-04-07
shared_flag and static_flag don't work


glandium:
The `-shared` and `-static` flags are linkage flags, and don't do anything when passed alongside `-c`.

Taking the example from shared_flag:
```
cc::Build::new()
    .file("src/foo.c")
    .shared_flag(true)
    .compile("libfoo.so");
```

What that does is:
```
cc $OTHER_FLAGS -shared -o foo.o -c src/foo.c
ar cq libfoo.so.a foo.o
ar s libfoo.so.a
```
The first command is strictly equivalent to `cc $OTHER_FLAGS -o foo.o -c src/foo.c`. So in practice, it makes no difference whether shared_flag was passed or not, and `compile("libfoo.so")` would make you think you'd end up with a dynamic library, but you don't.

jerry73204:
Is the issue intended to be fixed?

ssh352:
how can I build dylib with cc crate?

zhangzhuang15:
Same Problem.

XuKeli8080174:
Same Problem too

zhangzhuang15:
你好 ，您发给我的邮件我已经收到。

sladg:
Any solution to this one?

NobodyXu:
This issue is due to cc always run ar, changing it should be possible.

cc @ChrisDenton @thomcc What's your thoughts on this?

silvanshade:
Here's a workaround for this:

```rust
let out_dir = Path::new(std::env::var("OUT_DIR").unwrap());
let builder = cc::Build::new().file("src/foo.c");
let objects = builder.compile_intermediates();
builder
    .get_compiler()
    .to_command()
    .args(["-shared", "-o"])
    .arg(out_dir.join("libfoo.so"))
    .args(&objects)
    .status()
    .unwrap();
```

Walter-Reactor:
Just following up here, this is doubly true on MSVC where the -shared flag actually triggers an error

zhangzhuang15:
I've received your message, and will respond  as soon as possible.

madsmtm:
Heads up: This is being discussed in https://github.com/rust-lang/cc-rs/pull/1444.

[...]

PR #1444, opened 2025-03-30
feat: support build and link dynamic library

westhide:
Close [shared_flag and static_flag don't work](https://github.com/rust-lang/cc-rs/issues/594)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -332,6 +332,7 @@ pub struct Build {
     shell_escaped_flags: Option<bool>,
     build_cache: Arc<BuildCache>,
     inherit_rustflags: bool,
+    link_shared_flag: bool,
 }
 
 /// Represents the types of errors that may occur while using cc-rs.
@@ -459,6 +460,7 @@ impl Build {
             shell_escaped_flags: None,
             build_cache: Arc::default(),
             inherit_rustflags: true,
+            link_shared_flag: false,
         }
     }
 
@@ -1227,6 +1229,14 @@ impl Build {
         self
     }
 
+    /// Configure whether cc should build dynamic library and link with `rustc-link-lib=dylib`
+    ///
+    /// This option defaults to `false`.
+    pub fn link_shared_flag(&mut self, link_shared_flag: bool) -> &mut Build {
+        self.link_shared_flag = link_shared_flag;
+        self
+    }
+
     #[doc(hidden)]
     pub fn __set_env<A, B>(&mut self, a: A, b: B) -> &mut Build
     where
@@ -1380,6 +1390,22 @@ impl Build {
         Ok(is_supported)
     }
 
+    fn get_canonical_library_names(name: &str) -> (&str, String, String) {
+        let lib_striped = name.strip_prefix("lib").unwrap_or(name);
+        let lib_name = if lib_striped.ends_with(".a") {
+            &lib_striped[..lib_striped.len() - 2]
+        } else if lib_striped.ends_with(".so") {
+            &lib_striped[..lib_striped.len() - 3]
+        } else {
+            lib_striped
+        };
+        (
+            lib_name,
+            format!("lib{lib_name}.a"),
+            format!("lib{lib_name}.so"),
+        )
+    }
+
     /// Run the compiler, generating the file `output`
     ///
     /// This will return a result instead of panicking; see [`Self::compile()`] for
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1396,21 +1418,25 @@ impl Build {
...
+
+            let mut command = self.try_get_compiler()?.to_command();
+            let cmd = command
+                .args(["-shared", "-o"])
+                .arg(dst.join(dynlib_name))
+                .args(&objects);
+            run(cmd, &self.cargo_output)?;
+        }
+
+        self.assemble(lib_name, &dst.join(static_name), &objects)?;
```
NobodyXu, line 1439: I think we don't need to run assemble if it's already dynamically linked?

tests/test.rs
```diff
@@ -532,8 +532,8 @@ fn gnu_apple_sysroot() {
         test.shim("fake-gcc")
             .gcc()
             .compiler("fake-gcc")
-            .target(&target)
-            .host(&target)
+            .target(target)
+            .host(target)
             .file("foo.c")
             .compile("foo");
```
NobodyXu, line 539: Can we have a test for dynlib?

Just one simple test case would be enough

NobodyXu (changes requested):
Thank you!

There are two places where I would like to see some changes

tests/test.rs
westhide: Test added. 
BTW, compile dynamic library relays on the output objects files, `cc-shim` test tool will cause link error, `gnu_link_shared` test case use native `cc::Build`.

src/lib.rs
westhide: Should we consider support `dll` on `windows`?

westhide: > I think we don't need to run assemble if it's already dynamically linked?

Sure, fixed.

NobodyXu: > Should we consider support `dll` on `windows`?

Can you elaborate on that please?

src/lib.rs
```diff
@@ -1227,6 +1229,14 @@ impl Build {
         self
     }
 
+    /// Configure whether cc should build dynamic library and link with `rustc-link-lib=dylib`
+    ///
+    /// This option defaults to `false`.
```
NobodyXu, line 1234: I think we should also ask user to set the `out_dir`, it's in `target` by default so once it's removed it will stop functioning

tests/test.rs
```diff
@@ -354,6 +354,37 @@ fn gnu_shared() {
     test.cmd(0).must_have("-shared").must_not_have("-static");
 }
 
+#[test]
+#[cfg(target_os = "linux")]
+fn gnu_link_shared() {
+    use std::process::Command;
+
+    let output = Command::new("rustc").arg("-vV").output().unwrap();
```
NobodyXu, line 362:
```suggestion
    let output = Command::new(env::var_os("RUSTC").unwrap()).arg("-vV").output().unwrap();
```

This would make sure it uses the same rustc as the cargo process that runs the tests

NobodyXu (changes requested):

src/lib.rs
westhide: provide `shared_lib_out_dir` for better UX

tests/test.rs
NobodyXu: Got it, in that case let's keep it as-is

NobodyXu (approved):
Thanks LGTM!

cc @madsmtm can you review this PR please?

madsmtm (changes requested):

madsmtm:
In general, I'm a bit inclined to think that `-shared` is an anti-feature of `cc-rs`, since:
- You never want this in a `build.rs` script, since then you need to distribute the shared library.
- Outside of `build.rs` it _may_ be valuable, but it also introduces a big cost, since now `cc-rs` has to know about linker flags as well (see e.g. https://github.com/rust-lang/cc-rs/pull/1424). I'd be more in favour of deprecating `shared_flag`, and instead recommending a `linker-rs` crate or similar.

westhide:
> In general, I'm a bit inclined to think that `-shared` is an anti-feature of `cc-rs`, since:
> 
> * You never want this in a `build.rs` script, since then you need to distribute the shared library.
> * Outside of `build.rs` it _may_ be valuable, but it also introduces a big cost, since now `cc-rs` has to know about linker flags as well (see e.g. [Fix wasm32-unknown-unknown by passing -c #1424](https://github.com/rust-lang/cc-rs/pull/1424)). I'd be more in favour of deprecating `shared_flag`, and instead recommending a `linker-rs` crate or similar.

So should i close this PR and use another crate to support dynamic link? The propose to submit this commit is I want to implement an C/C++ UDF with rust compute engine. I fine for me to just call bash command in `build.rs`.

madsmtm:
> The purpose of submitting this commit was to implement a C/C++ UDF with Rust's compute engine. It's fine for me to just call a bash command in build.rs.

Hmm, I guess I don't understand why that has to be dynamically linked?

> Should I close this PR and use another crate to support dynamic linking? 

We haven't actually decided to deprecate `.shared_flag` yet, so I'm still unsure what our recommendation should be.

NobodyXu:
I think this is mostly an ALI issue, since it doesn't add a lot of code we just want the right API.

madsmtm:
> Since I didn’t want to introduce a breaking change, I added the link_shared_flag function to properly support dynamic linking.

I don't think this should _really_ be seen as a breaking change, since `.shared_flag(true)` is broken today anyhow.

Alternatively, we deprecate both `.shared_flag` and `.static_flag`, and introduce `.shared` or `.dynamic`.

> In our use case, we have C/C++ UDF code that changes frequently. Static linking isn’t ideal because users would need to download and update the entire binary every time the UDF code changes, which can be time-consuming. With dynamic linking, as long as we keep the Rust FFI interface stable, users only need to:
> 
> * Update the latest UDF `.so` file,
> 
> * Reset the `LD_LIBRARY_PATH` environment variable.
> 
> This approach significantly reduces update overhead.

Thanks for the context. Would you mind adding something to this effect to the method documentation, as a use-case example?

tests/test.rs
```diff
@@ -532,8 +563,8 @@ fn gnu_apple_sysroot() {
         test.shim("fake-gcc")
             .gcc()
             .compiler("fake-gcc")
-            .target(&target)
-            .host(&target)
+            .target(target)
+            .host(target)
             .file("foo.c")
             .compile("foo");
```
madsmtm, line 570: Documentation on `compile` should also be updated to reflect the new renaming rules.

madsmtm, line 570: If we _actually_ want to support shared libraries, we should consider a `.linker` and `.get_linker` similar to `.archiver` and `.get_archiver`, as well as maybe allowing overriding the linker with an `LD` env var (?)

Though if we allow using a raw linker (e.g. invoke `rust-lld` binary instead of `cc`), we're getting into hairy territory of needing to know how to invoke the linker if it's not a compiler driver.

madsmtm: (Which again is why I'm really not in favour of it).

PR #1447, opened 2025-04-02
Use `std::thread::available_parallelism` for determining the default number of jobs

Kobzol:
The function was stabilized in 1.59, while the current MSRV is 1.63.

### DIFF
--- a/src/parallel/job_token.rs
+++ b/src/parallel/job_token.rs
@@ -233,19 +233,14 @@ mod inprocess_jobserver {
     impl JobServer {
         pub(super) fn new() -> Self {
             // Use `NUM_JOBS` if set (it's configured by Cargo) and otherwise
-            // just fall back to a semi-reasonable number.
-            //
-            // Note that we could use `num_cpus` here but it's an extra
-            // dependency that will almost never be used, so
-            // it's generally not too worth it.
-            let mut parallelism = 4;
-            // TODO: Use std::thread::available_parallelism as an upper bound
-            // when MSRV is bumped.
-            if let Ok(amt) = var("NUM_JOBS") {
-                if let Ok(amt) = amt.parse() {
-                    parallelism = amt;
-                }
-            }
+            // just fall back to the number of cores on the local machine, or a reasonable
+            // default if that cannot be determined.
+
+            let parallelism = var("NUM_JOBS")
+                .ok()
+                .and_then(|j| j.parse::<u32>().ok())
+                .or_else(|| Some(std::thread::available_parallelism().ok()?.get() as u32))
+                .unwrap_or(4);
 
             Self(AtomicU32::new(parallelism))
         }
--- a/src/parallel/job_token.rs
+++ b/src/parallel/job_token.rs
@@ -80,7 +80,7 @@ mod inherited_jobserver {
 
     pub(super) struct JobServer {
         /// Implicit token for this process which is obtained and will be
-        /// released in parent. Since JobTokens only give back what they got,
+        /// released in parent. Since `JobTokens` only give back what they got,
         /// there should be at most one global implicit token in the wild.
         ///
         /// Since Rust does not execute any `Drop` for global variables,
@@ -164,7 +164,7 @@ mod inherited_jobserver {
         helper_thread: Option<HelperThread>,
     }
 
-    impl<'a> ActiveJobServer<'a> {
+    impl ActiveJobServer<'_> {
         pub(super) async fn acquire(&mut self) -> Result<JobToken, Error> {
             let mut has_requested_token = false;
 
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -46,7 +46,7 @@ impl CargoOutput {
             warnings: true,
             output: OutputKind::Forward,
             debug: match std::env::var_os("CC_ENABLE_DEBUG_OUTPUT") {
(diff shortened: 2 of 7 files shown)


src/parallel/job_token.rs
```diff
@@ -238,9 +238,10 @@ mod inprocess_jobserver {
...
             // dependency that will almost never be used, so
             // it's generally not too worth it.
-            let mut parallelism = 4;
-            // TODO: Use std::thread::available_parallelism as an upper bound
-            // when MSRV is bumped.
+
+            let mut parallelism = std::thread::available_parallelism()
+                .map(|v| usize::from(v) as u32)
+                .unwrap_or(4);
             if let Ok(amt) = var("NUM_JOBS") {
```
madsmtm, line 245: Thanks! Could I get you to:
1. Not invoke `available_parallelism` if `NUM_JOBS` is set.
2. Update the comments above (they're now outdated, as we wouldn't want to use `num_cpus`).

Kobzol: Sure, will do! I tried to write the whole function as one long functional chain, but it wasn't super pretty :D I'll try to find some middle ground.

Kobzol:
Changed the code, let me know what do you think :) I used `ok()` because getting an environment variable and parsing a number have different `Error` types, so they don't compose well.

src/parallel/job_token.rs
```diff
@@ -233,19 +233,16 @@ mod inprocess_jobserver {
...
-            // when MSRV is bumped.
-            if let Ok(amt) = var("NUM_JOBS") {
-                if let Ok(amt) = amt.parse() {
-                    parallelism = amt;
-                }
-            }
+            // just fall back to the number of cores on the local machine.
+
+            let parallelism =
+                if let Some(num_jobs) = var("NUM_JOBS").ok().and_then(|j| j.parse::<u32>().ok()) {
```
madsmtm, line 239: Nit: Do something like:
```suggestion
            let num_jobs = var("NUM_JOBS").ok().and_then(|j| j.parse::<u32>().ok())
            let parallelism = if let Some(num_jobs) = num_jobs {
```

(Formats better).

madsmtm (approved):
Looks fine after formatting nit

src/parallel/job_token.rs
madsmtm: (At least I just really dislike the extra indentation the entire `if` block gets.

Kobzol: Yeah I also wasn't a fan of that indentation :grin:

Kobzol:
Reformatted.

src/parallel/job_token.rs
NobodyXu: IMO maybe better to use chained or_else/* calls

Kobzol:
CI failure is unrelated, due to stable Rust being bumped :) Might be better to pin Rust versions on CI.

src/parallel/job_token.rs
```diff
@@ -233,19 +233,14 @@ mod inprocess_jobserver {
...
+            // just fall back to the number of cores on the local machine.
+
+            let num_jobs = var("NUM_JOBS").ok().and_then(|j| j.parse::<u32>().ok());
+            let parallelism = num_jobs.unwrap_or_else(|| {
+                std::thread::available_parallelism()
+                    .map(|v| usize::from(v) as u32)
+                    .unwrap_or(4)
+            });
 
             Self(AtomicU32::new(parallelism))
```
NobodyXu, line 245:
```suggestion
            let parallelism = var("NUM_JOBS")
                .ok()
                .and_then(|j| j.parse::<u32>().ok())
                .or_else(|| Some(std::thread::available_parallelism().ok()?.get() as u32))
                .unwrap_or(4);
```

@madsmtm wdyt

NobodyXu (changes requested):

NobodyXu:
> CI failure is unrelated, due to stable Rust being bumped :) Might be better to pin Rust versions on CI.

I think they are fixed on main, so if you rebase it should be fine?

If my memory is wrong and it isn't fixed, can you please fix it?

I prefer not to pin rust versions and use latest releases

src/parallel/job_token.rs
Kobzol: I think that the amount of useful bikesheddin on a PR of this size has been spent xD This looks fine, I'll apply that.

Kobzol:
They don't look to be solved, I'll fix them in this PR then.

Kobzol:
I fixed most of the issues, but there is a problem with using the `env` stdlib function in the `JobServer::new` function. The project wants me to use `Build` instead, but it isn't clear to me how to thread a `Build` instance to this function. Should I just do `Build::new().getenv()`?

Kobzol:
Hmm, CI is green, but it failed locally :man_shrugging: Good enough for me I guess, maybe I ran Clippy in a different way than was expected.

NobodyXu:
> I fixed most of the issues, but there is a problem with using the `env` stdlib function in the `JobServer::new` function. The project wants me to use `Build` instead, but it isn't clear to me how to thread a `Build` instance to this function. Should I just do `Build::new().getenv()`?

Ohh it's a clippy we setup for our project, if you run `cargo clippy --no-deps` it should fix it

NobodyXu (approved):
Thank you!

PR #1453, opened 2025-04-06
Fix musl compilation: Add `musl` as a prefix fallback

### LINKED ISSUE #1452: fallback for the x86-64-linux-musl-gcc is not available and cause build failures
Build env (docker image):

> cat /etc/*-release
> PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
> NAME="Debian GNU/Linux"
> VERSION_ID="11"
> VERSION="11 (bullseye)"
> VERSION_CODENAME=bullseye
> ID=debian

Installed musl packages:
> musl/oldstable,now 1.2.2-1 amd64 [installed,automatic]
> musl-dev/oldstable,now 1.2.2-1 amd64 [installed]
> musl-tools/oldstable,now 1.2.2-1 amd64 [installed]
Cargo version:

> cargo 1.85.0 (d73d2caf9 2024-12-31)
> release: 1.85.0
> commit-hash: d73d2caf9e41a39daf2a8d6ce60ec80bf354d2a7
> commit-date: 2024-12-31
> host: x86_64-unknown-linux-gnu
> libgit2: 1.8.1 (sys:0.19.0 vendored)
> libcurl: 8.9.0-DEV (sys:0.4.74+curl-8.9.0 vendored ssl:OpenSSL/1.1.1w)
> ssl: OpenSSL 1.1.1w 11 Sep 2023
> os: Linux [64-bit]

After update of cc to 1.2.18, the build for the target x86_64-unknown-linux-musl fails with the error message 

> /bin/sh: 1: x86_64-linux-musl-gcc: not found
>   make[1]: *** [Makefile:3058: apps/lib/libapps-lib-app_libctx.o] Error 127
>   make: *** [Makefile:1577: build_libs] Error 2

According to the https://github.com/rust-lang/cc-rs/pull/1443, the /usr/bin/musl-gcc should be the symlink, but it's actually not:

> file /usr/bin/musl-gcc
> /usr/bin/musl-gcc: POSIX shell script, ASCII text executable

The issue first appeared with the cc crate 1.2.18, it was not present in 1.2.17 and below

Looks like the fallback for the musl-gcc should be preserved

UPD: workaround which works for me: creating the symlink to musl-gcc:

> ln -s /usr/bin/musl-gcc /usr/bin/x86_64-linux-musl-gcc

NobodyXu:
Fixed #1452

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2884,14 +2884,21 @@ impl Build {
                         "qcc".to_string()
                     }
                 } else if self.get_is_cross_compile()? {
-                    let prefix = self.prefix_for_target(&raw_target);
-                    match prefix {
-                        Some(prefix) => {
+                    self.prefix_for_target(&raw_target)
+                        .filter_map(|prefix| {
                             let cc = if target.abi == "llvm" { clang } else { gnu };
-                            format!("{}-{}", prefix, cc)
-                        }
-                        None => default.to_string(),
-                    }
+                            let prefixed_cc = format!("{}-{}", prefix, cc);
+                            let status = Command::new(&prefixed_cc)
+                                .arg("--version")
+                                .stdin(Stdio::null())
+                                .stdout(Stdio::null())
+                                .stderr(Stdio::null())
+                                .status()
+                                .ok()?;
+                            status.success().then_some(prefixed_cc)
+                        })
+                        .next()
+                        .unwrap_or_else(|| default.to_string())
                 } else {
                     default.to_string()
                 };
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3514,8 +3520,13 @@ impl Build {
                     "x86_64-unknown-netbsd" => Some("x86_64--netbsd"),
                     _ => None,
```
madsmtm, line 3521: I think all of this is unnecessary; we should just use:
```rust
"x86_64-unknown-linux-musl" => self.find_working_gnu_prefix(&["x86_64-linux-musl", "musl"]),
```

NobodyXu: Thanks, I will open another PR for this (since it's easier to start from scratch)

PR #1466, opened 2025-05-07
Add `flags` method to `cc::Build` for adding multiple flags

### LINKED ISSUE #1465: Add `.flags()` to `cc::Build` for passing multiple compiler flags
`cc::Build` has `.flag()` but lacks a `.flags()` method.

This becomes verbose when integrating with tools like `llvm-config`. For example:

```rust
let output = Command::new("llvm-config")
    .arg("--cxxflags")
    .output()
    .unwrap();
let flags = shell_words::split(std::str::from_utf8(&output.stdout).unwrap()).unwrap();

for flag in &flags {
    build.flag(flag);
}
```

With `.flags()`, this becomes simpler:

```rust
cc::Build::new().flags(&flags);
```

I think it is reasonable to add .flags() just like .files() method. I'm sorry if I am just missing something.

fiveseven-lambda:
Closes #1465. Thank you @NobodyXu.

My concerns:

- Should we have some tests?
- Should the `flags` method also handle the case where a single string of flags (e.g., `"flag1 flag2"`) is passed and split it into individual flags? Alternatively, should we delegate this responsibility to an external crate like [shell-words](https://crates.io/crates/shell-words) to handle more complex cases like quoted strings with spaces?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -561,6 +561,19 @@ impl Build {
         self
     }
 
+    /// Add multiple flags to the invocation of the compiler.
+    /// This is equivalent to calling `.flag()` for each item in the iterator.
+    pub fn flags<Iter>(&mut self, flags: Iter) -> &mut Build
+    where
+        Iter: IntoIterator,
+        Iter::Item: AsRef<OsStr>,
+    {
+        for flag in flags {
+            self.flag(flag);
+        }
+        self
+    }
+
     /// Removes a compiler flag that was added by [`Build::flag`].
     ///
     /// Will not remove flags added by other means (default flags,


src/lib.rs
```diff
@@ -561,6 +561,19 @@ impl Build {
         self
     }
 
+    /// Add multiple flags to the invocation of the compiler.
+    /// This is equivalent to calling `.flag()` for each item in the iterator.
```
NobodyXu, line 565: Having a small no_run example would be great

NobodyXu (changes requested):
Thanks!

> Should we have some tests?

IMO having a no_run example is enough

> Should the flags method also handle the case where a single string of flags (e.g., "flag1 flag2") is passed and split it into individual flags? 

No that should not be part of this function.

Having another function for this is acceptable, as we already do that kinds of parsing for environment variables.

src/lib.rs
```diff
@@ -561,6 +561,27 @@ impl Build {
...
     }
 
+    /// Add multiple flags to the invocation of the compiler.
+    /// This is equivalent to calling [`flag`](Self::flag) for each item in the iterator.
+    ///
+    /// # Example
+    /// ```no_run
+    /// cc::Build::new()
+    ///     .file("src/foo.c")
+    ///     .flag("-Wall -Wextra")
```
NobodyXu, line 571:
```suggestion
    ///     .flags(["-Wall, "-Wextra"])
```

NobodyXu (changes requested):

PR #1475, opened 2025-05-27
Also set `SDKROOT` when building apple platforms

chyyran:
This also has the downstream effect of properly setting `CMAKE_OSX_ARCHITECTURES` and `CMAKE_OSX_SYSROOT` automatically for non-macOS apple targets for cmake projects

Picks up where https://github.com/rust-lang/cmake-rs/pull/197 leaves off, obsoletes https://github.com/rust-lang/cmake-rs/pull/244


chyyran:
cc @madsmtm

NobodyXu:
Thanks, LGTM

PR #1479, opened 2025-06-03
use windows 2022 in CI

marcoieni:
I received this email from GitHub:
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
             win64,
             mingw32,
             mingw64,
-            windows-2019,
+            windows-2022,
           ]
         cargo_flags: ['', '--release', '--features parallel']
         include:
@@ -102,7 +102,7 @@ jobs:
             target: aarch64-pc-windows-msvc
             no_run: --no-run
           - build: win32
-            os: windows-2019
+            os: windows-2022
             rust: stable-i686-msvc
             target: i686-pc-windows-msvc
           - build: win64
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -33,7 +33,7 @@ jobs:
             win64,
             mingw32,
             mingw64,
-            windows-2019,
+            windows-2022,
```
NobodyXu, line 36: Could you please update the dependency of test-pass job at the end of the workflow?

Updating it would fix the pipeline

NobodyXu (changes requested):
Thanks, just have a suggested change to fix the pipeline

marcoieni:
I realized I forget to upload many other instances of windows-19.

I didn't get what I have to update 🤔

.github/workflows/main.yml
```diff
@@ -102,37 +102,33 @@ jobs:
...
             target: i686-pc-windows-msvc
           - build: win64
             os: windows-latest
             rust: stable
             target: x86_64-pc-windows-msvc
           - build: mingw32
-            # windows-latest, a.k.a. windows-2022, runner is equipped with
-            # a newer mingw toolchain, which appears to produce unexecutable
-            # mixed-language binaries in debug builds. Fall back to
-            # windows-2019 for now and revisit it later...
```
marcoieni, line 116: can you check if this is still true?

NobodyXu: If the test succeeds, then maybe it's fine

cc @ChrisDenton wdyt

ChrisDenton: I think this was to do with https://github.com/rust-lang/cc-rs/pull/694#issuecomment-1179797916. If the tests pass then it would appear that issue has been fixed. But @mati865 is the expert here.

mati865: I don't remember much from this issue, I guess we're fine now if it passes.

.github/workflows/main.yml
NobodyXu: Sorry my brain wasn't working, this obviously isn't an issue

PR #1488, opened 2025-06-30
Recognize `mlibc` environment

no92:
This PR serves to upstream the mlibc environment so that the `x86_64-unknown-managarm-mlibc` target can be upstreamed into rustc (see rust-lang/rust#123319) after this PR appears in a tagged release (and rustc bumps its cc-rs version for bootstrap).

This is the triple upstreamed into LLVM (as of commit llvm/llvm-project@0f302f38b0014a0018031ffb3cb898fdc7d90880), so no further changed are needed. Running the `rustc_target_test` with a rustc with this and the rustc PR applied succeeds, as do all other tests.


no92:
The clippy warnings about `format!` look unrelated to this PR.

no92:
Rebased on main.

PR #1490, opened 2025-07-03
Fix target parsing for powerpc

NobodyXu:
Fix https://github.com/rust-lang/cc-rs/actions/runs/16040831581/job/45261866398#step:6:56

### DIFF
--- a/src/target/parser.rs
+++ b/src/target/parser.rs
@@ -402,10 +402,7 @@ impl<'a> TargetInfo<'a> {
         if vendor == "uwp" {
             abi = "uwp";
         }
-        if [
-            "powerpc64-unknown-linux-gnu",
-            "powerpc64-wrs-vxworks",
-        ].contains(&target) {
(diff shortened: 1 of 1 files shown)


ChrisDenton:
It surprises me that this works. I'd expect there to be other ABIs too. Have you tested this locally with the latest nightly?

ChrisDenton:
Hm, yeah, I don't think overriding the ABI like this will work:

```
> rustc +nightly --target powerpc64-ibm-aix --print cfg | rg abi
target_abi="vec-extabi"
```

NobodyXu:
maybe we need more fine grand override?

like doing a target lookup

NobodyXu:
Ran the tests again and now it succeeds

https://github.com/rust-lang/cc-rs/actions/runs/16073912807/job/45364533415

NobodyXu:
cc @ChrisDenton 

BTW do we want another cc release for this fix?

ChrisDenton:
> BTW do we want another cc release for this fix?

I think so. Unless there's going to be another release soon anyway.

PR #1495, opened 2025-07-09
Stop passing an invalid target to `llvm-mingw`'s cross-compilation wrappers

AlexTMjugador:
The cross-compilation toolchains distributed by the `llvm-mingw` project for building code for the `*-pc-windows-gnullvm` targets have a wrapper script in place of their `clang` binaries that calls them with a specific target, much like it has been the case with the Android NDK. When `cc-rs` sets a explicit `--target` flag, the wrapper script breaks down, and is no longer able to use the proper linker and libraries.

To improve on this situation and make it possible to cross-compile Rust projects that depend on building C code with `cc-rs`, let's skip passing this flag when cross-compilation with such a wrapper script is attempted.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3003,6 +3003,23 @@ impl Build {
             };
         }
 
+        // Under cross-compilation scenarios, llvm-mingw's clang executable is just a
+        // wrapper script that calls the actual clang binary with a suitable `--target`
+        // argument, much like the Android NDK case outlined above. Passing a target
+        // argument ourselves in this case will result in an error, as they expect
+        // targets like `x86_64-w64-mingw32`, and we can't always set such a target
+        // string because it is specific to this MinGW cross-compilation toolchain.
+        //
+        // For example, the following command will always fail due to using an unsuitable
+        // `--target` argument we'd otherwise pass:
+        // $ /opt/llvm-mingw-20250613-ucrt-ubuntu-22.04-x86_64/bin/x86_64-w64-mingw32-clang --target=x86_64-pc-windows-gnu dummy.c
+        //
+        // Code reference:
+        // https://github.com/mstorsjo/llvm-mingw/blob/a1f6413e5c21fd74b64137b56167f4fba500d1d8/wrappers/clang-target-wrapper.sh#L31
+        if !cfg!(windows) && target.os == "windows" && is_llvm_mingw_wrapper(&tool.path) {
+            tool.has_internal_target_arg = true;
+        }
+
         // If we found `cl.exe` in our environment, the tool we're returning is
         // an MSVC-like tool, *and* no env vars were set then set env vars for
         // the tool that we're returning.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -4171,6 +4188,17 @@ fn android_clang_compiler_uses_target_arg_internally(clang_path: &Path) -> bool
     false
 }
 
+fn is_llvm_mingw_wrapper(clang_path: &Path) -> bool {
+    if let Some(filename) = clang_path
+        .file_name()
+        .and_then(|file_name| file_name.to_str())
```
NobodyXu, line 4194: Is there a more robust way of testing?

I.e. checking version etc?

AlexTMjugador: Thanks for the review! I agree that a more robust detection method could be nice.

After taking a closer look at the `llvm-mingw` repository, it appears that [this cross-compilation toolchain has been setting a target since its very first tagged release](https://github.com/mstorsjo/llvm-mingw/commit/b5deb9fc1ff16dd795261373991c5c0a240d6dc9), [`20181017`](https://github.com/mstorsjo/llvm-mingw/releases/tag/20181017).

For Unix host platforms, the wrapper script has consistently been symlinked from `x86_64-w64-mingw32-clang` and similar names, so the toolchain has been quite consistent in this regard. As a result, checking its version doesn't seem necessary.

Interestingly, running `x86_64-w64-mingw32-clang -v` on any version of the toolchain reports the default target as `x86_64-w64-windows-gnu`. This doesn't exactly match either `x86_64-pc-windows-gnu` or `x86_64-w64-mingw32`, but it still seems to work just fine. (Why are LLVM's target names so inconsistent? 😅 I don't think I'd want to rely on the reported target for any important decisions on `cc-rs`...)

That said, aside from the target string oddities, this is a fairly standard LLVM toolchain with varying standard LLVM versions that aren't particularly distinct from stock LLVM. I don't think it's worth running any command with potentially brittle output or implement any extra complexity to determine whether to skip passing a target parameter in this case.

Instead, I figured we could check whether the compiler path is a symlink to a `clang-target-wrapper.sh` script. If and only if it is, we return `true` in this function. This approach should have a negligible chance of false positives, and in the unforeseen case that something goes wrong, users still have a fallback option other than fiddling with compiler flags: modifying the wrapper script directly.

What do you think?

NobodyXu: I'm worried about something else pretending to be it...or the script gets renamed in the future

cc @madsmtm @ChrisDenton do you think there's a more robust way?

NobodyXu: I'm worried about something else pretending to be it...or the script gets renamed in the future

cc @madsmtm @ChrisDenton do you think there's a more robust way?

NobodyXu (commented):
(Sorry my comment is pending for a while and forgot I need to post them)

src/lib.rs
NobodyXu: cc @thomcc

NobodyXu (approved):
Thanks!

while it is unfortunate we have to use filename to detect it, we can't find a better way for now, so I'll merge this

AlexTMjugador:
Awesome, thanks a lot @NobodyXu! Feel free to ping me if you could use any help with further tweaks to this :+1:

PR #1497, opened 2025-07-16
Proposal: ability to statically link against C++ stdlib

Vaider7:
Hi everyone!

I want to create a little proposal. It's very common situation when you need to statically link C++ stdlib to your binary. Usually it makes your binary a little bit more portable, especially if your app will be distributed to old systems. So this PR introduce an ability to explicitly say that you want to statically link C++ stdlib

Also maybe it will be useful to have a function to set a path where stdlib in .a format can be found, but I'm not sure about this one

### DIFF
--- a/dev-tools/cc-test/build.rs
+++ b/dev-tools/cc-test/build.rs
@@ -181,7 +181,7 @@ fn main() {
         // it doesn't require any additional steps to link against system
         // `libstdc++.so`, but if we emit `cargo:rustc-link-lib=static=stdc++`, it will
         // not be able to find `libstdc++.a` file despite it almost always located next to
-        // `libstdc++.so`. So providing explicit `rustc-link-search` solves the error
+        // `libstdc++.so`. So symlinking to OUT dir solves the problem
 
         let mut cmd = Command::new("g++");
         cmd.arg("--print-file-name=libstdc++.a");


NobodyXu (commented):
Thanks, would you be able to add a test case for it?

Or just test that it works for your local project?

Vaider7:
I would be happy to add test case for it, but It will be highly depending on host environment. I mean for environment where this test will be run, we need to be sure that somewhere in the system there's a file `libstdc++.a` (as example), which is available for linking, otherwise test will fail.

If it's not a problem to set CI to meet this requirement, I'll do that

Also I tested it locally for my project and it works as expected

NobodyXu:
Yes you can change our ci to install the libstdc++.a

Vaider7:
Took a look at tests, If I got it right, the main way of testing is checking emitted C compiler flags, but I'm a little bit confused how to test static linking of C++ stdlib, because it goes to stdout as `cargo:rustc-link-lib=static=stdc++` 

@NobodyXu Could you approximately describe how do you see the test case please? I'm stuck at this point

NobodyXu:
> Took a look at tests, If I got it right, the main way of testing is checking emitted C compiler flags, but I'm a little bit confused how to test static linking of C++ stdlib, because it goes to stdout as `cargo:rustc-link-lib=static=stdc++`
> 
> @NobodyXu Could you approximately describe how do you see the test case please? I'm stuck at this point

https://github.com/rust-lang/cc-rs/blob/main/dev-tools/cc-test/build.rs

we have a build script to test cc functionality, I think you can add another workspace there

Vaider7:
Wrote the test for Linux. I do not like the part with invoking g++ to find libstdc++.a file, but it's the best I can come up with. Unfortunately, linker will not automatically find static .a file like it does for shared .so file

Hope you have some ideas about that

NobodyXu (approved):
Thank you, I think the test is good enough to ensure it's working!

Issue #1501, opened 2025-07-26
Skip recompile unchanged source files


stepinto:
I found cc-rs seems to always recompile source files even if they are unchanged.

For example, in a project there are `main.rs`, `foo.c` and `bar.c`.  In `build.rs`, it compiles `foo.c` and `bar.c` into `libfoobar.a`.

```rs
    cc::Build::new()
        .file("foo.c")
        .file("bar.c")
        .compile("foobar");
```

After the initial full build, I change `foo.c`, and then `cargo build`.  I found `bar.c` is recompiled too, even it is not changed.

This is okey for small projects, but will waste large amount of time in large projects.  In my other project, where cuda is used, the compilation of a single `.cu` file takes minutes.  I think we should skip such unnecessary recompilation. 

The widely-used technique to solve this issue is to produce a `.d` file along the compilation.  That file describes a source file depeneds on what header files.  The build program checks the timestamp of the source file and the headers.  If the outcome is newer, then compilation can be safely skipped.

I can help if needed.

NobodyXu:
Yeah I agree that'd be nice to have, however there's a simpler way:

You could try using sccache by setting CC to `sccache cc`

This is under documented and we probably should add that to crate doc.

stepinto:
It works.  Thanks you @NobodyXu .  Just a quick update: setting NVCC to `sccache nvcc` does not work.  Instead, I did `export RUSTC_WRAPPER=sccache`, which works.  I created a MR #1503 for this.

PR #1503, opened 2025-07-27
doc: Add instructions on using sccache

stepinto:
Add document on how to use `sccache`, as described by issue #1501.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,7 +95,12 @@
 //!   dependencies. Note that operations that return [`Tool`]s such as
 //!   [`Build::get_compiler`] may produce less accurate results as in some cases `cc` runs
 //!   commands in order to locate compilers. Additionally, this does nothing to prevent
-//!   users from running [`Tool::to_command`] and executing the [`Command`] themselves.//!
+//!   users from running [`Tool::to_command`] and executing the [`Command`] themselves.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -95,7 +95,10 @@
 //!   dependencies. Note that operations that return [`Tool`]s such as
 //!   [`Build::get_compiler`] may produce less accurate results as in some cases `cc` runs
 //!   commands in order to locate compilers. Additionally, this does nothing to prevent
-//!   users from running [`Tool::to_command`] and executing the [`Command`] themselves.//!
+//!   users from running [`Tool::to_command`] and executing the [`Command`] themselves.
+//! * `RUSTC_WRAPPER` - If set, the specified command will be prefixed to the compiler
+//!   command. This is useful for projects that want to use
+//!   [sccache](https://github.com/mozilla/sccache) or other compiler wrappers.
```
NobodyXu: cc also support buildcache and cachepot, I think we should list them here just in case anyone needs it

NobodyXu (changes requested):
Thanks, just one nit

src/lib.rs
stepinto: > cc also support buildcache and cachepot, I think we should list them here just in case anyone needs it

Fixed.

NobodyXu (approved):
Thanks!

PR #1506, opened 2025-08-03
Windows `find_tools`: add support for finding Clang

dennisameling:
This PR extends `windows_registry::find` to be able to find LLVM/Clang tools [that can be installed as of Visual Studio 2019 version 16.2](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170). 

It can be used as follows:

```rust
use cc::windows_registry::find;

let clang = find("x64", "clang.exe");
```

... which for example resolves to `C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\LLVM\x64\bin\clang.exe` on a x64 host, or `C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\LLVM\ARM64\bin\clang.exe` on an arm64 host. The architecture in the folder is the **host** architecture. `clang.exe` by default can cross-compile to x64, x86 and arm64 targets on Windows.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -96,11 +96,19 @@ jobs:
             rust: stable
             target: aarch64-apple-ios
             no_run: --no-run
-          - build: windows-aarch64
+          - build: cross-windows-aarch64
             os: windows-latest
             rust: stable
             target: aarch64-pc-windows-msvc
             no_run: --no-run
+          - build: windows-aarch64
+            os: windows-11-arm
+            rust: stable
+            target: aarch64-pc-windows-msvc
+          - build: cross-win64
+            os: windows-11-arm
+            rust: stable
+            target: x86_64-pc-windows-msvc
           - build: win32
             os: windows-2022
             rust: stable-i686-msvc
--- a/src/windows/find_tools.rs
+++ b/src/windows/find_tools.rs
@@ -168,6 +168,15 @@ pub(crate) fn find_tool_inner(
         return impl_::find_devenv(target, env_getter);
     }
 
+    // Clang/LLVM isn't located in the same location as other tools like
+    // cl.exe and lib.exe.
+    if ["clang", "lldb", "llvm", "ld", "lld"]
+        .iter()
+        .any(|&t| tool.contains(t))
+    {
+        return impl_::find_llvm_tool(tool, target, env_getter);
+    }
+
     // Ok, if we're here, now comes the fun part of the probing. Default shells
     // or shells like MSYS aren't really configured to execute `cl.exe` and the
     // various compiler tools shipped as part of Visual Studio. Here we try to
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -96,11 +96,19 @@ jobs:
...
             target: aarch64-pc-windows-msvc
             no_run: --no-run
+          - build: windows-aarch64
+            os: windows-11-arm
+            rust: stable
+            target: aarch64-pc-windows-msvc
+          - build: cross-win64
+            os: windows-11-arm
+            rust: stable
+            target: x86_64-pc-windows-msvc
```
dennisameling, line 111: GitHub Actions has Windows 11 ARM runners nowadays, so let's leverage them to build and test on native arm64 hosts.

src/windows/find_tools.rs
```diff
@@ -509,6 +518,45 @@ mod impl_ {
         find_tool_in_vs16plus_path(r"MSBuild\Current\Bin\MSBuild.exe", target, "16", env_getter)
     }
 
+    pub(super) fn find_llvm_tool(
+        tool: &str,
+        target: TargetArch,
+        env_getter: &dyn EnvGetter,
+    ) -> Option<Tool> {
+        find_llvm_tool_vs17(tool, target, env_getter)
```
dennisameling, line 530: While Clang/LLVM [can be installed as of Visual Studio 2019 version 16.2](https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=msvc-170), I just went with Visual Studio 2022 support for now.

src/windows/find_tools.rs
```diff
@@ -509,6 +518,45 @@ mod impl_ {
...
+    ) -> Option<Tool> {
+        vs16plus_instances(target, "17", env_getter)
+            .filter_map(|path| {
+                let mut base_path = path.join(r"VC\Tools\LLVM");
+                let host_folder = match host_arch() {
+                    // The default LLVM bin folder is x86, and there's separate subfolders
+                    // for the x64 and ARM64 host tools.
+                    X86 => "",
+                    X86_64 => "x64",
+                    AARCH64 => "ARM64",
```
dennisameling, line 546: Here's what the folder structure looks like on the filesystem:

afbeelding

src/windows/find_tools.rs
```diff
@@ -1058,6 +1106,153 @@ mod impl_ {
...
+    mod tests {
+        use super::*;
+        use std::path::Path;
+        // Import the find function from the module level
+        use crate::windows::find_tools::find;
+        // Import StdEnvGetter from the parent module
+        use crate::windows::find_tools::StdEnvGetter;
+
+        #[test]
+        fn test_find_cl_exe() {
```
dennisameling, line 1131: Also added some tests for finding `cl.exe` while at it 😊

.github/workflows/main.yml
NobodyXu: Thanks, TIL

src/windows/find_tools.rs
```diff
@@ -1058,6 +1110,153 @@ mod impl_ {
...
+
+            let target_architectures = ["x64", "x86", "arm64"];
+            let mut found_any = false;
+
+            // Determine the host architecture
+            let host_arch_value = host_arch();
+            let host_name = match host_arch_value {
+                PROCESSOR_ARCHITECTURE_INTEL => "x86",
+                PROCESSOR_ARCHITECTURE_AMD64 => "x64",
+                PROCESSOR_ARCHITECTURE_ARM64 => "arm64",
```
NobodyXu: This seems to be different from the code above, I think we probably only want one version of it, and maybe put it into a function so that we can use it in tests directly?

src/windows/find_tools.rs
```diff
@@ -1058,6 +1110,153 @@ mod impl_ {
...
+                        "cl.exe should exist at: {:?}",
+                        cmd.get_program()
+                    );
+
+                    // Verify the path contains the correct host-target combination
+                    // Use case-insensitive comparison since VS IDE uses "Hostx64" while Build Tools use "HostX64"
+                    let path_str = cmd.get_program().to_string_lossy();
+                    let path_str_lower = path_str.to_lowercase();
+                    let expected_host_target_path =
+                        format!("\\bin\\host{}\\{}", host_name, target_arch);
```
NobodyXu:
```suggestion
                        format!("\\bin\\host{host_name}\\{target_arch}");
```

src/windows/find_tools.rs
```diff
@@ -509,6 +522,45 @@ mod impl_ {
...
+    }
+
+    fn find_llvm_tool_vs17(
+        tool: &str,
+        target: TargetArch,
+        env_getter: &dyn EnvGetter,
+    ) -> Option<Tool> {
+        vs16plus_instances(target, "17", env_getter)
+            .filter_map(|path| {
+                let mut base_path = path.join(r"VC\Tools\LLVM");
```
NobodyXu:
```suggestion
            .filter_map(|mut base_path| {
                base_path.push(r"VC\Tools\LLVM");
```

src/windows/find_tools.rs
```diff
@@ -509,6 +522,45 @@ mod impl_ {
...
+                    // E.g. C:\...\VC\Tools\LLVM\x64
+                    base_path.push(host_folder);
+                }
+                // E.g. C:\...\VC\Tools\LLVM\x64\bin\clang.exe
+                let tool_path = base_path.join("bin").join(tool);
+                if !tool_path.is_file() {
+                    return None;
+                }
+                let tool = Tool::with_family(tool_path, MSVC_FAMILY);
+                Some(tool)
```
NobodyXu:
```suggestion
                // E.g. C:\...\VC\Tools\LLVM\x64\bin\clang.exe
                base_path.push("bin");
                base_path.push(tool);
                base_path.is_file()
                    .then(|| Tool::with_family(base_path, MSVC_FAMILY))
```

NobodyXu (changes requested):
Thank you for the great work!

I just have some minor NIT for the PR

src/windows/find_tools.rs
dennisameling: Good catch! Updated ✅

NobodyXu (approved):
Thanks!

Next release is scheduled for this Friday

PR #1511, opened 2025-08-08
[win][arm64ec] Fix finding assembler and setting is_arm for Arm64EC

dpaoliello:


src/lib.rs
```diff
@@ -1399,7 +1399,7 @@ impl Build {
         }
 
         let mut cmd = compiler.to_command();
-        let is_arm = matches!(target.arch, "aarch64" | "arm");
+        let is_arm = matches!(target.arch, "aarch64" | "arm64ec" | "arm");
```
NobodyXu: Can we extract a function for this please?

dpaoliello: Done

NobodyXu:
Seems like it break the ci on main

https://github.com/rust-lang/cc-rs/actions/runs/16855945795/job/47749185799#step:6:1

dpaoliello:
```
clang: error: unknown argument: '-?'
```

Errr... doesn't seem related to my change. It shouldn't have affected non-MSVC targets...

dpaoliello:
Hmm... This is for build-std with nightly, did something break in the Rust project?

NobodyXu:
Maybe? i was thinking if there's some
new commits on main, combine with this PR, cause the failure

NobodyXu:
Took another look, you are right, it's probably a nightly issue

dpaoliello:
Looks like those compiler warnings were red herrings, the real error is:

```
note: ld: building for 'visionOS-simulator', but linking in object file (/Users/runner/work/cc-rs/cc-rs/target/aarch64-apple-visionos-sim/debug/build/cc-test-55b62b3e97617bab/out/libOptLinkage.a[2](ea708c7824d36062-opt_linkage.o)) built for 'visionOS'
```

dpaoliello:
I believe the root cause is <https://github.com/rust-lang/rust/pull/139451> (cc @madsmtm)

Looking at: <https://github.com/rust-lang/rust/pull/139451/files#diff-e8d43d302ad15ae5c24ca635b7a9abd544a4bac8b12b086b3f71c8da49e68538>

Seems that `sim` was changed from an `abi` to `env`, which breaks the checks in `cc`:
https://github.com/rust-lang/cc-rs/blob/ca81dcc1a77a0894b45e336957ed7a772494cee8/src/target/apple.rs#L33-L47

@wesleywiser can we do a crater run with nightly? Seems like this is a breaking change...

Guessing the fix in cc is to check both `abi` AND `env`?

NobodyXu:
Fixed in #1517

NobodyXu:
Thank you for finding the rust PR for me!

wesleywiser:
I don't think a crater run will really find anything as it effectively can only test `{i686,x86_64}-unknown-linux-{gnu,musl}`. Even the breakage detected in CI here would not have been uncovered by crater as far as I can tell.

I agree there's a compatibility concern here but that was disclosed on https://github.com/rust-lang/rust/pull/139451 and the compiler team FCP'd the change. I'll make sure we have a release note for that.

PR #1512, opened 2025-08-08
[win][arm64ec] Add testing for Arm64EC Windows

dpaoliello:
* Add a native and cross-build Arm64EC job, using beta as stable doesn't have <https://github.com/rust-lang/rust/pull/143387>.
* Add a dedicated arm64ec.asm test file with correctly mangled names.
* Pass the correct cl and lib flags to the NMake test to build as Arm64EC.


NobodyXu (approved):
Thank you!

PR #1516, opened 2025-08-10
Add prefer_clang_cl_over_msvc

dennisameling:
Closes https://github.com/rust-lang/cc-rs/issues/882

This could probably replace https://github.com/rust-lang/cc-rs/pull/1367. Now that https://github.com/rust-lang/cc-rs/pull/1506 has been merged, `windows_registry_find_tool` can find `clang-cl.exe` correctly.

This PR also adds a fix for something I missed in `find_tools` as part of https://github.com/rust-lang/cc-rs/pull/1506: `clang_cl` should be set to `true` when using that toolchain.

Lastly, this PR adds some tests to ensure that the logic that @briansmith described [here](https://github.com/rust-lang/cc-rs/pull/1367#issuecomment-3156808856) works as expected.


src/windows/find_tools.rs
```diff
@@ -555,7 +556,7 @@ mod impl_ {
                 base_path.push(tool);
                 base_path
                     .is_file()
-                    .then(|| Tool::with_family(base_path, MSVC_FAMILY))
+                    .then(|| Tool::with_family(base_path, ToolFamily::Msvc { clang_cl: true }))
```
dennisameling: I missed this bit in my other PR https://github.com/rust-lang/cc-rs/pull/1506. `MSVC_FAMILY` basically sets `ToolFamily::Msvc { clang_cl: false }`, which isn't accurate when looking for `clang-cl.exe`.

Question: 

```rust
use cc::windows_registry::find;

let clang = find("x64", "clang.exe"); // Will return `ToolFamily::Msvc { clang_cl: true }`. Is that OK, or should it only be set to `true` when we're explicitly looking for `clang-cl.exe`?
let clang = find("x64", "clang-cl.exe"); // Will return `ToolFamily::Msvc { clang_cl: true }`.
```

tests/support/mod.rs
```diff
@@ -69,6 +71,14 @@ impl Test {
         t
     }
 
+    // For msvc_autodetect, don't explicitly set the compiler - let the build system discover it
```
dennisameling, line 74: Lines [115-118](https://github.com/rust-lang/cc-rs/pull/1516/files#diff-55805875b5522b6f9fb284322ec0d3768f380a00d50f3089999942ed9a73ba65R115-R118) explicitly set the compiler to `cl` when `msvc` is set to `true`, which makes our use case much harder to test. We could just rely on the build system to discover it, given that GitHub Actions runners have MSVC + Clang installed anyway. I think this would make the tests fail on Windows systems that don't have MSVC + Clang installed though. Is that an acceptable trade-off, or should I try to mock things better and not rely on MSVC + Clang _actually_ being installed when running the tests?

tests/test.rs
```diff
@@ -851,3 +851,97 @@ fn clang_android() {
...
+#[cfg(windows)]
+fn msvc_prefer_clang_cl_over_msvc_respects_explicit_cc_env() {
+    reset_env();
+
+    let test = Test::msvc_autodetect();
+    let compiler = test
+        .gcc()
+        // We can't set the CC=cl.exe environment variable directly in the test as it's removed
+        // in mod.rs, so we simulate it by setting the compiler directly
+        .compiler("cl.exe")
```
dennisameling: [This logic](https://github.com/dennisameling/cc-rs/blob/add-clang-cl-preference-tests/src/lib.rs#L2907) already ensures that the script first looks for an explicit `CC` or `CXX` env variable, and if not present, fall back to the default detection logic. That's fully in line with the expected logic that @briansmith defined [here](https://github.com/rust-lang/cc-rs/pull/1367#issuecomment-3156808856).

dennisameling:
The failing tests in CI seem to be caused by https://github.com/rust-lang/cc-rs/commit/ca81dcc1a77a0894b45e336957ed7a772494cee8, which has the same failing tests on the main branch.

NobodyXu:
https://github.com/rust-lang/cc-rs/issues/1426

Confirmed that nightly has removed env in favor of abi, so it's not caused by this PR or that one

NobodyXu:
Fixed in #1517, just rebasing would fix it

src/windows/find_tools.rs
NobodyXu: `clang_cl` should only set to true for clang-cl

For `clang`, I think it'd count as family clang instead of msvc?

tests/support/mod.rs
NobodyXu: You could add a `cfg(not(CC_TEST_DISABLE_CLANG_CL_TESTS))` just in case someone needs it

tests/test.rs
```diff
@@ -851,3 +851,97 @@ fn clang_android() {
...
+#[test]
+#[cfg(windows)]
+fn msvc_prefer_clang_cl_over_msvc_respects_explicit_cc_env() {
+    reset_env();
+
+    let test = Test::msvc_autodetect();
+    let compiler = test
+        .gcc()
+        // We can't set the CC=cl.exe environment variable directly in the test as it's removed
+        // in mod.rs, so we simulate it by setting the compiler directly
```
NobodyXu: We have a hidden [`cc::Build::__set_env`](https://docs.rs/cc/latest/src/cc/lib.rs.html#1294) function

tests/test.rs
```diff
@@ -851,3 +851,97 @@ fn clang_android() {
         test.cmd(0).must_not_have("--target=arm-linux-androideabi");
     }
 }
+
+#[test]
+#[cfg(windows)]
```
NobodyXu, line 856: Might be better to put them in a `mod` to avoid repeated cfg

tests/test.rs
dennisameling: Thanks for the suggestion. I just tried using that as follows:

```rust
    let compiler = test
        .gcc()
        // If the user has set CC to cl.exe, we should respect that
        .__set_env("CC", "cl.exe")
        .prefer_clang_cl_over_msvc(true)
        .try_get_compiler()
        .expect("Failed to get compiler");
```

However, it's not being picked up because `get_base_compiler` [calls](https://github.com/dennisameling/cc-rs/blob/add-clang-cl-preference-tests/src/lib.rs#L2910) `env_tool` which - if I understand the code correctly - [looks at the _actual_ environment variables](https://github.com/dennisameling/cc-rs/blob/add-clang-cl-preference-tests/src/lib.rs#L3140) that are available to the current process, and not at the ones that were set through the hidden `__set_env` method. So even though it looks for the `CC` env var, it doesn't find anything because it's not set at the process level.

I'm curious if you're aware of any alternatives though! I also tried `std::env::set_var("CC", "cl.exe");` which works when I run the test in isolation, but not when running the entire test suite, which runs many tests in parallel and probably shares the same process-level env vars.

dennisameling: We _could_ update the `getenv` [function](https://github.com/dennisameling/cc-rs/blob/add-clang-cl-preference-tests/src/lib.rs#L3798) to first look at `self.env` if set, and otherwise fall back to process-level env variables using `env::var_os`, which it was already doing. I'm by no means proficient in Rust so I'm sure there might be a more efficient way to do this, but the code below seems to work and the entire test suite (ran with `cargo test`) also passes with it:

```diff
diff --git a/src/lib.rs b/src/lib.rs
index 8a74141..24a7bbb 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3795,7 +3795,11 @@ impl Build {
             self.cargo_output
                 .print_metadata(&format_args!("cargo:rerun-if-env-changed={v}"));
         }
-        let r = env::var_os(v).map(Arc::from);
+        let r = self.env.iter().find(|(k, _)| k.as_ref() == v).cloned().map(|(_, v)| v);
+        let r = match r {
+            Some(ref val) if !val.is_empty() => Some(val.clone()),
+            _ => env::var_os(v).map(Arc::from),
+        };
         self.cargo_output.print_metadata(&format_args!(
             "{} = {}",
             v,
```

NobodyXu: It looks pretty good except for the empty checking and unnecessary cloning.

```rust
        let r = self.env
            .iter()
            .find(|(k, value)| k.as_ref() == v)
            .map(|(_, value)| value.clone())
            .or_else(|| env::var_os(v).map(Arc::from));
```

src/windows/find_tools.rs
dennisameling: I think that makes sense actually. It [looks like](https://clang.llvm.org/docs/UsersManual.html#clang-cl) `clang-cl.exe` supports MSVC-style flags like `/std:c++20`, while `clang.exe` only supports flags like `-std:c++20` and therefore should be API-compatible with non-MSVC Clang.

Updated ✅

dennisameling:
Looks like some (seemingly unrelated) tests are now [failing](https://github.com/rust-lang/cc-rs/actions/runs/17022681511/job/48253924403?pr=1516#step:13:434) with:

```
 thread 'apple_sdkroot_wrong' panicked at tests/support/mod.rs:143:13:
didn't find "/Applications/Xcode_15.4.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS17.5.sdk" in ["-O2", "-fPIC", "--target=arm64-apple-ios", "-miphoneos-version-min=17.5", "-isysroot", "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk", "-Wall", "-Wextra", "-o", "/Users/runner/work/cc-rs/cc-rs/target/aarch64-apple-darwin/debug/cc-shim-teste6yNGK/db3b6bfb95261072-foo.o", "-c", "foo.c"]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

... which is interesting because the documentation for `macos-14-arm64` shows that the iOS 17.5 SDK should be installed. Any thoughts on this?

NobodyXu:
It seems that the path to the sdj has changed.

We expect to find `/Applications/Xcode_15.4.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS17.5.sdk`, but found:

```
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```

Maybe we are using the wrong sdk?

Cargo.toml
```diff
@@ -33,6 +33,10 @@ parallel = ["dep:libc", "dep:jobserver"]
 # This is a placeholder feature for people who incorrectly used `cc` with `features = ["jobserver"]`
 # so that they aren't broken. This has never enabled `parallel`, so we won't do that.
 jobserver = []
+# Visual Studio doesn't have clang-cl (Clang/LLVM) installed by default as it's an optional module.
+# While GitHub Actions runners have it installed, it might cause some failing tests if you don't
+# have it. Run the tests with "--features disable-clang-cl-tests" if you want to disable related tests.
+disable-clang-cl-tests = []
```
NobodyXu: I think having a cfg is better than feature, given that it doesn't actually change any prod code, plus feature is supposed to be additive

```rust
cfg(disable_clang_cl_tests)
```

src/windows/find_tools.rs
```diff
@@ -553,9 +554,10 @@ mod impl_ {
                 // E.g. C:\...\VC\Tools\LLVM\x64\bin\clang.exe
                 base_path.push("bin");
                 base_path.push(tool);
+                let clang_cl = tool.contains("clang-cl");
```
NobodyXu, line 557: For clang-cl detection I recommend using this

https://github.com/rust-lang/cc-rs/blob/a92892393c948d1a69777a90b9dc48271f4c5c87/src/tool.rs#L123

Extracting a function and then use it

dennisameling: I'm confused as to the value that would add to this logic. In the `find_llvm_tool_vs17` function, it:
- Takes the requested `tool` (e.g. `clang.exe` or `clang-cl.exe`).
- Builds the host arch LLVM folder base path (e.g. `C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\ARM64\bin`)
- Does the `is_file()` check below to check whether the executable path exists (e.g. `C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\ARM64\bin\clang-cl.exe`).
  - If found, returns `ToolFamily::Msvc { clang_cl: true }`
  - If not found, simply return `None`

Visual Studio's LLVM/Clang has different binaries for `clang.exe` and `clang-cl.exe`, and given that this `find_llvm_tool_vs17` function expects a `tool` to be provided, we'll already know if someone is requesting `clang.exe`, `clang-cl.exe` or another related tool, so there shouldn't be a need for running the command to extract its capabilities. If I understand things correctly, looking at the [comment](https://github.com/dennisameling/cc-rs/blob/add-clang-cl-preference-tests/src/tool.rs#L493) on the `clang_cl` field, it should only be set to `true` when `clang-cl.exe` is the requested tool, because it's the only executable capable of handling those cl-specific flags.

Please let me know if I'm misunderstanding things here. I found some related logic [here](https://github.com/rust-lang/cc-rs/blob/a92892393c948d1a69777a90b9dc48271f4c5c87/src/tool.rs#L267), but extracting that single line into a function to re-use it here feels a bit like overkill to me.

In any case, thank you very much for all your continued feedback and support. Really appreciate the guidance as I'm new to both Rust and this project 🙏🏼

NobodyXu: Thanks maybe I was the one who were confused 😅 in the past I'm been very pedantic about checking filename for tool.

But here we are requesting tools by name, not filename, so it's ok

src/lib.rs
```diff
@@ -3785,7 +3802,12 @@ impl Build {
             self.cargo_output
                 .print_metadata(&format_args!("cargo:rerun-if-env-changed={v}"));
         }
-        let r = env::var_os(v).map(Arc::from);
+        let r = self
+            .env
+            .iter()
+            .find(|(k, _)| k.as_ref() == v)
+            .map(|(_, value)| value.clone())
+            .or_else(|| env::var_os(v).map(Arc::from));
```
dennisameling, line 3810: Just tested and these changes are what's causing the macOS tests to fail. When I remove [this](https://github.com/dennisameling/cc-rs/blob/add-clang-cl-preference-tests/tests/test.rs#L794) line, the tests work again on macOS, but I have a feeling it might defeat the purpose of that `apple_sdkroot_wrong()` test. WDYT?

NobodyXu: My thoughts is to comment this line for the PR to merge, and we will have to deal with it in another PR

https://github.com/dennisameling/cc-rs/blob/add-clang-cl-preference-tests/tests/test.rs

PR #1524, opened 2025-08-16
fix: add `-mcpu=mvp` for `wasm32v1-none`

### LINKED ISSUE #1523: Suspect `cc-rs` added extra instructions to `.wasm`
https://github.com/rust-lang/cc-rs/blob/a92892393c948d1a69777a90b9dc48271f4c5c87/src/target/generated.rs#L255
`wasm32v1-none` target should probably use `-C target-cpu=mvp` when compiling
https://doc.rust-lang.org/beta/rustc/platform-support/wasm32v1-none.html#enabled-webassembly-features

ref https://github.com/rust-lang/rust/issues/140174

StackOverflowExcept1on:
Resolves #1523
@NobodyXu

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2308,9 +2308,12 @@ impl Build {
                             None
                         };
 
-                    let clang_target =
-                        target.llvm_target(&self.get_raw_target()?, version.as_deref());
+                    let rustc_target = self.get_raw_target()?;
+                    let clang_target = target.llvm_target(&rustc_target, version.as_deref());
                     cmd.push_cc_arg(format!("--target={clang_target}").into());
+                    if rustc_target == "wasm32v1-none" {
+                        cmd.push_cc_arg("-mcpu=mvp".into());
+                    }
                 }
             }
             ToolFamily::Msvc { clang_cl } => {
@@ -2335,13 +2338,13 @@ impl Build {
                         // <https://github.com/microsoft/STL/pull/4741>.
                         cmd.push_cc_arg("-arch:SSE2".into());
                     } else {
+                        let rustc_target = self.get_raw_target()?;
                         cmd.push_cc_arg(
-                            format!(
-                                "--target={}",
-                                target.llvm_target(&self.get_raw_target()?, None)
-                            )
-                            .into(),
+                            format!("--target={}", target.llvm_target(&rustc_target, None)).into(),
                         );
+                        if rustc_target == "wasm32v1-none" {
+                            cmd.push_cc_arg("-mcpu=mvp".into());
+                        }
                     }
                 } else if target.full_arch == "i586" {
                     cmd.push_cc_arg("-arch:IA32".into());


src/lib.rs
```diff
@@ -2335,13 +2338,13 @@ impl Build {
...
+                        let rustc_target = self.get_raw_target()?;
                         cmd.push_cc_arg(
-                            format!(
-                                "--target={}",
-                                target.llvm_target(&self.get_raw_target()?, None)
-                            )
-                            .into(),
+                            format!("--target={}", target.llvm_target(&rustc_target, None)).into(),
                         );
+                        if rustc_target == "wasm32v1-none" {
```
StackOverflowExcept1on: I haven't checked this on clang-cl MSVC

StackOverflowExcept1on: It works, but I had to download clang from LLVM releases, not from VS2022, because MS does not provide wasm32 target

StackOverflowExcept1on: I tested `CC = clang` and `CC = clang-cl` using the FFI example for `wasm32v1-none` in our repository: https://github.com/gear-tech/gear/blob/master/stack-buffer/build.rs. I also had to disable stack checks via `/GS-` on `clang-cl`, but I can confirm that it works and produces the correct instructions.

NobodyXu: > It works, but I had to download clang from LLVM releases, not from VS2022, because MS does not provide wasm32 target

You probably want clang-cl, which simulates msrv but is based on clang, so it might support wasm?

StackOverflowExcept1on: It supports wasm, but MS compiled their clang-cl without wasm

NobodyXu: Thanks, I guess it's ok

src/lib.rs
```diff
@@ -2308,9 +2308,12 @@ impl Build {
...
                         };
 
-                    let clang_target =
-                        target.llvm_target(&self.get_raw_target()?, version.as_deref());
+                    let rustc_target = self.get_raw_target()?;
+                    let clang_target = target.llvm_target(&rustc_target, version.as_deref());
                     cmd.push_cc_arg(format!("--target={clang_target}").into());
+                    if rustc_target == "wasm32v1-none" {
+                        cmd.push_cc_arg("-mcpu=mvp".into());
+                    }
```
NobodyXu: It'd be great if we can avoid having duplicate code.

Maybe we can have another match and if below this block?

StackOverflowExcept1on: I'm afraid it will be difficult to debug if someone needs it

StackOverflowExcept1on: done

StackOverflowExcept1on: I also think it's worth adding `-mmutable-globals`

StackOverflowExcept1on:
Tag new release plz

NobodyXu:
Mext release is scheduled for Friday, if you need it I can cut one tomorrow

StackOverflowExcept1on:
I will wait

PR #1526, opened 2025-08-21
find_tool: Fallback to manually searching for tool dir if it can't otherwise be found

### LINKED ISSUE #1525: Could not find local Visual Studio install on x86_64-pc-windows-msv if C++ x64/x86 Build Tools (Latest) is not installed
cc-rs seems to be unable to locate local VS install and build tools unless *Latest* build tools component is installed from the Visual Studio Installer. Any other MSVC build tool packages (out of supported or pinned version build tool packages that can be installed from Individual Components tab) will result in cc-rs being unable to find and use Visual Studio install.

Steps to reproduce:
1. Open Visual Studio Installer
2. Install Visual Studio Community 2022
3. Select "Desktop Development with C++" workload
4. Unselect "MSVC v143 - VS 2022 C++ x64/x86 build tools (Latest)" component
5. Go to individual components tab and select other version of MSVC build tools here ("MSVC v143 - VS 2022 C++ x64/x86 build tools (v14.38 - v17.8)" in my case).
6. Install the selected components
7. Try compiling a native crate ("`libsqlite3-sys v0.27.0" in my case).
8. Observe that cc-rs cannot find installation path for cl.exe

ChrisDenton:
If all other methods fail, this attempts to find the MSVC tools bin path by checking for the existence of the bin subdirectory. As with other code, this picks the highest version from the possible candidates.

Ideally we'd also update the code for the recent Visual Studio changes but that would need more investigation and in any case this fallback will continue to be useful as a hedge against other changes that may break cc-rs.

Fixes #1525

### DIFF
--- a/src/windows/find_tools.rs
+++ b/src/windows/find_tools.rs
@@ -815,7 +815,22 @@ mod impl_ {
                 }
             }
             if version_file.is_empty() {
-                return None;
+                // If all else fails, manually search for bin directories.
+                let tools_dir: PathBuf = dir.join(r"VC\Tools\MSVC");
+                return tools_dir
+                    .read_dir()
+                    .ok()?
+                    .filter_map(|file| {
+                        let file = file.ok()?;
+                        let name = file.file_name().into_string().ok()?;
+
+                        file.path().join("bin").exists().then(|| {
+                            let version = parse_version(&name);
+                            (name, version)
+                        })
+                    })
+                    .max_by(|(_, a), (_, b)| a.cmp(b))
+                    .map(|(version, _)| version);
             }
             version_path.push(version_file);
             File::open(version_path).ok()?


src/windows/find_tools.rs
```diff
@@ -815,7 +815,20 @@ mod impl_ {
...
+                    let file = file.ok()?;
+                    let name = file.file_name();
+                    let name = name.to_str()?;
+                    if parse_version(name) > parse_version(&version)
+                        && file.path().join("bin").exists()
+                    {
+                        version.replace_range(.., name);
+                    }
+                }
+                return (!version.is_empty()).then_some(version);
```
NobodyXu:
```suggestion
                let version = tools_dir
                    .read_dir()
                    .ok()?
                    .into_iter()
                    .filter_map(Result::Ok)
                    .filter_map(|file| {
                        let name = file.file_name().into_string().ok()?;

                        file.path().join("bin").exists().then(|| (name, parse_version(&name)))
                    })
                    .max_by_key(|(_, parsed_version)| parsed_version)
                    .map(|(version, _)| version);
```

src/windows/find_tools.rs
```diff
@@ -815,7 +815,20 @@ mod impl_ {
...
-                return None;
+                // If all else fails, manually search for bin directories.
+                let tools_dir: PathBuf = dir.join(r"VC\Tools\MSVC");
+                let version = tools_dir
+                    .read_dir()
+                    .ok()?
+                    .into_iter()
+                    .filter_map(Result::Ok)
+                    .filter_map(|file| {
+                        let name = file.file_name().into_string().ok()?;
```
NobodyXu, line 825:
```suggestion
                        let name = OsString::into_string(file.file_name().).ok()?;
```

NobodyXu (changes requested):
Sorry about this

src/windows/find_tools.rs
ChrisDenton: The problem was the `file` is a `Result`. I also had to change `max_by_key` to `max_by` due to lifetime issues.

NobodyXu: Yeah once we bumped the msrv we can change it back

src/windows/find_tools.rs
```diff
@@ -815,7 +815,23 @@ mod impl_ {
                 }
             }
             if version_file.is_empty() {
-                return None;
+                // If all else fails, manually search for bin directories.
+                let tools_dir: PathBuf = dir.join(r"VC\Tools\MSVC");
+                return tools_dir
+                    .read_dir()
+                    .ok()?
+                    .filter_map(Option::Some)
```
NobodyXu:
```suggestion
```

I think we don't need this line?

NobodyXu (approved):

NobodyXu:
Thank you!

PR #1527, opened 2025-08-21
Optimize parse_version in find_tools.rs

NobodyXu:
Avoid heap alloc given that it can have at most 4 numbers

### DIFF
--- a/src/windows/find_tools.rs
+++ b/src/windows/find_tools.rs
@@ -633,9 +633,12 @@ mod impl_ {
     // Inspired from official microsoft/vswhere ParseVersionString
     // i.e. at most four u16 numbers separated by '.'
     fn parse_version(version: &str) -> Option<[u16; 4]> {
-        let mut iter = version.split('.').map(|chunk| u16::from_str(chunk)).fuse();
-        let mut get_next_number =
-            move || iter.next().transpose().ok().map(|digit| digit.unwrap_or(0));
+        let mut iter = version.split('.').map(u16::from_str).fuse();
+        let mut get_next_number = move || match iter.next() {
+            Some(Ok(version_part)) => Some(version_part),
+            Some(Err(_)) => None,
+            None => Some(0),
+        };
         Some([
             get_next_number()?,
             get_next_number()?,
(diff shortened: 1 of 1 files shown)


NobodyXu:
cc @ChrisDenton

src/windows/find_tools.rs
```diff
@@ -632,11 +632,15 @@ mod impl_ {
...
-            .collect()
+    fn parse_version(version: &str) -> Option<[Option<u16>; 4]> {
+        let mut iter = version.split('.').map(|chunk| u16::from_str(chunk)).fuse();
+        let mut get_next_number = move || iter.next().transpose().ok();
+        Some([
+            get_next_number()?,
+            get_next_number()?,
+            get_next_number()?,
+            get_next_number()?,
+        ])
```
ChrisDenton: Hm, could this just return `Option<[u16; 4]>` (the array is then the same size as a `u64`)? Using zero for the missing parts is fine.

src/windows/find_tools.rs
```diff
@@ -632,16 +632,21 @@ mod impl_ {
 
     // Inspired from official microsoft/vswhere ParseVersionString
     // i.e. at most four u16 numbers separated by '.'
-    fn parse_version(version: &str) -> Option<Vec<u16>> {
-        version
-            .split('.')
-            .map(|chunk| u16::from_str(chunk).ok())
-            .collect()
+    fn parse_version(version: &str) -> Option<[u16; 4]> {
+        let mut iter = version.split('.').map(|chunk| u16::from_str(chunk)).fuse();
```
ChrisDenton:
```suggestion
        let mut iter = version.split('.').map(u16::from_str).fuse();
```

src/windows/find_tools.rs
```diff
@@ -632,16 +632,21 @@ mod impl_ {
...
+            get_next_number()?,
+            get_next_number()?,
+            get_next_number()?,
+        ])
     }
 
     pub(super) fn find_msvc_15plus(
         tool: &str,
-        target: TargetArch,
+        target: TargetArch, cc
```
ChrisDenton: I don't think you meant to do that?

NobodyXu: Thanks reverted

src/windows/find_tools.rs
NobodyXu: Thanks

PR #1529, opened 2025-08-23
Create rust.yml

alexjalonso7777:


NobodyXu:
Thanks but we already have a working pipeline

PR #1531, opened 2025-08-24
Refactor: Extract windows registry

NobodyXu:
By Extracting it out as a separate crate, we will have
 - more freedom of changing APIs to adopt to different needs of API without breaking cc API
 - users who only need `windows_registry` doesn't have to pull in entire cc
 - faster compilation since windows-registry codegen can be done in parallel to cc compilation

image


NobodyXu:
CI semver failed due to false positive obi1kenobi/cargo-semver-checks#638 , temporarily disabling it until another release is made

.github/workflows/main.yml
```diff
@@ -402,13 +402,13 @@ jobs:
...
-        uses: actions/checkout@v5
-      - name: Check semver
-        uses: obi1kenobi/cargo-semver-checks-action@v2
+  #semver-checks:
+  #  runs-on: ubuntu-latest
+  #  steps:
+  #    - name: Checkout
+  #      uses: actions/checkout@v5
+  #    - name: Check semver
+  #      uses: obi1kenobi/cargo-semver-checks-action@v2
```
madsmtm, line 411: Nit: I think we could use `continue-on-error: true` instead of commenting out, that way, one can still inspect the error.

windows-registry/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+repository = "https://github.com/rust-lang/cc-rs"
+homepage = "https://github.com/rust-lang/cc-rs"
+documentation = "https://docs.rs/windows-registry"
+description = """
+A build-time dependency for Cargo build scripts to assist in invoking the native
+C compiler to compile native C code into a static archive to be linked into Rust
+code.
+"""
+keywords = ["build-dependencies"]
+readme = "../README.md"
```
madsmtm: Nit:

```suggestion
name = "windows-registry"
version = "0.1.0"
edition = "2018"
license = "MIT OR Apache-2.0"
repository = "https://github.com/rust-lang/cc-rs"
documentation = "https://docs.rs/windows-registry"
description = "Read MSVC versions from the registry and from COM interfaces"
keywords = ["build-dependencies"]
readme = "../README.md"
```

Also, the crate should probably have its own README if we're to treat it as a public crate.

src/tool.rs
```diff
@@ -39,6 +39,25 @@ pub struct Tool {
...
+
+        cc_tool.env = tool
+            .env()
+            .into_iter()
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect();
+
+        cc_tool
+    }
+}
```
madsmtm: If the plan is to bump `windows-registry` versions without bumping `cc-rs`, then to _actually_ be SemVer compatible, we can't have this impl.

src/lib.rs
```diff
@@ -261,11 +261,23 @@ use shlex::Shlex;
...
-// it has been since 2015, so don't break it.
-pub use windows::find_tools as windows_registry;
+/// A helper module to looking for windows-specific tools:
+/// 1. On Windows host, probe the Windows Registry if needed;
+/// 2. On non-Windows host, check specified environment variables.
+pub mod windows_registry {
+    // Regardless of whether this should be in this crate's public API,
+    // it has been since 2015, so don't break it.
+
+    pub use ::windows_registry::{find, find_vs_version, VsVers};
```
madsmtm: Similar to above, if the plan is to bump `windows-registry` versions without bumping `cc-rs`, then to be SemVer compatible, we can't re-export the `VsVers` type like this. Instead, we should do something like:
```suggestion
    pub use ::windows_registry::find; // Fine, only uses std types

    /// A version of Visual Studio
    #[derive(Debug, PartialEq, Eq, Copy, Clone)]
    #[non_exhaustive]
    pub enum VsVers {
        /// Visual Studio 12 (2013)
        #[deprecated = "Visual Studio 12 is no longer supported. cc will never return this value."]
        Vs12,
        /// Visual Studio 14 (2015)
        Vs14,
        /// Visual Studio 15 (2017)
        Vs15,
        /// Visual Studio 16 (2019)
        Vs16,
        /// Visual Studio 17 (2022)
        Vs17,
    }

    pub fn find_vs_version() -> Result<VsVers, String> {
        ::windows_registry::find_vs_version().map(|vers| match vers {
            #[allow(deprecated)]
            ::windows_registry::VsVers::Vs12 => VsVers::Vs12,
            ::windows_registry::VsVers::Vs14 => VsVers::Vs14,
            ::windows_registry::VsVers::Vs15 => VsVers::Vs15,
            ::windows_registry::VsVers::Vs16 => VsVers::Vs16,
            ::windows_registry::VsVers::Vs17 => VsVers::Vs17,
            _ => unreachable!("unknown VS version"),
        })
    }
```

windows-registry/src/find_tools.rs
```diff
@@ -137,7 +134,7 @@ pub fn find(arch_or_target: &str, tool: &str) -> Option<Command> {
 
 /// Similar to the `find` function above, this function will attempt the same
 /// operation (finding a MSVC tool in a local install) but instead returns a
-/// `Tool` which may be introspected.
+/// `Tool` which may bpub mod windows_registrye introspected.
```
madsmtm: Nit:

```suggestion
/// `Tool` which may be introspected.
```

src/lib.rs
```diff
@@ -261,11 +261,23 @@ use shlex::Shlex;
...
+
+    /// Similar to the `find` function above, this function will attempt the same
+    /// operation (finding a MSVC tool in a local install) but instead returns a
+    /// [`Tool`](crate::Tool) which may bpub mod [`windows_registry`] introspected.
```
madsmtm: Nit:

```suggestion
    /// [`Tool`](crate::Tool) which may be introspected.
```

madsmtm (commented):
Looks good with comments resolved.

.github/workflows/main.yml
NobodyXu: https://github.com/rust-lang/cc-rs/actions/runs/17190176566/job/48764717848#step:3:1

Got a link for the error, I will quickly re-install as my first prioroty it as soon as release is cut

src/tool.rs
NobodyXu: Aha right, I need to make this a private fn instead of public trait impl

NobodyXu: Thanks fixed

windows-registry/Cargo.toml
NobodyXu: added a readme

NobodyXu:
I have to rename the crate to windows-find-tools since windows-registry is already taken

https://crates.io/crates/windows-registry

windows-find-tools/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "windows-find-tools"
```
madsmtm: A few alternative naming ideas:
- `windows-tools`
- `find-msvc-tools`
- `msvc-tools`

I'd defer to @ChrisDenton for which name actually makes the most sense.

windows-find-tools/README.md
```diff
@@ -0,0 +1,23 @@
+# windows-find-tools
```
madsmtm: As [per discussion on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/351149-t-libs.2Fcrates/topic/interior.20decorating.20with.20cc-rs/near/536114787), please state that this is an "internal use" crate.

windows-find-tools/Cargo.toml
ChrisDenton: - Having `registry` in the name would have been a bad regardless, a) because the main API doesn't even use the registry and b) because that's an implementation detail in any case.
- I don't think having `windows` in the name makes sense either. This is specific to finding msvc (and not, e.g., mingw) and also it could (in principle) be used to find tools to cross-compile from another OS (though iirc we don't support that atm)
- I do think `msvc` should be in the name because that's ultimately what this is about

So of the names currently suggested I think `find-msvc-tools` is the best, if perhaps a little verbose.

NobodyXu: Thanks renamed it to find-msvc-tools @ChrisDenton @madsmtm

windows-find-tools/README.md
NobodyXu: Marked it as internal use

find-msvc-tools/README.md
```diff
@@ -1,6 +1,6 @@
 # find-msvc-tools
```
madsmtm, line 2: Please add the wording from [the crate ownership policy](https://forge.rust-lang.org/policies/crate-ownership.html):

```suggestion

> This crate is maintained by the library team, primarily for use by the `cc` crate and not intended for external use (except as a transitive dependency). This crate may make major changes to its APIs or be deprecated without warning.

```

madsmtm (approved):
Thanks, looks good now (after nit)!

PR #1541, opened 2025-09-03
Replace periods with underscores as well when parsing env variables

### LINKED ISSUE #1540: Selecting compiler specifically for `thumbv8m.main-none-eabi` and other `.` containing targets does not work
`cc-rs` allows users to override the compiler used for a specific target triple by setting the environment variable `CC_<triple>`, but for targets such as `thumv8m.main-none-eabi` this doesn't work as `.`-s are not allowed to be used in the name of an environment variable in (most?) shells, so setting e.g. `CC_thumbv8m.main_none_eabi` is not allowed.

It would be great if this could be supported, either by replacing the `.` with an `_`, or some other solution that allows us to actually set this environment variable.

datdenkikniet:
Fixes #1540


NobodyXu (approved):
Thank you!

datdenkikniet:
As a question, if you happen to have an answer: is there any sort of release cadence/ETA for when we might expect a version of `cc` to be released with this change merged in? We're not in a horrible hurry to get this working, but putting an ETA on our "update `cc` to latest` work item would be neat :)

If not, we will be as patient as necessary. Thank you for your efforts in maintaining this crate!

ETA: whoops, just saw you have a bot that performs semi-automatic upgrades in #1542. So it looks like it will be Pretty Quick :tm: Very cool, we'll be patient

PR #1553, opened 2025-09-09
[win] Search the Windows SDK for tools as well

dpaoliello:
The Rust Compiler needs to find `rc.exe` from the Windows SDK, which it does by finding the linker, then searching through the `PATH` env var to see if `rc.exe` exists anywhere there:
https://github.com/rust-lang/rust/blob/364da5d88d772fa40fb20353443595385443ac25/compiler/rustc_windows_rc/src/lib.rs#L141-L158

Instead, we can have `find-msvc-tools` search the Windows SDK paths for tools if that tool is not found in the VC path.

### DIFF
--- a/find-msvc-tools/src/find_tools.rs
+++ b/find-msvc-tools/src/find_tools.rs
@@ -300,6 +300,13 @@ mod impl_ {
         include: Vec<PathBuf>,
     }
 
+    #[derive(Default)]
+    struct SdkInfo {
+        libs: Vec<PathBuf>,
+        path: Vec<PathBuf>,
+        include: Vec<PathBuf>,
+    }
+
     struct LibraryHandle(HMODULE);
 
     impl LibraryHandle {
@@ -373,6 +380,12 @@ mod impl_ {
             }
         }
 
+        fn add_sdk(&mut self, sdk_info: SdkInfo) {
+            self.libs.extend(sdk_info.libs);
+            self.path.extend(sdk_info.path);
+            self.include.extend(sdk_info.include);
+        }
+
         fn into_tool(self, env_getter: &dyn EnvGetter) -> Tool {
             let MsvcTool {
                 tool,
@@ -389,6 +402,12 @@ mod impl_ {
             add_env(&mut tool, "PATH", path, env_getter);
             add_env(&mut tool, "INCLUDE", include, env_getter);
             tool
+        }
+    }
+
+    impl SdkInfo {
+        fn find_tool(&self, tool: &str) -> Option<PathBuf> {
+            self.path.iter().map(|p| p.join(tool)).find(|p| p.exists())
         }
     }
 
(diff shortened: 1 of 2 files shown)


find-msvc-tools/src/find_tools.rs
```diff
@@ -963,22 +986,26 @@ mod impl_ {
...
             .filter(|(path, _)| path.is_file())
-            .map(|(path, host)| {
-                let mut tool = MsvcTool::new(path);
-                tool.path.push(host);
-                tool
-            })
-            .filter_map(|mut tool| {
+            .next()
+            .or_else(|| Some((sdk_info.find_tool(tool)?, None)))
+            .map(|(tool_path, host)| {
```
NobodyXu:
```suggestion
            .chain(iter::once(|| Some((sdk_info.find_tool(tool)?, None))))
            .find_map(|(tool_path, host)| {
```

dpaoliello: `clippy` prefers `map` + `next`, so following that recommendation.

find-msvc-tools/src/find_tools.rs
```diff
@@ -963,22 +986,25 @@ mod impl_ {
...
                 )
             })
             .filter(|(path, _)| path.is_file())
-            .map(|(path, host)| {
-                let mut tool = MsvcTool::new(path);
-                tool.path.push(host);
-                tool
-            })
-            .filter_map(|mut tool| {
+            .chain(iter::from_fn(|| Some((sdk_info.find_tool(tool)?, None))))
```
NobodyXu: Sorry about this, from_fn is clearly wrong as it would create an infinite iterator (though we only want the first one so in reality it's ok, but it's still bad), it should be

```suggestion
            .chain(iter::once_with(|| sdk_info.find_tool(tool).map(|path| Some((path, None))).flatten())
```

dpaoliello: But, is that going to eagerly evaluate it?

Also, guess I should add a negative test where we look for a tool that doesn't exist...

NobodyXu: Aha yes...we can use [`iter::once_with`](https://doc.rust-lang.org/stable/std/iter/fn.once_with.html)

Edit:

@dpaoliello updated suggestion

dpaoliello: `once_with` doesn't work, as it requires a function that returns exactly one item whereas we produce zero-or-one. Either we stick with `from_fn` or go back to `next` + `or_else`.

Really wish there was a lazy `chain_with` function.

NobodyXu: We could use `once_with` to return `Option<>` and then use `flatten()` as the [`Option` can be converted to a an iterator](https://doc.rust-lang.org/std/option/enum.Option.html#impl-IntoIterator-for-Option%3CT%3E)

Edit: updated suggestions

dpaoliello: Nice, `once_with` + `flatten` works, and I didn't need to use `map` either.

NobodyXu (approved):
Thank you!

dpaoliello:
If you create a release for `find-msvc-tools`, then I can switch rustc over to using it (without the cc bump)

NobodyXu:
Unfortunately our release pipeline would create release for cc and find-msvc-tools

It is scheduled to run on this Friday, so if you want it befre Friday, I can cut an early release of find-msvc-tools and cc for you

dpaoliello:
Friday is fine: I'm in no rush

NobodyXu:
@dpaoliello new release is cut!

PR #1559, opened 2025-09-19
Fix atomicity violations in concurrent cache operations

quantpoet:
Replace load-then-store patterns with atomic compare_exchange operations to eliminate race conditions. 

The changes ensure thread-safe initialization and prevent duplicate operations in multi-threaded scenarios.


NobodyXu (approved):
Thanks!

PR #1562, opened 2025-09-22
Use `$WASI_SDK_PATH` on WASI targets by default

alexcrichton:
This commit updates compiler detection logic to use the `$WASI_SDK_PATH` variable, if set, by default when looking up a compiler for WASI targets. This is commonly set in installations of wasi-sdk and it can be helpful to avoid the need to further configure how to find a WASI-specific compiler.

CI for WASI was expanded here from `wasm32-wasip1-threads` to other WASI targets as well to ensure it works on all of Rust's targets.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -152,6 +152,8 @@ jobs:
             CXX: clang-cl
     steps:
       - uses: actions/checkout@v5
+      - run: curl -vv https://static.rust-lang.org/dist/channel-rust-1.87.0.toml.sha256
+        shell: bash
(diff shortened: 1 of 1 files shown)


alexcrichton:
I think github actions is having a bad day today, and in my experience it takes ~24 hours for an error like this to go away (maybe a bad cache somewhere? iunno). I'll try to remember to push some time tomorrow to kick CI and will ping here when green.

alexcrichton:
Oh yay, tests look green now!

alexcrichton:
er, sorry, that was the wrong button

PR #1564, opened 2025-09-23
Fix rustcflags mapping: require -Clinker-plugin-lto for -flto

### LINKED ISSUE #1463: Arguably improper -Clto -> -flto flag mapping
cc-rs appears to parse RUSTFLAGS and map -Clto rustc codegen option to clang's -flto=full. It should be noted that on the rustc side the -Clto is required to be complemented with -Cembed-bitcode. This means that rustc's preferred object format is one with embedded bitcode. But on the clang side -flto makes the compiler emit raw bitcode, a non-object file. One can argue that it would be more appropriate to match the rustc's preferred format. But before one rushes to map -Cembed-bitcode to -fembed-bitcode, one should recognize that -flto overrides -fembed-bitcode. In the sense that clang -flto -fembed-bitcode will still generate raw bitcode. So that if one aims to match the formats, one would have to map -Cembed-bitcode alone and let -Clto go unmapped. Alternatively one can wonder if -Cembed-bitcode will be implied with -Clto in the future and become optional. If there is a possibility, then one can make a case for mapping -Clto to -fembed-bitcode instead of -flto.

NobodyXu:
Only when -Clto and -Clinker-plugin-lto are both specified, would we pass -flto, making sure that we won't break environments which has ar doesn't support llvm ir bitcode and linker that doesn't have llvm lto plugin.

Fixed #1463

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -380,7 +380,7 @@ jobs:
       - uses: actions/checkout@v5
       - name: Install Rust
         run: |
-          rustup toolchain install stable --no-self-update --profile minimal --component rustfmt
+          rustup toolchain install stable --no-self-update --profile minimal --component clippy
           rustup default stable
         shell: bash
       - uses: Swatinem/rust-cache@v2


src/flags.rs
```diff
@@ -139,7 +139,7 @@ impl<'this> RustcCodegenFlags<'this> {
             // https://doc.rust-lang.org/rustc/codegen-options/index.html#control-flow-guard
             "-Ccontrol-flow-guard" => self.control_flow_guard = value.or(Some("true")),
             // https://doc.rust-lang.org/rustc/codegen-options/index.html#lto
-            "-Clto" => self.lto = value.or(Some("true")),
+            "-Clinker-plugin-lto" => self.lto = value.or(Some("true")),
```
dot-asm: User can assign a value to the flag in question, and it would be a **file path** to a shared library that implements the plugin. As it stands, it would be misinterpreted when the -flto mapping is attempted elsewhere, where the expected value is "yes"/"on"/etc.

NobodyXu: Thanks, was in the process of updating it, I didn't know it could pass a path to linker plugin.

So it's either used like
 - `-Clinker-plugin-lto`
 - `-Clinker-plugin-lto=/path/to/linker/lto`

but not a yes/no flag?

edit:

i want to double check this because in our tests there's a `-Clinker-plugin-lto=yes`, just want to make sure I understand what it means before changing it

dot-asm: That's what the documentation says. Note that clang has ~to~ **no** use for the path, so the mapping should be -Clinker-plugin-lto[=/what/ever] -> -flto. It's unclear to me what's best, just -flto, -flto=thin or -flto=full. All generate raw bitcode. I'd go with the default.

src/flags.rs
```diff
@@ -324,7 +327,9 @@ impl<'this> RustcCodegenFlags<'this> {
                         _ => None,
                     };
                     if let Some(cc_val) = cc_val {
-                        push_if_supported(format!("-flto={cc_val}").into());
+                        if self.linker_plugin_lto.unwrap_or(false) {
+                            push_if_supported(format!("-flto={cc_val}").into());
+                        }
```
dot-asm: ??? So it's conditional on .lto. I thought we were in agreement that at the very least -Clink-plugin-lto would be an independent mapping.

src/flags.rs
NobodyXu: Updated the code, it would parse both -Clto and -Clinker-plugin-lto, however it would only pass -flto is both are specified, and the value for -flto will be from -Clto

src/flags.rs
NobodyXu: We would need the value from -Clto to defermine the lto mode.

And it's possible for users to pass -Clinker-plugin-lto but not -Clto

dot-asm: > And it's possible for users to pass -Clinker-plugin-lto but not -Clto

No, it's not. If you pass -Clinker-plugin-lto **alone**, then this line is not reached, it's conditional on self.lto being Some(value).

dot-asm:
While you're at it https://github.com/rust-lang/cc-rs/blob/9a2e7ce93df9e30be1a4d0ad43f8665e17946fb6/src/flags.rs#L516 is non-representative of reality.

src/flags.rs
NobodyXu: What I meant is that user can pass `-Clinker-plugin-lto` without `-Clto`, hence that's why it'll only add a `-flto` if both flags are passed

NobodyXu: What I meant is that user can pass `-Clinker-plugin-lto` without `-Clto`, hence that's why it'll only add a `-flto` if both flags are passed

dot-asm:
```
[bar 0.1.0] OUT_DIR = Some(/home/foo/bar/target/release/build/bar-3725da7400c82368/out)
[bar 0.1.0] OPT_LEVEL = Some(3)
[bar 0.1.0] TARGET = Some(x86_64-unknown-linux-gnu)
[bar 0.1.0] HOST = Some(x86_64-unknown-linux-gnu)
[bar 0.1.0] cargo:rerun-if-env-changed=CC_x86_64-unknown-linux-gnu
[bar 0.1.0] CC_x86_64-unknown-linux-gnu = None
[bar 0.1.0] cargo:rerun-if-env-changed=CC_x86_64_unknown_linux_gnu
[bar 0.1.0] CC_x86_64_unknown_linux_gnu = None
[bar 0.1.0] cargo:rerun-if-env-changed=HOST_CC
[bar 0.1.0] HOST_CC = None
[bar 0.1.0] cargo:rerun-if-env-changed=CC
[bar 0.1.0] CC = Some(clang)
[bar 0.1.0] cargo:rerun-if-env-changed=CC_KNOWN_WRAPPER_CUSTOM
[bar 0.1.0] CC_KNOWN_WRAPPER_CUSTOM = None
[bar 0.1.0] RUSTC_WRAPPER = None
[bar 0.1.0] cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
[bar 0.1.0] running: "clang" "-E" "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/17775671041449616416detect_compiler_family.c"
[bar 0.1.0] # 1 "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/17775671041449616416detect_compiler_family.c"
[bar 0.1.0] # 1 "<built-in>" 1
[bar 0.1.0] # 1 "<built-in>" 3
[bar 0.1.0] # 389 "<built-in>" 3
[bar 0.1.0] # 1 "<command line>" 1
[bar 0.1.0] # 1 "<built-in>" 2
[bar 0.1.0] # 1 "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/17775671041449616416detect_compiler_family.c" 2
[bar 0.1.0] 
[bar 0.1.0] #pragma message("clang")
[bar 0.1.0] 
[bar 0.1.0] 
[bar 0.1.0] 
[bar 0.1.0] #pragma message("gcc")
[bar 0.1.0] 
[bar 0.1.0] running: "clang" "-?"
[bar 0.1.0] cargo:warning=clang: error: unknown argument: '-?'
[bar 0.1.0] cargo:warning=clang: error: no input files
[bar 0.1.0] exit status: 1
[bar 0.1.0] running: "clang" "--version"
[bar 0.1.0] exit status: 0
[bar 0.1.0] cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
[bar 0.1.0] CRATE_CC_NO_DEFAULTS = None
[bar 0.1.0] DEBUG = Some(false)
[bar 0.1.0] cargo:rerun-if-env-changed=CFLAGS
[bar 0.1.0] CFLAGS = None
[bar 0.1.0] cargo:rerun-if-env-changed=HOST_CFLAGS
[bar 0.1.0] HOST_CFLAGS = None
[bar 0.1.0] cargo:rerun-if-env-changed=CFLAGS_x86_64_unknown_linux_gnu
[bar 0.1.0] CFLAGS_x86_64_unknown_linux_gnu = None
[bar 0.1.0] cargo:rerun-if-env-changed=CFLAGS_x86_64-unknown-linux-gnu
[bar 0.1.0] CFLAGS_x86_64-unknown-linux-gnu = None
[bar 0.1.0] CARGO_ENCODED_RUSTFLAGS = Some(-Clinker-plugin-lto)
[bar 0.1.0] OUT_DIR = Some(/home/foo/bar/target/release/build/bar-3725da7400c82368/out)
[bar 0.1.0] cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
[bar 0.1.0] running: "clang" "-E" "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/7281727231767827925detect_compiler_family.c"
[bar 0.1.0] # 1 "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/7281727231767827925detect_compiler_family.c"
[bar 0.1.0] # 1 "<built-in>" 1
[bar 0.1.0] # 1 "<built-in>" 3
[bar 0.1.0] # 389 "<built-in>" 3
[bar 0.1.0] # 1 "<command line>" 1
[bar 0.1.0] # 1 "<built-in>" 2
[bar 0.1.0] # 1 "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/7281727231767827925detect_compiler_family.c" 2
[bar 0.1.0] 
[bar 0.1.0] #pragma message("clang")
[bar 0.1.0] 
[bar 0.1.0] 
[bar 0.1.0] 
[bar 0.1.0] #pragma message("gcc")
[bar 0.1.0] 
[bar 0.1.0] running: "clang" "-?"
[bar 0.1.0] cargo:warning=clang: error: unknown argument: '-?'
[bar 0.1.0] cargo:warning=clang: error: no input files
[bar 0.1.0] exit status: 1
[bar 0.1.0] running: "clang" "--version"
[bar 0.1.0] exit status: 0
[bar 0.1.0] cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
[bar 0.1.0] CRATE_CC_NO_DEFAULTS = None
[bar 0.1.0] TARGET = Some(x86_64-unknown-linux-gnu)
[bar 0.1.0] HOST = Some(x86_64-unknown-linux-gnu)
[bar 0.1.0] cargo:rerun-if-env-changed=CFLAGS
[bar 0.1.0] CFLAGS = None
[bar 0.1.0] cargo:rerun-if-env-changed=HOST_CFLAGS
[bar 0.1.0] HOST_CFLAGS = None
[bar 0.1.0] cargo:rerun-if-env-changed=CFLAGS_x86_64_unknown_linux_gnu
[bar 0.1.0] CFLAGS_x86_64_unknown_linux_gnu = None
[bar 0.1.0] cargo:rerun-if-env-changed=CFLAGS_x86_64-unknown-linux-gnu
[bar 0.1.0] CFLAGS_x86_64-unknown-linux-gnu = None
[bar 0.1.0] running: LC_ALL="C" "clang" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "--target=x86_64-unknown-linux-gnu" "-Wall" "-Wextra" "-o" "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/ea708c7824d36062-a.o" "-c" "src/a.c"
[bar 0.1.0] exit status: 0
[bar 0.1.0] running: LC_ALL="C" "clang" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "--target=x86_64-unknown-linux-gnu" "-Wall" "-Wextra" "-o" "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/ea708c7824d36062-b.o" "-c" "src/b.c"
[bar 0.1.0] exit status: 0
[bar 0.1.0] cargo:rerun-if-env-changed=AR_x86_64-unknown-linux-gnu
[bar 0.1.0] AR_x86_64-unknown-linux-gnu = None
[bar 0.1.0] cargo:rerun-if-env-changed=AR_x86_64_unknown_linux_gnu
[bar 0.1.0] AR_x86_64_unknown_linux_gnu = None
[bar 0.1.0] cargo:rerun-if-env-changed=HOST_AR
[bar 0.1.0] HOST_AR = None
[bar 0.1.0] cargo:rerun-if-env-changed=AR
[bar 0.1.0] AR = None
[bar 0.1.0] cargo:rerun-if-env-changed=ARFLAGS
[bar 0.1.0] ARFLAGS = None
[bar 0.1.0] cargo:rerun-if-env-changed=HOST_ARFLAGS
[bar 0.1.0] HOST_ARFLAGS = None
[bar 0.1.0] cargo:rerun-if-env-changed=ARFLAGS_x86_64_unknown_linux_gnu
[bar 0.1.0] ARFLAGS_x86_64_unknown_linux_gnu = None
[bar 0.1.0] cargo:rerun-if-env-changed=ARFLAGS_x86_64-unknown-linux-gnu
[bar 0.1.0] ARFLAGS_x86_64-unknown-linux-gnu = None
[bar 0.1.0] running: ZERO_AR_DATE="1" "ar" "cq" "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/liba.a" "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/ea708c7824d36062-a.o" "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/ea708c7824d36062-b.o"
[bar 0.1.0] exit status: 0
[bar 0.1.0] running: "ar" "s" "/home/foo/bar/target/release/build/bar-3725da7400c82368/out/liba.a"
[bar 0.1.0] exit status: 0
[bar 0.1.0] cargo:rustc-link-lib=static=a
[bar 0.1.0] cargo:rustc-link-search=native=/home/foo/bar/target/release/build/bar-3725da7400c82368/out
warning: bar@0.1.0: clang: error: unknown argument: '-?'
warning: bar@0.1.0: clang: error: no input files
warning: bar@0.1.0: clang: error: unknown argument: '-?'
warning: bar@0.1.0: clang: error: no input files
```

NobodyXu: Seems to be working fine, the `-?` error is from the compiler detection stage, which is expected.

There is no `-flto`, I think the default is thin local lto, so this PR's working as expected?

NobodyXu:
> While you're at it
> 
> https://github.com/rust-lang/cc-rs/blob/9a2e7ce93df9e30be1a4d0ad43f8665e17946fb6/src/flags.rs#L516
> 
> is non-representative of reality.

Thanks, fixed

src/flags.rs
dot-asm: > There is no `-flto`,

Exactly.

> I think the default is thin local lto,

No clang flag -> no lto for C.

> so this PR's working as expected?

The #1463 boils down to if user passes -Clinker-plugin-lto [through RUSTFLAGS], then cc-rs would add -flto to clang command, a.k.a. -Clinker-plugin-lto -> -flto mapping. It was also suggested to drop -Clto mapping.

NobodyXu: > It was also suggested to drop -Clto mapping.

I did drop the mapping, now it's linker-plugin-lto + lto mapping.

We only take the value from -Clto (if it's not present then it's disabled)

I think this would fix compilation for environments and the original issue?

dot-asm: Just align with [the official suggestion](https://doc.rust-lang.org/rustc/linker-plugin-lto.html).

NobodyXu: > Just align with [the official suggestion](https://doc.rust-lang.org/rustc/linker-plugin-lto.html).

Yeah I think the current impl does align with that.

If it looks good to you, I will merge it

dot-asm: > Yeah I think the current impl does align with that.

The official suggestion is that -Clinker-plugin-lto is **sufficient** to perform cross-language LTO. Or in other words users who aim for cross-language LTO are effectively entitled to achieving the goal with -Clinker-plugin-lto **alone**. If a user chooses to throw in -Clto to affect Rust code generation, it should be between them and the Rust compiler. At least for now.

On a related note, it was also speculated in #1463 that it might be appropriate to emit cargo:rustc-link-arg to smooth things out. I reckon that `cargo:rustc-link-arg=-fuse-ld=lld` would be in alignment with the official suggestion. Though Rust 1.90 does it by itself.

NobodyXu: Thanks, I've changed it to use fat lto if linker-plugin-lto is passed but lto does not have value thin.

src/flags.rs
```diff
@@ -140,6 +141,8 @@ impl<'this> RustcCodegenFlags<'this> {
             "-Ccontrol-flow-guard" => self.control_flow_guard = value.or(Some("true")),
             // https://doc.rust-lang.org/rustc/codegen-options/index.html#lto
             "-Clto" => self.lto = value.or(Some("true")),
+            // https://doc.rust-lang.org/beta/rustc/linker-plugin-lto.html
```
dot-asm: /beta appears unnecessary in the URL.

NobodyXu: Thanks fixed

PR #1569, opened 2025-09-25
Fix cross compilation to xtensa-esp32s3-espidf

### LINKED ISSUE #1552: CC not resolve to correct arch for xtensa espidf arch by default even if explicitly build targetted
This actually threw me off and made me go on a wild goose chase. So the packages compiled by setting `--target` to viz xtensa-esp32s3-espidf, compiles rust code for the target arch but not the dependencies built using C via the cc crate.

So, at the linker stage I was facing issues with mismatch arch for rlib.

Managed to workaround this by explicitly setting env var `export CROSS_COMPILE="xtensa-esp32s3-elf"`, but IMO this should be the default behavior, use the target arch's compiler rather than the host's compiler.

To further make this seamless, it should use the resolved binary name from PATH var, and if not found have some sort of autodiscovery (like how Windows MSVC is done), failing which throw some sort of error message that the required compiler was not found, rather than using a wrong compiler 

Ref: https://github.com/mlua-rs/mlua/issues/642

NobodyXu:
Fixed #1552

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2944,7 +2944,13 @@ impl Build {
                             let cc = if target.abi == "llvm" { clang } else { gnu };
                             format!("{prefix}-{cc}").into()
                         }
-                        None => default.into(),
+                        None => {
+                            if raw_target == "xtensa-esp32s3-espidf" {
+                                "xtensa-esp32s3-elf".into()
+                            } else {
+                                default.into()
+                            }
+                        }
                     }
                 } else {
                     default.into()


Vollbrecht:
@NobodyXu thanks for fixing that stuff. A bit of background i may add here. The espressif [provided C toolchain](https://github.com/espressif/crosstool-NG) did go through some naming schema changes. E.g starting with 13.2.0 they introduced a multilib for all xtensa targets. Previously it was needed to specifically call  xtensa-esp32s3-elf xtensa-esp32s2-elf etc.

But since then this should not be needed anymore technically. They still provided wrappers for the "old" naming convention and ship them. They are actually "rust" wrapper to make it extra fun if you have a look [here](https://github.com/espressif/esp-toolchain-bin-wrappers/tree/master).

So with respect to the fix i am not so sure if hard-coding this in the long run is even the correct thing here. I think something with OP's setup was otherwise wonky, e.g explicitly hard coding this stuff now is not a proper fix. 

Still thank you for your work on the `cc` crate. =) 

If you run into problems with the xtensa / riscv espressif targets feel free to ping me or @ivmarkov in the future btw.

NobodyXu:
That'd be great!

i can update https://github.com/rust-lang/cc-rs/blob/main/src/lib.rs#L3459 to have this prefix, and then fallback to the old name if the new name doesn't exist.

cc @Vollbrecht what if the new name of the compiler?

sansmoraxz:
@NobodyXu isn't it better to put the toolchain name in `fn prefix_for_target`?

This approach would make growing the function too clunky. As there are other compiler targets.

PR #1573, opened 2025-09-26
Prefer clang if linker-plugin-lto specified

NobodyXu:
Follow up for #1564

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -220,22 +220,22 @@ jobs:
   test-linker-plugin-lto:
      name: Test linker-plugin-lto
      runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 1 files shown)


NobodyXu:
Merging in since there's no objection.

kamulos:
In the Job:

```yml
  test-linker-plugin-lto:
    ...
         env:
           RUSTCFLAGS: -C linker-plugin-lto
```

Is `RUSTCFLAGS` correct or is it a typo of `RUSTFLAGS`? A quick local test only seems to use the variable `RUSTFLAGS`

NobodyXu:
Thanks, it's indeed a typo, I'll open a PR to fix it

NobodyXu:
Opened #1665

PR #1574, opened 2025-09-26
Improve ESP microcontroller targets

sansmoraxz:
Targets for _ESP32_, _ESP32-S2_, _ESP32-S3_.

**Note:** _ESP32-C6_ is covered by target `riscv32imac`

**Improves:** #1569

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2944,7 +2944,7 @@ impl Build {
                             let cc = if target.abi == "llvm" { clang } else { gnu };
                             format!("{prefix}-{cc}").into()
                         }
-                        None => default.into()
+                        None => default.into(),
                     }
                 } else {
                     default.into()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3589,10 +3589,10 @@ impl Build {
                     "x86_64-unknown-netbsd" => Some("x86_64--netbsd"),
                     "xtensa-esp32-espidf" => Some("xtensa-esp32-elf"),
                     "xtensa-esp32-none-elf" => Some("xtensa-esp32-elf"),
-                    "xtensa-esp32s2-espidf" => Some("xtensa-esp32s2-elf"),
-                    "xtensa-esp32s2-none-elf" => Some("xtensa-esp32s2-elf"),
-                    "xtensa-esp32s3-espidf" => Some("xtensa-esp32s3-elf"),
-                    "xtensa-esp32s3-none-elf" => Some("xtensa-esp32s3-elf"),
+                    "xtensa-esp32s2-espidf" => Some("xtensa-esp32-elf"),
+                    "xtensa-esp32s2-none-elf" => Some("xtensa-esp32-elf"),
+                    "xtensa-esp32s3-espidf" => Some("xtensa-esp32-elf"),
+                    "xtensa-esp32s3-none-elf" => Some("xtensa-esp32-elf"),
                     _ => None,
                 }
                 .map(Cow::Borrowed)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3587,12 +3587,12 @@ impl Build {
                         self.find_working_gnu_prefix(&["x86_64-linux-musl", "musl"])
                     }
                     "x86_64-unknown-netbsd" => Some("x86_64--netbsd"),
-                    "xtensa-esp32-espidf" => Some("xtensa-esp-elf"),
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3593,6 +3587,12 @@ impl Build {
                         self.find_working_gnu_prefix(&["x86_64-linux-musl", "musl"])
                     }
                     "x86_64-unknown-netbsd" => Some("x86_64--netbsd"),
+                    "xtensa-esp32-espidf" => Some("xtensa-esp32-elf"),
```
Vollbrecht: Thanks for working on this. As i described earlier in my comment in https://github.com/rust-lang/cc-rs/pull/1569#issuecomment-3335217457 the esp32/esep32s2/esp32s3 schema was droped from the toolchain and is now only a unified "esp" for all xtensa target. While there exist "wrappers" for backwards compatibility we could just call the non wrapper directly

Vollbrecht: E.g the tools should all be found under `xtensa-esp-elf` for the non-gdb versions. gdb is not part of the shipped toolchain, and is available separately.

Vollbrecht: And this is not a "new" thing, this changes were introduced two years ago with gcc13

sansmoraxz: OK was worried about the backwards compatibility, but I guess it works out, if it's too old

sansmoraxz: Was worried cause I once tried to write a esp32 target into the s3 and espflash refused to do so.

src/lib.rs
```diff
@@ -3589,10 +3589,10 @@ impl Build {
                     "x86_64-unknown-netbsd" => Some("x86_64--netbsd"),
                     "xtensa-esp32-espidf" => Some("xtensa-esp32-elf"),
                     "xtensa-esp32-none-elf" => Some("xtensa-esp32-elf"),
-                    "xtensa-esp32s2-espidf" => Some("xtensa-esp32s2-elf"),
-                    "xtensa-esp32s2-none-elf" => Some("xtensa-esp32s2-elf"),
-                    "xtensa-esp32s3-espidf" => Some("xtensa-esp32s3-elf"),
-                    "xtensa-esp32s3-none-elf" => Some("xtensa-esp32s3-elf"),
+                    "xtensa-esp32s2-espidf" => Some("xtensa-esp32-elf"),
```
Vollbrecht: sorry if that is a bit confusing but its not xtensa-esp32-elf but just xtensa-esp-elf :D and all the wrappers for esp32 (legacy), esp32s2 and esp32s3 points to the xtensa-esp-elf variant. Hope that makes sense

Vollbrecht: you can confirm that by looking in the toolchains folder you have

src/lib.rs
```diff
@@ -2944,13 +2944,7 @@ impl Build {
...
                             format!("{prefix}-{cc}").into()
                         }
-                        None => {
-                            if raw_target == "xtensa-esp32s3-espidf" {
-                                "xtensa-esp32s3-elf".into()
-                            } else {
-                                default.into()
-                            }
-                        }
+                        None => default.into(),
```
NobodyXu, line 2947: We could keep this as a fallback for these targets, by using `self.which(cross_compile_cc, None).is_some()` to check for existence of the new compiler

sansmoraxz: Perhaps. I thought it would be better to group all of these in one place. Also `xtensa-esp32s3-espidf` is covered in below function

src/lib.rs
```diff
@@ -3593,6 +3587,12 @@ impl Build {
                         self.find_working_gnu_prefix(&["x86_64-linux-musl", "musl"])
                     }
                     "x86_64-unknown-netbsd" => Some("x86_64--netbsd"),
+                    "xtensa-esp32-espidf" => Some("xtensa-esp-elf"),
+                    "xtensa-esp32-none-elf" => Some("xtensa-esp-elf"),
+                    "xtensa-esp32s2-espidf" => Some("xtensa-esp-elf"),
+                    "xtensa-esp32s2-none-elf" => Some("xtensa-esp-elf"),
+                    "xtensa-esp32s3-espidf" => Some("xtensa-esp-elf"),
+                    "xtensa-esp32s3-none-elf" => Some("xtensa-esp-elf"),
```
NobodyXu: You can merge all of them into one branch with `|`

NobodyXu (changes requested):

NobodyXu (approved):
Thank you!

PR #1577, opened 2025-09-27
Fix building for Mac Catalyst

madsmtm:
Without this change, we end up passing `aarch64-apple-ios26.0-macabimacabi` to the compiler, which is definitely wrong.

The core change here is to handle this in `from_cargo_environment_variables` as well, the rest is just a cleanup now that `target.env` contains the canonical value (and `target.abi`).

See also the previous https://github.com/rust-lang/cc-rs/pull/1517 and https://github.com/rust-lang/cc-rs/pull/1534.

I really should've checked that this worked in `cc-rs` before landing https://github.com/rust-lang/rust/pull/139451, sorry about that!

### DIFF
--- a/src/target/parser.rs
+++ b/src/target/parser.rs
@@ -1,4 +1,4 @@
-use std::env;
+use std::{env, mem};
 
 use crate::{target::TargetInfo, utilities::OnceLock, Error, ErrorKind};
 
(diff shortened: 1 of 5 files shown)


src/target/parser.rs
```diff
@@ -67,13 +67,19 @@ impl TargetInfoParserInner {
...
         // MSRV, so it may not always be available; In that case, fall back to
         // `""`, which is _probably_ correct for unknown target names.
-        let abi = cargo_env("CARGO_CFG_TARGET_ABI", ft.map(|t| t.abi))
+        let mut abi = cargo_env("CARGO_CFG_TARGET_ABI", ft.map(|t| t.abi))
             .unwrap_or_else(|_| String::default().into_boxed_str());
 
+        // Remove `macabi` and `sim` from `target_abi` (if present), it's been moved to `target_env`.
+        // TODO: Remove once MSRV is bumped to 1.91 and `rustc` removes these from `target_abi`.
+        if matches!(&*abi, "macabi" | "sim") {
+            env = mem::replace(&mut abi, String::default().into_boxed_str());
```
NobodyXu, line 86: do we want to have a debug assert that the previous value of `env` is either an empty string or same as `abi`?

```rust
debug_assert!(matches!(env, "" | "macbi" | "sim"), env);
```

madsmtm: I've added a debug assertion now.

madsmtm:
> Can you run the "Test nightly `rustc` targets and add issue comment if changed" just to make sure there's no regression?

I ran the following locally, and that worked:
```sh
RUSTFLAGS=--cfg=rustc_target_test cargo +nightly test --lib
```

I don't know how to trigger a workflow on a PR?

NobodyXu (approved):
Thank you!

As oong as it can pass all tests, I'm be ok with it

PR #1578, opened 2025-09-29
Add missing C++ ABI linker flag for AIX

uyoyo0:
This PR resolves a linking failure when building C++ code on AIX. For example, when working with a downstream crate like `cxx`, `cargo test` fails during the linking stage with the following error:

```
ld: 0711-317 ERROR: Undefined symbol: __xlcxx_personality_v1
```

The current implementation links against libc++ but we additionally need it to link against libc++abi which is what this patch does.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1632,6 +1632,11 @@ impl Build {
                     ));
                 }
             }
+            // On aix we must also link libc++abi when using libc++
+            if self.get_target()?.os == "aix" {
+                self.cargo_output
+                    .print_metadata(&"cargo:rustc-link-lib=c++abi");
+            }
         }
 
         let cudart = match &self.cudart {


NobodyXu:
Hmm does enabling [`Build::cpp`](https://docs.rs/cc/latest/cc/struct.Build.html#method.cpp) not working?

uyoyo0:
> Hmm does enabling [`Build::cpp`](https://docs.rs/cc/latest/cc/struct.Build.html#method.cpp) not working?

It correctly identifies libc++ but the issue is that it fails to identify and link libc++abi which contains symbols for things such as exception handling.

NobodyXu:
> It correctly identifies libc++ but the issue is that it fails to identify and link libc++abi which contains symbols for things such as exception handling.

That's strange, based on my experience of using c++ with gcc/clang, using the c++ compiler automatically link in stdc++ and other stuff required, including exception handling.

cc @madsmtm wdyt

uyoyo0:
Hi @NobodyXu @madsmtm, just a friendly ping to see if you've had the chance to look further into this.

NobodyXu:
Still not sure if aix needs cxxabi, because usually linking with libstdc++ or libcxx is suffice

https://stackoverflow.com/questions/20357033/usr-lib-x86-64-linux-gnu-libstdc-so-6-version-cxxabi-1-3-8-not-found

Maybe you didn't add cxxabi to LD_LIBRARY_PATH or it's not in the path where the dynamic linkr can find?

src/lib.rs
```diff
@@ -1632,6 +1632,11 @@ impl Build {
                     ));
                 }
             }
+            // On aix we must also link libc++abi when using libc++
+            if self.get_target()?.os == "aix" {
```
madsmtm, line 1636: Nit:

```suggestion
            if target.os == "aix" {
```

PR #1579, opened 2025-09-29
Reorder changelog and remove duplicate Unreleased section

jeremycline:
It seems some automation ran amok and began inserting release notes in the middle of the changelog. Presumably this was caused by the duplicate header part way through the log. This orders the release notes by version again.


NobodyXu (approved):
Thank you!

PR #1582, opened 2025-10-02
Remove shared_flag and static_flag

### LINKED ISSUE #594: shared_flag and static_flag don't work

yshui:
They are meaningless since cc-rs only generates static libraries. Having them does nothing and only causes confusion.

Closes #1444
Closes #594

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -413,8 +413,6 @@ pub struct Build {
     pic: Option<bool>,
     use_plt: Option<bool>,
     static_crt: Option<bool>,
-    shared_flag: Option<bool>,
-    static_flag: Option<bool>,
     warnings_into_errors: bool,
     warnings: Option<bool>,
     extra_warnings: Option<bool>,
@@ -519,8 +517,6 @@ impl Build {
             asm_flags: Vec::new(),
             no_default_flags: false,
             files: Vec::new(),
-            shared_flag: None,
-            static_flag: None,
             cpp: false,
             cpp_link_stdlib: None,
             cpp_link_stdlib_static: false,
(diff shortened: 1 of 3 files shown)


NobodyXu (changes requested):
We can have these functions deprecated, but we cannot just remove them as it would break compatibility and release a 2.0.0 release

yshui:
emmm, can we create a version 2.0 branch then? i can change the PR target.

NobodyXu:
We have absolutely no plan for 2.0, that'll just render too many build script to stick to a 1.0 version so we'd either backport to maintain 1.0 or give them up, which is not realistic

yshui:
huh, and people really are [using](https://github.com/search?q=shared_flag+lang%3Arust+path%3Abuild.rs&type=code&p=1) them even if they do nothing.

how about i also add a `doc(hidden)`?

NobodyXu:
> huh, and people really are [using](https://github.com/search?q=shared_flag+lang%3Arust+path%3Abuild.rs&type=code&p=1) them even if they do nothing.

Thanks this reminds me, some does use it to build shared lib I think?

> how about i also add a `doc(hidden)`?

Hmmm no that's bad, I'd prefer deprecated if we really want to do that

yshui:
> use it to build shared lib I think?

it won't work though. all they would get is `liblibfoo.so.a`.

NobodyXu:
This PR just need a rebase

NobodyXu:
Thanks

Issue #1565, opened 2025-09-24
Support c/c++ library internal LTO


NobodyXu:
In #1564 we disabled LTO if linker-plugin-lto is not passed in to fix compilation on certain environment.

We could reclaim back the performance, by doing a c/c++ library internal LTO, similar to rustc's crate internal LTO.

To find these tools, we might want 'clang -print-resource-dir', or combined with env var lookup LLVM_INSTALL_DIR

> Did some research, for clang it seems to support this
> 
> llvm-link file1.bc file2.bc file3.bc -o merged.bc
> llc merged.bc -filetype=obj -o merged.o
> I didn't find anyway to achieve that for gcc.
> 
> Ok so the [opt command from llvm](https://llvm.org/docs/CommandGuide/opt.html) can actually do the optimization, might worth trying to see if it works. 

We could do

    llvm-link file1.bc file2.bc file3.bc -o merged.bc
    opt -O{level} merged.bc
    llc merged.bc -filetype=obj -o merged.o

 _Originally posted by @NobodyXu in [#1463](https://github.com/rust-lang/cc-rs/issues/1463#issuecomment-3327314742)_

NobodyXu:
Seems that we can use `clang -O3 -c merged.bc` to do the optimization and codegen, that'd makes things much easier

dot-asm:
> Seems that we can use `clang -O3 merged.bc` to do the optimization and codegen, that'd makes things much easier

I don't follow. I thought that the goal was to produce an .a that is IPO-ed, or "thin-lto-ed" in Rust terms. clang -O3 will attempt to link a final application.

NobodyXu:
Thanks I forgot to put `-c` in it, I need to find some time to experiment it on my PC and see if it can do the job

dot-asm:
I would advocate in favour of implementing this as a **feature** and even make it independent of Rust -> clang flag mappings. Indeed, it would more than natural to give crate maintainers a way to perform localized IPO independent of user setting environment variables. Just like it's more natural to configure Rust lto through crate's Cargo.toml.

NobodyXu:
So tested this out and we just need llvm-link:

```bash
clang -O3 -flto -c a.c b.c
llvm-link a.o b.o -o merged.o
clang -O3 -x ir -c merged.o -o final.o
```

We could use [ar_archive_writer](https://docs.rs/ar_archive_writer) instead of llvm-link but it'd be more complicated, plus we will pull in more dependenciesz

[...]

PR #1587, opened 2025-10-06
Fix `get_base_archiver_variant` for clang-cl: use `--print-search-dirs`

NobodyXu:
Fix https://github.com/rust-lang/cc-rs/issues/1565#issuecomment-3369224944

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3326,12 +3326,8 @@ impl Build {
                     let compiler = self.get_base_compiler().ok()?;
                     if compiler.is_like_clang() {
                         name = format!("llvm-{tool}").into();
-                        self.search_programs(
-                            &compiler.path,
-                            &name,
-                            &self.cargo_output,
-                        )
-                        .map(|name| self.cmd(name))
+                        self.search_programs(&compiler.path, &name, &self.cargo_output)
+                            .map(|name| self.cmd(name))
                     } else {
                         None
                     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3365,7 +3365,7 @@ impl Build {
                     let lib = if compiler.family == (ToolFamily::Msvc { clang_cl: true }) {
                         self.search_programs(
                             &mut self.cmd(&compiler.path),
-                            "llvm-lib.exe",
+                            Path::new("llvm-lib.exe"),
                             &self.cargo_output,
                             true,
                         )
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3327,10 +3327,9 @@ impl Build {
                     if compiler.is_like_clang() {
                         name = format!("llvm-{tool}").into();
                         self.search_programs(
-                            &mut self.cmd(&compiler.path),
+                            &compiler.path,
                             &name,
                             &self.cargo_output,
-                            false,
                         )
                         .map(|name| self.cmd(name))
                     } else {
@@ -3364,10 +3363,9 @@ impl Build {
                     let compiler = self.get_base_compiler()?;
                     let lib = if compiler.family == (ToolFamily::Msvc { clang_cl: true }) {
                         self.search_programs(
-                            &mut self.cmd(&compiler.path),
-                            Path::new("llvm-lib.exe"),
+                            &compiler.path,
+                            Path::new("llvm-lib"),
                             &self.cargo_output,
-                            true,
                         )
                         .or_else(|| {
                             // See if there is 'llvm-lib' next to 'clang-cl'
@@ -4141,14 +4139,12 @@ impl Build {
     /// search for |prog| on 'programs' path in '|cc| -print-search-dirs' output
     fn search_programs(
         &self,
-        cc: &mut Command,
+        clang: &Path,
         prog: &Path,
         cargo_output: &CargoOutput,
-        is_clang_cl: bool,
     ) -> Option<PathBuf> {
-        let arg = "--print-search-dirs";
         let search_dirs = run_output(
-            cc.arg(if is_clang_cl { arg } else { &arg[1..] }),
+            self.cmd(clang).arg("--print-search-dirs"),
             // this doesn't concern the compilation so we always want to show warnings.
             cargo_output,
         )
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3361,23 +3362,31 @@ impl Build {
...
-                        // the -print-search-dirs option.
-                        if let Some(mut cmd) = self.which(&compiler.path, None) {
-                            cmd.pop();
-                            cmd.push("llvm-lib.exe");
-                            if let Some(llvm_lib) = self.which(&cmd, None) {
-                                llvm_lib.to_str().unwrap().clone_into(&mut lib);
+                    let lib = if compiler.family == (ToolFamily::Msvc { clang_cl: true }) {
+                        self.search_programs(
+                            &mut self.cmd(&compiler.path),
+                            Path::new("llvm-lib.exe"),
```
dot-asm: "llvm-lib" might be more appropriate. Here and elsewhere. While I for one wouldn't even attempt it, people seem to be talking about using --target=\*-windows-msvc on non-Windows platforms. Problem with explicit .exe is that it would break this probe on non-Windows platforms, while the lack of .exe won't break it on Windows (because `fn which` adds std::env::consts::EXE_EXTENSION).

src/lib.rs
```diff
@@ -4138,9 +4144,11 @@ impl Build {
         cc: &mut Command,
         prog: &Path,
         cargo_output: &CargoOutput,
+        is_clang_cl: bool,
     ) -> Option<PathBuf> {
+        let arg = "--print-search-dirs";
         let search_dirs = run_output(
-            cc.arg("-print-search-dirs"),
+            cc.arg(if is_clang_cl { arg } else { &arg[1..] }),
```
dot-asm: As mentioned elsewhere, double-dash works universally. In other words there is no need to make it conditional on clang-cl. Again, "universally" means "at least since clang 3.8 and gcc 4.6." One can argue that it's not reassuring enough, because  Rust 1.0 predates clang 3.8. Fair enough, I've tested with 3.0 and it works.

src/lib.rs
NobodyXu: Thanks!

src/lib.rs
NobodyXu: Thank you, I've also tested it out and it's perfect

src/lib.rs
```diff
@@ -4135,12 +4139,12 @@ impl Build {
     /// search for |prog| on 'programs' path in '|cc| -print-search-dirs' output
```
dot-asm: I'd harmonize the commentary with the update.

src/lib.rs
```diff
@@ -4135,12 +4139,12 @@ impl Build {
...
     fn search_programs(
         &self,
-        cc: &mut Command,
+        clang: &Path,
```
dot-asm: As mentioned elsewhere, I advocate in favour of using this method even with gcc. In other words I'd keep `cc` as variable name.

src/lib.rs
NobodyXu: THanks updated

src/lib.rs
NobodyXu: Thanks, renamed to `cc`

src/lib.rs
```diff
@@ -3361,23 +3357,30 @@ impl Build {
...
+                        self.search_programs(
+                            &compiler.path,
+                            Path::new("llvm-lib"),
+                            &self.cargo_output,
+                        )
+                        .or_else(|| {
+                            // See if there is 'llvm-lib' next to 'clang-cl'
+                            if let Some(mut cmd) = self.which(&compiler.path, None) {
+                                cmd.pop();
+                                cmd.push("llvm-lib.exe");
```
dot-asm: The original remark was "here and elsewhere." This line is the "elsewhere."

Just in case, again, I personally have no problem with explicit .exe, but others might appreciate the consideration.

NobodyXu: Thanks updated, didn't realize it means this

dot-asm:
On a tangential, yet related note. Related to the remark about possibility of using search_programs even with gcc. The commentary reads "clang driver appears to be forcing UTF-8 output **even on Windows**." This doesn't seem to apply to gcc driver. I reckon it's not fatal, because gcc on Windows people are not that likely to end up with non-ASCII path in the --print-search-dirs output. And if they do a) there normally is a fallback for search_programs; b) in gcc context it would be most critical in cross-compilation case, while on Windows cross-compile toolchain is more likely to be self-contained, so that fallback is virtually guaranteed to work.

dot-asm:
Another related note. The --print-search-dirs output depends on platform flags, such as --target=\* and -m32/-m64. So that formally speaking search_programs should be probed with the said flags. However, one can make a case that omitting them is not fatal, because a) llvm tools are all multi-platform, so that any found on the the search path would do the job; b) in gcc case one would only look up ar, and the associated ar should handle the target.

NobodyXu:
Thank you dot-asm!

so to sum up, if we want to support gcc for search_programs we might want
 - no  utf-8 support
 - passing target flags

dot-asm:
UTF-8 problem is specific to gcc on **Windows**. And it's not given that one can actually solve it, not in most general case. Because gcc driver gets an opportunity to replace non-ASCII character with '?', which would be irreversible. I personally wouldn't bother actually doing something about this. Being aware of the issue and explaining it to affected users should be sufficient. It should be noted that at least MSYS2 installer **refuses** to proceed with installation to a directory with non-ASCII characters. It's probably safe to assume that other installers do the same. Or at least if they don't, one can argue that they should. In other words if somebody manages to install gcc in a directory with non-ASCII characters, they should be advised to reconsider.

As for target flags. In gcc context it's ~about~ exclusively about -m64/-m32. Omitting them when probing shouldn't be a problem, because it's practically guaranteed that accompanying 'ar' can handle either. --target=\* is significant in clang context, but even here omitting it shouldn't be a problem, because accompanying llvm tools can handle either target gracefully. The keywords in this paragraph is "should" as in "shouldn't be a problem."

To summarize it's not much of "might want," but rather be aware and ready to tell users what's wrong with their installations.

NobodyXu:
Got it, thank you for the detailed explanation.

Basically it's something to be put into comment for debugging if gcc is used

PR #1589, opened 2025-10-06
Allow using VCToolsVersion to request a specific msvc version

danielframpton:
Currently, cc-rs always returns the latest version of the msvc tools it can find, including preview versions.

For scenarios (such as we have at Microsoft) where we want to ensure we are building with a specific version of the tools, this change makes it possible to request a specific version of the msvc tools even if a developer has other versions on their machine.

### DIFF
--- a/find-msvc-tools/src/find_tools.rs
+++ b/find-msvc-tools/src/find_tools.rs
@@ -128,6 +128,13 @@ impl EnvGetter for StdEnvGetter {
 /// return `Some(cmd)` which represents a command that's ready to execute the
 /// tool with the appropriate environment variables set.
 ///
+/// To find MSVC tools, this function will first attempt to detect if we are
+/// running in the context of a developer command prompt, and then use the tools
+/// as found in the current `PATH`. If that fails, it will attempt to locate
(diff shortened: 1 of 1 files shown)


find-msvc-tools/src/find_tools.rs
```diff
@@ -843,7 +843,13 @@ mod impl_ {
         ))
     }
 
-    fn vs15plus_vc_read_version(dir: &Path) -> Option<String> {
+    fn vs15plus_vc_read_version(dir: &Path, env_getter: &dyn EnvGetter) -> Option<String> {
+        if let Some(version) = env_getter.get_env("VCToolsVersion") {
+            // Restrict the search to a specific msvc version; if it doesn't exist then
+            // our caller will fail to find the tool for this instance and move on.
+            return version.to_str().map(|s| s.to_string());
```
NobodyXu:
```suggestion
            return version.to_str().map(ToString::to_string)
```

danielframpton:
Thanks @NobodyXu. I added a doc comment and made the change you suggested; please let me know if there was something else you were looking for in terms of comments.

PR #1594, opened 2025-10-13
Fix linker-plugin-lto: use `-flto=thin`

NobodyXu:
The linker-plugin/lld uses thin LTO by default, and it requires the archive to be compiled with thin LTO as well, since thin LTO generates function summary which is needed by thin LTO.

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -324,8 +324,9 @@ impl<'this> RustcCodegenFlags<'this> {
                 // https://doc.rust-lang.org/rustc/linker-plugin-lto.html
                 if self.linker_plugin_lto.unwrap_or(false) {
                     // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-flto
-                    // It has to be thin LTO because llvm linker plugin/lld uses thin LTO by default.
-                    // And for thin LTO to work, the archive also has to be compiled using thin LTO,
+                    // In order to use linker-plugin-lto to achieve cross-lang lto, cc has to use thin LTO
+                    // to compile the c/c++ libraries because llvm linker plugin/lld uses thin LTO by default.
+                    // And for thin LTO in linker plugin to work, the archive also has to be compiled using thin LTO,
                     // since thin LTO generates extra information that fat LTO does not generate that
                     // is required for thin LTO process.
                     push_if_supported("-flto=thin".into());


src/flags.rs
```diff
@@ -322,11 +322,10 @@ impl<'this> RustcCodegenFlags<'this> {
...
                 if self.linker_plugin_lto.unwrap_or(false) {
                     // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-flto
-                    let cc_val = match self.lto {
-                        Some("thin") => "thin",
-                        _ => "full",
-                    };
-                    push_if_supported(format!("-flto={cc_val}").into());
+                    // It has to be thin LTO because llvm linker plugin/lld uses thin LTO by default.
+                    // And for thin LTO to work, the archive also has to be compiled using thin LTO,
+                    // since thin LTO generates function summary that fat LTO does not generate.
```
dot-asm: I'm not sure if this is the correct description. For example, if you compile b.c from https://github.com/rust-lang/cc-rs/issues/1463#issuecomment-3393079597 with `-flto=fat -S` and `-flto=thin -S` and compare outputs, you'll see both have function "summaries." The differences are a) function's `notEligibleToImport` attribute; b) module's `ThinLTO` and `EnableSplitLTOUnit` attributes.

src/flags.rs
```diff
@@ -322,11 +322,10 @@ impl<'this> RustcCodegenFlags<'this> {
...
+                    push_if_supported("-flto=thin".into());
```
dot-asm, line 332: This effectively renders -Clto mapping ineffective. Incidentally that's what I ended up suggesting in #1463 :-) Since we ended up here it would be appropriate to either actually drop the mapping or add commentary that it's ignored.

dot-asm: > actually drop the mapping

Just in case to eliminate the ambiguity, drop or add a remark to the -Clto mapping in line 143.

src/flags.rs
NobodyXu: Thanks fixed comment to just say that thin lto generates additional information required by thin lto

src/flags.rs
NobodyXu: thanks, added a comment saying that it's not used

src/flags.rs
```diff
@@ -322,11 +324,11 @@ impl<'this> RustcCodegenFlags<'this> {
...
                     // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-flto
-                    let cc_val = match self.lto {
-                        Some("thin") => "thin",
-                        _ => "full",
-                    };
-                    push_if_supported(format!("-flto={cc_val}").into());
+                    // It has to be thin LTO because llvm linker plugin/lld uses thin LTO by default.
+                    // And for thin LTO to work, the archive also has to be compiled using thin LTO,
+                    // since thin LTO generates extra information that fat LTO does not generate that
+                    // is required for thin LTO process.
```
dot-asm, line 331: Six "thin LTO" in two sentences. Well, not that I care that much, but it does sting the eye. It also feels that it would look worse in isolation, i.e. not in the diff context. I mean
```
- push_if_supported(format!("-flto={cc_val}").into());
+ // It has to be thin LTO because llvm linker plugin/lld uses thin LTO by default.
```
looks like meaningful "rebuttal," while
```
// https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-flto
// It has to be thin LTO because llvm linker plugin/lld uses thin LTO by default.
```
feels like "wow, where does it come from?"

NobodyXu: Thanks updated

dot-asm: Five "thin LTO"-s, only one down... But what does second sentence add now? It's basically the first sentence... Again, not that I care that much :-) Just do you :-)

NobodyXu: hmm if it can be understood by another maintainer with no context before hand, then I think it's good

thesamesam: FWIW, I think this might become problematic with `rustc_codegen_gcc`, as GCC has `-flto` but not `-flto=thin` (the way GCC does LTO is different). I believe `rustc_codegen_gcc` already has some support for cross-language LTO so it's not just theoretical. cc @antoyo

NobodyXu: -flto is only enabled for clang, it's currently always disabled for gcc

PR #1598, opened 2025-10-22
chore: minor improvement for docs

spuradage:
minor improvement for docs


NobodyXu (approved):
Thanks, unfortunately the current nightly breaks our pipeline so we're waiting for nightly to fix it before merging it.

LI might mark that particular pipeline as flaky if I get time to do it

spuradage:
> Thanks, unfortunately the current nightly breaks our pipeline so we're waiting for nightly to fix it before merging it.
> 
> LI might mark that particular pipeline as flaky if I get time to do it

No matter, thank you.

PR #1615, opened 2025-11-20
Add HelenOS GCC toolchain prefixes

mvolfik:
HelenOS support was recently merged to the compiler, this adds the toolchain prefixes. The toolchains are not available in any standard distribution, but can be built from source as described in HelenOS docs: https://www.helenos.org/index.fcgi/wiki/UsersGuide/CompilingFromSource#a2.Buildasupportedcross-compiler

For reference, the toolchain identifiers are defined at https://github.com/HelenOS/helenos/blob/3a4c6d93a7348ed56f730d1686fe37062c67f586/tools/toolchain.sh#L316


NobodyXu (approved):
Thanks!

PR #1624, opened 2025-12-06
Support more of Cargo's debug levels with Build::debug_str

### LINKED ISSUE #1622: More nuanced debug info choice
Today, `cc` checks whether `DEBUG` is set, and if so directs the underlying compiler to produce debug info. However, `DEBUG` is not just a binary option; it's the ["[value] of the corresponding [variable] for the profile currently being built"](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts), which in turn may be [one of several options](https://doc.rust-lang.org/cargo/reference/profiles.html#debug). I know this won't ever be perfect, but it would be great if e.g. setting `debug = "line-tables-only"` actually translated into passing `-gline-tables-only` when using Clang and `-g1` when using GCC (and some nearby thing on MSVC). Or at least some more constrained upper bound than "full debug info".

jrose-signal:
...at least when using GCC or Clang. (MSVC doesn't seem to have a `-g1` equivalent from a cursory search, and I'm not an MSVC expert. This can always be improved later.)

I modeled this after the handling for optimization levels, making sure not to break the existing `debug` and `get_debug` functionality.

There are two cases where this will behave differently from before (besides the cases that now produce `-g1`):
- `"none"` is now treated as "no debug info" in addition to `"0"` and `"false"` and `""`, for consistency with Cargo
- If the `DEBUG` environment variable is set, but its contents are not valid UTF-8, we now produce "no debug info". Previously that would count as "include debug info" for being non-empty. I don't expect this to break anyone in practice; it certainly won't ever happen from a Cargo build script.

I could not figure out how to test the special case of `-gline-directives-only`, which is supported by Clang and *not* GCC; the test utility's `gnu` mode doesn't have a way to force one or the other. It seems unlikely to be a problem, but if someone can describe the right test incantation I can add it!

Fixes #1622.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -405,7 +405,7 @@ pub struct Build {
     host: Option<Arc<str>>,
     out_dir: Option<Arc<Path>>,
     opt_level: Option<Arc<str>>,
-    debug: Option<bool>,
+    debug: Option<Arc<str>>,
     force_frame_pointer: Option<bool>,
     env: Vec<(Arc<OsStr>, Arc<OsStr>)>,
     compiler: Option<Arc<Path>>,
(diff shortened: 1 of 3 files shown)


NobodyXu (approved):
Thank you!

jrose-signal:
Alas, while I believe I implemented this correctly on the cc-rs side, Cargo doesn't actually do what it says! Filed https://github.com/rust-lang/cargo/issues/16363. (Anyone using the new `debug_str` API directly will get the behavior they ask for, though.)

PR #1629, opened 2025-12-08
Add support for targets like x86_64-unknown-linux-gnu-asan

jakos-sec:
Currently we are not able to create targets that have specific sanitizers enabled by default, since having five components gives:

```
cc-rs: too many components in target x86_64-unknown-linux-gnu-asan
```

This allows currently parsing such targets and simply strips the final part before turning it into a `TargetInfo`.

I'm opening this since I was suggested to use `x86_64-unknown-linux-gnu-asan` as the appropriate name for the target I'm creating (context: https://github.com/rust-lang/rust/pull/149644#discussion_r2598243549).

However I'd love to get some input on how we can solve this in the best way.

Right now I'm basically just throwing away the 5th component. Should we also parse this into the TargetInfo, do we want this sanitizer specific or give this a more generic name?

I'm opening this to hopefully get some input on what the most desirable way to solve this is. Thank you!

### DIFF
--- a/src/target/parser.rs
+++ b/src/target/parser.rs
@@ -337,6 +337,16 @@ impl<'a> TargetInfo<'a> {
                 })?;
                 (*vendor, *os, env, abi)
             }
+            // Five components; format is `arch-vendor-os-env+abi-sanitizer`.
+            [vendor, os, envabi, _sanitizer] => {
+                let (env, abi) = parse_envabi(envabi).ok_or_else(|| {
+                    Error::new(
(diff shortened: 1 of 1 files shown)


madsmtm:
Let's wait for [the MCP](https://github.com/rust-lang/compiler-team/issues/951) to conclude before settling on this one

madsmtm:
Superseded by https://github.com/rust-lang/cc-rs/pull/1661.

PR #1632, opened 2025-12-16
Fix is_flag_supported for assembly-specific flags

### LINKED ISSUE #359: is_flag_supported only works for C sources
`is_flag_supported` is implemented by attempting to compile a C file. This will obviously fail to correctly work when check is done in a non-C context.

For example:

```
    let mut cfg = cc::Build::new();
    cfg.flag_if_supported("-xassembler-with-cpp");
    cfg.file(asm);
    cfg.compile("libanan.a");
```

will fail to correctly identify that the flag is supported, because `-xassembler-with-cpp` will try to compile the C file as assembly, which will obviously fail.

This failure condition is not documented.

For at least gcc and clang the file compiled does not need to have any contents within it at all. For example:

```
echo "" | $CC $TESTED_FLAG -c - -o /tmp/test.o
```

will correctly test the `$TESTED_FLAG`.

naoNao89:
Fixed a 7-year-old bug where `is_flag_supported()` failed with assembly flags like `-xassembler-with-cpp`. 

The issue? It was testing flags against C files. Now it uses stdin compilation for GNU/Clang, avoiding file extension problems entirely.

 Added regression test to prevent this from breaking again.

Fixes #359

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1425,25 +1425,6 @@ impl Build {
         )
     }
 
-    fn ensure_check_file(&self) -> Result<PathBuf, Error> {
-        let out_dir = self.get_out_dir()?;
-        let src = if self.cuda {
-            assert!(self.cpp);
-            out_dir.join("flag_check.cu")
-        } else if self.cpp {
-            out_dir.join("flag_check.cpp")
-        } else {
-            out_dir.join("flag_check.c")
-        };
-
-        if !src.exists() {
-            let mut f = fs::File::create(&src)?;
-            write!(f, "int main(void) {{ return 0; }}")?;
-        }
-
-        Ok(src)
-    }
-
     fn is_flag_supported_inner(
         &self,
         flag: &OsStr,
@@ -1467,7 +1448,6 @@ impl Build {
         }
 
         let out_dir = self.get_out_dir()?;
-        let src = self.ensure_check_file()?;
         let obj = out_dir.join("flag_check");
 
         let mut compiler = {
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix `is_flag_supported` to work with assembly-specific flags by using stdin compilation instead of file-based compilation ([#359](https://github.com/rust-lang/cc-rs/issues/359))
```
NobodyXu: The changelog is auto generated by release-plz, I don't think it'd work with manual edits.

src/lib.rs
```diff
@@ -1539,7 +1521,37 @@ impl Build {
             }
         }
 
-        let output = cmd.current_dir(out_dir).output()?;
+        // Use stdin compilation for all flag tests to avoid issues with
+        // language-specific flags like -xassembler-with-cpp.
+        // See: https://github.com/rust-lang/cc-rs/issues/359
+        let flag_str = flag.to_str().unwrap_or("");
```
NobodyXu: I think using to_string_lossy here is better:

```suggestion
        let flag_str = flag.to_string_lossy();
```

src/lib.rs
```diff
@@ -1425,25 +1425,6 @@ impl Build {
         )
     }
 
-    fn ensure_check_file(&self) -> Result<PathBuf, Error> {
```
NobodyXu, line 1428: It seems that msvc/cl.exe may not support compiling file from stdin? cc @ChrisDenton

naoNao89: i have updated, to use a hybrid approach:

For MSVC: Keep the filebased compilati on (creates `flag_check.c/cpp/cu files`)

For GCC/Clang: Use stdin compilation with `-x <lang>`  ( like `-xassembler-with-cpp`)

NobodyXu: Thanks that'd absolutely work for major use cases, though it'd be great if it's more like a dynamic fallback.

From the top of my head, the nvcc/sccache/ccache maybe a problem.

For nvcc, it seems that

```bash
nvcc -x cu -
```

Is needed.

And sccache does not support it at all, it's possible to use sccache with this crate
 - by specifying an env, or 
 - configuring cargo to use it
 - symlink/hardlink sccache to cc/gcc/clang etc in /usr/local/bin

ChrisDenton: > It seems that msvc/cl.exe may not support compiling file from stdin? cc @ChrisDenton

That is correct. I don't if there's some clever way to trick it into doing so but I doubt it.

NobodyXu: Yes basically we can only do this for clang/gcc, if there's no wrapper applied.

And if it fails, fallback to temp file is required.

PR #1641, opened 2025-12-24
Update Readmes

atouchet:
- Update MIT license URL
- Use consistent line widths


NobodyXu:
Thanks!

I checked the license link, if worked well, is there any reason to change that?

atouchet:
I updated it as I noticed that opensource.org appears to have changed their URLs. The old one still works however so I don't think it makes much difference either way.

NobodyXu (approved):
Thanks

PR #1647, opened 2026-01-04
Have CUDA compilaion check for sbsa-linux when targeting aarch64.

dot-asm:
Originally, CUDA placed aarch64 libraries in a "natural" path, {target_arch}-linux, but then changed it to sbsa-linux.


NobodyXu (approved):
Thank you!

PR #1649, opened 2026-01-05
inhert_rustflags: Pass -Ctarget-cpu as -mcpu

evanj:
This is probably not quite correct, but it "works for me" on aarch64 Linux. I think this should not cause errors, since the flag detection tests if clang/gcc supports the string before it uses it.

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -21,6 +21,7 @@ pub(crate) struct RustcCodegenFlags<'a> {
     dwarf_version: Option<u32>,
     stack_protector: Option<&'a str>,
     linker_plugin_lto: Option<bool>,
+    target_cpu: Option<&'a str>,
 }
 
 impl<'this> RustcCodegenFlags<'this> {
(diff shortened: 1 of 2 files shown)


evanj:
This is probably wrong! In particular: the test is failing on x86 due to the flag detection. I'll probably have to do something gross to fix it.

I'd love feedback on how to test this, and how to improve it. This is fairly important for ensuring the C/C++ code being compiled has the correct CPU features (e.g. CRC32, AES, etc). If this is accepted, I'll also make a similar attempt to support `-Ctarget-feature`, but the target-cpu flag is probably more impactful.

NobodyXu:
Hmmm I think the -mcpu value might be slightly different msvc and clang/gcc at the very least?

cc @madsmtm @ChrisDenton

ChrisDenton:
MSVC's `cl` doesn't have a `mcpu` or equivalent flag.

PR #1663, opened 2026-01-27
Add `find_windows_sdk` API

ChrisDenton:
This essentially exposes `get_sdks` so that it can be used even if Visual Studio is not installed.

My original motivation was so that rustup can detect which parts of a toolchain install are broken (e.g. MSVC tools are installed but the SDK isn't or vice versa). But I think this can be useful in more general cases too.

### DIFF
--- a/find-msvc-tools/src/find_tools.rs
+++ b/find-msvc-tools/src/find_tools.rs
@@ -18,7 +18,7 @@ use std::{
     env,
     ffi::{OsStr, OsString},
     ops::Deref,
-    path::PathBuf,
+    path::{Path, PathBuf},
     process::Command,
     sync::Arc,
 };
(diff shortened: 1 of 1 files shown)


find-msvc-tools/src/find_tools.rs
```diff
@@ -195,6 +195,46 @@ pub fn find_tool_with_env(full_arch: &str, tool: &str, env_getter: &dyn EnvGette
...
+    /// Directories containing source headers (e.g. `.h` files)
+    pub fn include(&self) -> impl Iterator<Item = &Path> {
+        self.include.iter().map(PathBuf::as_path)
+    }
+
+    /// The Windows SDK version that was found.
+    pub fn sdk_version(&self) -> &str {
+        &self.version
+    }
+}
```
NobodyXu, line 227: Thank you, LGTM

PR #1671, opened 2026-02-01
Fix `-I` and double quotes when using `clang-cl`

ArchangelX360:


src/lib.rs
```diff
@@ -1965,8 +1965,13 @@ impl Build {
             cmd.push_cc_arg(format!("-std{separator}{std}").into());
         }
         for directory in self.include_directories.iter() {
-            cmd.args.push("-I".into());
-            cmd.args.push(directory.as_os_str().into());
+            if cmd.is_like_clang_cl() {
+                cmd.args
+                    .push(format!("-Xclang=-I{}", directory.display()).into());
```
madsmtm, line 1991: `clang-cl` supports `/I` (and thereby also `-I`)? At least the user manual says so:
https://clang.llvm.org/docs/UsersManual.html#clang-cl

It'd help if you could add a test or describe the problem in more detail?

ArchangelX360: IIRC, I tried `/I` first but encounter other issues, it was a few months ago so I don't remember them and forgot to take a note 😦, that's why I fell back to use `-Xclang=-I` instead

ArchangelX360: > It'd help if you could add a test or describe the problem in more detail?

I'll see what I can do later this week yep.

ArchangelX360:
I tried to reproduce the issue to give more details, I don't manage to do it.
I recompiled from scratch my monorepo on all three OSes with the content of that PR reverted from my `cc-rs` fork.

Seems like the issue I observed a few months ago was either related to Bazel's `rules_rust` (https://github.com/bazelbuild/rules_rust/pull/3832), or specific crates' incorrect implementation of their `build.rs` that I fixed since that time (https://github.com/tikv/jemallocator/pull/155).
Another hypothesis is that we dropped our indirect way of starting the compilation on Windows in Bazel, we were using a wrapper which introduced some double quote interpretation that needed to be handled.

I apologize deeply for the noise, I'm closing that PR.

PR #1672, opened 2026-02-05
add `-fsanitize=...` flag if sanitizer is enabled in rust

### LINKED ISSUE #1666: Feature: Add `-fsanitize=address` for `x86_64-unknown-linux-gnuasan`
I think it could make sense to have `cc` automatically set `-fsanitize=address`, when the new target `x86_64-unknown-linux-gnuasan` is selected in Rust. This would make enabling ASAN in mixed code bases incredibly easy. And I think this would be what the user intends in every case when selecting the target.

I wanted to open a PR for this, but I am too unfamiliar with the code base to do it properly

kamulos:
resolves #1666 

I tried this mainly using ASAN. It works for me on gnu linux using `clang` and `gcc`. It triggers when the `x86_64-unknown-linux-gnuasan` target is used, and also when `RUSTFLAGS="-Z sanitizer=address"` is set.

I don't know this crate well, so I am unsure about a few things:

1. Where is the correct place in the source code to put this?
2. Hopefully `CARGO_CFG_SANITIZE` is only set on targets, that support a specific sanitizer. Does this mean, there are no additional checks of the target necessary?
3. Especially with MSVC I have no idea if this would work or what would need to be changed
4. Should it be this general (just passing the sanitizer string to the C compiler) or should it be more like a whitelist (for example enabling it only for asan and then extending it in the future)?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2026,6 +2026,12 @@ impl Build {
             self.add_inherited_rustflags(&mut cmd, &target)?;
         }
 
+        if let Some(sanitizer) = self.getenv("CARGO_CFG_SANITIZE") {
+            let mut sanitizer_flag = OsString::from("-fsanitize=");
+            sanitizer_flag.push(sanitizer);
+            cmd.args.push(sanitizer_flag);
+        }
+
         // Set flags configured in the builder (do this second-to-last, to allow these to override
         // everything above).
         for flag in self.flags.iter() {


src/lib.rs
```diff
@@ -2026,6 +2026,12 @@ impl Build {
             self.add_inherited_rustflags(&mut cmd, &target)?;
         }
 
+        if let Some(sanitizer) = self.getenv("CARGO_CFG_SANITIZE") {
+            let mut sanitizer_flag = OsString::from("-fsanitize=");
```
NobodyXu, line 2030: It seems that rust and clang support `hwaddress` sanitizer, but gcc doesn't.

So we probably want to map it to normal `address` right?

cc @madsmtm

NobodyXu (changes requested):

src/lib.rs
```diff
@@ -2026,6 +2026,12 @@ impl Build {
             self.add_inherited_rustflags(&mut cmd, &target)?;
         }
 
+        if let Some(sanitizer) = self.getenv("CARGO_CFG_SANITIZE") {
```
madsmtm, line 2029: The ordering here is a bit significant. I think it should be before `add_inherited_rustflags`?

(Probably also before `envflags("CFLAGS")` above, but `add_inherited_rustflags` isn't, so let's not do that just yet).

src/lib.rs
madsmtm: I agree that we should parse the value here, and only emit the flag on tested configurations

(Which also means I think we should have some sort of test for this in CI).

src/lib.rs
madsmtm: Or maybe place it'd make sense to place in `add_default_flags`?

PR #1675, opened 2026-02-11
Add `CC_PREFER_CLANG_CL_OVER_MSVC` environment variable to override `prefer_clang_cl_over_msvc`

### LINKED ISSUE #1673: `prefer_clang_cl_over_msvc` is only tweakable from crates directly using `cc-rs`
`prefer_clang_cl_over_msvc` is only tweakable from crates directly using `cc-rs`, however, this is a "toolchain-specific" setting which is usually the responsibility of a higher scope of the build.
If I missed something, I apologize in advance and please point me to the right direction.

Use case: I want/can to only use `clang-cl` in my toolchain, some of my crates are depending on a dependency like `zstd-sys` which uses `cc-rs` under the hood to compile some C code. I did not find any way to propagate my desire to prefer `clang-cl` over `MSVC` to the transitive`cc-rs` dependency.

We could in theory ask every single author of crates that use `cc-rs` to expose some parameter in their `build.rs` too which propagates that boolean to their usage of `cc-rs`.
But it seems quite tedious, e.g. my build has 4-5 dependencies using `cc-rs`, it would require a tremendous effort to convince/contribute to all crates that does not allow such propagation (all of them currently 😄).

Maybe something like an env var would work?

As a mitigation, I depend on a fork with that patch applied, but this is of course not ideal for us 😅:
```
Index: src/lib.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- a/src/lib.rs	(revision df67ed3948377535664362bac6ccb9ad6af4ee67)
+++ b/src/lib.rs	(revision 6cf7186175096ee63b34a22debe06cf988a9a437)
@@ -551,7 +551,7 @@
             shell_escaped_flags: None,
             build_cache: Arc::default(),
             inherit_rustflags: true,
-            prefer_clang_cl_over_msvc: false,
+            prefer_clang_cl_over_msvc: true,
         }
     }
 ```

### PR DESCRIPTION
**Pull request by ArchangelX360**
Fixes #1673

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1386,8 +1386,17 @@ impl Build {
     /// Prefer to use clang-cl over msvc.
     ///
     /// This option defaults to `false`.
+    ///
+    /// This option is always overridden by the environment variable `CC_PREFER_CLANG_CL_OVER_MSVC`, if set
     pub fn prefer_clang_cl_over_msvc(&mut self, prefer_clang_cl_over_msvc: bool) -> &mut Build {
-        self.prefer_clang_cl_over_msvc = prefer_clang_cl_over_msvc;
+        match get_prefer_clang_cl_over_msvc_from_env() {
+            Some(b) => {
+                self.cargo_output.print_warning(&format_args!("`prefer_clang_cl_over_msvc` is overridden by {:?} to {:?}, the value set by this call is ignored", CC_PREFER_CLANG_CL_OVER_MSVC_ENV_VAR, b));
+            }
+            None => {
+                self.prefer_clang_cl_over_msvc = prefer_clang_cl_over_msvc;
+            }
+        }
         self
     }
 
@@ -2869,7 +2878,8 @@ impl Build {
         let target = self.get_target()?;
         let raw_target = self.get_raw_target()?;
 
-        let msvc = if self.prefer_clang_cl_over_msvc {
+        let prefer_clang_cl_over_msvc_env = get_prefer_clang_cl_over_msvc_from_env();
+        let msvc = if prefer_clang_cl_over_msvc_env.unwrap_or(self.prefer_clang_cl_over_msvc) {
             "clang-cl.exe"
         } else {
             "cl.exe"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/lib.rs [hunk 1]
```diff
@@ -4475,6 +4484,19 @@ fn check_exe(mut exe: PathBuf) -> Option<PathBuf> {
     check.then_some(exe)
 }
 
+const CC_PREFER_CLANG_CL_OVER_MSVC_ENV_VAR: &str = "CC_PREFER_CLANG_CL_OVER_MSVC";
+fn get_prefer_clang_cl_over_msvc_from_env() -> Option<bool> {
+    #[allow(clippy::disallowed_methods)]
```
ArchangelX360, line 4490: If I understood correclty, `getenv` or `getenv_boolean` won't default to the host environment, so it's not applicable to us here as we actually want to have the host environment leak to us if I'm not mistaken.

However, how can I test that? Because tests are run in parallel, so in my current test implementation it just does not work as tests are leaking that env to one another most probably.

src/lib.rs
```diff
@@ -552,7 +552,7 @@ impl Build {
             shell_escaped_flags: None,
             build_cache: Arc::default(),
             inherit_rustflags: true,
-            prefer_clang_cl_over_msvc: false,
+            prefer_clang_cl_over_msvc: get_prefer_clang_cl_over_msvc_from_env().unwrap_or(false),
```
madsmtm: I think I'd prefer to do things at the usage site instead, something like:
```rust
let prefer_clang_cl_over_msvc_env = self.getenv_boolean("CC_PREFER_CLANG_CL_OVER_MSVC");
let msvc = if prefer_clang_cl_over_msvc_env.unwrap_or(self.prefer_clang_cl_over_msvc) {
    "clang-cl.exe"
} else {
    "cl.exe"
};
```

src/lib.rs
madsmtm: See `tests/cc_env.rs`

ArchangelX360: Thanks I'll implement it that way for now.

Then we can discuss the following here: would it be an acceptable design for you to always propagate `CC_PREFER_CLANG_CL_OVER_MSVC` to `self.env` in the `Build::new` constructor?

There is two advantages to that:
- I don't have to use the `main` function trick in tests and rely on setting `self.env`
- I can use `self.getenv_boolean` in the implementation of `Build`

`CC_FORCE_DISABLE` would be under the same regime too, for example.

madsmtm: I'm not entirely sure I understand the question, but you might wanna wait for https://github.com/rust-lang/cc-rs/pull/1656 before making too many bold changes to how environment variables work?

PR #1700, opened 2026-04-06
fix(ar): suppress warnings from `D` modifier probe

weihanglo:
On macOS, the system `ar` doesn't support the `D` modifier. It causes `illegal option -- D` to be emitted as `cargo:warning=` during the probe.

Add a `run_silent_on_error()` helper that buffers stderr and only forwards it when the command succeeds.

Repro script:

```bash
#!/usr/bin/env bash
# Reproduce the "illegal option -- D" cargo:warning= issue on macOS.
# Requires macOS system ar (which does not support the D modifier).
set -euo pipefail

DIR=$(mktemp -d)
trap 'rm -rf "$DIR"' EXIT

mkdir -p "$DIR/src"

cat > "$DIR/Cargo.toml" <<'TOML'
[package]
name = "cc-repro"
version = "0.0.0"
edition = "2021"

[build-dependencies]
#cc = { path = "/path/local/cc-rs" }
cc = "=1.2.59"
TOML

echo 'fn main() {}' > "$DIR/src/main.rs"
echo 'void foo(void) {}' > "$DIR/foo.c"
echo 'fn main() { cc::Build::new().file("foo.c").compile("foo"); }' > "$DIR/build.rs"

echo "==> Building in $DIR ..."
cd "$DIR"
OUTPUT=$(cargo build 2>&1)

if echo "$OUTPUT" | grep -qi "illegal option"; then
    echo "BUG: found spurious 'illegal option' warning:"
    echo "$OUTPUT" | grep -i "illegal option"
    exit 1
else
    echo "OK: no spurious warnings."
fi
```

Address <https://github.com/rust-lang/cc-rs/pull/1697#issuecomment-4189000198>

### DIFF
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -345,6 +345,39 @@ pub(crate) fn run(cmd: &mut Command, cargo_output: &CargoOutput) -> Result<(), E
     wait_on_child(cmd, &mut child, cargo_output)
 }
 
+/// Like [`run`], but stderr is only forwarded as `cargo:warning=` when the
+/// command succeeds. On failure, stderr is silently discarded.
+///
+/// Useful for probe commands where failure is expected and the error
+/// message is not actionable.
+pub(crate) fn run_silent_on_error(
+    cmd: &mut Command,
+    cargo_output: &CargoOutput,
+) -> Result<(), Error> {
+    let Output {
+        status,
+        stdout: _,
+        stderr,
+    } = spawn_and_wait_for_output(cmd, cargo_output)?;
+
+    cargo_output.print_debug(&status);
+
+    if status.success() {
+        if cargo_output.warnings {
+            stderr
+                .split(|&b| b == b'\n')
+                .filter(|line| !line.is_empty())
+                .for_each(write_warning);
+        }
+        Ok(())
+    } else {
+        Err(Error::new(
+            ErrorKind::ToolExecError,
+            format!("command did not execute successfully (status code {status}): {cmd:?}"),
+        ))
+    }
+}
+
 pub(crate) fn spawn_and_wait_for_output(
     cmd: &mut Command,
     cargo_output: &CargoOutput,
--- a/src/command_helpers.rs
+++ b/src/command_helpers.rs
@@ -366,6 +366,7 @@ pub(crate) fn run_silent_on_error(
         if cargo_output.warnings {
             stderr
                 .split(|&b| b == b'\n')
+                .map(|line| line.strip_suffix(b"\r").unwrap_or(line))
                 .filter(|line| !line.is_empty())
                 .for_each(write_warning);
         }
--- a/src/bin/cc-shim.rs
+++ b/src/bin/cc-shim.rs
@@ -84,6 +84,14 @@ fn main() -> ExitCode {
         }
     }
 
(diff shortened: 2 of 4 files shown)


weihanglo:
cc @jieyouxu

Let me know if there is a better way to handle this, or you want a test (though that would be trick to write)

src/command_helpers.rs
```diff
@@ -345,6 +345,37 @@ pub(crate) fn run(cmd: &mut Command, cargo_output: &CargoOutput) -> Result<(), E
...
+        stderr,
+    } = spawn_and_wait_for_output(cmd, cargo_output)?;
+
+    cargo_output.print_debug(&status);
+
+    if status.success() {
+        stderr
+            .split(|&b| b == b'\n')
+            .filter(|part| !part.is_empty())
+            .for_each(write_warning);
```
NobodyXu: We should only write warning if `CargoOutput` requires warning

NobodyXu (changes requested):
Thanks, just some feedback onn print warning

src/command_helpers.rs
weihanglo: Updated. Thanks for the feedback!

weihanglo: Just realized I pinged the wrong person 😓 
(only remembered the last name Xu, sorry for you both)

src/command_helpers.rs
```diff
@@ -345,6 +345,39 @@ pub(crate) fn run(cmd: &mut Command, cargo_output: &CargoOutput) -> Result<(), E
...
+        stderr,
+    } = spawn_and_wait_for_output(cmd, cargo_output)?;
+
+    cargo_output.print_debug(&status);
+
+    if status.success() {
+        if cargo_output.warnings {
+            stderr
+                .split(|&b| b == b'\n')
+                .filter(|part| !part.is_empty())
```
NobodyXu:
```suggestion
                .lines()
                .map(str::strip)
                .filter(|part| !part.is_empty())
```

You might want to use `lines()` which handles `\r\n` correctly

NobodyXu (changes requested):
Identified a bug in line splitting, missing `\r\n` handling

Regarding testing, you can write a rust binary under `bin/` and use it during testing

src/command_helpers.rs
weihanglo: https://github.com/rust-lang/cc-rs/blob/f4c5ac7a7e7917660d2068ecb5fd015a11ffb3d6/src/command_helpers.rs?plain=1#L372-L375

Other places are still using `\n`. I guess that we might want to have a separate PR. Do you want the new code here done correctly this time?

weihanglo: Also `Vec<u8>` does not have `lines()` btw.

NobodyXu: Yes I'd want it to get it right, for other places we do need a separate pr

weihanglo: Done. And add tests.

I added tests first as a bug repro. The next `fix` commit also updates test assertion, so that the assertion change show the behavior change. This ensures bugs really exist. See <https://epage.github.io/dev/pr-style/#c-test> about this pattern.

NobodyXu (approved):
Thank you, LGTM!

Release scheduled on this Friday

PR #1702, opened 2026-04-10
CI: Hash-pin all actions, apply other suggestions from zizmor

woodruffw:
Hello! Apologies for the cold PR.

I'm opening this in my capacity as one of [uv](https://github.com/astral-sh/uv)'s maintainers; we have a set of downstreams (including `cc`!) that we depend on, and we'd like to ensure their CI/CD processes are as hermetic and secure as possible (within the limits of GitHub's platform).

To that effect, this PR contains a few different commits that aim to make `cc`'s CI more secure. None of these changes fix _vulnerabilities_; they're purely defense-in-depth changes that will make a future [Trivy-style compromise](https://www.wiz.io/blog/trivy-compromised-teampcp-supply-chain-attack) less fruitful for an attacker. 

To summarize:

- I've hash-pinned all of your action dependencies with `pinact run -v`. Dependabot will keep these action references up to date (and will bump the version comments too), so this shouldn't add any additional maintenance burden.
- I've enabled [cooldowns](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-) on your Dependabot ecosystem groups, to reduce the "window of opportunity" for compromise if/when a downstream of `cc`'s is compromised.
- I've disabled `actions/checkout`'s default credential-persistence behavior with `persist-credentials: false`, where possible. You have a few workflows that intentionally push back to the repo (or create issues/PRs), so for those workflows I've ensured that `persist-credentials: true` is explicitly set.
- I've dropped your default `permissions` to `{}`, wherever possible. I've also moved all non-empty permissions into their respective jobs rather than setting them at the entire workspace level.

_Most_ of the above was detected automatically with [zizmor](https://docs.zizmor.sh), which you can [integrate](https://docs.zizmor.sh/integrations/#github-actions) into GitHub Actions if you'd like. I've left that out of this PR however, since not every project wants another thing running in CI. But let me know if you'd like it and I'd be happy to send a follow-up PR!

Last but not least, please let me know if there's any other information I can provide. All of the above was 100% human written and reviewed 🙂

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,9 +6,13 @@ updates:
   directory: "/"
   schedule:
     interval: "daily"
+  cooldown:
+    default-days: 7
 - package-ecosystem: cargo
   directory: "/"
   schedule:
     interval: daily
     time: "08:00"
   open-pull-requests-limit: 10
+  cooldown:
+    default-days: 7
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref || github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
+permissions: {}
+
 jobs:
   test:
     name: Test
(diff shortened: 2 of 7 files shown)


NobodyXu (approved):
Thank you, LGTM!

Shame that there isn't a tool that I can automatically apply to other repos, or that GHA having a safer default

woodruffw:
Thanks @NobodyXu!

> Shame that there isn't a tool that I can automatically apply to other repos, or that GHA having a safer default

FYI you _can_ use [zizmor](https://docs.zizmor.sh) for this, it's how I found almost all of these. I recommend trying that out if you haven't. I'm the maintainer of it, so I can answer any questions you have if any come up 🙂

NobodyXu:
Thanks I'd take a look at it later when I got time

PR #1704, opened 2026-04-17
Add `p3` to the list of known LLVM environments

alexcrichton:
Needed for upcoming support for the `wasm32-wasip3` target.

### DIFF
--- a/src/target/llvm.rs
+++ b/src/target/llvm.rs
@@ -89,8 +89,8 @@ impl TargetInfo<'_> {
         };
         let version = version.unwrap_or("");
         let env = match self.env {
-            "newlib" | "nto70" | "nto71" | "nto71_iosock" | "p1" | "p2" | "relibc" | "sgx"
-            | "uclibc" => "",
+            "newlib" | "nto70" | "nto71" | "nto71_iosock" | "p1" | "p2" | "p3" | "relibc"
+            | "sgx" | "uclibc" => "",
             "sim" => "simulator",
             env => env,
         };


NobodyXu:
Thank you!

Can we have more info on how wasm-wasip3 is planned and probably some comment?

I'm really confused as when I googled germini told me wasi-sdk will provide wasip3 target that is in transitional stages.

alexcrichton:
Happy to yeah, are you thinking something in the code or just documented here in the comments of the PR? The preexisting strings here don't otherwise have documentation so I'm not sure if there's a better place in the code to put some words.

For some background: wasip3 is going to have a `wasm32-wasip3` target. The WASIp3 specifications themselves are not finalized and thus there is no target in existence in this time, hence why Google's a bit confused. The spec is about to be finalized, however, so I'm going through and doing some legwork to prepare various ecosystem integration points to support the target when it exists.

To that extent I surfaced this PR's issue when I was running `./x.py test test/ui --target wasm32-wasip3` in the Rust repository. Without this PR it would pass the target string `wasm32-unknown-wasip3` to Clang where the target string we want to pass is `wasm32-wasip3`. With the "unknown" part in the middle it wasn't able to find headers/libraries for the sysroot, but with this PR the `cc` crate will correctly pass the target string as `wasm32-wasip3`.

Wanted to at least explain here a bit, but I'm also happy to write down words in the code if you can recommend a good place to put them.

NobodyXu:
Thanks! You could put the comment as a link in the code though in case anyone wondered, otherwise it looks good to me!

alexcrichton:
Investigating a bit more I'll need to do some more legwork here I think. Rust's locked verison of `cc` wasn't compatible with testing wasm32-wasip3, but turns out if I just update it to the latest version on crates.io it is compatible (this code I'm adding here never runs). I'll try to bottom out what's going on, but otherwise I'm not actually sure if this is needed.

alexcrichton:
Ok this ended up getting fixed in #1592, so I don't think that the special case is needed here. Apologies for the noise!

PR #1705, opened 2026-04-19
command_helpers.rs: fix `OutputKind::Capture` documentation

DanielEScherzer:
Either both backticks should be inside of the brackets, or neither; go with both to match the other enum variants


NobodyXu (approved):
Thank you!

PR #1713, opened 2026-04-30
Add support for `aarch64-unknown-linux-pauthtest` target

jchlanda:
This PR adds support for `aarch64-unknown-linux-pauthtest`, a target that enables Pointer Authentication Code (PAC) support in Rust on AArch64 ELF based Linux systems. It uses the `aarch64-unknown-linux-pauthtest` LLVM target and a pointer-authentication-enabled sysroot with a custom [musl](https://github.com/access-softek/musl) as a reference libc implementation. Dynamic linking is required, with a dynamic linker acting as the ELF interpreter that can resolve pauth relocations and enforce pointer authentication constraints.

Please note, that while `cc-rs` is able to compile c/c++ sources and link against the target binary successfully this is not sufficient for `aarch64-unknown-linux-pauthtest`. The target has a hard requirement on shared objects and dynamic linker must be used. This seems to be incompatible with the [current state](https://github.com/jchlanda/cc-rs/blob/dd545adfde3725e876257f9113f651e36de14fb4/src/lib.rs#L795) of `cc-rs`.

Please consult a rust-lang PR for the details on the target: https://github.com/rust-lang/rust/pull/155722

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1669,6 +1669,14 @@ impl Build {
                     ));
                 }
             }
+            // Pauthtest needs LLVM's libc++, libc++abi provided by the sysroot.
+            if target.abi == "pauthtest" {
+                let pauthtest_sysroot = self.pauthtest_sysroot()?;
+                self.cargo_output.print_metadata(&format_args!(
+                    "cargo:rustc-flags=-L {}/lib -lc++ -lc++abi",
+                    Path::new(&pauthtest_sysroot).display(),
+                ));
+            }
         }
 
         let cudart = match &self.cudart {
@@ -2280,6 +2288,34 @@ impl Build {
                             format!("--sysroot={}", Path::new(&musl_sysroot).display()).into(),
                         );
                         cmd.push_cc_arg("-pthread".into());
+                    } else if target.abi == "pauthtest" {
+                        let pauthtest_sysroot = self.pauthtest_sysroot()?;
+                        let pauthtest_resource_dir = self.pauthtest_resource_dir()?;
+                        cmd.push_cc_arg(
+                            format!("--sysroot={}", Path::new(&pauthtest_sysroot).display()).into(),
+                        );
+                        cmd.push_cc_arg(
+                            format!(
+                                "-resource-dir={}",
+                                Path::new(&pauthtest_resource_dir).display()
+                            )
+                            .into(),
+                        );
+                        cmd.push_cc_arg("-march=armv8.3-a+pauth".into());
+                        if self.cpp && self.cpp_set_stdlib.is_none() {
+                            cmd.push_cc_arg("-stdlib=libc++".into());
+                            cmd.push_cc_arg(
+                                format!(
+                                    "-I{}/include/c++/v1",
+                                    Path::new(&pauthtest_sysroot).display()
+                                )
+                                .into(),
+                            );
+
+                            cmd.push_cc_arg(
+                                format!("-L{}/lib", Path::new(&pauthtest_sysroot).display()).into(),
+                            );
+                        }
                     }
                     // Pass `--target` with the LLVM target to configure Clang for cross-compiling.
                     //
@@ -2987,7 +3023,7 @@ impl Build {
             // and many modern illumos distributions today ship GCC as "gcc" without
             // also making it available as "cc".
             Cow::Borrowed(Path::new(gnu))
-        } else if self.prefer_clang() {
+        } else if self.prefer_clang() || target.abi == "pauthtest" {
             self.which(Path::new(clang), None)
                 .map(Cow::Owned)
                 .unwrap_or(fallback)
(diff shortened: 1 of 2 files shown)


jchlanda:
Although the parent change ([introduction](https://github.com/rust-lang/rust/pull/155722) of aarch64-unknown-linux-pauthtest) has not landed yet, it seems this should not prevent this PR from being marked as ready for review.

NobodyXu (commented):
Looks good to me, cc @madsmtm since you have also reviewed the corresponding rustc PR, can you please also take a look at this?

jchlanda:
@davidtwco, @folkertdev, @tgross35, @madsmtm FWI this is a follow up to https://github.com/rust-lang/rust/pull/155722

src/lib.rs
```diff
@@ -1598,6 +1598,12 @@ impl Build {
         self.assemble(lib_name, &dst.join(gnu_lib_name), &objects)?;
 
         let target = self.get_target()?;
+        if target.abi == "pauthtest" {
+            self.cargo_output.print_warning(
+                &"cc-rs shoutl not be used with `pauthtest` target: it only builds \
```
```suggestion
                &"cc-rs should not be used with `pauthtest` target: it only builds \
```

NobodyXu (changes requested):
Minor typo

NobodyXu (approved):
Thanks lgtm, just waiting for others to double check

jchlanda:
Hi all (@davidtwco, @folkertdev, @tgross35, @madsmtm, @bjorn3),

The PR [introducing the pauthtest](https://github.com/rust-lang/rust/pull/155722) target has been merged into main. I'd be very grateful if we could move this one forward as well.

Thank you!

jchlanda:
@NobodyXu, would you be able to suggest any reviewers for this PR, or help move it forward? Now that the [library PR](https://github.com/rust-lang/rust/pull/156548) has been earmarked for merging, this will be the last missing piece for pauthtest. Thank you!

src/target/parser.rs
```diff
@@ -225,14 +225,14 @@ fn parse_envabi(last_component: &str) -> Option<(&str, &str)> {
         // Environments
         "msvc" => ("msvc", ""),
         "ohos" => ("ohos", ""),
+        "pauthtest" => ("musl", "pauthtest"),
```
NobodyXu: I believe I've fixed this manually on main, can you please rebase, and remove it if it's indeed duplicate?

src/lib.rs
```diff
@@ -1598,6 +1598,12 @@ impl Build {
         self.assemble(lib_name, &dst.join(gnu_lib_name), &objects)?;
 
         let target = self.get_target()?;
+        if target.abi == "pauthtest" {
+            self.cargo_output.print_warning(
+                &"cc-rs should not be used with `pauthtest` target: it only builds \
+                static libraries, while `pauthtest` requires shared objects.",
+            );
```
NobodyXu, line 1606: This is a bit confusing to me, can you please elaborate?

How would users of cc use this target if it's "should not be used"?

NobodyXu (changes requested):
Hello!

I'm willing to merge it if you rebase and addressed my two comments, given that there's no objection and rust already has it, delaying this PR adds no value, even if I don't have the knowledge to check every detail, merging it is better than not

src/target/parser.rs
jchlanda: Yes, that's a duplicate, removed.

src/lib.rs
jchlanda: `pauthtest` must be linked [dynamically](https://github.com/rust-lang/rust/blob/main/src/doc/rustc/src/platform-support/aarch64-unknown-linux-pauthtest.md):

> Dynamic linking is required, with a dynamic linker acting as the ELF interpreter that can resolve pauth relocations and enforce pointer authentication constraints.

The parser must be extended to recognize the new target; otherwise, it'll crash. I actually didn't realise you've already done it, thank you! For example, during bootstrap:

> error occurred in cc-rs: unknown environment/ABI pauthtest in target aarch64-unknown-linux-pauthtest

I thought I could go one step further and provide the scaffolding for the future, when `cc-rs` starts supporting dynamic linking. As a result, this PR does more than just introduce the new target (also: LLVM triple construction, sysroot discovery, etc.). At the same time, I didn't want to give the false impression that `pauthtest` would work "out of the box", hence the warning.

My intention was to kick off a discussion about [adding support for dynamic linking](https://github.com/rust-lang/cc-rs/blob/main/src/lib.rs#L796) once this is merged.

NobodyXu (approved):
Thank you, LGTM, would merge in this PR

Scheduled release is around this Friday, however I can cut an early release as we also have a nightly/1.95 deprecation fix (by ignoring the deprecation)

src/lib.rs
jchlanda: As a side note, I’ve just read through all the issues mentioning shared objects/dynamic linking in the project, and it seems there is a clear consensus that `cc-rs` won’t support it in the foreseeable future. I still think it makes sense to merge this PR, especially since it warns users about static linking, it'll make extending the support in a downstream fork easier (should anyone attempt it ever).
While not relevant to this PR, it does change expectations on our side going forward.

PR #1726, opened 2026-05-14
Remove the shlex dependency in favor of an in-tree implementation

### LINKED ISSUE #1725: Consider removing the shlex dependency in favor of an in-tree implementation
## Context

`shlex` is used in one place in cc-rs: parsing `*FLAGS` environment variables (`CFLAGS`, `CXXFLAGS`, `ARFLAGS`, `RANLIBFLAGS`) when shell-style splitting is opted into via `Build::shell_escaped_flags` or `CC_SHELL_ESCAPED_FLAGS=1`. The single call site is in `Build::envflags`.

## Reasoning

cc-rs sits in the build-dep closure of nearly every Rust crate that touches native code, so any transitive dep is effectively part of the supply chain for a large portion of the ecosystem.

Of the ~1,300 LOC in `shlex` 1.3.0, cc-rs only exercises the iterator (~120 LOC). The remaining ~700 LOC is the quoting-side API, which cc-rs never calls. Vendoring the small subset cc-rs actually uses would:

- Remove `shlex` from the dep graph of every crate that uses cc as a build-dep.
- Shrink the audit surface to code that lives in this repo.
- Avoid pulling in API surface (the `Quoter`/escape side).

The trade-off is taking on ~120 LOC of vendored parsing code (with attribution preserved).

Would the maintainers be open to this? If so, I can put up a PR.

zerosnacks:
Closes #1725

Note: _this PR was created with assistance of Amp (AI) and carefully manually reviewed and locally differentially tested w/ fuzzed inputs against Shlex 1.3.0_

## What

Replaces the `shlex` crate dependency with a ~100-line vendored word splitter in `src/parse.rs`. `shlex` is used in exactly one place in cc-rs: parsing `*FLAGS` environment variables (`CFLAGS`, `CXXFLAGS`, `ARFLAGS`, `RANLIBFLAGS`) when shell-style splitting is opted into via `Build::shell_escaped_flags` or `CC_SHELL_ESCAPED_FLAGS=1`. The single call site is in `Build::envflags`.

## Why

cc-rs sits in the build-dep closure of nearly every Rust crate that touches native code, so any transitive dep is effectively part of the supply chain for a large portion of the ecosystem.

Of the ~1,300 LOC in `shlex` 1.3.0, cc-rs only exercises the iterator (~120 LOC). The remaining ~700 LOC is the quoting-side API (`Quoter`/escape side), which cc-rs never calls. Vendoring the small subset cc-rs actually uses:

- removes `shlex` from the dep graph of every crate that uses cc as a build-dep,
- shrinks the audit surface to code that lives in this repo,
- avoids pulling in API surface (the quoting side) that is irrelevant to cc-rs.
  
## Commit structure

The PR is split into commits for easier review:

1. **Vendor shlex 1.3.0 iterator path; drop shlex dependency.** The parser state machine in `src/parse.rs` is byte-identical to upstream `bytes::Shlex`, with attribution preserved.
2. **Drop dead `line_no` and `had_error`.** cc-rs never reads them; the iterator stops at exactly the same input positions because the parse helpers already return `None`/`Err(())` on EOF inside quotes or after a trailing backslash.
3. **rustfmt + clippy fixes.** Block-form match arms (rustfmt), `'\n' as u8` → `b'\n'` and `'\\' as u8` → `b'\\'` (`clippy::char_lit_as_u8`), and dropping a no-op `ch as u8` cast (`clippy::unnecessary_cast`). All bit-identical for ASCII inputs; no behavior change.
4. **Em-dash → colon in a test comment.** Trivial nit.
5. **Expand attribution header to upstream's full license notice.** Reproduces upstream's MIT-or-Apache-2.0 dual-license disclaimer verbatim and links to the specific upstream source files we adapted.

## Note

The parser state machine is kept byte-for-byte identical to upstream rather than refactored, so equivalence with the original `shlex::Shlex` iterator is easy to audit. The upstream `SPLIT_TEST_ITEMS` corpus is preserved verbatim as the primary equivalence proof, with additional targeted tests layered on top. Stylistic modernization was deliberately kept out of scope for this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ rust-version = "1.63"
 
 [dependencies]
 jobserver = { version = "0.1.30", default-features = false, optional = true }
-shlex = "1.3.0"
 find-msvc-tools = { version = "0.1.9", path = "find-msvc-tools" }
 
 [target.'cfg(unix)'.dependencies]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -251,10 +251,11 @@ use std::path::{Component, Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::sync::{Arc, RwLock};
 
-use shlex::Shlex;
-
 #[cfg(feature = "parallel")]
 mod parallel;
+
+mod parse;
+use self::parse::Shlex;
 
 mod target;
 use self::target::*;
@@ -1361,7 +1362,8 @@ impl Build {
         self
     }
 
-    /// Configure whether *FLAGS variables are parsed using `shlex`, similarly to `make` and
+    /// Configure whether *FLAGS variables are parsed using shell-style word
+    /// splitting (with quote and backslash handling), similarly to `make` and
     /// `cmake`.
     ///
     /// This option defaults to `false`.
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -6,23 +6,17 @@
 
 /// An iterator that takes an input string and splits it into the words using the same syntax as
 /// the POSIX shell.
+///
+/// If the input ends while inside a quotation or right after an unescaped backslash, the
+/// in-progress token is dropped and iteration ends.
 pub struct Shlex<'a> {
     in_iter: core::slice::Iter<'a, u8>,
-    /// The number of newlines read so far, plus one.
-    pub line_no: usize,
-    /// An input string is erroneous if it ends while inside a quotation or right after an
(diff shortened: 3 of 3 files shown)


xtqqczze:
It seems like it would be better to refactor the [`rust-shlex`](https://github.com/comex/rust-shlex) crate to use features so the the quoting functionality can be disabled.

PR #1732, opened 2026-05-21
apply LDFLAGS env var to compiler commands as well

comicfans:
LDFLAGS is also used in similar environment to pass additional flags to linker, something like 

```
LDFLAGS="--target=x86_64-unknown-linux-gnu -no-canonical-prefixes -lm -fuse-ld=lld -Wl,--build-id=md5 -Wl,--hash-style=gnu -Wl,-z,relro,-z,now -l:libstdc++.a --sysroot=....."
```

this PR makes LDFLAGS  part of envflags so link flags also included

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,6 +70,8 @@
 //! * `CFLAGS` - a series of space separated flags passed to compilers. Note that
 //!   individual flags cannot currently contain spaces, so doing
 //!   something like: `-L=foo\ bar` is not possible.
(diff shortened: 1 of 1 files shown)


NobodyXu (changes requested):
We don't perform linking in cc, it's mostly compile + assembly into an archive

comicfans:
my usecase is that ffmpeg-sys-next using cc-rs to perform both compile and link,  if it's better to handle LDFLAGS out of cc-rs's scope, then I close this

NobodyXu:
Thanks, I think handling there is better outside cc-rs

PR #1742, opened 2026-06-02
Add default compiler prefixes for thumbv7a/thumbv7r/thumbv8r bare-metal targets

cezarbbb:
These five `Thumb-mode` bare-metal Arm targets were recently [promoted to Tier 2](https://github.com/rust-lang/rust/pull/155763) in
rust-lang/rust([MCP](https://github.com/rust-lang/compiler-team/issues/985) is here). Without entries in `get_default_prefix`, the `cc` crate falls back to
the host compiler, which errors on `-mthumb`.


src/lib.rs
```diff
@@ -3715,6 +3715,11 @@ impl Build {
                     "armv7r-none-eabi" => Some("arm-none-eabi"),
                     "armv7r-none-eabihf" => Some("arm-none-eabi"),
                     "armv8r-none-eabihf" => Some("arm-none-eabi"),
+                    "thumbv7a-none-eabi" => Some("arm-none-eabi"),
+                    "thumbv7a-none-eabihf" => Some("arm-none-eabi"),
```
NobodyXu, line 3719: The description of the pr linked seems to match it to a `arm-none-eabihf`, as well as any target ends with `hf`

NobodyXu (changes requested):
Thanks! Just a question

cezarbbb:
Thank you so much for your review! I thought so too before, but the existing entries in this match arm already map all eabihf targets to `arm-none-eabi` — for example `armv7a-none-eabihf`, `armv7r-none-eabihf`, `armv8r-none-eabihf`, `thumbv7em-none-eabihf`, `thumbv8m.main-none-eabihf` all use `arm-none-eabi`. In fact I found that `arm-none-eabihf` is never used as a prefix anywhere in this crate. It seems the bare-metal toolchain ships as a single `arm-none-eabi-gcc` binary and handles hard-float via compiler flags (`-mfloat-abi=hard`) rather than a separate prefix. Therefore, I simply referred to the processing methods of other existing `*-none-eabihf` targets.

NobodyXu:
Thank you!

 I just remember that we use `mfloat-abi` instead of `hf` suffiz

https://github.com/rust-lang/cc-rs/blob/b49589a6a1d207d9d61079ddf599c03ef89b3b55/src/lib.rs#L2384

NobodyXu (approved):

PR #1748, opened 2026-06-05
Remove unnecessary clone

FranciscoTGouveia:
This clone can be safely removed as `args` will not be used later in the function.

CC @nunoplopes


NobodyXu (approved):
Thank you!

PR #1770, opened 2026-06-27
Add kache to the supported compiler wrappers

jleni:
Adds [kache](https://github.com/kunobi-ninja/kache) to the supported compiler wrappers, alongside sccache, cachepot and buildcache.

kache is a `RUSTC_WRAPPER`-first, content-addressed build cache for Rust and C/C++. Cache keys are blake3 hashes of normalized compiler inputs, and hits restore zero-copy: a reflink where the filesystem supports it (APFS, btrfs, XFS), otherwise a hardlink or copy, with identical blobs stored once and shared. It can also sync artifacts across machines via S3. It wraps C/C++ object compiles in addition to rustc, which is what motivates adding it here.

- `known_wrappers`, so `CC="kache cc"` is parsed with the real compiler introspected for its family, instead of needing `CC_KNOWN_WRAPPER_CUSTOM=kache`.
- `VALID_WRAPPERS`, so `RUSTC_WRAPPER=kache` also wraps cc-rs C/C++ compiles with no extra configuration, matching the existing behavior for the other caching wrappers.

Requested in kunobi-ninja/kache#460 (thanks @DaniPopes).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,7 +74,7 @@
 //!   wrapper via `sccache cc`. This compiler must understand the `-c` flag. For
 //!   certain `TARGET`s, it also is assumed to know about other flags (most
 //!   common is `-fPIC`).
-//!   ccache, distcc, sccache, icecc, cachepot and buildcache are supported,
+//!   ccache, distcc, sccache, icecc, cachepot, buildcache and kache are supported,
 //!   for sccache, simply set `CC` to `sccache cc`.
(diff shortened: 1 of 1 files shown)


NobodyXu (approved):
Thank you!

Scheduled release is on next Friday, but if this is urgent we can cut a new one today or tomorrow

jleni:
Thanks @NobodyXu !

not urgent! but very happy to see kache included!

hopefully, more people using AI + worktrees will find kache useful

PR #1782, opened 2026-07-07
Fix: replace deprecated Atomic::fetch_update with try_update

NobodyXu:


NobodyXu:
Superseded by #1783, as our MSRV is way too old for try_update

PR #1808, opened 2026-07-29
chore: update rust-version to 1.64

xtqqczze:
use `core::ffi` types

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -429,7 +429,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
     env:
-      MSRV: 1.63.0
+      MSRV: 1.64.0
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ categories = ["development-tools::build-utils"]
 # The binary target is only used by tests.
 exclude = ["/.github", "tests", "src/bin"]
 edition = "2018"
-rust-version = "1.63"
+rust-version = "1.64.0"
 
 [dependencies]
 jobserver = { version = "0.1.30", default-features = false, optional = true }
--- a/dev-tools/gen-target-info/src/read.rs
+++ b/dev-tools/gen-target-info/src/read.rs
@@ -4,7 +4,7 @@ use crate::{RustcTargetSpecs, TargetSpec};
 
 pub fn get_targets_msrv() -> Vec<u8> {
     let mut cmd = process::Command::new("rustc");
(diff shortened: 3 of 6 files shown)


xtqqczze:
@NobodyXu let's bump MSRV over a sequence of PRs, using new features as they become available

NobodyXu (approved):
thanks, LGTM

xtqqczze:
@NobodyXu We can use Cargo workspace inheritance for `edition` and `rust-version`?

NobodyXu:
> @NobodyXu We can use Cargo workspace inheritance for `edition` and `rust-version`?

Yes, we definitely want it

And for other scripts/ci, I wish we could read from the Cargo.toml, rather than having to scattered around to bump it

xtqqczze:
For CI, GitHub Actions runners have `jq`, so we can use:

```sh
msrv=$(cargo metadata --no-deps --format-version 1 \
      | jq -r '.packages[0].rust_version')
```

xtqqczze:
> For CI, GitHub Actions runners have `jq`, so we can use:
> 
> ```shell
> msrv=$(cargo metadata --no-deps --format-version 1 \
>       | jq -r '.packages[0].rust_version')
> ```

Actually, like other `rust-lang` repositories, we should use [`cargo-hack`](https://github.com/taiki-e/cargo-hack) instead. It has a `--rust-version` option that we can use.

xtqqczze:
We should definitely investigate reading `rust-version` from Cargo.toml for CI, because it's easy to miss an instance when bumping:

https://github.com/rust-lang/cc-rs/blob/228195a71a99dfa247e77aed371680612f96650e/.github/workflows/regenerate-target-info.yml#L35

NobodyXu:
Ahaa oops..yeah should 100% switch to `cargo-hack`

xtqqczze:
> https://github.com/rust-lang/cc-rs/blob/228195a71a99dfa247e77aed371680612f96650e/.github/workflows/regenerate-target-info.yml#L35

It seems like an oversight that we install MSRV and nightly here, but then fail to use them.

NobodyXu:
@xtqqczze it's used within the gen-target-info

https://github.com/rust-lang/cc-rs/blob/228195a71a99dfa247e77aed371680612f96650e/dev-tools/gen-target-info/src/read.rs#L5

This is the reason why I want it to be moved into one place...

xtqqczze:
So why isn't https://github.com/rust-lang/cc-rs/actions/workflows/regenerate-target-info.yml failing?

NobodyXu:
Well it currently generates target data for 1.63 and nightly, and gen-target-info, like other dev-tools doesn't have a msrv.

Basically it's generating data for previous msrv 1.63 instead of new 1.64

xtqqczze:
> Basically it's generating data for previous msrv 1.63 instead of new 1.64

Fixed by https://github.com/rust-lang/cc-rs/pull/1837

PR #1814, opened 2026-07-30
fix `-Cforce-frame-pointers`'s corresponding `CFLAGS`

usamoi:
1. `-Cforce-frame-pointers` should imply `-mno-omit-leaf-frame-pointer` to align `rustc`'s behavior. See also [rustc's command-line argument -> rustc's `FramePointer`](https://github.com/rust-lang/rust/blob/be3d26db984c6f96335faca1f254dc04873cb1c1/compiler/rustc_session/src/options.rs#L1218-L1228).

2. `-Cforce-frame-pointers=no` should not imply `-fomit-frame-pointer`. [The rustc book](https://doc.rust-lang.org/rustc/codegen-options/index.html#force-frame-pointers) says:

> `n`, `no`, `off` or `false`: do not force-enable frame pointers. This does not necessarily mean frame pointers will be removed.

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -244,15 +244,13 @@ impl<'this> RustcCodegenFlags<'this> {
                 }
             }
             // https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fno-omit-frame-pointer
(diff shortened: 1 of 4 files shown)


NobodyXu (approved):
Thank you, LGTM!

PR #1815, opened 2026-07-30
cargo: do not inherit rust-version for dev-tools

xtqqczze:


NobodyXu (changes requested):
Honestly we don't need a msrv for these internal tools - we can always assume rust +stable

Btw merge conflicts need to be resolved

xtqqczze:
right, so let's delete `rust-version` field for dev-tools?

NobodyXu:
> right, so let's delete `rust-version` field for dev-tools?

Yes, just do not inherit rust-version

And it can always use the latest edition

xtqqczze:
```log
  env -u CARGO_REGISTRIES_CRATES_IO_PROTOCOL cargo check --lib -p cc --locked
  shell: /usr/bin/bash -e {0}
  env:
    CARGO_INCREMENTAL: 0
    CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
    CC_ENABLE_DEBUG_OUTPUT: true
    MSRV: 1.64.0
    CACHE_ON_FAILURE: false
error: failed to load manifest for workspace member `/home/runner/work/cc-rs/cc-rs/dev-tools/cc-test`

Caused by:
  failed to parse manifest at `/home/runner/work/cc-rs/cc-rs/dev-tools/cc-test/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2024` edition, and only supports `2015`, `2018`, and `2021` editions.
Error: Process completed with exit code 101.
```

let's just keep edition 2021

dev-tools/gen-target-info/Cargo.toml
```diff
@@ -2,8 +2,6 @@
 name = "gen-target-info"
 version = "0.1.0"
 edition.workspace = true
-rust-version.workspace = true
-publish = false
```
NobodyXu, line 6: Why is it removed? It's not intended to be published

NobodyXu (changes requested):

dev-tools/gen-target-info/Cargo.toml
xtqqczze: bad merge

NobodyXu (approved):
Thanks!

PR #1823, opened 2026-07-31
cc: fix compilation on Rust 1.64

xtqqczze:


tests/cflags.rs
```diff
@@ -1,5 +1,8 @@
 //! This test is in its own module because it modifies the environment and would affect other tests
 //! when run in parallel with them.
+
+#![allow(deprecated)]
```
NobodyXu, line 4: I think we should use expect deprecated to only catch deprecation from our crate's API

```suggestion
#![expect(deprecated, reason = "...")]
```

NobodyXu (changes requested):

tests/cflags.rs
xtqqczze: We can't use `expect` due to MSRV, it's a Rust 1.81 feature.

NobodyXu: Msrv is not applicable to our tests though, we never run msrv checks on tests

xtqqczze: Sorry, I forgot that we only consider `rust-version` to apply to the libraries. I still think it would be worth documenting that somewhere.

xtqqczze: To be honest, I think it’s less confusing to apply `rust-version` consistently throughout the `cc` crate. I've updated the PR.

NobodyXu (approved):
LGTM, given the size of the file allow deprecated probably fine

PR #1825, opened 2026-07-31
cargo: fix for minimal-versions

xtqqczze:


.github/workflows/main.yml
```diff
@@ -437,17 +437,16 @@ jobs:
...
-          rustup default $MSRV
-        shell: bash
-      - name: Create Cargo.lock with minimal version
-        run: cargo +nightly update -Zminimal-versions
-      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
-      - run: env -u CARGO_REGISTRIES_CRATES_IO_PROTOCOL cargo check --lib -p cc --locked
-      - run: env -u CARGO_REGISTRIES_CRATES_IO_PROTOCOL cargo check --lib -p cc --locked --all-features
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
+        with:
+          toolchain: $MSRV
```
NobodyXu: Now that we use cargo-hack, do we still need this?

.github/workflows/main.yml
```diff
@@ -437,17 +437,16 @@ jobs:
...
+      - uses: taiki-e/install-action@6a1bd70eaac3c8bdf093356838d7ee09fda951cf # v2.85.5
+        with:
+          tool: cargo-minimal-versions,cargo-hack
+          fallback: none
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+      - run: env -u CARGO_REGISTRIES_CRATES_IO_PROTOCOL cargo minimal-versions check --workspace --all-targets --ignore-private
+      - run: env -u CARGO_REGISTRIES_CRATES_IO_PROTOCOL cargo minimal-versions check --workspace --all-targets --all-features --ignore-private
```
NobodyXu: Seems we are missing cargo-hack

```suggestion
      - run: |
        env -u CARGO_REGISTRIES_CRATES_IO_PROTOCOL cargo hack --rust-version --feature-powerset --no-private minimal-versions check --workspace --all-targets

```

NobodyXu (changes requested):

.github/workflows/main.yml
xtqqczze: `cargo-minimal-versions` already uses `cargo-hack` internally, so it supports `--feature-powerset` as well.

NobodyXu (approved):
Thank you, LGTM

.github/workflows/main.yml
xtqqczze: Let's keep this rather then using the `--rust-version` option for now, I'd like to investigate the impact on the GitHub Actions cache, as `msrv` job appears to be using a huge amount of cache storage: https://github.com/rust-lang/cc-rs/actions/caches?query=sort%3Asize-desc

NobodyXu: Oh yeah...i forgot that

The rust-cache we use indeed depend on it

Maybe we can have an action defined in `.github/actions` that sets `MSRV` env?

```yaml
name: "Get msrv"

runs:
  using: "composite"
  steps:
    - name: get msrv
      shell: bash
      run: |
        MSRV="$(cargo metadata --no-deps --format-version 1 | jq -r '.packages[0].rust_version')"
        echo "MSRV=${MSRV}" >> "$GITHUB_ENV"

```

I'm thinking of opening a PR for this

xtqqczze: Yeah composite action would be perfect, something like: https://github.com/actions-rust-lang/msrv/blob/main/action.yml

NobodyXu: Opened #1837

PR #1829, opened 2026-08-01
Use lp64d ABI for Managarm riscv64

no92:
We already use cc-rs on Managarm since #1488 is upstream, however this change is needed to support riscv64.


NobodyXu (approved):
Thank you!

PR #1832, opened 2026-08-01
ci: run clippy in more cases

xtqqczze:


dev-tools/gen-target-info/src/main.rs
```diff
@@ -27,8 +27,10 @@ fn generate_target_mapping(f: &mut File, target_specs: &RustcTargetSpecs) -> std
 #[allow(clippy::disallowed_methods)]
```
NobodyXu: Do we still need this if it's not doing anything?

Or we switch to a mod-level allow?

NobodyXu (changes requested):

NobodyXu (approved):
Thank you!

Sorry somehow I missed this PR in the swarm of notifications

PR #1845, opened 2026-08-07
do not emit -mno-omit-leaf-frame-pointer if unsupported

### LINKED ISSUE #1844: New error from v1.4.1 due to `-mno-omit-leaf-frame-pointer`
It seems this flag may be unconditionally getting emitted somewhere while it apparently isn't supported on all compilers. From https://github.com/rust-lang/compiler-builtins/actions/runs/31182294873/job/92878294046?pr=1255#step:16:1605:

```text
04
  cargo:warning=arm-linux-gnueabi-gcc: error: unrecognized command-line option '-mno-omit-leaf-frame-pointer'; did you mean '-fno-omit-frame-pointer'?
```

Presumably part of https://github.com/rust-lang/cc-rs/pull/1814, cc @usamoi

usamoi:
fix https://github.com/rust-lang/cc-rs/issues/1844

A patch version bump may be needed to eliminate the impact.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2264,7 +2264,19 @@ impl Build {
 
         if self.get_force_frame_pointer() {
             let family = cmd.family;
-            family.add_force_frame_pointer(cmd);
+            match family {
+                ToolFamily::Gnu | ToolFamily::Clang { .. } => {
+                    cmd.push_cc_arg("-fno-omit-frame-pointer".into());
+                    let flag = OsString::from("-mno-omit-leaf-frame-pointer");
+                    if self
+                        .is_flag_supported_inner(&flag, cmd, target)
+                        .unwrap_or(false)
+                    {
+                        cmd.push_cc_arg(flag);
+                    }
+                }
+                _ => (),
+            }
         }
 
         if !cmd.is_like_msvc() {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2264,18 +2264,15 @@ impl Build {
 
         if self.get_force_frame_pointer() {
             let family = cmd.family;
-            match family {
-                ToolFamily::Gnu | ToolFamily::Clang { .. } => {
-                    cmd.push_cc_arg("-fno-omit-frame-pointer".into());
-                    let flag = OsString::from("-mno-omit-leaf-frame-pointer");
-                    if self
-                        .is_flag_supported_inner(&flag, cmd, target)
-                        .unwrap_or(false)
-                    {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -2264,7 +2264,19 @@ impl Build {
 
         if self.get_force_frame_pointer() {
             let family = cmd.family;
-            family.add_force_frame_pointer(cmd);
+            match family {
+                ToolFamily::Gnu | ToolFamily::Clang { .. } => {
```
NobodyXu: We could use if-let here

```suggestion
            if let ToolFamily::Gnu | ToolFamily::Clang { .. } = family {
```

src/lib.rs
```diff
@@ -2264,7 +2264,19 @@ impl Build {
...
+                    cmd.push_cc_arg("-fno-omit-frame-pointer".into());
```
NobodyXu: Do we want a clone of the cmd before modifying it?

src/lib.rs
NobodyXu: BTW, can clang-cl also support this?

NobodyXu (changes requested):
Thank you!

Some question regarding whether clang-cl can support this and whether mutating `Tool` inline is ok

src/lib.rs
usamoi: Why do we need a clone of `cmd`?

NobodyXu: I suspect it's used elsewhere, and it unconditionally push the nee flags into the `Tool`, regardless of whether it is supported

NobodyXu: Oops sorry, brain rot

This flag pushed is always supported 

Closing

src/lib.rs
usamoi: I'm not familiar with `clang-cl`. After adding the `/Oy-` option, the generated code still doesn't appear to have a frame pointer (at least for x86_64).

https://godbolt.org/z/7c3eW7dMd

And its documentation says:

```
>   /Oy-                    Disable frame pointer omission (x86 only, default)
>   /Oy                     Enable frame pointer omission (x86 only)
```

Code also shows it's unsupported on x86_64.

https://github.com/llvm/llvm-project/blob/6716e3108af4eba29b28140d17d1f022eb7adb64/clang/lib/Driver/ToolChains/MSVC.cpp#L1158

NobodyXu (approved):
Thanks LGTM

Will cut a new release tomorrow (too late at my Tz)

PR #1852, opened 2026-08-11
ci: replace rustup installation with actions-rust-lang/setup-rust-toolchain

xtqqczze:
There are a few advantages of this action:

- It handles setting `CARGO_REGISTRIES_CRATES_IO_PROTOCOL` appropriately (e.g. enabling the sparse registry).
- It sets a few useful CI defaults such as `CARGO_INCREMENTAL=0`, `RUST_BACKTRACE=short`, and `CARGO_TERM_COLOR=always`.
- It provides a Rust-specific problem matcher, so compiler and `rustfmt` errors are surfaced in the GitHub Actions UI.

https://github.com/actions-rust-lang/setup-rust-toolchain/blob/main/action.yml

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,6 @@ on:
 
 env:
   CARGO_INCREMENTAL: 0
-  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
   CC_ENABLE_DEBUG_OUTPUT: true
 
 concurrency:
@@ -164,12 +163,12 @@ jobs:
           persist-credentials: false
       - run: curl -vv https://static.rust-lang.org/dist/channel-rust-1.87.0.toml.sha256
         shell: bash
-      - name: Install Rust (rustup)
-        run: |
-          set -euxo pipefail
-          rustup toolchain install ${{ matrix.rust }} --no-self-update --profile minimal --target ${{ matrix.target }}
-          rustup default ${{ matrix.rust }}
-        shell: bash
+      - name: Install Rust
+        uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
+        with:
+          cache: false
+          target: ${{ matrix.target }}
+          toolchain: ${{ matrix.rust }}
       - name: Install g++-multilib
         run: |
           set -e
@@ -251,12 +250,12 @@ jobs:
        - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
          with:
            persist-credentials: false
-       - name: Install Rust (rustup)
-         run: |
-           set -euxo pipefail
-           rustup toolchain install stable --no-self-update --profile minimal
-           rustup default stable
-         shell: bash
+
+       - name: Install Rust
+         uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
+         with:
+           cache: false
+           toolchain: stable
 
        - uses: taiki-e/install-action@6a1bd70eaac3c8bdf093356838d7ee09fda951cf # v2.85.5
          with:
@@ -300,13 +299,12 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - name: Install Rust (rustup)
-        run: |
-          set -euxo pipefail
-          rustup toolchain install nightly --no-self-update --profile minimal
-          rustup component add rust-src --toolchain nightly
-          rustup default nightly
-        shell: bash
+      - name: Install Rust
+        uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
+        with:
+          cache: false
+          components: rust-src
+          toolchain: nightly
       - run: cargo update
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
       - run: cargo test -Z build-std=std --no-run --workspace --target ${{ matrix.target }}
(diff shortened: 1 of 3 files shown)


xtqqczze:
[CI / MSRV (ubuntu-latest) (pull_request)](https://github.com/rust-lang/cc-rs/actions/runs/31495651584/job/93792563630?pr=1852) takes a long time: "Updating crates.io index"

NobodyXu:
> [CI / MSRV (ubuntu-latest) (pull_request)](https://github.com/rust-lang/cc-rs/actions/runs/31495651584/job/93792563630?pr=1852) takes a long time: "Updating crates.io index"

Yeah because it still doesn't support sparse index

Iirc it'd have to be 1.68 or later which support it

.github/workflows/main.yml
```diff
@@ -8,8 +8,6 @@ on:
     - main
 
 env:
-  CARGO_INCREMENTAL: 0
```
NobodyXu, line 11: We do want to keep this one because rust-cache doesn't work very well with incremental and incremental just do extra work we don't need and debug builds default to incremental

.github/workflows/main.yml
```diff
@@ -8,8 +8,6 @@ on:
...
-  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
```
NobodyXu, line 12: Iirc on rust 1.68 sparse is supported but not default to, and using sparse would speed it up

ChrisDenton:
What's the advantage of using the github action over using rustup directly?

.github/workflows/main.yml
```diff
@@ -463,20 +463,14 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - name: Get msrv
-        uses: ./.github/actions/get-msrv
-      - name: Install Rust
-        run: |
-          rustup toolchain install $MSRV --no-self-update --profile minimal
-          rustup default $MSRV
```
NobodyXu, line 471: We still need this before rust-cache, because rust-cache uses the rust version during caching

NobodyXu (changes requested):
Thank you, some feedback on env/rust-cache

NobodyXu:
> What's the advantage of using the github action over using rustup directly?

Well...shorter to parse for humans?

.github/workflows/main.yml
xtqqczze: [`setup-rust-toolchain`](https://github.com/actions-rust-lang/setup-rust-toolchain) does set `CARGO_INCREMENTAL=0`, do we want to set it explicitly anyway?

.github/workflows/main.yml
xtqqczze: The action has some logic to set `CARGO_REGISTRIES_CRATES_IO_PROTOCOL` if appropriate:

https://github.com/actions-rust-lang/setup-rust-toolchain/blob/8439c15d249cd2e2a5c80ad3da200c3d8c553ae4/action.yml#L240

xtqqczze:
> [CI / MSRV (ubuntu-latest) (pull_request)](https://github.com/rust-lang/cc-rs/actions/runs/31495651584/job/93792563630?pr=1852) takes a long time: "Updating crates.io index"

I rebased and run again, it's much faster when there is a cache hit.

.github/workflows/main.yml
xtqqczze: That would mean we can't use `--rust-version`, because `cargo-hack` doesn't necessarily resolve the version to the exact toolchain specified in `rust-version`.

For example, if `rust-version` is `1.66.0`, `cargo-hack` will attempt to install/use `1.66`, which is not necessarily the same toolchain as `1.66.0` (or `1.66.1`).

I guess that’s fine, though.

.github/workflows/main.yml
NobodyXu: Yeah let's keep it explicit

.github/workflows/main.yml
NobodyXu: Well but it doesn't seem yo set sparse protocol for rust supporting it?

.github/workflows/main.yml
NobodyXu: Yeah that's a limitation of rust-cache or any caching system, because rust has no abi or target dir stability

xtqqczze:
> What's the advantage of using the github action over using rustup directly?

There are a few advantages:

- It handles setting `CARGO_REGISTRIES_CRATES_IO_PROTOCOL` appropriately (e.g. enabling the sparse registry).
- It sets a few useful CI defaults such as `CARGO_INCREMENTAL=0`, `RUST_BACKTRACE=short`, and `CARGO_TERM_COLOR=always`.
- It provides a Rust-specific problem matcher, so compiler and `rustfmt` errors are surfaced in the GitHub Actions UI.

https://github.com/actions-rust-lang/setup-rust-toolchain/blob/main/action.yml

.github/workflows/main.yml
xtqqczze: https://github.com/actions-rust-lang/setup-rust-toolchain/blob/8439c15d249cd2e2a5c80ad3da200c3d8c553ae4/action.yml#L150-L152

.github/workflows/main.yml
```diff
@@ -466,17 +467,18 @@ jobs:
       - name: Get msrv
         uses: ./.github/actions/get-msrv
       - name: Install Rust
-        run: |
-          rustup toolchain install $MSRV --no-self-update --profile minimal
-          rustup default $MSRV
+        uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
+        with:
+          cache: false
+          toolchain: ${{ steps.get-msrv.outputs.msrv }}
```
NobodyXu: It sete the env MSRV but no output, I suspect it probably isn't working?

NobodyXu (changes requested):

.github/workflows/main.yml
xtqqczze: Right, it's installing `stable`?

NobodyXu: Yeah it's weird it's failing silently...

NobodyXu (approved):
Thank you! LGTM

PR #1855, opened 2026-08-13
cc-test: add loongarch64 assembly shim

tangaac:
## Problem

`dev-tools/cc-test`'s `build.rs` compiles `src/<arch>.S` based on the target architecture. `aarch64.S`, `armv7.S`, `i686.S`, `riscv64gc.S`, and `x86_64.S` are all present, but there is no `loongarch64.S`. On `loongarch64-unknown-linux-gnu` (stable Rust target since 1.75) this fails with:

```
cc1: fatal error: src/loongarch64.S: No such file or directory
```

The `cc` crate itself already declares support for LoongArch64 (`loongarch64-unknown-linux-gnu => loongarch64-linux-gnu` in `src/lib.rs`), so this is purely a missing dev-tool shim.

## Fix

Add `dev-tools/cc-test/src/loongarch64.S`, modeled on `riscv64gc.S`, exporting both `asm` and `_asm` symbols. Verified on real LoongArch64 hardware: `cargo build` succeeds and `cargo test` passes 9/9 — the `asm_here` test asserts the shim returns 7.

### DIFF
--- a/dev-tools/cc-test/src/loongarch64.S
+++ b/dev-tools/cc-test/src/loongarch64.S
@@ -1,0 +1,9 @@
+.globl asm
+asm:
+    li.w $a0, 7
+    jirl $r0, $r1, 0
+
+.globl _asm
+_asm:
+    li.w $a0, 7
+    jirl $r0, $r1, 0


NobodyXu (commented):
Thank you!

Would you like to add a CI job to compile the test on loongarch?

It's ok if you don't want to and I can get this merged

tangaac:
I dropped the `check-loongarch64` CI job — cross-linking the test suite on an x86_64 runner hits an unrelated problem. `build.rs` has a `cfg(target_os = "linux")` block (for the static-libstdc++ link test) that runs the *host* `g++ --print-file-name=libstdc++.a` and symlinks the host `libstdc++.a` into OUT_DIR. On an x86_64 runner that is an x86_64 archive, and rustc embeds its objects (e.g. `del_ops.o`) into the test rlib, so the loongarch64 linker fails with `Relocations in generic ELF (EM: 62)`. This only surfaces on a linux cross target — the existing cross jobs are Apple/Windows, so this path was never exercised.

So this PR is just the shim for now. It is verified on a real LoongArch64 machine: the assembly compiles and `asm_here` passes.

NobodyXu:
Thanks, merged

Issue #1859, opened 2026-08-15
`is_flag_supported` ignores `Build::env` overrides; `gnu_debug_fp`/`gnu_debug_fp_auto` fail on Windows hosts without a system `cc` (since #1845)


MohammedAlkindi:
**Environment**: Windows 11, rustup `stable-x86_64-pc-windows-gnu`, mingw-w64 gcc 16.2 on `PATH` as `gcc.exe` — no `cc.exe` (`where cc` finds nothing), no MSVC installed. I'd expect this to be a common shape for Windows contributors using rustup plus a mingw toolchain.

**Symptom**: on a fresh clone of `main`, `cargo test --test test` fails:

```
thread 'gnu_debug_fp' panicked at tests\test.rs:119:17:
didn't find "-mno-omit-leaf-frame-pointer" in ["-O2", "-ffunction-sections", "-fdata-sections",
"-fPIC", "-g", "-gdwarf-4", "-fno-omit-frame-pointer", "-m64", "-o",
"...\\db3b6bfb95261072-foo.o", "-c", "foo.c"]
```

`gnu_debug_fp_auto` fails the same way. This started when #1845 made `-mno-omit-leaf-frame-pointer` conditional on `is_flag_supported`. CI stays green, but for a reason that I think is itself a bug — see below.

**Mechanism** (traced with `CC_ENABLE_DEBUG_OUTPUT=1`):

The test framework only exposes the shim directory through `Build::env`:

```rust
// tests/support/mod.rs
.env("PATH", self.path())   // shim dir prepended
```

`Build::env` overrides are applied to the `Tool`'s spawn environment at the end of `try_get_compiler` ("Do this last, to allow overwriting the other values above"), and the main compile honors them. But `is_flag_supported_inner` (src/lib.rs:1488) constructs a fresh `Build` for the probe and copies `compiler` / `opt_level` / `debug` / `cpp` / `cuda` / … — everything except `self.env`. So the probe resolves and spawns `cc` against the real process environment:

```
running: "cc" "-E" "...\detect_compiler_family.c"
cargo:warning=Compiler family detection failed due to error: ToolNotFound: failed to find tool "cc": program not found
running: "cc" "-O2" ... "-g" "-gdwarf-4" "-fno-omit-frame-pointer" "-m64" ... "-c" "foo.c"   <-- flag missing
exit code: 0
```

Consequences:

[...]

PR #1866, opened 2026-08-20
fix: honor Build::env overrides in flag support probes

MohammedAlkindi:
Implements the design settled in #1859.

`is_flag_supported_inner` builds its probe `Build` without copying `self.env`, so `Build::env` overrides are ignored by flag probes: the probe resolves a bare `cc` through the ambient `PATH` and answers about a different compiler than the one being built with. Since #1845 made `-mno-omit-leaf-frame-pointer` conditional on `is_flag_supported`, that silently changes flags. Family detection in `src/tool.rs` had the same gap at four more spawns, which applied no environment at all.

Copying `self.env` alone shifts the shim's `out{N}` indices, so this adds `CC_SHIM_OUT_FILES_FOR_FAMILY_DETECTION` and `CC_SHIM_OUT_FILES_FOR_FLAG_SUPPORT_CHECK`. cc renames the class-appropriate one to `CC_SHIM_OUT_FILES` before spawning that probe; when it is absent the probe records nothing, per your call in #1859. `cached_compiler_family` is now keyed on the probe environment too, so two builds differing only in `Build::env` cannot share a family answer.

Baseline `caf7609` vs this branch, `cargo nextest run --no-fail-fast --workspace`, x86_64-pc-windows-msvc:

|  | baseline | branch |
|---|---|---|
| total | 115 run, 112 passed, 3 failed, 2 skipped | 116 run, 115 passed, 1 failed, 2 skipped |
| `gnu_debug_fp` | FAIL | PASS |
| `gnu_debug_fp_auto` | FAIL | PASS |
| `clang_android` | FAIL | FAIL (unchanged) |
| `probe_env_overrides` | n/a | PASS (new) |

Doc tests 28 to 28, no regressions. `clippy --no-deps --workspace --all-targets -- -D warnings` clean, `fmt --check` clean, and `cargo check -p cc` passes on 1.65.0.

Three things worth your attention:

Once detection runs under `Build::env` it reaches the shim, which emitted no family macros but exited 0 on `-?`, so every `Test::gnu()` compiler classified as MSVC. Previously detection failed only because the shim panicked on a missing `CC_SHIM_OUT_DIR`. The shim now declines `-E` explicitly rather than learning to preprocess. Family answers match baseline for every shim name in the suite.

`clang_android` is the same root cause at a third site — `src/lib.rs:3723` probes `llvm-ar` with a bare `Command::new`, bypassing `Build::env`'s `PATH`. I left it out: fixing it needs a third recording class, which expands a design you settled at two, and it needs its own test since your CI runners ship a real `llvm-ar`. Happy to send it separately.

Everything above is Windows-only. I audited the `cfg(not(windows))` tests for dependence on a real compiler and converted the one that needed a probe rejection, but your CI settles Unix.

#851 and #1632 touch the same region of `src/lib.rs`; neither implements this design and this does not subsume them.

### DIFF
--- a/src/bin/cc-shim.rs
+++ b/src/bin/cc-shim.rs
@@ -59,7 +59,7 @@ fn out_file(program: &str) -> Option<PathBuf> {
 }
 
 /// Record the args passed to the command, if this invocation records at all.
-fn record(program: &str, args: &[&String]) {
+fn record(program: &str, args: &[String]) {
     let Some(candidate) = out_file(program) else {
         return;
     };
@@ -96,12 +96,10 @@ fn record(program: &str, args: &[&String]) {
 }
 
 fn main() -> ExitCode {
-    let args = env::args().collect::<Vec<_>>();
-    let mut args = args.iter();
-    let program = args.next().expect("Unexpected empty args");
-    let args = args.collect::<Vec<_>>();
+    let argv = env::args().collect::<Vec<_>>();
+    let (program, args) = argv.split_first().expect("Unexpected empty args");
 
-    record(program, &args);
+    record(program, args);
 
     // Compiler family detection preprocesses a probe file with `-E` and reads
     // the family out of the expansion. The shim is not a preprocessor and
--- a/src/bin/cc-shim.rs
+++ b/src/bin/cc-shim.rs
@@ -10,30 +10,43 @@ use std::io::{self, prelude::*};
 use std::path::PathBuf;
 use std::process::ExitCode;
 
-fn main() -> ExitCode {
-    let args = env::args().collect::<Vec<_>>();
-    let mut args = args.iter();
-    let program = args.next().expect("Unexpected empty args");
+/// Test-only environment variable naming a directory to record invocations in.
+///
+/// Arguments are written to the first `out{i}` in it that does not exist yet, so
+/// a test reads the `i`-th invocation back with `Test::cmd(i)`. `Test::gcc()`
+/// sets this through `Build::env`, and the compile and archive commands that a
+/// build actually runs inherit it.
+const OUT_DIR: &str = "CC_SHIM_OUT_DIR";
 
-    let out_dir = env::var_os("CC_SHIM_OUT_DIR")
-        .map(PathBuf::from)
-        .or_else(|| {
-            // Flag probes use a fresh `Build`, so they don't inherit the
-            // test-only `CC_SHIM_OUT_DIR`. `is_flag_supported_inner` runs
-            // them with `out_dir` as their working directory, so we can
-            // record their arguments there.
-            //
-            // We only use this fallback for `-c` the compile-only flag probe
-            // in this flow. Family detection uses `-E` without `-c` and
-            // it must keep failing.
-            args.clone()
-                .any(|arg| arg == "-c")
-                .then(|| env::current_dir().expect("failed to get probe working directory"))
-        })
-        .unwrap_or_else(|| panic!("{}: CC_SHIM_OUT_DIR not found", program));
+/// Test-only environment variable naming an explicit `PATH`-separated list of
+/// files to record invocations in, the first that does not exist yet winning.
+///
+/// cc sets this on its own probing invocations by renaming the dedicated
+/// `CC_SHIM_OUT_FILES_FOR_*` variable for the class of probe it is about to
+/// spawn, and clears [`OUT_DIR`] for them at the same time. Which probes run at
+/// all depends on the compiler, the target and cc's internal caching, so a test
+/// opts one class in by name; every probe it did not ask about records nothing
+/// and cannot shift the `out{i}` numbering of the invocations it is asserting
+/// on. See `set_probe_env` in `src/command_helpers.rs`.
+const OUT_FILES: &str = "CC_SHIM_OUT_FILES";
 
+/// Pick the file this invocation should record its arguments in, if any.
+fn out_file(program: &str) -> Option<PathBuf> {
+    if let Some(files) = env::var_os(OUT_FILES) {
+        let candidate = env::split_paths(&files)
+            .filter(|file| !file.as_os_str().is_empty())
+            .find(|file| !file.exists());
+        return Some(candidate.unwrap_or_else(|| {
+            panic!(
+                "{}: every file named by {} has already been written to: {:?}",
+                program, OUT_FILES, files
+            )
+        }));
+    }
+
+    let out_dir = PathBuf::from(env::var_os(OUT_DIR)?);
     // Find the first nonexistent candidate file to which the program's args can be written.
-    let candidate = (0..).find_map(|i| {
+    Some((0..).find_map(|i| {
         let candidate = out_dir.join(format!("out{i}"));
 
         if candidate.exists() {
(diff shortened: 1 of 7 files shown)


src/bin/cc-shim.rs
```diff
@@ -74,13 +93,36 @@ fn main() -> ExitCode {
             e
         )
     });
+}
+
+fn main() -> ExitCode {
+    let args = env::args().collect::<Vec<_>>();
```
NobodyXu: Nit: I think we don't need this `.collect()` here?

src/lib.rs
```diff
@@ -1520,9 +1520,19 @@ impl Build {
...
                 .inherit_trim_paths(false)
                 .emit_rerun_if_env_changed(self.emit_rerun_if_env_changed);
+            // The probe has to see the environment the compiler is invoked in,
+            // or it answers a question about a different compiler than the one
+            // being built with: a bare compiler name resolves through this
+            // `PATH`, not the ambient one. Also share the caches, so that a
+            // compiler family already worked out is not worked out again.
+            for (key, value) in self.env.iter() {
+                cfg.env(key, value);
+            }
```
NobodyXu: Nit: I think we can just clone it:

```suggestion
            cfg.env.clone_from(&self.env);
```

src/command_helpers.rs
```diff
@@ -340,6 +340,70 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
...
+/// distinguishes: they are set through `Build::env` but read by cc itself, and
+/// rewritten for exactly one child. cc renames the one matching `kind` to
+/// `CC_SHIM_OUT_FILES` and otherwise clears it, and clears `CC_SHIM_OUT_DIR`
+/// either way, instead of copying `Build::env` over blindly. A probe a test did
+/// not ask about then records nothing at all, rather than taking an `out{i}`
+/// slot and shifting the invocations the test is asserting on. They are
+/// test-only, like `CC_SHIM_OUT_DIR`, and so are documented in
+/// `src/bin/cc-shim.rs` rather than in the table of public variables in
+/// `src/lib.rs`.
+pub(crate) fn set_probe_env<K, V>(cmd: &mut Command, env: &[(K, V)], kind: ProbeKind)
```
NobodyXu: Given how we use it in the `tool` mod, mayeb we can make this an ext method?

```rust
trait CommandExt {
     fn set_family_detection_env<K, V>(
&mut self, env: &[(K, V)]) -> Self

     fn set_flag_supported_env<K, V>(
&mut self, env: &[(K, V)]) -> Self
}
```


Then we also don't have to expose `ProbeKind` or `set_probe_env`

tests/test.rs
```diff
@@ -396,23 +396,54 @@ fn gnu_shared() {
...
+/// and record only the class of probe a test asks for by name.
+/// <https://github.com/rust-lang/cc-rs/issues/1859>
+#[test]
+fn probe_env_overrides() {
+    let test = Test::gnu();
+    test.gcc()
+        .env(
+            "CC_SHIM_OUT_FILES_FOR_FAMILY_DETECTION",
+            test.probe_out_files(&["family-probe"]),
+        )
```
NobodyXu: Shall we have a dedicated function for this?

I.e.

```rust
impl Test {
    pub fn collect_family_detection_probes(&mut self) -> &mut Self;

    pub fn get_family_detection_probes(&self) -> Execution;
}
```

NobodyXu (changes requested):
Thank you so much!

Just some feedback on code but should be quite to settle

NobodyXu:
> clang_android is the same root cause at a third site — src/lib.rs:3723 probes llvm-ar with a bare Command::new, bypassing Build::env's PATH. I left it out: fixing it needs a third recording class, which expands a design you settled at two, and it needs its own test since your CI runners ship a real llvm-ar. Happy to send it separately.



Yeah that indeed worth a separate PR, thank you for this PR!

MohammedAlkindi:
All four done in 94eca3a. Thanks, the `CommandExt` one in particular made the tool module read better.

**The extra collect.** Gone. `split_first` gives the program and the rest in one step, so `record` now takes `&[String]` and there is only the one collect of `env::args()`.

**`clone_from`.** Taken as suggested.

**`CommandExt`.** `ProbeKind` and `set_probe_env` are private again. The trait is on `Command` with `set_family_detection_env` and `set_flag_supported_env`, so `tool.rs` and `lib.rs` name the class rather than passing an enum.

**The harness helper.** `collect_family_detection_probes` and `collect_flag_supported_probes`, with `get_family_detection_probes` and `get_flag_supported_probes` to read them back. `CC_SHIM_OUT_FILES_FOR_*` and the file names now live entirely inside `Test`, so a test opts a class in by name.

One deviation from your sketch, and it is the only place I did not follow it exactly. The getters take an index:

```rust
pub fn get_flag_supported_probes(&self, i: usize) -> Execution
```

`clang_cl_scope_all` records two flag support probes in one build, one for `-fmacro-prefix-map` and one for `-fdebug-prefix-map`, so a single file per class could not hold both. Each class now gets four slots, filled in the order the probes run, and the test asks for 0 and 1. If you would rather have the singular signature with a separate accessor for the multi-probe case, say so and I will split it.

Verified on Windows 11, `x86_64-pc-windows-msvc`, from this branch:

- every test binary green, and doc-tests 28 passed
- `tests/test.rs` run one test per process: 58 of 59 pass, the exception being `clang_android`, which is the pre-existing third probe site we agreed belongs in its own PR
- `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean
- MSRV `check -p cc` on 1.65.0 clean

One honest note on the test run. `cargo test --test archiver` failed on this machine with `An Application Control policy has blocked this file (os error 4551)`, which is a Windows policy blocking cargo's spawn of the freshly built binary, not anything in the code. Running that binary directly gives `3 passed`, and it passes on the unmodified branch too, so it is my box rather than this change.

And agreed on `clang_android`, I will send that separately.

src/tool.rs
```diff
@@ -108,13 +124,20 @@ impl Tool {
...
         cuda: bool,
+        env: &BuildEnv,
         cached_compiler_family: &RwLock<CompilerFamilyLookupCache>,
         cargo_output: &CargoOutput,
         out_dir: Option<&Path>,
     ) -> Self {
-        fn is_zig_cc(path: &Path, cargo_output: &CargoOutput) -> bool {
+        /// Set up a compiler family detection probe to run in `env`.
+        fn probe<'cmd>(cmd: &'cmd mut Command, env: &BuildEnv) -> &'cmd mut Command {
+            cmd.set_family_detection_env(env);
```
NobodyXu: Do we still need this function? I think we can remove it?

tests/support/mod.rs
```diff
@@ -139,10 +163,56 @@ impl Test {
         env::join_paths(path).unwrap()
     }
 
+    /// Read back the `i`-th invocation a build actually performed.
+    ///
+    /// cc's own probing invocations are not recorded unless a test asks for
+    /// them by name with [`Test::probe_out_files`], so this numbering covers
+    /// the compile and archive commands only.
     pub fn cmd(&self, i: u32) -> Execution {
+        self.execution(self.td.path().join(format!("out{}", i)))
```
NobodyXu: Can we use 

```suggestion
        self.execution(self.probe_slot("out", i))
```

NobodyXu (changes requested):
Thanks, just two final nits, and it'd be ready for merge

MohammedAlkindi:
Both done in f028de4.

**The `probe` wrapper is gone.** `set_family_detection_env` already returns `&mut Command`, so the four call sites now chain it directly and read better for it, for example:

```rust
Command::new(path).arg("--version").set_family_detection_env(env)
```

**`Test::cmd` reuses `probe_slot`**, as suggested. It takes `usize` now rather than `u32` so the two share a signature.

Verified on Windows 11, `x86_64-pc-windows-msvc`: every other test binary green, doc-tests 28 passed, `tests/test.rs` 58 of 59 one test per process, clippy `-D warnings` clean, fmt clean, MSRV `check -p cc` on 1.65.0 clean. The one exception is `clang_android`, which is the third probe site and is fixed in #1867 rather than here, so it is expected to abort on this branch.

One piece of sequencing, since you mentioned merging this one first. **#1867 still points at the pre-nit commit**, because it was already approved and I did not want to force-push a rebase under the approval. If this lands as a squash, #1867's branch will carry the old versions of `tool.rs` and `support/mod.rs` and would undo f028de4 on merge. I will rebase it onto `main` as soon as this merges, so it arrives as just the one `llvm-ar` commit. Worth not merging #1867 until I have done that.

NobodyXu (approved):
Thank you, LGTM! Will wait for rebase on the follow-up PR

PR #1875, opened 2026-08-28
fix: probe flag support without OUT_DIR via tempfile

### LINKED ISSUE #1765: Fix flag_if_supported when OUT_DIR not set
Reference rust-lang/rust#158169

Use a tempfile if not exists

cestercian:
`Build::flag_if_supported` queues flags that are later applied via `is_flag_supported_inner(...).unwrap_or(false)`. The probe writes `flag_check.c` under `get_out_dir()`, which only succeeds when `Build::out_dir()` was set or `OUT_DIR` is in the environment. Outside a Cargo build script (rustc bootstrap, rust-lang/rust#158169), that lookup errors, every flag is treated as unsupported, and flags such as `-gz` are silently dropped.

`get_out_dir` is unchanged: `compile()` / archive output still require `OUT_DIR`. Only the flag-support probe falls back to unique `NamedTempfile`s under `std::env::temp_dir()` when the dir is missing.

## Changes

- `is_flag_supported_inner`: tempfile fallback when `OUT_DIR` / `out_dir` is unset; existing `flag_check.c` reuse when `OUT_DIR` is present
- Test helper `Test::gcc_without_out_dir` plus `flag_if_supported_without_out_dir` covering unset `OUT_DIR`

## Related

Fixes #1765. Bootstrap symptom: rust-lang/rust#158169.

## Testing

```
cargo test --locked --test test flag_if_supported
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Probe flag support without `OUT_DIR` via tempfile, so `flag_if_supported` no longer silently drops flags outside Cargo build scripts
+
 ## [1.4.4](https://github.com/rust-lang/cc-rs/compare/cc-v1.4.3...cc-v1.4.4) - 2026-08-21
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Probe flag support without `OUT_DIR` via tempfile, so `flag_if_supported` no longer silently drops flags outside Cargo build scripts
+- Probe flag support without `OUT_DIR` via tempfile, so `flag_if_supported` no longer silently drops flags outside Cargo build scripts ([#1875](https://github.com/rust-lang/cc-rs/pull/1875))
 
 ## [1.4.4](https://github.com/rust-lang/cc-rs/compare/cc-v1.4.3...cc-v1.4.4) - 2026-08-21
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1491,7 +1491,9 @@ impl Build {
     }
 
     fn write_flag_check_src(file: &mut fs::File) -> io::Result<()> {
-        write!(file, "int main(void) {{ return 0; }}")
+        write!(file, "int main(void) {{ return 0; }}")?;
+        file.flush()?;
+        file.sync_data()
     }
 
     fn ensure_check_file(&self) -> Result<PathBuf, Error> {
@@ -1533,8 +1535,6 @@ impl Build {
                 Self::write_flag_check_src(&mut tmp_file)?;
                 // Close the handle before invoking the compiler; Windows
                 // cannot open a file that another handle still holds.
-                tmp_file.flush()?;
-                tmp_file.sync_data()?;
                 drop(tmp_file);
 
                 let mut tmp_obj = crate::tempfile::NamedTempfile::new(&dir, "flag_check")?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1507,9 +1507,51 @@ impl Build {
             return Ok(is_supported);
         }
 
-        let out_dir = self.get_out_dir()?;
-        let src = self.ensure_check_file()?;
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Probe flag support without `OUT_DIR` via tempfile, so `flag_if_supported` no longer silently drops flags outside Cargo build scripts
```
NobodyXu: Release changelog is auto generated, if you want to manually create one, please include the PR number as `#1875`

src/lib.rs
```diff
@@ -1507,9 +1507,51 @@ impl Build {
...
 
-        let out_dir = self.get_out_dir()?;
-        let src = self.ensure_check_file()?;
-        let obj = out_dir.join("flag_check");
+        // Cargo build scripts have `OUT_DIR`; reuse `flag_check.c` there so
+        // probes stay cheap. Callers such as rustc bootstrap do not, and
+        // treating a missing dir as "unsupported" silently drops flags.
+        // Fall back to unique tempfiles instead of a shared name in `/tmp`.
+        // Held until `cmd.output()` returns so Drop can remove the temp files.
+        let mut temp_files = None;
```
NobodyXu: Can you extract this to a function please?

I think it may get reused in the future

src/lib.rs
```diff
@@ -1507,9 +1507,51 @@ impl Build {
...
+                fs::create_dir_all(&tmp_dir)?;
+
+                let suffix = if self.cuda {
+                    assert!(self.cpp);
+                    "flag_check.cu"
+                } else if self.cpp {
+                    "flag_check.cpp"
+                } else {
+                    "flag_check.c"
+                };
```
NobodyXu: It seems to be mostly the same code as the `ensure_check_file`, can we extract a fn?

NobodyXu (changes requested):
Thank you! Some feedback on the code

CHANGELOG.md
cestercian: Will add `#1875` to the Unreleased entry.

src/lib.rs
cestercian: Will extract the tempfile fallback into a function.

src/lib.rs
cestercian: Will share the check-file write with `ensure_check_file`.

CHANGELOG.md
cestercian: Landed in aa1f6a94: Unreleased entry now includes `#1875`.

src/lib.rs
cestercian: Landed in aa1f6a94: probe dir/tempfile setup is `flag_support_probe_files`.

src/lib.rs
cestercian: Landed in aa1f6a94: `flag_check_src_name` / `write_flag_check_src` shared with `ensure_check_file`.

src/lib.rs
```diff
@@ -1466,25 +1479,81 @@ impl Build {
...
+                fs::create_dir_all(&dir)?;
+
+                let mut tmp_src =
+                    crate::tempfile::NamedTempfile::new(&dir, self.flag_check_src_name())?;
+                let mut tmp_file = tmp_src.take_file().unwrap();
+                Self::write_flag_check_src(&mut tmp_file)?;
+                // Close the handle before invoking the compiler; Windows
+                // cannot open a file that another handle still holds.
+                tmp_file.flush()?;
+                tmp_file.sync_data()?;
```
NobodyXu: Can you move this to `write_flag_check_src` please?

NobodyXu (changes requested):
Thank you, just one last comment to be resolved!

src/lib.rs
cestercian: Will move flush/sync into `write_flag_check_src`.

cestercian: Landed in 2fe48916: flush/sync now inside `write_flag_check_src`; NamedTempfile drop stays at the call site for Windows.

cestercian: Landed in 2fe48916: `flush`/`sync_data` now live inside `write_flag_check_src`. `drop(tmp_file)` stays at the NamedTempfile call site.

NobodyXu (approved):
Thank you!

top2tanvir123-ai:
#1875

PR #1882, opened 2026-09-04
fix: update fd handling to use BorrowedFd

xtqqczze:


src/parallel/stderr.rs
```diff
@@ -8,14 +10,14 @@ use crate::{Error, ErrorKind};
 compile_error!("Only unix and windows support non-blocking pipes! For other OSes, disable the parallel feature.");
 
 #[cfg(unix)]
-fn get_flags(fd: std::os::unix::io::RawFd) -> Result<i32, Error> {
-    let flags = unsafe { libc::fcntl(fd, libc::F_GETFL, 0) };
+fn get_flags(fd: std::os::unix::io::BorrowedFd) -> Result<i32, Error> {
```
NobodyXu:
```suggestion
fn get_flags(fd: std::os::unix::io::BorrowedFd<'_>) -> Result<i32, Error> {
```

I think we'd want explicit lifetime parameter

NobodyXu (changes requested):
Thank you! Just one feedback

src/parallel/stderr.rs
xtqqczze: It seems unnecessary, is there a lint that requires this?

NobodyXu: Yes `#![deny(elided_lifetimes_in_paths)]`

xtqqczze: I see:

```
error: hidden lifetime parameters in types are deprecated
   --> src/parallel/stderr.rs:30:37
    |
 30 | fn set_flags(fd: std::os::unix::io::BorrowedFd, flags: core::ffi::c_int) -> Result<(), Error> {
    |                  -------------------^^^^^^^^^^
    |                  |
    |                  expected lifetime parameter
    |
```

NobodyXu (approved):
Thank you

PR #1884, opened 2026-09-06
fix: Prevent check_exe from dropping flags on the floor.

### LINKED ISSUE #1883: #1646 drops flags after executable.
#1646 started running `check_exe` in a way that breaks some of Firefox's windows builds.

Those builds run with:

```
CC=/path/to/clang.exe --driver-mode=cl
```

It seems cc-rs is dropping the `--driver-mode=cl`, because:

```rs
            let exe_ext = std::env::consts::EXE_EXTENSION;
            let check =
                exe.exists() || (!exe_ext.is_empty() && exe.set_extension(exe_ext) && exe.exists());
            check.then_some(exe)
```

Since `exe.exists()` is false, `set_extension` turns  `/path/to/clang.exe --driver-mode=cl` to `/path/to/clang.exe`.

Without `--driver-mode=cl` a bunch of the cflags are not valid and the build fails as seen here:

https://treeherder.mozilla.org/logviewer?job_id=590609849&repo=try&task=IOIQGr1fS36665mXz39wuQ.0&lineNumber=46524-46537

`check_exe` should probably not clobber an existing extension (or only append or so?)

emilio:
This prevents CC=/path/to/clang.exe --flags from losing flags.

Fixes #1883

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4857,10 +4857,23 @@ impl AsmFileExt {
     }
 }
 
-fn check_exe(mut exe: PathBuf) -> Option<PathBuf> {
+fn check_exe(exe: PathBuf) -> Option<PathBuf> {
+    if exe.exists() {
+        return Some(exe);
+    }
     let exe_ext = std::env::consts::EXE_EXTENSION;
-    let check = exe.exists() || (!exe_ext.is_empty() && exe.set_extension(exe_ext) && exe.exists());
-    check.then_some(exe)
+    if exe_ext.is_empty() {
+        return None;
+    }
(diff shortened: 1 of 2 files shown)


emilio:
@NobodyXu could you take a look whenever you have some time? Thanks in advance!

emilio:
We had a test for the exact situation that broke, but it didn't check that the flag was there unfortunately, I think the test change I made should cover it, lmk if there's some better check.

emilio:
Gah, `add_extension` is too new for this crate's MSRV...

NobodyXu (approved):
Thank you

PR #1893, opened 2026-09-10
Add support for CUDA on Windows on ARM.

dot-asm:
Side note. Depending on what one does one might have to set NVCC_PREPEND_FLAGS environment varialbe to "-std c++20 -Xcompiler /Zc:preprocessor".

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1817,6 +1817,10 @@ impl Build {
                 } else if cfg!(target_env = "msvc") {
                     libdir.push("lib");
                     match target.arch {
+                        "aarch64" => {
+                            libdir.push("arm64");
+                            libtst = true;
+                        }
                         "x86_64" => {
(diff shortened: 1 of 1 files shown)


NobodyXu (changes requested):
Thank you!

Is it possible to run CI for it?

https://github.com/rust-lang/cc-rs/blob/29acb758e4843920b606c59f207c70e20de7395b/.github/workflows/main.yml#L454

dot-asm:
> Thank you!
> 
> Is it possible to run CI for it?
> 
> https://github.com/rust-lang/cc-rs/blob/29acb758e4843920b606c59f207c70e20de7395b/.github/workflows/main.yml#L454

I'm not aware of any non-GUI method to install CUDA toolkit on Windows. I did look for it.

NobodyXu (approved):
Thank you!

PR #1899, opened 2026-09-13
fix(deps): update minimal-versions for wasi

xtqqczze:
Fix minimal version requirements for wasi and add a check to CI.

This also refactors the MSRV workflow configuration in CI by avoiding a matrix build, reducing GitHub Actions cache usage.


NobodyXu (approved):
Thank you

xtqqczze:
@taiki-e Is there currently a better way to do this using `cargo-minimal-versions`? It seems like having support for `generate-lockfile` would be useful. This looks related to https://github.com/taiki-e/cargo-minimal-versions/issues/17.

taiki-e:
The correct approach is to call cargo-minimal-versions multiple times. Instead of simply calling `generate-lockfile` and then `cargo-hack`, `cargo-minimal-versions` performs various operations to avoid false negatives. `generate-lockfile` + `cargo-hack`, which does not accurately emulate this behavior, is always worse than `cargo-minimal-versions`. See https://github.com/taiki-e/cargo-minimal-versions#details for details.

.github/workflows/main.yml
```diff
@@ -499,15 +497,24 @@ jobs:
...
-            cargo-hack
-            cargo-minimal-versions
+          tool: cargo-hack
           fallback: none
+      # Cache the crates.io registry because older Cargo versions do not support the sparse registry protocol.
+      # This can be removed once the MSRV is >= 1.68.
       - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
-      - run: cargo minimal-versions check --workspace --all-targets --feature-powerset --ignore-private
+      - run: RUSTC_BOOTSTRAP=1 cargo -Zminimal-versions generate-lockfile
+      - run: cargo hack check --locked --workspace --all-targets --feature-powerset --ignore-private --target aarch64-apple-darwin
```
taiki-e, line 514: --all-targets is usually a bad idea here since it enables dev-deps. This often hides compile errors that exist in lib/bin which will be released on crates.io. See also the quoted comment in https://github.com/taiki-e/cargo-minimal-versions#details.

xtqqczze:
@taiki-e Is there a way to run `cargo minimal-versions check` for multiple targets while resolving the minimal-version `Cargo.lock` only once?

taiki-e:
Does multi-target (https://github.com/taiki-e/cargo-hack/issues/167) not work?

xtqqczze:
> Does multi-target ([taiki-e/cargo-hack#167](https://github.com/taiki-e/cargo-hack/issues/167)) not work?

I didn't know about Cargo's multi-target builds! Since that feature was implemented in 1.64, we should be good to use it with our MSRV of 1.65.

taiki-e:
> Since that feature was implemented in 1.64

Btw, cargo-hack (used inside cargo-minimal-versions) emulates multi-target builds on pre-1.64 Rust (https://github.com/taiki-e/cargo-hack/pull/168).

.github/workflows/main.yml
xtqqczze: Is `--no-dev-deps` necessary with `cargo-minimal-versions`?

taiki-e: cargo-minimal-versions automatically do the same processing as --no-dev-deps when `--all-targets`/`--tests`/etc. are not used.

PR #1902, opened 2026-09-13
fix: strip OUT_DIR from the object file name hash too

### LINKED ISSUE #1901: Object names still hash the absolute OUT_DIR for generated sources (gap in #1270)
cc hashes the source file's directory into the object file name. #1270 strips CARGO_MANIFEST_DIR from that directory first, which handles in-tree sources. It misses sources a build script generates into OUT_DIR: if the target dir is outside the package (CARGO_TARGET_DIR set elsewhere, as out-of-tree/distro builds do), OUT_DIR is not under CARGO_MANIFEST_DIR, so that strip doesn't help and the OUT_DIR path goes into the hash. The object name then changes with the build location.

Repro: a crate that generates a .c into OUT_DIR and compiles it, built twice with different target dirs.

```toml
# Cargo.toml
[package]
name = "ccrepro"
version = "0.1.0"
edition = "2021"
build = "build.rs"
[build-dependencies]
cc = "1"
```
```rust
// build.rs
use std::{env, fs, path::Path};
fn main() {
    let out = env::var("OUT_DIR").unwrap();
    let src = Path::new(&out).join("gen.c");
    fs::write(&src, "int gen_symbol(void){return 42;}\n").unwrap();
    cc::Build::new().file(&src).compile("gen");
    for e in fs::read_dir(&out).unwrap() {
        let p = e.unwrap().path();
        if p.extension().is_some_and(|x| x == "o") {
            println!("cargo:warning={}", p.file_name().unwrap().to_string_lossy());
        }
    }
}
```
```
$ cd /tmp/ccrepro
$ CARGO_TARGET_DIR=/tmp/ccrepro-a cargo build 2>&1 | grep -E 'Compiling cc v|gen.o'
   Compiling cc v1.4.6
warning: ccrepro@0.1.0: e8a56c71b9820449-gen.o
$ CARGO_TARGET_DIR=/tmp/ccrepro-b cargo build 2>&1 | grep gen.o
warning: ccrepro@0.1.0: 334cae97e1210068-gen.o
```
Same source and flags, only the build path differs, but the object names differ (cc 1.4.6). Same result with a .S file instead of the .c; the language doesn't matter, only the directory. A source kept under CARGO_MANIFEST_DIR is
already stable since #1270, so this is specifically the OUT_DIR case.

Where it bites: rustc's compiler-builtins generates its outline-atomics sources (`lse_*.S`) into OUT_DIR, so on out-of-tree builds (e.g. Yocto) libcompiler_builtins and librustc_driver end up with a build-path-dependent prefix on the `lse_*.o` names and aren't reproducible.

#1277 reported the same thing and was closed as a dup of #1270, but #1270 doesn't cover OUT_DIR. Fix is to strip OUT_DIR too (falling back to CARGO_MANIFEST_DIR) before hashing. With that change, both builds above give
db3b6bfb95261072-gen.o. I can send a PR.

windsunil:
Fixes #1901

`objects_from_files` hashes the source dir into the object name and strips `CARGO_MANIFEST_DIR` first (#1270). A source a build script writes into `OUT_DIR` is not under the manifest dir when the target dir is elsewhere, so the absolute path still goes into the hash. This strips `OUT_DIR` first and falls back to `CARGO_MANIFEST_DIR`; in-tree sources are unaffected.

Repro from the issue, before/after:

```
1.4.6:  e8a56c71b9820449-gen.o   334cae97e1210068-gen.o
fixed:  db3b6bfb95261072-gen.o   db3b6bfb95261072-gen.o
```

The new test fails on 1.4.6 and passes with this change. #1277 was an earlier attempt at the same thing, closed as covered by #1270.


src/command_helpers.rs
```diff
@@ -308,9 +308,19 @@ pub(crate) fn objects_from_files(files: &[Arc<Path>], dst: &Path) -> Result<Vec<
...
-            let root = root.to_string_lossy();
-            Cow::Borrowed(dirname.strip_prefix(&*root).unwrap_or(&dirname))
+        // Sources generated by a build script live under OUT_DIR, which is not under
+        // CARGO_MANIFEST_DIR when the target dir is elsewhere, so try OUT_DIR first.
+        let strip_root = |var| {
+            cargo_env_var_os(var).and_then(|root| {
+                dirname
+                    .strip_prefix(&*root.to_string_lossy())
+                    .map(str::to_owned)
+            })
```
NobodyXu, line 318:
```suggestion
            Cow::Owned(cargo_env_var_os(var).and_then(|root| {
                dirname
                    .strip_prefix(&*root.to_string_lossy())
                    .map(str::to_owned)
            }))
```

NobodyXu (changes requested):
Thank you just one minor nit

src/command_helpers.rs
windsunil: Applied in 1b6d242, thanks. I kept the Option on the outside so it
  type-checks: the closure maps into Cow::Owned and the call site is
  now an or_else/unwrap_or chain.

NobodyXu (approved):
Thank you! LGTM

PR #1904, opened 2026-09-13
WIP: Update MSRV to 1.68

xtqqczze:


xtqqczze:
Now we can use the Cargo sparse registry protocol, we don't benefit from caching in the MSRV job. We can remove the `setup-rust-toolchain` action and use `--rust-version` in `cargo-minimal-versions` instead.

xtqqczze:
@NobodyXu I'm not sure the matrix build is a significant improvement here. The main benefit would be making `minimal-build` failures easier to attribute to a specific target, but that seems fairly marginal compared with the additional CI jobs and complexity.

NobodyXu:
Yeah we don't have to use matrix, previously it's mostly for setting runner-os and caching, but now we donit all on ubuntu-latest so it'd be fine

xtqqczze:
@NobodyXu This is done, but I’d like to update 1.65 → 1.66 → 1.67 first. Closing this for now.

PR #1909, opened 2026-09-14
chore: rename clippy.toml

xtqqczze:
Rename `clippy.toml` to `.clippy.toml`

Just to fix ordering with other config like `.tombi.toml`.


NobodyXu (approved):
Thanks

PR #1930, opened 2026-09-19
fix: pass -Tp on MSVC for .cc so they are not treated as objects

### LINKED ISSUE #1877: Compile error in mimalloc 0.1.52
System: Windows 11 Home 25H2, OS Version: 26200.9168
cc version: 1.4.4
mimalloc version: 0.1.52
Cargo.toml config:
[dev-dependencies]
mimalloc = "0.1"

Codes:
`
use mimalloc::MiMalloc;

#[global_allocator]
static GLOBAL_ALLOC: MiMalloc = MiMalloc;

fn main() {
    test1_basic();
}

fn test1_basic() {
    let mut v = Vec::with_capacity(1024);
    v.push(666);
    println!("v: {:?}", v);
}
`

In Visual Studio Code and I use cargo run or cargo build, it report the error below:

error: failed to run custom build command for `libmimalloc-sys v0.1.49`
Caused by:
  process didn't exit successfully: `C:\prjs_rust\_git_src\rust_learning\mytest02\target\debug\build\libmimalloc-sys-3169b469e5474a5c\build-script-build` (exit code: 1)
  --- stderr

  error occurred in cc-rs: command did not execute successfully (status code exit code: 2): "C:\\Program Files (x86)\\Microsoft Visual Studio\\18\\BuildTools\\VC\\Tools\\MSVC\\14.51.36231\\bin\\HostX64\\x64\\cl.exe" "-nologo" "-MD" "-Z7" "-Brepro" "-std:c++17" "-I" "c_src/mimalloc/v3/include" "-I" "c_src/mimalloc/v3/src" "-W4" "/Zc:__cplusplus" "-DMI_DEBUG=0" "-FoC:\\prjs_rust\\_git_src\\rust_learning\\mytest02\\target\\debug\\build\\libmimalloc-sys-0b3f78552cb42b55\\out\\2a4e75ae9feb7516-mimalloc-static.o" "-c" "C:\\prjs_rust\\_git_src\\rust_learning\\mytest02\\target\\debug\\build\\libmimalloc-sys-0b3f78552cb42b55\\out\\mimalloc-static.cc"

I looked for the file based on the error message and found that the file 2a4e75ae9feb7516-mimalloc-static.o is missing. But why?

cestercian:
## Summary

MSVC does not treat `.cc` as C++. With `cpp(true)`, mimalloc 0.1.52 writes `OUT_DIR/mimalloc-static.cc`; without `/TP`, `cl.exe` assumes that file is an object (D9024) and never produces the `-Fo` output (`*-mimalloc-static.o`).

Pass `-TP` on MSVC-like compilers when `cpp(true)`.

Fixes #1877

## Test plan

- [x] `cargo +stable test --locked --test test` — 59 passed (includes `msvc_cpp_cc_source_not_treated_as_object`)
- [x] clippy on `cc` clean


NobodyXu (changes requested):
Thank you!

Is it possible to use `-Tp` instead, to recognise the immediate following `.cc` as c++ source files?

`/TP` could potentially break c code as C++ as slightly different rules, sometimes stricter

cestercian:
@NobodyXu Switched to per-file `-Tp` immediately before the `.cc` path (no compiler-wide `/TP` / `-TP`), so C inputs and flag probes stay C. Regression test now asserts that shape. Head: `de5683e`.

src/lib.rs
```diff
@@ -1999,6 +1997,34 @@ impl Build {
...
+            // assumed to be an object unless `-Tp` forces C++ compilation
+            // (#1877).
+            cmd.arg("-Tp");
+        } else if compiler.supports_path_delimiter() && !is_assembler_msvc {
+            // #513: For `clang-cl`, separate flags/options from the input file.
+            // When cross-compiling macOS -> Windows, this avoids interpreting
+            // common `/Users/...` paths as the `/U` flag and triggering
+            // `-Wslash-u-filename` warning.
+            cmd.arg("--");
+        }
```
NobodyXu: Ok there's a few issues we need to solve here
 - we shouldn't need to check `!is_assembler_msvc` for `/Tp`
 - `--` is used for `clang-cl` to signal end of the options here so we cannot pass `/Tp` afterwards, however the good news is, we don't need it because clang-cl can recognise `.cc`
 - we need to pass `/Tp` only for `.cc` files


so we need to do something like this:


```suggestion
        if
            self.cpp &&
            src.extension == Some("cc") &&
            matches!(compiler.family, ToolFamily::Msvc { clang_cl: false })
        {
            // MSVC recognizes only `.c` / `.cpp` / `.cxx` as source. A `.cc`
            // file (mimalloc 0.1.52 writes `OUT_DIR/mimalloc-static.cc`) is
            // assumed to be an object unless `-Tp` forces C++ compilation
            // (#1877).
            cmd.arg("-Tp");
        }

        if compiler.supports_path_delimiter() && !is_assembler_msvc {
            // #513: For `clang-cl`, separate flags/options from the input file.
            // When cross-compiling macOS -> Windows, this avoids interpreting
            // common `/Users/...` paths as the `/U` flag and triggering
            // `-Wslash-u-filename` warning.
            cmd.arg("--");
        }
```

src/lib.rs
```diff
@@ -866,6 +866,11 @@ impl Build {
     /// 2. Else if the `CXXSTDLIB` environment variable is set, use its value.
     /// 3. Else the default is `c++` for OS X and BSDs, `c++_shared` for Android,
     ///    `None` for MSVC and `stdc++` for anything else.
+    ///
+    /// On MSVC this also passes `-Tp` immediately before each source file, so
+    /// inputs such as `.cc` are compiled as C++ rather than assumed to be
```
```suggestion
    /// On MSVC this also passes `-Tp` immediately before each `.cc` source file to ensure that
    /// they are compiled as C++ rather than assumed to be
```

NobodyXu (changes requested):
Thank you, just want the implementation to be changed as described and have the doc updated

You would probably want to update regression test to ensure C files aren't interpreted as C++, and that `/Tp` isn't used for `clang-cl`

src/lib.rs
cestercian: makes sense, narrowed it to -Tp only for .cc on MSVC (not clang-cl), and left the -- path alone

src/lib.rs
cestercian: updated the cpp() docs to match

NobodyXu (approved):
Thank you!

A new release is scheduled on next weekend, if you need it early, I can cut an early release

PR #1931, opened 2026-09-19
chore: keep CI-only commits out of the changelog

### LINKED ISSUE #1898: Changelog should not include CI-only changes
CI-only changes shouldn't appear in the user-facing changelog.

For example, GitHub Actions dependency bumps were included in the [`1.4.6` changelog](https://github.com/rust-lang/cc-rs/blob/main/CHANGELOG.md#146---2026-09-11), despite having no impact on the published `cc` crate.

I'd suggest excluding changes that are purely related to CI or development infrastructure, while continuing to include anything that can affect crate users.

A simple rule could be:

> Only include changes in the changelog if they affect the published crate or are otherwise relevant to its users.

cestercian:
## Summary

Configure release-plz changelog parsers to skip CI-only subjects (`ci:`, `ci(…):`) and Dependabot Actions bumps (`chore(deps): bump …`). Point Dependabot Actions commits at `ci(deps):` so they hit the same skip.

User-facing `chore:` and crate dep updates still land in the changelog.

Fixes #1898

## Test plan

- [x] Parser simulation against 24 historical subjects — 24/24
- [x] `tombi format --check release-plz.toml`


release-plz.toml
```diff
@@ -0,0 +1,17 @@
...
+# CI-only commits are excluded from the user-facing changelog (#1898).
+
+[changelog]
+# Custom parsers replace release-plz defaults; keep Keep a Changelog groups
+# and skip CI commits first so they are not caught by the catch-all.
+commit_parsers = [
+    { message = "^ci(?:\\(.*\\))?:", skip = true },
+    { message = "^chore\\(deps\\): bump ", skip = true },
+    { message = "^feat", group = "added" },
+    { message = "^changed", group = "changed" },
```
xtqqczze, line 12: Many of these we don't use. For example, there are no commits with a title matching `^changed`:

```sh
git log --format='%s' | grep '^changed'
```

I guess there's no harm in keeping it though.

.github/dependabot.yml
```diff
@@ -8,6 +8,10 @@ updates:
     interval: "daily"
   cooldown:
     default-days: 7
+  # Prefix GHA bumps as ci(deps): so release-plz can skip them (#1898).
```
xtqqczze: No need to include issue number here

release-plz.toml
```diff
@@ -0,0 +1,17 @@
...
+    { message = "^deprecated", group = "deprecated" },
+    { message = "^removed", group = "removed" },
+    { message = "^fix", group = "fixed" },
```
xtqqczze: This will match "fixup! " commits generated by `git commit --fixup`, which is probably undesirable.

cestercian: Good catch — narrowed `^fix` to `^fix(?:\(.*\))?:` so `fixup!` from `git commit --fixup` is no longer grouped as “fixed”. No `^fixup!` skip parser.

.github/dependabot.yml
cestercian: Dropped the issue number from the Dependabot comment. Prefix is still `ci` / `ci(deps):`.

release-plz.toml
cestercian: Agreed they’re unused today — left them as soft forward-compat for conventional-commit titles. Happy to drop if you’d rather keep the file leaner.

release-plz.toml
```diff
@@ -0,0 +1,17 @@
+# Changelog generation for release-plz / git-cliff.
+# CI-only commits are excluded from the user-facing changelog (#1898).
```
xtqqczze: Please exclude the issue number from the changes.

.github/dependabot.yml
```diff
@@ -8,6 +8,10 @@ updates:
     interval: "daily"
   cooldown:
     default-days: 7
+  # Prefix GHA bumps as ci(deps): so release-plz can skip them.
```
NobodyXu:
```suggestion
  # Prefix GHA bumps as `ci(deps):` so release-plz can skip them.
```

NobodyXu (changes requested):
Thank you! One nit feedback in additional to xtqqczze's feedback

.github/dependabot.yml
cestercian: Thanks 😊

.github/dependabot.yml
```diff
@@ -8,6 +8,10 @@ updates:
     interval: "daily"
   cooldown:
     default-days: 7
+  # Prefix GHA bumps as `ci(deps):` so release-plz can skip them.
```
xtqqczze: Might be unnecessary to mention `release-plz` here. We probably want this commit convention anyway.

cestercian: fine by me, dropped the release-plz mention

.github/dependabot.yml
```diff
@@ -8,6 +8,10 @@ updates:
     interval: "daily"
   cooldown:
     default-days: 7
+  # Prefix GHA bumps as `ci(deps):` so they follow the repo commit convention.
```
xtqqczze: This seems tautological, just remove the comment?

xtqqczze:
@NobodyXu Comparing the changelog for [v1.4.7](https://github.com/rust-lang/cc-rs/releases/tag/cc-v1.4.7) with the auto-generated changelog shown in https://github.com/rust-lang/cc-rs/pull/1928, there are quite a few commits that would not be excluded correctly. If we want this to work reliably, we’ll need to follow the Conventional Commits format going forward. It might therefore be worth having a separate PR to document this in `CONTRIBUTING.md`.

release-plz.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[changelog]
+# Custom parsers replace release-plz defaults; keep Keep a Changelog groups
+# and skip CI commits first so they are not caught by the catch-all.
+commit_parsers = [
+    { message = "^ci(?:\\(.*\\))?:", skip = true },
+    { message = "^chore\\(deps\\): bump ", skip = true },
```
xtqqczze: @NobodyXu Should we exclude all commits matching `^chore:`? This would also cover hypothetical commits like `chore: rename clippy.toml`.

xtqqczze: @NobodyXu Also what is your opinion on including this recent commit in the changelog: `refactor(dev-tools): update prelude comment for bindgen`?

NobodyXu: I'd say we should exclude both of them, with the exception of regeneration of target-info/windows-sys which could be of interest to the end users

.github/dependabot.yml
cestercian: yep, dropped it

release-plz.toml
cestercian: sounds good, skipping chore/refactor broadly now and keeping target-info/windows-sys regen in the changelog

release-plz.toml
cestercian: done, issue number is out of that comment

NobodyXu (approved):
Thank you!

If this looks good to @xtqqczze I'll merge it

release-plz.toml
```diff
@@ -0,0 +1,21 @@
+# Changelog generation for release-plz / git-cliff.
+# CI-only commits are excluded from the user-facing changelog.
+
+[changelog]
+# Custom parsers replace release-plz defaults; keep Keep a Changelog groups.
+# Match target-info / windows-sys regeneration first so later skip rules
+# do not hide those user-facing commits. Then skip CI, chore, and refactor.
+commit_parsers = [
+    { message = "(?i)^(?:(?:chore|refactor)(?:\\(.*\\))?: )?regenerat(?:e|ion).*(?:target.?info|windows.?sys)", group = "other" },
```
xtqqczze: I think this is overfitting, we don't have any commits that would match the `chore: regenerat` case.

xtqqczze:
@NobodyXu As a follow-up, we may want to document in CONTRIBUTING.md that we follow Conventional Commits and add a check to CI to enforce this, what do you think?

release-plz.toml
cestercian: fair, dropped that regenerat exception

release-plz.toml
xtqqczze: can we tweak this like the others

release-plz.toml
```diff
@@ -0,0 +1,19 @@
+# Changelog generation for release-plz / git-cliff.
+# CI-only commits are excluded from the user-facing changelog.
+
+[changelog]
+# Custom parsers replace release-plz defaults; keep Keep a Changelog groups.
+# Skip CI, chore, and refactor conventional subjects.
+commit_parsers = [
+    { message = "^ci(?:\\(.*\\))?:", skip = true },
+    { message = "^chore\\(deps\\): bump ", skip = true },
```
xtqqczze: this one is now redundant, no?

cestercian: yep, redundant under the broad chore skip, removed

release-plz.toml
cestercian: tweaked, feat and security now use the same scoped pattern

xtqqczze (approved):
Looks fine to me; this can always be tweaked later. Could you please rebase this to a clean history of commits?

cestercian:
thanks, rebased to a single clean commit on latest main

xtqqczze (approved):

NobodyXu:
> @NobodyXu As a follow-up, we may want to document in CONTRIBUTING.md that we follow Conventional Commits and add a check to CI to enforce this, what do you think?

That's a good idea, especially on the generated files and changelog, to remind people they don't have to manually write changelog

xtqqczze:
@NobodyXu This might require a change in merge strategy, looks like a squash merge is used currently?

NobodyXu:
Yes we currently use squash merge for a clean history

xtqqczze:
> Yes we currently use squash merge for a clean history

We could require a clean history and switch to Rebase and Merge instead.

NobodyXu:
> We could require a clean history and switch to Rebase and Merge instead.

Is there any particular benefit to that?

xtqqczze:
> Is there any particular benefit to that?

If a PR contains more than one type of commit, release-plz won’t see those individual commit types if it categorises the change solely based on the PR title.

NobodyXu:
I see...personally I think squash merge does make web based editing easier

For pr, I prefer it to focus on one type (fix/feature) not multiple of them, I think it's rare for one to have multiple and even in that case, there's a major and minor

